PR #11, opened 2023-11-07
test(e2e): update to add approval comment workflow for e2e trigger

charliedmcb:
Fixes # 

**Description**
Currently for PR check-in the E2E suite is run via an approval comment directly for the e2e trigger workflow.

This works to test things, but overall has some quality of life issues, as if another approval comes in the trigger will be checked again, un-linking the previous e2e run from the PR regardless of if a new E2E suite run was requested or not.

This PR is step one of two meant to break apart the approval comment check so that the E2E suite run will not be overridden by a new review. This one added the new trigger scenario. Another follow up will be made once tested to remove the old trigger.

Not able to test since this trigger only works when in the default branch. Due to this I'm only adding the new trigger, and not removing the old one. Once we've tested that it works correctly in the `main` branch than I'll remove the old trigger scenario.
*

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
NONE
```

### DIFF
--- a/.github/workflows/approval-comment.yaml
+++ b/.github/workflows/approval-comment.yaml
@@ -1,0 +1,13 @@
+name: ApprovalComment
+on:
+  pull_request_review:
+    types: [submitted]
+
+jobs:
+  approval-comment:
+    if: startsWith(github.event.review.body, '/e2e')
+    runs-on: ubuntu-latest
+    steps:
+      - name: run e2e
+        run: |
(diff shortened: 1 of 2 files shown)


charliedmcb (commented):
testing no e2e requested

charliedmcb (commented):
/e2e (testing E2E requested)

tallaxes (approved):

PR #13, opened 2023-11-08
chore(armclient): removing 429 sdk retries in favor of controller retries

Bryce-Soghigian:
Fixes # 

**Description**
Karpenter is a reconciler, and will retry the scale up or delete request if it fails. Also note it only retries on 429 and some other options by default. 

Instead of retrying with SDK we should retry with karpenter, and if we need additional retry flexiblity with backoff for throttling aim to get this change into karpenter core. 

Scale up scale down 
make az-presubmit 
e2etests

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/utils/opts/armopts.go
+++ b/pkg/utils/opts/armopts.go
@@ -34,6 +34,10 @@ func DefaultArmOpts() *arm.ClientOptions {
 
 func DefaultRetryOpts() policy.RetryOptions {
 	return policy.RetryOptions{
+		// MaxRetries specifies the maximum number of attempts a failed operation will be retried
+		// before producing an error.
+		// The default value is three.  A value less than zero means one try and no retries.
+		// See Reference here: https://github.com/Azure/azure-sdk-for-go/blob/v61.4.0/sdk/azcore/policy/policy.go#L73
 		MaxRetries: -1,
 	}
 }
--- a/pkg/utils/opts/armopts.go
+++ b/pkg/utils/opts/armopts.go
@@ -18,7 +18,6 @@ package opts
 
 import (
 	"net/http"
-	"time"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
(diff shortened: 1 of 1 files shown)


Bryce-Soghigian (commented):
/test

pkg/utils/opts/armopts.go
```diff
@@ -35,21 +34,7 @@ func DefaultArmOpts() *arm.ClientOptions {
...
-		//   http.StatusRequestTimeout      408
-		//   http.StatusTooManyRequests     429
-		//   http.StatusInternalServerError 500
-		//   http.StatusBadGateway          502
-		//   http.StatusServiceUnavailable  503
-		//   http.StatusGatewayTimeout      504
-		// Specifying values will replace the default values.
-		// Specifying an empty slice will disable retries for HTTP status codes.
-		// StatusCodes: nil,
+		MaxRetries: -1,
```
matthchr, line 37: Consider leaving a comment here explaining what `-1` means, as there are APIs where this could easily mean "retry forever", but in the Go SDK it means no retries.

matthchr (dismissed):
Approved but we should make sure we have test coverage that demonstrates that the controller-runtime loop-based backoffs do the right thing here. Do we?

pkg/utils/opts/armopts.go
Bryce-Soghigian: yes, can update with comment. 

> Approved but we should make sure we have test coverage that demonstrates that the controller-runtime loop-based backoffs do the right thing here. Do we?

I would like to spend some time seperately investigating Performance and potentailly looking at controller options for retries separately on 429. Ideally karpenter core would stop sending requests for some timestamp when seeing 429, and wait to requeue until that interval passes, but need to think through this more. Will add when im at that stage. 

for now just removing the 20 retries we do for each failure case to lower throttling.

matthchr: Consider that the RP can return `retryAfter` to you to, which you can extract from the response headers and flow upstream to Karpenter (via `ctrl.Result{}` with a specific time), which in the case of throttling especially may be worth doing.

So while karpenter-core should probably be doing some basic backoff, etc, it might still be worth thinking about how our code can do the right thing here (or give Karpenter core the tools to do the right thing).

matthchr (approved):

PR #23, opened 2023-11-10
fix: remove confidential VM SKUs (DC,EC) from consideration

tallaxes:
Fixes # 

**Description**

This removes confidential VM SKUs (DC,EC) from consideration, since the current provider does not support them (yet). The implementation looks for DC or EC prefix in the VM Size name, as apparently there is no direct confidential capability indication available in SKU resource API. Not filtering all *C, because NCs are supported GPU SKUs.

The above is the only functional change, the rest is refactoring and test. 
Some of the refactoring was needed to consolidate all filtering in one place, and without triggering gocyclo.
Consolidated filter also fixes the post-filter count of SKUs reported in the log. Expect it to be lower (and not just because confidential VMs are now filtered out).

* `make presubmit`

- [ ] Yes, issue opened: # 
- [x] No


tallaxes (commented):
/test

Bryce-Soghigian (approved):

PR #24, opened 2023-11-10
test(e2e): allow for manual triggering of E2E suite

charliedmcb:
Fixes # 

**Description**
Currently there is no way to test within the github action workflows a new E2E test suite. 

I'd like to be able to have it automatically pick up the change to the set of test suites and use that. However, the workflows are being run via their state on the `main` branch. This solution is meant as an alternative solution by allowing a developer to trigger a e2e suite run, and then they can share the link within their PR description.

Nothing additional
*

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
NONE
```

### DIFF
--- a/.github/workflows/e2e-matrix.yaml
+++ b/.github/workflows/e2e-matrix.yaml
@@ -17,6 +17,10 @@ on:
         required: true
       E2E_SUBSCRIPTION_ID:
         required: true
+  workflow_dispatch:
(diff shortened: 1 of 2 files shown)


charliedmcb:
Currently don't have a need for additional manual run controls for the E2E suites, so closing this.

PR #25, opened 2023-11-10
test(e2e): add git_ref for manual trigger to allow testing on branches

charliedmcb:
Fixes # 

**Description**
Triggering the `E2EMatrixTrigger` workflow for a non-main branch doesn't actually end up testing on the non-main branch. Instead it ends up still testing the main branch.

This is particularly an issue when someone is updating the e2e test matrix. We want to allow devs to test their changes to the e2e matrix before checking it in. 

By making the resolve args special case for manual triggers and grab the branch name this should allow for devs to manually trigger the workflows for their changes and get them tested before checking in.

Manually ran an E2E test on this branch:
https://github.com/Azure/karpenter/actions/runs/6882110677
Note: at the time I added an additional fake E2E test suite to the matrix called `Another` just to ensure things were being picked up correctly.

Also, note that confirming the git_ref on this run matches the correct commit.
*

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
NONE
```


charliedmcb (commented):
/test

tallaxes (approved):

PR #28, opened 2023-11-12
fix: consider NodeClaim linked annotation on Delete

tallaxes:
**Description**

This has `cloudprovider.Delete` use NodeClaim linked annotation (`karpenter.sh/linked`) for determining VM providerID, if `NodeClaim.providerID` is not set. 

This may fix "machine.termination" reconciler error of "parsing vm name from provider id", on trying to delete NodeClaim with empty providerID.

The above is the only functional change. The rest is refactoring, extracting the single point of reference to NodeClaim linked annotation.

* `make presubmit`

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note

```


rakechill (approved):
everything here lgtm and would make sense if some nodeclaims were linked. however, I didn't think we were explicitly linking nodeclaims for any purpose yet, so I'm a bit confused how people would be getting into a situation where nodeclaims were created w/o a providerID

tallaxes:
> I didn't think we were explicitly linking nodeclaims for any purpose yet, so I'm a bit confused how people would be getting into a situation where nodeclaims were created w/o a providerID

Link controller can create new NodeClaims for existing VMs, though not sure how one can end up without providerID

tallaxes (commented):
/test

PR #29, opened 2023-11-13
fix: remove support for HyperV generation selectors

tallaxes:
Fixes: https://github.com/Azure/karpenter/issues/32

**Description**

Removes support for selecting HyperV generation via requirements. The label is still used internally, e.g. for determining image compatibility, but is not available for explicit selection via NodePool (will be rejected if specified). As before, if a given VM SKU supports both generations, Gen2 will be used. Thus the ability to explicitly request Gen1 is the only capability removed, which is deemed acceptable.

This should also fix unexpected RequirementsDrift, due to incorrect HyperV generation label placed on NodeClaim, in turn likely reducing flakiness of E2E test runs.

**Background** 

Since HyperV generation is a potentially multi-valued requirement - there are VM SKUs that support both Gen1 and Gen2 - the actual NodeClaim label value needs to be fetched (by instanceToNodeClaim) from VM itself. This would require introducing VM GET for each provisioned VM, and the more expensive "instanceView" flavor. While this is indeed possible (and for List we can obtain the equivalent via Azure Graph), the alternative of not supporting explicit selection of HyperV generation appears to be a better option for now.
 
This is implemented by removing the corresponding label from WellKnownLabels, and including it in the new RestrictedLabels set. These restricted labels can still be used internally (and are included in AllowedUnknownLabels in requirement compatibility checks), but are not allowed to be used in NodePool requirements.

* `make presubmit`, E2E tests

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


tallaxes (commented):
/test

Bryce-Soghigian (approved):

PR #31, opened 2023-11-13
fix(bootstrap): kubernetes 1.28 support / remove outdated feature gates

tallaxes:
**Description**

Remove all feature gates passed to kubelet: `CSIMigrationAzureFile`, `DelegateFSGroupToCSIDriver`, `RotateKubeletServerCertificate`, fixing bootstrap on Kubernetes 1.28 nodes.

Adds 1.28 to test matrix (fixes https://github.com/Azure/karpenter/issues/6), and test coverage reporting via [coveralls](https://coveralls.io/builds/63958038).

Background:

`DelegateFSGroupToCSIDriver` feature flag is no longer supported in Kubernetes 1.28. More importantly, all of the currently set flags already default to true in all versions we care about:

Feature | Default | Stage | Since | Until
-- | -- | -- | -- | --
CSIMigrationAzureFile | true | Beta | 1.24 | 1.25
CSIMigrationAzureFile | true | GA | 1.26
DelegateFSGroupToCSIDriver | true | Beta | 1.23 | 1.25
DelegateFSGroupToCSIDriver | true | GA | 1.26 | 1.27
RotateKubeletServerCertificate | true | Beta | 1.12

Reference:
https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/
https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates-removed/

* `make presubmit`, E2E tests, manually creating 1.28 cluster & scaling up

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


tallaxes (commented):
/test

.github/workflows/ci-test.yml
```diff
@@ -18,7 +18,9 @@ jobs:
       with:
         k8sVersion: ${{ matrix.k8sVersion }}    
     - run: K8S_VERSION=${{ matrix.k8sVersion }} make ci-test
-#    - uses: shogo82148/actions-goveralls@v1
-#      if: env.K8S_VERSION == '1.25.x'
-#      with:
-#        path-to-profile: coverage.out
+    - name: Send coverage
+      # should only send converage once https://docs.coveralls.io/parallel-builds
+      if: matrix.k8sVersion == '1.27.x'
```
charliedmcb, line 23: Why only `1.27.x`? 

Just since we don't need coverage more than once?

If so would like a note calling out this reasoning.

charliedmcb (commented):
Overall, looks good, but a little confusing on the `1.27.x` coverage, so looking for an answer there

.github/workflows/ci-test.yml
tallaxes: Yes. It is called out in the comment. The only reason I picked 1.27 is that that's the latest of the ones we have been testing extensively.

Bryce-Soghigian (approved):

PR #35, opened 2023-11-15
fix: support non-zonal regions

tallaxes:
**Description**

Add support for non-zonal regions / VMs.
Will also correctly filter out VM SKUs with location restriction.

* `make presubmit`, manual testing in non-zonal region (westcentralus), E2E tests

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


tallaxes (commented):
/test

pkg/providers/instancetype/instancetypes.go
```diff
@@ -233,3 +237,49 @@ func MaxEphemeralOSDiskSizeGB(sku *skewer.SKU) float64 {
 	// convert bytes to GB
 	return maxDiskBytes / float64(units.Gigabyte)
 }
+
+var (
+	// https://learn.microsoft.com/en-us/azure/reliability/availability-zones-service-support#azure-regions-with-availability-zone-support
+	// (could also be obtained programmatically)
```
rakechill, line 243: Probably makes more sense to add a hack/codegen script for this similar to instance types, right?

tallaxes (commented):
/test

pkg/providers/instancetype/instancetypes.go
tallaxes: For production, we could just determine this directly by using the API on startup; only need to do it for the region in use.
For unit testing - Yes, it would be useful to have non-zonal region represented, though codegen would be an overkill, I think.

tallaxes: Thinking about this some more - codegen for _instance types_ themselves for a non-zonal region would actually make sense, and may be needed for proper unit test coverage

Bryce-Soghigian (approved):
E2E Tests pass, ran tests locally and zonal behavior seems to behave as we expect. 

There is a bit of concern about the hardcoded regions. Basically for each new azure region we will need to bump this list. But thats true as well for our image gallery, and we will have a clear heads up when to do so. 

given we follow up with an item for making this process programatic in the future, and this is certainly a step forward, lets approve and follow up with work to see if a programatic approach to getting the regions is better.

PR #40, opened 2023-11-17
test(e2e): fix drift E2E to use ReplaceRequirements

charliedmcb:
Fixes # 

**Description**
Drift E2E has been a bit flaky with overprovisioning. While investigating I noticed that there was an issue where the drift E2E test was using spot instances:
image

This should not be the case. We should be using on-demand for drift.

Found there was an issue where we were replacing the Requirements on the nodepool, rather than using the override func. 

Nothing additional.
*

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
NONE
```


charliedmcb (commented):
/test

tallaxes (approved):
Good catch!

PR #43, opened 2023-11-22
fix: handle SkuNotAvailable provisioning error

tallaxes:
**Description**

Handle `SkuNotAvailable` provisioning error, representing out of capacity for Spot, by taking the SKU out of consideration for 1hr. (In case for some reason this is observed for on-demand - should not happen, see below - it will still defensively take SKU out of consideration, but for 23hrs.)

Based on https://aka.ms/azureskunotavailable, it can happen if the SKU is either not available for a location or zone, or out of capacity for Spot. We only expect to observe the Spot case, not location or zone restrictions, because:
- SKUs with location restriction are already filtered out via sku.HasLocationRestriction, see https://github.com/Azure/karpenter/pull/35. (Originally SkuNotAvailable was observed prior to this filter).
- Zonal restrictions are filtered out internally by sku.AvailabilityZones, and don't get offerings

* `make presubmit`, manual testing (have not yet observed Spot out-of-capacity)

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


pkg/providers/instance/instance.go
```diff
@@ -452,14 +462,20 @@ func (p *Provider) handleResponseErrors(ctx context.Context, instanceType *corec
...
-		p.unavailableOfferings.MarkUnavailable(ctx, ZonalAllocationFailure, instanceType.Name, zone, corev1beta1.CapacityTypeOnDemand)
-		p.unavailableOfferings.MarkUnavailable(ctx, ZonalAllocationFailure, instanceType.Name, zone, corev1beta1.CapacityTypeSpot)
+		p.unavailableOfferings.MarkUnavailable(ctx, ZonalAllocationFailureReason, instanceType.Name, zone, corev1beta1.CapacityTypeOnDemand)
+		p.unavailableOfferings.MarkUnavailable(ctx, ZonalAllocationFailureReason, instanceType.Name, zone, corev1beta1.CapacityTypeSpot)
+	} else if isSKUNotAvailable(err) {
+		// SKU not available (usually as a result of restriction),
+		// mark the instance type as unavailable for all offerings
+		for _, offering := range instanceType.Offerings {
+			p.unavailableOfferings.MarkUnavailableWithTTL(ctx, SKUNotAvailableReason, instanceType.Name, offering.Zone, capacityType, SKUNotAvailableTTL)
+		}
```
Bryce-Soghigian, line 478: We seem to be developing a pattern where we mark all offerings unavailable for a SKU and a given reason. Should we add a helper for this?

pkg/providers/instance/instance.go
```diff
@@ -440,10 +444,16 @@ func (p *Provider) launchInstance(
 	return resp, instanceType, nil
 }
 
+// isSKUNotAvailable - to be moved to azure-sdk-for-go-extensions
+func isSKUNotAvailable(err error) bool {
+	azErr := sdkerrors.IsResponseError(err)
+	return azErr != nil && azErr.ErrorCode == "SkuNotAvailable"
+}
```
Bryce-Soghigian, line 451: Can you share the arm call and the response we get back here? Just so others(Me) can be aware of the full shape of the SkuNotAvailable Subcode?

pkg/providers/instance/instance.go
tallaxes: For now there are two cases, sightly different ones (the other is sensitive to CapacityType), localized and very readable, so maybe later?

pkg/providers/instance/instance.go
tallaxes: > Code="SkuNotAvailable" Message="The requested size for resource '(redacted)' is currently not available in location '(redacted)' zones '1,2,3' for subscription '(redacted)'. Please try another size or deploy to a different location or zones. See https://aka.ms/azureskunotavailable for details."

tallaxes: This actually makes me think that maybe we can be smarter about the zones here? Though I have not observed this with other than all zones in the region, it may well be possible ...

Bryce-Soghigian (approved):

pkg/providers/instance/instance.go
tallaxes: Giving this some more thought, it seems we don't expect to observe SkuNotAvailable for regular instances once we introduce filtering based on SKU location restriction in https://github.com/Azure/karpenter/pull/35. Refocusing this to handle the remaining case of out-of-capacity for Spot, see updated comments in code.

pkg/providers/instancetype/suite_test.go
```diff
@@ -489,6 +491,34 @@ var _ = Describe("InstanceType Provider", func() {
 			node := ExpectScheduled(ctx, env.Client, pod)
 			Expect(node.Labels).To(HaveKeyWithValue(v1.LabelInstanceTypeStable, "Standard_D2_v2"))
 		})
+
+		Context("on SkuNotUnavailable, should cache SKU as unavailable in all zones", func() {
+			AssertUnavailable := func(sku string, capacityType string) {
+				// fake a SKU not available error
+				azureEnv.VirtualMachinesAPI.VirtualMachinesBehavior.VirtualMachineCreateOrUpdateBehavior.Error.Set(
+					&azcore.ResponseError{ErrorCode: instance.SKUNotAvailableErrorCode},
+				)
```
tallaxes, line 500: In the future this should be replaced with more capable helpers requesting specific errors for given SKUs. Should also likely include a valid http.Response ...

Bryce-Soghigian (approved):
LGTM, please remove SKUNotAvailable into azure-sdk-for-go-extensions when you get a chance. 


Also it might be nice to add some test functions for response errors in extensions as well.

PR #45, opened 2023-11-23
feat: supporting additional handing for TotalRegionalCores quota error, and adding better customer facing error messages on nodeclaim

Bryce-Soghigian:
Fixes # 

**Description**
This PR adds support for TotalRegionalCores errors and makes it more apparent for the customer that this is the root cause of the problem for why their scale up stopped. 

This also modifies the code to use a less generic check for subscription level quota errors, and specifically only marking instances as unavailable if we see a SKUFamily quota error instead of just all subscription level quota errors. 

We also are only logging the Azure Errors and are returning better customer facing errors in this case since they will be viewed in the nodeclaim. 

- Scale Up observing better errors.
- make presubmit

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/cloudprovider/cloudprovider.go
+++ b/pkg/cloudprovider/cloudprovider.go
@@ -103,7 +103,7 @@ func (c *CloudProvider) Create(ctx context.Context, nodeClaim *corev1beta1.NodeC
 	}
 	instance, err := c.instanceProvider.Create(ctx, nodeClass, nodeClaim, instanceTypes)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("creating instance, %w", err)
 	}
 	instanceType, _ := lo.Find(instanceTypes, func(i *cloudprovider.InstanceType) bool {
 		return i.Name == string(*instance.Properties.HardwareProfile.VMSize)
--- a/pkg/cloudprovider/cloudprovider.go
+++ b/pkg/cloudprovider/cloudprovider.go
@@ -103,7 +103,7 @@ func (c *CloudProvider) Create(ctx context.Context, nodeClaim *corev1beta1.NodeC
 	}
 	instance, err := c.instanceProvider.Create(ctx, nodeClass, nodeClaim, instanceTypes)
 	if err != nil {
-		return nil, fmt.Errorf("creating instance, %w", err)
+		return nil, err
 	}
 	instanceType, _ := lo.Find(instanceTypes, func(i *cloudprovider.InstanceType) bool {
 		return i.Name == string(*instance.Properties.HardwareProfile.VMSize)
--- a/pkg/providers/instance/instance.go
+++ b/pkg/providers/instance/instance.go
@@ -503,7 +503,7 @@ func (p *Provider) handleResponseErrors(ctx context.Context, instanceType *corec
 	if sdkerrors.RegionalQuotaHasBeenReached(err) {
 		logging.FromContext(ctx).Error(err)
 		//nolint:stylecheck // Ignore ST1005: error strings should not be capitalized. This error message will pop up in the machine CRD and is intended to be read directly by the customer
-		return corecloudprovider.NewInsufficientCapacityError(errors.New("The regional capacity limit for your subscription has been reached. To scale beyond this limit, please review the quota increase process here: https://learn.microsoft.com/en-us/azure/quotas/regional-quota-requests"))
+		return corecloudprovider.NewInsufficientCapacityError(errors.New("regional quota limit for subscription has been reached. To scale beyond this limit, please review the quota increase process here: https://learn.microsoft.com/en-us/azure/quotas/regional-quota-requests"))
 	}
 	return err
 }
--- a/pkg/providers/instance/instance.go
+++ b/pkg/providers/instance/instance.go
@@ -454,7 +454,7 @@ func isSKUNotAvailable(err error) bool {
 
 func (p *Provider) handleResponseErrors(ctx context.Context, instanceType *corecloudprovider.InstanceType, zone, capacityType string, err error) error {
 	if sdkerrors.SKUFamilyQuotaHasBeenReached(err) {
-		// Subscription quota has reached for this vm sku, mark the instance type as unavailable in all zones available to the offering
+		// Subscription quota has been reached for this VM SKU, mark the instance type as unavailable in all zones available to the offering
 		// This will also update the TTL for an existing offering in the cache that is already unavailable
 
 		logging.FromContext(ctx).Error(err)
--- a/pkg/providers/instance/instance.go
+++ b/pkg/providers/instance/instance.go
@@ -461,9 +461,7 @@ func (p *Provider) handleResponseErrors(ctx context.Context, instanceType *corec
 		for _, offering := range instanceType.Offerings {
 			if offering.CapacityType != capacityType {
 				continue
-			}
-
-			// If we have a quota limit of 0 vcpus, we mark the offerings unavailable for an hour.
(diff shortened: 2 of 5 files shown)


Bryce-Soghigian:
@juan-lee  Let me know if you think this leads to a better experience in more clearly communicating you hit TotalRegionCores errors!

Bryce-Soghigian (commented):
/test

pkg/providers/instance/instance.go
```diff
@@ -457,9 +460,21 @@ func (p *Provider) handleResponseErrors(ctx context.Context, instanceType *corec
 				p.unavailableOfferings.MarkUnavailable(ctx, SubscriptionQuotaReached, instanceType.Name, offering.Zone, capacityType)
 			}
 		}
-	} else if sdkerrors.ZonalAllocationFailureOccurred(err) {
+		//nolint:stylecheck // Ignore ST1005: error strings should not be capitalized. This error message will pop up in the machine CRD and is intended to be read directly by the customer
+		return fmt.Errorf("We've encountered the subscription level quota on the resources available for the selected instance type (%v). Karpenter will try to provision using an alternative instance type.", instanceType.Name)
```
tallaxes, line 464:
```suggestion
		return fmt.Errorf("subscription level quota for %s has been reached (may try provision an alternative instance type): %w", instanceType.Name, err)
```
- suggest avoiding capitalization based on intended use. (Also, wouldn't this error be further wrapped anyway?)
- can't know it will try provisioning alternative (may not have options)
- best avoid using "Karpenter" in this context
- any benefit to using %v over just %s here?
- any reason not to wrap/include the original error?

pkg/providers/instance/instance.go
```diff
@@ -457,9 +460,21 @@ func (p *Provider) handleResponseErrors(ctx context.Context, instanceType *corec
...
+	}
+	if sdkerrors.ZonalAllocationFailureOccurred(err) {
+		logging.FromContext(ctx).With("zone", zone).Error(err)
 		p.unavailableOfferings.MarkUnavailable(ctx, ZonalAllocationFailure, instanceType.Name, zone, corev1beta1.CapacityTypeOnDemand)
 		p.unavailableOfferings.MarkUnavailable(ctx, ZonalAllocationFailure, instanceType.Name, zone, corev1beta1.CapacityTypeSpot)
+
+		//nolint:stylecheck // Ignore ST1005: error strings should not be capitalized. This error message will pop up in the machine CRD and is intended to be read directly by the customer
+		return fmt.Errorf("We're currently unable to allocate resources in the selected zone (%s). Karpenter will try a different zone to fulfill your request.", zone)
```
tallaxes, line 472: Same comments here

pkg/providers/instance/instance.go
```diff
@@ -457,9 +460,21 @@ func (p *Provider) handleResponseErrors(ctx context.Context, instanceType *corec
...
+	}
+	if sdkerrors.RegionalQuotaHasBeenReached(err) {
+		logging.FromContext(ctx).Error(err)
+		//nolint:stylecheck // Ignore ST1005: error strings should not be capitalized. This error message will pop up in the machine CRD and is intended to be read directly by the customer
+		return corecloudprovider.NewInsufficientCapacityError(errors.New("The regional capacity limit for your subscription has been reached. To scale beyond this limit, please review the quota increase process here: https://learn.microsoft.com/en-us/azure/quotas/regional-quota-requests"))
```
tallaxes, line 477:
```suggestion
		return corecloudprovider.NewInsufficientCapacityError(errors.New("regional quota limit for subscription has been reached. To scale beyond this limit, please review the quota increase process here: https://learn.microsoft.com/en-us/azure/quotas/regional-quota-requests"))
```
* Same comments as elsewhere
* Let's add an explicit comment on why InsufficientCapacityError is appropriate here, but not elsewhere

tallaxes (changes requested):
LGTM, mostly minor feedback in comments. Could we also have tests for these cases? Could follow example in https://github.com/Azure/karpenter/pull/43/commits/58586db60e22f7cab940b79b9033f9755d29b528

pkg/providers/instance/instance.go
Bryce-Soghigian: Q: any reason not to wrap/include the original error?
A: The character limit for errors included in the nodeclaim is 300 characters. With this message, the only parts included are the separator bits `-----------------------------------`. I figured it would be a better user experience to only see the relevant bit. 
Q: any benefit to using %v over just %s here?
A: Good shout will change.
Q: Wouldn't this error be wrapped further anyway? 
A: Yes, but the wrapping is [minimal](https://github.com/Azure/karpenter/blob/main/pkg/cloudprovider/cloudprovider.go#L106)
Q: suggest avoiding capitalization based on intended use
A: Sure I can.
Q: can't know it will try provisioning alternative (may not have options)
A: Yes, but it will clearly communicate that in the next error message? But I can see the case against because customers sometimes use error messages as a contract or promise so your verbage might make sense in that case.

pkg/providers/instance/instance.go
```diff
@@ -452,26 +453,25 @@ func isSKUNotAvailable(err error) bool {
...
 				p.unavailableOfferings.MarkUnavailable(ctx, SubscriptionQuotaReachedReason, instanceType.Name, offering.Zone, capacityType)
 			}
 		}
-	} else if sdkerrors.ZonalAllocationFailureOccurred(err) {
-		p.unavailableOfferings.MarkUnavailable(ctx, ZonalAllocationFailureReason, instanceType.Name, zone, corev1beta1.CapacityTypeOnDemand)
-		p.unavailableOfferings.MarkUnavailable(ctx, ZonalAllocationFailureReason, instanceType.Name, zone, corev1beta1.CapacityTypeSpot)
-	} else if isSKUNotAvailable(err) {
+		return fmt.Errorf("subscription level quota for %s has been reached (may try provision an alternative instance type)", instanceType.Name)
+	}
+	if isSKUNotAvailable(err) {
```
tallaxes, line 474: Wrapping the original error is a better pattern, I think, but if the conclusion is it makes for unacceptable status error UX, let's be consistent and replace it with logging in this codepath as well.

tallaxes (changes requested):

pkg/providers/instance/instance.go
```diff
@@ -452,26 +453,25 @@ func isSKUNotAvailable(err error) bool {
...
 		// This will also update the TTL for an existing offering in the cache that is already unavailable
+
+		logging.FromContext(ctx).Error(err)
 		for _, offering := range instanceType.Offerings {
 			if offering.CapacityType != capacityType {
 				continue
-			}
-
-			// If we have a quota limit of 0 vcpus, we mark the offerings unavailable for an hour.
+			} // If we have a quota limit of 0 vcpus, we mark the offerings unavailable for an hour.
```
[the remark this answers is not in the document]
Bryce-Soghigian: good catch was not intentional.

pkg/cloudprovider/cloudprovider.go
```diff
@@ -103,7 +103,7 @@ func (c *CloudProvider) Create(ctx context.Context, nodeClaim *corev1beta1.NodeC
 	}
 	instance, err := c.instanceProvider.Create(ctx, nodeClass, nodeClaim, instanceTypes)
 	if err != nil {
-		return nil, fmt.Errorf("creating instance, %w", err)
+		return nil, err
```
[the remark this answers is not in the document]
Bryce-Soghigian: Sounds good, returning it again.

Bryce-Soghigian (commented):
/test

tallaxes (approved):
/test

PR #58, opened 2023-12-12
test: unit test provisioning in non-zonal regions

tallaxes:
**Description**

Update unit test environment to support testing zonal and non-zonal regions. Capture and use test SKU data from `eastus` and `westcentralus` (non-zonal). Adapt existing zonal tests to run for both zonal and non-zonal environments. Add unit test of provisioning in non-zonal region. 

**Details**

* Refactor pre-generated SKU data to support multiple regions 
   * Uses map with region -> SKUs
   * Expect file per-region, with map entry populated via init()
   * Update resource SKU API fake to be region-aware and use the above

* Update the SKU generation code for the above
   * Add support for generating SKU restriction data
   * Reduce auth configuration needs (use `NewDefaultAzureCredential`, [docs](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azidentity/README.md#defaultazurecredential)), only `SUBSCRIPTION_ID` is required
   * Minor fixes

* Update codegen (skugen) scripts to generate the above, 
  * Generate for both `eastus` and `westcentralus` (non-zonal)
  * Remove the unnecessary auth
  * Check that subscription used is not too restrictive

* Update unit test environment and test for non-zonal
  * Add support for non-zonal test environment
  * Adapt existing zone-aware tests to run for both zonal and non-zonal regions/environments
  * Add non-zonal provisioning test

* `make presubmit`

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note

```


hack/codegen.sh
```diff
@@ -19,54 +19,38 @@ pricing() {
 }
 
 skugen() {
-   SUBSCRIPTION_ID=$(az account show --query 'id' --output tsv)
-  export SUBSCRIPTION_ID
-  if [ -z "${SUBSCRIPTION_ID}" ]; then
-    echo "No subscription is set. Please login and set a subscription."
-    return 1
+  location=${1:-eastus}
```
rakechill, line 22: what does this do? does it basically say "use args[1] or default to eastus"? not sure i understand :)

pkg/test/environment.go
```diff
@@ -60,19 +60,29 @@ type Environment struct {
...
+	// Settings
+	nonZonal bool
 }
 
 func NewEnvironment(ctx context.Context, env *coretest.Environment) *Environment {
+	return NewRegionalEnvironment(ctx, env, fake.Region, false)
+}
+
+func NewEnvironmentNonZonal(ctx context.Context, env *coretest.Environment) *Environment {
+	return NewRegionalEnvironment(ctx, env, "westcentralus", true)
```
rakechill, line 73: nit: maybe have a `fake.NonZonalRegion`?

pkg/providers/instance/suite_test.go
```diff
@@ -107,24 +108,34 @@ var _ = Describe("InstanceProvider", func() {
...
+		azureEnv.Reset()
+		azureEnvNonZonal.Reset()
 	})
+
+	var ZonalAndNonZonalRegions = []TableEntry{
+		Entry("zonal", azureEnv, cloudProvider),
+		Entry("non-zonal", azureEnvNonZonal, cloudProviderNonZonal),
+	}
+
+	DescribeTable("should return an ICE error when all attempted instance types return an ICE error",
```
rakechill, line 121: Do we plan to make more of our tests table-driven?

pkg/providers/instancetype/suite_test.go
```diff
@@ -71,31 +71,38 @@ import (
 var ctx context.Context
 var stop context.CancelFunc
 var env *coretest.Environment
-var azureEnv *test.Environment
+var azureEnv, azureEnvNonZonal *test.Environment
```
rakechill, line 74: I know we need separate vars for nonzonal vs. zonal, but it feels a bit messy to have to make two of each of these in all the test files. I don't have a good suggestion for how to fix, so I'll still approve. Just wanted to call it out in case you have some ideas for easier extensibility.

rakechill (dismissed):
Basically just had some questions sprawled throughout for my own understanding. One real suggestion is to add a `fake.NonZonalRegion`. Overall looks good, so I'll approve

hack/code/instancetype_testdata_gen.go
```diff
@@ -32,23 +33,21 @@ import (
 
 func main() {
 	fmt.Println("starting generation of sku data...")
-	tenantID := os.Getenv("TENANT_ID")
```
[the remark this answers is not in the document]
tallaxes: Just reading the docs on [`DefaultAzureCredential`](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azidentity/README.md#defaultazurecredential) (updated PR description with doc link). I suspect we just did not have a good example of using it.

hack/codegen.sh
tallaxes: Yes, exactly. This is just idiomatic bash.

Explaining something like this is where Copilots shine, BTW. Use your favorite integration, or just from directly from GitHub:

image

You will get this:

image

pkg/providers/instance/suite_test.go
tallaxes: No specific plans, but we can definitely use this pattern where it makes sense.

pkg/providers/instancetype/suite_test.go
tallaxes: Agreed. I did not see an obvious better (and still clean) way to implement non-zonal vs zonal, besides a separate environment. However, what I think we could (and probably should) do, is isolate the tests that need both, so that we only need to use this pattern in one place, and the rest of tests can stay unaffected.

pkg/test/environment.go
tallaxes: Extracted const. (May still need a better location for it.)

hack/codegen.sh
rakechill: that's awesome

rakechill (approved):

PR #60, opened 2023-12-13
chore(pricing): cleanup pricing duplication

charliedmcb:
Fixes # 

**Description**
In the pricing provider it was fetching from the pricing API twice (once for spot, and another for on-demand)

Refactored the logic to only fetch once.

Note: the locking feels a little messy as during the price update we grab the lock around the fetch, release it, and then grab it again when updating the prices. However, I don't see this actually causing issues. I think this could still be cleaned up in another PR, but think it'd add a little more refactoring I want to leave out of this one.

Nothing additional
*

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
NONE
```


charliedmcb (commented):
/test

charliedmcb (commented):
/test

charliedmcb (commented):
/test

charliedmcb (commented):
/test

charliedmcb (commented):
/test

charliedmcb (commented):
/test

charliedmcb (commented):
/test

PR #68, opened 2023-12-15
fix: removing nic after vm creation fails

### LINKED ISSUE #67: if NodeClaims are not satisfied, Karpenter still create the NIC and does no delete it
### Version

__Karpenter Version:__ 0.29.2

__Kubernetes Version:__ v1.27.7

### Expected Behavior

When a NodeClaim cant be satisfied and is removed, the NIC should not be created or deleted afterwards.

### Actual Behavior

Karpenter creates and deletes the NodeClaims but does not remove the NIC. So you end up with a lot of NICs in the `MC_` resource group.

run` kubectl scale deploy inflate --replicas 300 -n default` and  watch the Karpenter pods logs, you will see for example this error: `\"error\": {\n    \"code\": \"OperationNotAllowed\",\n    \"message\": \"Operation could not be completed as it results in exceeding approved Total Regional Cores quota. Additional details - Deployment Model: Resource Manager, Location: northeurope, Current Limit: 65, Current Usage: 54, Additional Required: 96, (Minimum) New Limit Required: 150.`

Karpenter creates the NodeClaim and you will see them appear and disappear (`kubectl get nodeclaim -w`). For every NodeClaim a NIC is created but not deleted. So you end up with a lot of NICs in the `MC_` resource group.

### Resource Specs and Logs

{"level":"ERROR","time":"2023-12-15T21:40:08.562Z","logger":"controller.nodeclaim.lifecycle","message":"creating instance, insufficient capacity, regional on-demand vCPU quota limit for subscription has been reached. To scale beyond this limit, please review the quota increase process here: https://learn.microsoft.com/en-us/azure/quotas/regional-quota-requests","commit":"76a3b5a-dirty","nodeclaim":"general-purpose-rmshj","nodepool":"general-purpose"}
{"level":"INFO","time":"2023-12-15T21:40:08.601Z","logger":"controller.nodeclaim.termination","message":"deleted nodeclaim","commit":"76a3b5a-dirty","nodeclaim":"general-purpose-rmshj","nodepool":"general-purpose","node":"","provider-id":""}

### Community Note

Bryce-Soghigian:
Fixes #67

**Description**
When we issue a VM.Create call, and the vm fails to create (happens frequently at the beginning of karpenter when its trying to find unavailable offerings), we don't delete the Network Interface. 

This pr simply adds that delete call

- make az-all, paired with a scale up that hit quota errors. 

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
fix: removing nic after vm creation fails
```

### DIFF
--- a/pkg/providers/instance/instance.go
+++ b/pkg/providers/instance/instance.go
@@ -400,6 +400,10 @@ func (p *Provider) createVirtualMachine(ctx context.Context, vm armcompute.Virtu
 		if vmErr != nil {
 			logging.FromContext(ctx).Errorf("virtualMachine.Delete for %s failed: %v", vmName, vmErr)
 		}
+		nicErr := deleteNicIfExists(ctx, p.azClient.networkInterfacesClient, p.resourceGroup, vmName)
+		if nicErr != nil {
+			logging.FromContext(ctx).Errorf("networkInterface.Delete for %s failed: %v", vmName, nicErr) 
+		}
 		return nil, fmt.Errorf("virtualMachine.BeginCreateOrUpdate for VM %q failed: %w", vmName, err)
 	}
 	logging.FromContext(ctx).Debugf("Created virtual machine %s", *result.ID)


Bryce-Soghigian (commented):
/test

pkg/providers/instancetype/suite_test.go
```diff
@@ -146,7 +146,37 @@ var _ = Describe("InstanceType Provider", func() {
 		ExpectCleanedUp(ctx, env.Client)
 	})
 
-	Context("subscription level quota error responses", func() {
+	Context("vm creation error responses", func() {
+
+		It("should delete the network interface on failure to create the vm", func() {
+			genericAzureErrMsg := "some generic error message"
```
rakechill, line 152: nit: i personally like this fun variable naming, but maybe better to make it formal. 
```
errorMsg := "test error"
errorCode := "400"

File: pkg/providers/instancetype/suite_test.go [hunk 2]
```diff
@@ -146,7 +146,37 @@ var _ = Describe("InstanceType Provider", func() {
...
+				},
+			)
+			pod := coretest.UnschedulablePod()
+			ExpectProvisioned(ctx, env.Client, cluster, cloudProvider, coreProvisioner, pod)
+			ExpectNotScheduled(ctx, env.Client, pod)
+
+			// We should have created a nic for the vm
+			Expect(azureEnv.NetworkInterfacesAPI.NetworkInterfacesCreateOrUpdateBehavior.CalledWithInput.Len()).To(Equal(1))
+			// The nic we used in the vm create, should be cleaned up if the vm call fails
+			nic := azureEnv.NetworkInterfacesAPI.NetworkInterfacesCreateOrUpdateBehavior.CalledWithInput.Pop()
```
rakechill, line 170: This should be the `DeleteBehavior`, right?

rakechill: Or I guess do we want to check that the DeleteBehavior is called at all? Or just rely on the ok == false check from Load?

rakechill (commented):
Overall looks really good and I like your approach here. Just have a few questions about your unit tests. 

Also wondering if you've been able to test this in a cluster where you've previously seen NICs hanging around to demonstrate that this change is effective at preventing that.

pkg/providers/instancetype/suite_test.go
Bryce-Soghigian: We should test both delete and create. These tests were added before, will update them.

pkg/providers/instancetype/suite_test.go
Bryce-Soghigian: it was originally "genericDepressedAzureErrMsg" but I see i need to formalize it a bit more :)

Bryce-Soghigian:
@rakechill  i did a scale to 750 nodes, and a scale down, and nics were mostly cleaned up. We do see 4 nics that were not and all of them failed due to the same error `NicReservedForAnotherVm`. I have a doc i am working on for this. Its a bit complex because we need to consider retires, non blocking deletion/creation etc. 


So wanted to cut in this pr first.

pkg/providers/instancetype/suite_test.go
Bryce-Soghigian: fixed.

Bryce-Soghigian:
Did another run, scaled to 300 or so then deleted all the nodes, we are left with one nic that remains due to NicReservedForAnotherVm

rakechill (approved):

pkg/utils/opts/armopts.go
```diff
@@ -25,6 +25,8 @@ import (
 	"github.com/Azure/karpenter/pkg/auth"
 )
 
+var MaxRetries = 20
```
tallaxes, line 28: Is this used anywhere?

PR #71, opened 2023-12-17
docs: modifying v1 of docs to point to managed experience

Bryce-Soghigian:
Fixes # 

**Description**
This PR adds a base for the readme, to point people towards the managed offering for an easy experience getting things setup and configured.  

We still need to solve documentation for self-hosted Helm Chart experience.

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


README.md
```diff
@@ -6,67 +6,42 @@
...
+az aks create --name myFirstNap --resource-group napTest --node-provisioning-mode Auto --network-plugin azure --network-plugin-mode overlay --network-dataplane cilium
 ```
-ERRO[0007] gcloud binary not found
-...
-ERRO[0003] gcloud binary not found
-...
-ERRO[0187] walk.go:74: found symbolic link in path: /workspaces/karpenter/charts/karpenter/crds resolves to /workspaces/karpenter/pkg/apis/crds. Contents of linked file included and used  subtask=0 task=Render
-```
-- If you see platform architecture error during `skaffold debug`, adjust (or comment out) `--platform` argument.
-- If you are not able to set/hit breakpoints, it could be an issue with source paths mapping; see comments in debug launch configuration (`launch.json`)
```
Bryce-Soghigian, line 48: After going through a couple of live streams of people trying to get karpenter configured following these steps, I saw a fair bit of confusion. 

For now, until we have a proper helm chart to point users to, lets remove the developer setup wiki.

README.md
```diff
@@ -6,67 +6,42 @@
...
-
-#### FAQs
 
-Q: I was able to trigger Karpenter to execute scaling up nodes as expected, using my own customized deployment of pods. However, scaling down was not handled automatically when I removed the deployment. The two new nodes created by Karpenter were left around. What is going on?
+[View Limitations of the node autoprovisioning preview here](https://learn.microsoft.com/en-gb/azure/aks/node-autoprovision?tabs=azure-cli#limitations)
```
Bryce-Soghigian, line 35: Should we maintain the list of limitations here instead?

tallaxes (changes requested):
Reorganizing for clarity is good, but please keep the development section (can be moved down) until we have something better. Even better, extract it into CONTRIBUTING.md and link to it.

Bryce-Soghigian:
@tallaxes do we want to include it in both the readme and the website? I guess we can wait for website pr to be merged before removing contribution experience section

README.md
rakechill: Better not to keep two sources of truth here I'd say. That link doesn't seem to capture the aks karpenter provider API version, but I think that's fine. If we need to add self-hosted specific limitations, then we could add a few bullet points here as needed. 

Might be worth calling out that that links to some information only relevant for our managed experience, but that the   `Limitations` section is relevant to self-hosted.

Bryce-Soghigian: Right now we don't have a self hosted experience besides makefile cmmands, so we should probably wait until we do

rakechill (approved):

tallaxes (approved):

PR #74, opened 2023-12-19
chore: change tags to karpenter.azure.com_cluster

philwelz:
**Description**

If you run `make az-resc` is does not show all resources created by Karpenter as the `resk` var only show resources with the label `karpenter.sh_nodepool`. But as you can see on [#67 ](https://github.com/Azure/karpenter/issues/67) Ghost Nics for example do not have the labels `karpenter.sh_nodepool` but they do have the labels `karpenter.azure.com_cluster`. So this PR changes the label used by the `resk` call.

It seems also that NICs never get the tag `karpenter.sh_nodepool` assigned.

```
make az-resc
az resource list --tag=karpenter.sh_nodepool --query "[?resourceGroup=='MC_rg-karpenter-pwe-tmp1_aks-karpenter-pwe-tmp1_northeurope']" -o tsv | wc -l
1
```
vs

```
make az-resc
az resource list --tag=karpenter.azure.com_cluster --query "[?resourceGroup=='MC_rg-karpenter-pwe-tmp1_aks-karpenter-pwe-tmp1_northeurope']" -o tsv | wc -l
38
```

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note
NONE
```

### DIFF
--- a/Makefile-az.mk
+++ b/Makefile-az.mk
@@ -225,7 +225,7 @@ az-perftest1000: ## Test scaling out/in (1000 VMs)
 az-resg: ## List resources in MC rg
 	az resource list -o table -g $(AZURE_RESOURCE_GROUP_MC)
 
-RESK=az resource list --tag=karpenter.sh_nodepool --query "[?resourceGroup=='$(AZURE_RESOURCE_GROUP_MC)']"
+RESK=az resource list --tag=karpenter.azure.com_cluster --query "[?resourceGroup=='$(AZURE_RESOURCE_GROUP_MC)']"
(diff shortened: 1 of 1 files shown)


Bryce-Soghigian (approved):

tallaxes:
> It seems also that NICs never get the tag karpenter.sh_nodepool assigned.

That's what needs to be fixed, I think, instead of listing by `karpenter.azure.com_cluster` tag. It would be better not to rely on the latter; we should consider retiring it altogether.

tallaxes:
Closing in favor of properly tagging network interfaces, #75

PR #84, opened 2024-01-05
feat(identity): add support for using workload id for karpenter pod

charliedmcb:
Fixes # 

**Description**
Currently the makefile steps, and E2E are running the karpenter pod using the cluster's kubeletIdentity. This is not the identity we want to be using for e2e, or presenting to repo users to be using either. The karpenter pod should have its own workload identity, which this PR adds support for.

Follow ups:
- Address logging, and `klog`
- remove SAVM, or support workload id with it
- update to using track 2 sdk, rather than the workaround through another lib+track 1 

Within a codespace so far, but will continue testing before completion.
*

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
Adding support for workload identities which is the patterning that the karpenter pod should be using.
```


Bryce-Soghigian (commented):
/test

charliedmcb (commented):
/test

charliedmcb (commented):
/test

charliedmcb (commented):
/test

charliedmcb (commented):
/test

charliedmcb (commented):
/test

charliedmcb (commented):
/test

charliedmcb (commented):
/test

charliedmcb (commented):
/test

charliedmcb (commented):
/test

charliedmcb (commented):
/test

charliedmcb (commented):
/test

Makefile-az.mk
```diff
@@ -92,18 +117,18 @@ az-rmvmss-vms: ## Delete all VMs in VMSS Flex (use with care!)
...
-	# Note (charliedmcb): need to be objectId for E2E workflow as the pipeline identity doesn't have permissions to "query Graph API"
-	$(eval AZURE_OBJECT_ID=$(shell az aks show --name $(AZURE_CLUSTER_NAME) --resource-group $(AZURE_RESOURCE_GROUP) | jq  -r ".identityProfile.kubeletidentity.objectId"))
-	az role assignment create --assignee $(AZURE_OBJECT_ID) --scope /subscriptions/$(AZURE_SUBSCRIPTION_ID)/resourceGroups/$(AZURE_RESOURCE_GROUP_MC) --role "Virtual Machine Contributor"
-	az role assignment create --assignee $(AZURE_OBJECT_ID) --scope /subscriptions/$(AZURE_SUBSCRIPTION_ID)/resourceGroups/$(AZURE_RESOURCE_GROUP_MC) --role "Network Contributor"
-	az role assignment create --assignee $(AZURE_OBJECT_ID) --scope /subscriptions/$(AZURE_SUBSCRIPTION_ID)/resourceGroups/$(AZURE_RESOURCE_GROUP_MC) --role "Managed Identity Operator"
-	az role assignment create --assignee $(AZURE_OBJECT_ID) --scope /subscriptions/$(AZURE_SUBSCRIPTION_ID)/resourceGroups/$(AZURE_RESOURCE_GROUP)    --role "Network Contributor" # in some case we create vnet here
+	# Note (charliedmcb): need to be principalId for E2E workflow as the pipeline identity doesn't have permissions to "query Graph API"
```
Bryce-Soghigian, line 120: Don't see a reason for your alias to be here unless is there some followup work on the pipeline here? Generally like the pattern of naming a todo to indicate ownership. I don't love it for notes

pkg/auth/autorest_auth.go
```diff
@@ -23,10 +23,23 @@ import (
...
+	"k8s.io/klog/v2"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
+	"github.com/jongio/azidext/go/azidext"
 )
 
 func NewAuthorizer(config *Config, env *azure.Environment) (autorest.Authorizer, error) {
+	// TODO (charliedmcb): need to get track 2 support for the skewer API, and align all auth under workload identity in the same way within cred.go
+	if config.UseNewCredWorkflow {
+		klog.V(2).Infoln("auth: using workload identity for new authorizer")
```
Bryce-Soghigian, line 35: since we are touching the config code do you think we could remove instances of klog all together? Is there a reason we use klog rather than standard knativing contxt logging?

pkg/auth/cred.go
```diff
@@ -29,7 +30,13 @@ func NewCredential(cfg *Config) (azcore.TokenCredential, error) {
 		return nil, fmt.Errorf("failed to create credential, nil config provided")
 	}
 
+	if cfg.UseNewCredWorkflow {
+		klog.V(2).Infoln("cred: using workload identity for new credential")
+		return azidentity.NewDefaultAzureCredential(nil)
+	}
+
 	if cfg.UseManagedIdentityExtension || cfg.AADClientID == "msi" {
+		klog.V(2).Infoln("cred: using msi for new credential")
```
Bryce-Soghigian, line 39: Again don't like the klog usage

pkg/auth/cred.go
```diff
@@ -29,7 +30,13 @@ func NewCredential(cfg *Config) (azcore.TokenCredential, error) {
 		return nil, fmt.Errorf("failed to create credential, nil config provided")
 	}
 
+	if cfg.UseNewCredWorkflow {
```
Bryce-Soghigian, line 33: Can we name this something more fitting than NewCredWorkflow? Most auth code is ripped out by others and its probably the most likely candidate to get copy pasted from karpenter into another project, so naming moreso around workloadIdentity than UseNewCredWorkFlow makes sense to me

pkg/auth/autorest_auth.go
```diff
@@ -23,10 +23,23 @@ import (
 	"github.com/Azure/go-autorest/autorest"
 	"github.com/Azure/go-autorest/autorest/adal"
 	"github.com/Azure/go-autorest/autorest/azure"
-	klog "k8s.io/klog/v2"
+	"k8s.io/klog/v2"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
+	"github.com/jongio/azidext/go/azidext"
```
Bryce-Soghigian, line 29: Seems this whole library https://github.com/jongio/azidext/tree/master/go/azidext should be merged with the azure-sdk-for-go-extensions since they effectively serve the same purpose.

Bryce-Soghigian (changes requested):

charliedmcb (commented):
/test

Makefile-az.mk
charliedmcb: I actually quite like the idea of this practice for notes as it allows for a known reference point for more info if more info is needed on the note. However, I get how that could rot over time, so you could argue against it.

pkg/auth/autorest_auth.go
charliedmcb: I was just modeling off the surrounding code here. I'm not sure why klog was used around here before. I do think I'd save any change to the logging for a separate PR though for keeping focus/PRs small.

I did actually find that these log outputs were not being shown by default in standalone, which I'd personally consider a bug. I was going to surface fixing that after this PR.

pkg/auth/cred.go
charliedmcb: See other comment.

pkg/auth/cred.go
charliedmcb: Yea, this was just my general temp name, and happy to change, but didn't have a better one in mind.

I was thinking something like `UseWorkloadIdentity`. However, this path doesn't have to just be for workload identity is the issue, as the backing function I'm pretty sure could handle other cases too.

pkg/auth/cred.go
charliedmcb: Unless you have strong objections I plan on leaving this as is within this PR.

Makefile-az.mk
Bryce-Soghigian: Yeah I think more often then not it rots imo for any large codebase over a 4 year time scale

pkg/auth/autorest_auth.go
Bryce-Soghigian: Ok we can solve in a different PR

pkg/auth/cred.go
Bryce-Soghigian: UseWorkloadEntity makes a lot more sense to me, its not apparent how the backing function is being used for other cases.

pkg/auth/autorest_auth.go
Bryce-Soghigian: If its not a microsoft repo, and we aren't confident we can move the code to sdk for go extensions then we shouldn't be relying on it in the first place? 

We should be conservative when adding new dependencies to our codebase.

charliedmcb: Its the recommended approach from the `azure-sdk-for-go` repo:
> https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azidentity/MIGRATION.md#use-azidentity-credentials-with-older-packages

charliedmcb: It does have a Microsoft Corporation Copyright so looks like it is Microsoft:
> // Copyright (c) Microsoft Corporation. All rights reserved.
// Licensed under the MIT License.

charliedmcb: Not being under the `Azure` path was confusing me.

charliedmcb: Although, wait the top level License is not Microsft:
image

Makefile-az.mk
charliedmcb: I like having them there, but sure I'm fine with removing it.

pkg/auth/cred.go
matthchr: Since it's just calling the sdk default method, and AFAIK that default method is primarily based on checking the env for things such as:
```
	azureClientID                   = "AZURE_CLIENT_ID"
	azureClientSecret               = "AZURE_CLIENT_SECRET"
	azureFederatedTokenFile         = "AZURE_FEDERATED_TOKEN_FILE"
```

Maybe something around `UseCredentialFromEnvironment`? This path ought to work for more than just the workload identity flow, it works for SP too if you put the clientid/clientsecret into those env variables and pass them to this pod.

charliedmcb: Yea, I feel from env could be a good naming. Feels better than the others.

pkg/auth/autorest_auth.go
matthchr: > DISCLAIMER: The code in this repo is not officially supported or intended for production use. The intention of this repo it to unblock customers who would like to use the Azure.Identity capabilities in the Fluent, Resource Management, and Service Bus SDKs before they have been migrated to the new SDK Azure.Core and officially support TokenCredential. We have included minimal tests in this repo, so please take it upon yourself to fully test this code to ensure it works in your environment.

They do say that, which does make me hesitate too... but I am not sure you're worse off doing it this way vs what you had

Bryce-Soghigian (approved):
Approved given we properly followup with the items that were put aside in this pr

charliedmcb (commented):
/test

charliedmcb (commented):
/test

charliedmcb (commented):
/test

charliedmcb (commented):
/test

charliedmcb (commented):
/test

charliedmcb (commented):
/test

charliedmcb (commented):
/test

pkg/auth/cred.go
charliedmcb: Updated naming.

charliedmcb (commented):
/test

tallaxes (approved):

comtalyst (approved):

charliedmcb (commented):
/test

charliedmcb (commented):
/test

charliedmcb (commented):
/test

charliedmcb (commented):
/test

PR #86, opened 2024-01-08
chore: Small style + usability refactors for new contributors

rakechill:
**Description**
*These changes are based on feedback from @matthchr*

- [x] Update (and as a follow-up git ignore) `skaffold.yaml` so that `make az-all` automated local changes during development are less confusing.
- [x] Style refactors
  - [x] `NewAZClientFromAPI` vs. `NewAZClient`. Maybe one of these should call the other then? `NewAZClient` constructs all the clients + calls `NewAZClientFromAPI`
  - [x] Inconsistent with abbreviation casing (`CreateAzClient` vs. `NewAZClient`)
  - [x] Fakes: `var _ instance.NetworkInterfacesAPI = (*NetworkInterfacesAPI)(nil)` -> 
  `var _ instance.NetworkInterfacesAPI = &NetworkInterfacesAPI{}` 

* `make presubmit`
* `make az-all`

- [ ] Yes, issue opened: # 
- [ ] No

### DIFF
--- a/pkg/fake/azureresourcegraphapi.go
+++ b/pkg/fake/azureresourcegraphapi.go
@@ -39,7 +39,7 @@ type AzureResourceGraphBehavior struct {
 }
 
 // assert that the fake implements the interface
-var _ instance.AzureResourceGraphAPI = (*AzureResourceGraphAPI)(nil)
+var _ instance.AzureResourceGraphAPI = &AzureResourceGraphAPI{}
 
 type AzureResourceGraphAPI struct {
 	AzureResourceGraphBehavior
--- a/pkg/fake/communityimageversionsapi.go
+++ b/pkg/fake/communityimageversionsapi.go
@@ -30,7 +30,7 @@ type CommunityGalleryImageVersionsAPI struct {
 }
 
 // assert that the fake implements the interface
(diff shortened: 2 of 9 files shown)


Bryce-Soghigian:
@rakechill  there is more than a single test using normal go tests.

Bryce-Soghigian:
> networkinterfaceapi.go fake BeginDelete incorrectly returns a nil poller
this is addressed in my pr for networkinterface delete

rakechill:
> @rakechill there is more than a single test using normal go tests.
> 
> ./pkg/auth/cred_test.go ./pkg/auth/config_test.go ./pkg/providers/imagefamily/bootstrap/aksbootstrap_test.go ./pkg/fake/skus_test.go ./pkg/fake/azureresourcegraphapi_test.go ./pkg/utils/gpu_test.go
> 
> Are all the go tests. IMO sometimes its nice to write a very simple go test rather than setting up all the ginkgo shenanigans and they are better for fine grained unit tests vs the coarse unit tests we write through the fakes framework.

Ack. Maybe a better solution is to call out the different sets of testing we do instead. That could go somewhere on the contributing.md/website

Bryce-Soghigian (approved):

PR #93, opened 2024-01-17
fix: mark spot as unavailable on LowPriorityQuota

### LINKED ISSUE #91: LowPriorityCoresQuota error should mark spot as unavailable

Any roadmap to support scale spot node fail, and fallback to on-demand node? 

I saw Karpenter is AWS already easy to support.
https://github.com/aws/karpenter-provider-aws/blob/e00c97c3c834a1349a4126f69500fdcc4caf1344/website/content/en/v0.32/faq.md?plain=1#L82

but in Azure, if the spot instance cannot created succuess, that will crash, cannot failback to on-demand.

to save the cost, balance the spot and on-demand resources.

no

### Additional Context

image
Tested on Azure, I create a deployment with 4 gpu pod. And I only have 3 gpu spot quota.
As design, it will create the fourth node with on-demand, but now always crash in the fourth node in spot.

### Attachments

_No response_

### Community Note

Bryce-Soghigian:
Fixes #91 

**Description**
If spot capacity hits the spot regional cores count, karpenter should fall back onto on-demand. This PR adds that functionality

- make presubmt 
- Manual repro where i first reproed the issue presented in #91 , then proceeded to build a new image with this code and validate it will successfully choose on demand instances if spot capacity is fully used up. 

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
feat: fail-over to on-demand if spot capacity is unavailable
```

### DIFF
--- a/pkg/cache/unavailableofferings.go
+++ b/pkg/cache/unavailableofferings.go
@@ -69,15 +69,6 @@ func (u *UnavailableOfferings) MarkUnavailable(ctx context.Context, unavailableR
 	u.MarkUnavailableWithTTL(ctx, unavailableReason, instanceType, zone, capacityType, UnavailableOfferingsTTL)
 }
 
-func (u *UnavailableOfferings) Put(key string, value struct{}, ttl time.Duration) {
-	u.cache.Set(key, value, ttl)
-}
-
-func (u *UnavailableOfferings) Has(key string) bool {
-	_, found := u.cache.Get(key)
-	return found
-}
-
 func (u *UnavailableOfferings) Flush() {
 	u.cache.Flush()
 }
--- a/pkg/cache/unavailableofferings.go
+++ b/pkg/cache/unavailableofferings.go
@@ -57,7 +57,7 @@ func (u *UnavailableOfferings) IsUnavailable(instanceType, zone, capacityType st
 }
 
 // MarkSpotUnavailable communicates recently observed temporary capacity shortages for spot
-func (u *UnavailableOfferings) MarkSpotUnavailable(ctx context.Context) {
+func (u *UnavailableOfferings) MarkSpotUnavailableWithTTL(ctx context.Context, ttl time.Duration) {
 	u.MarkUnavailableWithTTL(ctx, "SpotUnavailable", "", "", v1beta1.CapacityTypeSpot, UnavailableOfferingsTTL)
 }
 
--- a/pkg/cache/unavailableofferings.go
+++ b/pkg/cache/unavailableofferings.go
@@ -32,25 +32,33 @@ import (
 type UnavailableOfferings struct {
 	// key: <capacityType>:<instanceType>:<zone>, value: struct{}{}
 	cache *cache.Cache
+	// pre-computed key for spot since we will be computing this key many times and it
+	// will always be the same.
+	spotKey string
 }
 
 func NewUnavailableOfferingsWithCache(c *cache.Cache) *UnavailableOfferings {
-	return &UnavailableOfferings{
-		cache: c,
+	u := &UnavailableOfferings{
+		cache: cache.New(UnavailableOfferingsTTL, DefaultCleanupInterval),
 	}
+	u.spotKey = u.key("", "", v1beta1.CapacityTypeSpot)
+	return u
 }
 
 func NewUnavailableOfferings() *UnavailableOfferings {
-	c := cache.New(UnavailableOfferingsTTL, DefaultCleanupInterval)
-	return &UnavailableOfferings{
-		cache: c,
+	u := &UnavailableOfferings{ 
+		cache: cache.New(UnavailableOfferingsTTL, DefaultCleanupInterval), 
 	}
+	u.spotKey = u.key("", "", v1beta1.CapacityTypeSpot)
+	return u
 }
 
 // IsUnavailable returns true if the offering appears in the cache
 func (u *UnavailableOfferings) IsUnavailable(instanceType, zone, capacityType string) bool {
-	if _, ok := u.cache.Get(u.key("", "", v1beta1.CapacityTypeSpot)); ok && capacityType == v1beta1.CapacityTypeSpot {
-		return true
+	if capacityType == v1beta1.CapacityTypeSpot {
+		if _, ok := u.cache.Get(u.spotKey); ok {
+			return true
+		}
 	}
 	_, found := u.cache.Get(u.key(instanceType, zone, capacityType))
 	return found
--- a/pkg/cache/unavailableofferings.go
+++ b/pkg/cache/unavailableofferings.go
@@ -69,6 +69,15 @@ func (u *UnavailableOfferings) MarkUnavailable(ctx context.Context, unavailableR
 	u.MarkUnavailableWithTTL(ctx, unavailableReason, instanceType, zone, capacityType, UnavailableOfferingsTTL)
 }
 
+func (u *UnavailableOfferings) Put(key string, value struct{}, ttl time.Duration) {
+	u.cache.Set(key, value, ttl)
+}
+
+func (u *UnavailableOfferings) Has(key string) bool {
+	_, found := u.cache.Get(key)
+	return found
+}
+
 func (u *UnavailableOfferings) Flush() {
 	u.cache.Flush()
 }
--- a/pkg/cache/unavailableofferings.go
+++ b/pkg/cache/unavailableofferings.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/aws/karpenter-core/pkg/apis/v1beta1"
 	"github.com/patrickmn/go-cache"
 	"knative.dev/pkg/logging"
 )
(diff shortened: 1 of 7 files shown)


Bryce-Soghigian (commented):
/test

pkg/cache/unavailableofferings.go
```diff
@@ -69,6 +69,15 @@ func (u *UnavailableOfferings) MarkUnavailable(ctx context.Context, unavailableR
...
 }
 
+func (u *UnavailableOfferings) Put(key string, value struct{}, ttl time.Duration) {
+	u.cache.Set(key, value, ttl)
+}
+
+func (u *UnavailableOfferings) Has(key string) bool {
+	_, found := u.cache.Get(key)
+	return found
+}
```
tallaxes, line 79: This seems to expose too much of the implementaional detail of UnavailableOfferings - and no reason to pass value? How about adding a method like MarkSpotUnavailableWithTTL, instead of opening the cache for direct manipulation? It could also  maybe use synthetic key("","",spot), instead of the specific error string?

pkg/providers/instance/instance.go
```diff
@@ -415,7 +415,14 @@ func (p *Provider) launchInstance(
 	return resp, instanceType, nil
 }
 
+// nolint:gocyclo
 func (p *Provider) handleResponseErrors(ctx context.Context, instanceType *corecloudprovider.InstanceType, zone, capacityType string, err error) error {
+	if sdkerrors.LowPriorityQuotaHasBeenReached(err) {
+		// Mark in cache that spot quota has been reached for this subscription
+		p.unavailableOfferings.Put(sdkerrors.LowPriorityQuotaExceededTerm, struct{}{}, SubscriptionQuotaReachedTTL)
```
tallaxes, line 422: See elsewhere on Put. I also slightly dislike the use of the error term as key, but this is less important (and best decided within unavailableOffereings implementation)

pkg/providers/instance/instance.go
```diff
@@ -415,7 +415,14 @@ func (p *Provider) launchInstance(
...
+		logging.FromContext(ctx).Error(err)
+		return fmt.Errorf("LowPriorityQuota Has been reached and this subscription has hit the regional vCPU quota for spot. To scale beyond this limit, please review the quota increase process here: https://docs.microsoft.com/en-us/azure/azure-portal/supportability/low-priority-quota")
```
tallaxes, line 424:
```suggestion
		return fmt.Errorf("this subscription has reached the regional vCPU quota for spot (LowPriorityQuota). To scale beyond this limit, please review the quota increase process here: https://docs.microsoft.com/en-us/azure/azure-portal/supportability/low-priority-quota")
```

pkg/providers/instancetype/instancetypes.go
```diff
@@ -136,7 +136,7 @@ func (p *Provider) createOfferings(sku *skewer.SKU, zones sets.Set[string]) []cl
 		onDemandPrice, onDemandOk := p.pricingProvider.OnDemandPrice(*sku.Name)
 		spotPrice, spotOk := p.pricingProvider.SpotPrice(*sku.Name)
 		availableOnDemand := onDemandOk && !p.unavailableOfferings.IsUnavailable(*sku.Name, zone, corev1beta1.CapacityTypeOnDemand)
-		availableSpot := spotOk && !p.unavailableOfferings.IsUnavailable(*sku.Name, zone, corev1beta1.CapacityTypeSpot)
+		availableSpot := spotOk && !p.unavailableOfferings.IsUnavailable(*sku.Name, zone, corev1beta1.CapacityTypeSpot) && !p.unavailableOfferings.Has(sdkerrors.LowPriorityQuotaExceededTerm)
```
tallaxes, line 139: Can the new check be added within IsUnavailable? Simpler and easier to use, better encapsulation / don't need Has ...

pkg/providers/instancetype/suite_test.go
```diff
@@ -172,6 +172,42 @@ var _ = Describe("InstanceType Provider", func() {
...
+						Body: createSDKErrorBody(sdkerrors.OperationNotAllowed, LowPriorityCoresQuotaErrorMessage),
+					},
+				},
+			)
+			// Create a pod that should fail to schedule
+			pod := coretest.UnschedulablePod()
+			ExpectProvisioned(ctx, env.Client, cluster, cloudProvider, coreProvisioner, pod)
+			ExpectNotScheduled(ctx, env.Client, pod)
+			azureEnv.VirtualMachinesAPI.VirtualMachineCreateOrUpdateBehavior.BeginError.Set(nil)
+			pod = coretest.UnschedulablePod()
```
tallaxes, line 198: Yikes. This clearly shows the limitations of current support for faking errors. We need something along the lines of [EC2Behavior.InsufficientCapacityPools](https://github.com/search?q=repo%3Aaws%2Fkarpenter-provider-aws%20InsufficientCapacityPools&type=code) ...

In any case, likely don't need a new unschedulable pod?

tallaxes (changes requested):
Like the direction, left some comments. 

May also want to rename this to something like "fix: mark spot as unavailable on LowPriorityQuota" (what actually happens after depends on other configuration)

pkg/cache/unavailableofferings.go
Bryce-Soghigian: synthetic key makes a lot of sense, will switch to that

pkg/providers/instance/instance.go
Bryce-Soghigian: Agree removed

pkg/providers/instancetype/instancetypes.go
Bryce-Soghigian: yep

pkg/providers/instancetype/suite_test.go
Bryce-Soghigian: removed the second pod

pkg/cache/unavailableofferings.go
tallaxes: The new changes go in the right direction I think, but I would suggest taking the next step and 

* Implement MarkSpotUnavailableWithTTL method on UnavailableOfferings (that would use the synthetic key internally) - this would remove from the consume the need to know anything about implementation; and 
* Add the check for this key into the already existing IsUnavailable (would check for this first, if asked about spot), which the consumer would call as before, without special casing / knowing anything about extra check for spot behind the scenes.

pkg/providers/instancetype/instancetypes.go
tallaxes: Sorry I did not make myself clear; I meant adding th check within IsUnavailable _implementation_ - so this call would not even need to change.

pkg/cache/unavailableofferings.go
Bryce-Soghigian: gotcha that makes sense

Bryce-Soghigian:
implemented all that was mentioned

pkg/cache/unavailableofferings.go
```diff
@@ -48,10 +49,18 @@ func NewUnavailableOfferings() *UnavailableOfferings {
 
 // IsUnavailable returns true if the offering appears in the cache
 func (u *UnavailableOfferings) IsUnavailable(instanceType, zone, capacityType string) bool {
+	if _, ok := u.cache.Get(u.key("", "", v1beta1.CapacityTypeSpot)); ok && capacityType == v1beta1.CapacityTypeSpot {
```
tallaxes, line 52: Checking capacityType first would be faster. Can also pre-compute the key, once. (To minimize performance impact we introduce here. Yes, slight, but called frequently ..) Otherwise 👍

Bryce-Soghigian: Is there a way to write 

``` 
if capacityType == v1beta1.CapacityTypeSpot {
		if _, ok := u.cache.Get(SpotUnavailableKey); ok {
			return true
		}
	}
``` 
As 
```
if capacityType == v1beta1.CapacityTypeSpot && _, ok := u.cache.Get(SpotUnavailableKey); ok {
```


```
The Go programming language does not allow declarations
 after &&. This is because && is a logical operator
 that returns true if both of its operands are true, and false otherwise. If a declaration were allowed after &&, it would be possible for the declaration to be executed before the && operator is evaluated, which could lead to sadness
For example, consider the following code
if true && var := 1; var > 0 {
  // This code will be executed, even if var is not initialized.
}

In this example, the variable var is declared after the && operator. This means that the declaration will be executed before the && operator is evaluated. If var is not initialized before the && operator is evaluated, it will be equal to the zero value of its type, which is 0. This means that the && operator will evaluate to false, and the code inside the if statement will not be executed.
To avoid this problem, declarations must be made before the && operator. This ensures that the declaration is executed before the && operator is evaluated, and that the variable is initialized before it is used.
```


I wanted to write the cpaaciyy type check before, but encountered this problem. I can use the first syntax to accomplish this just wondering if there is a better way i am missing.

tallaxes: Not that I know of; I would not worry about it

pkg/cache/unavailableofferings.go
```diff
@@ -31,27 +32,43 @@ import (
...
+	spotKey string
 }
 
 func NewUnavailableOfferingsWithCache(c *cache.Cache) *UnavailableOfferings {
-	return &UnavailableOfferings{
-		cache: c,
+	u := &UnavailableOfferings{
+		cache: cache.New(UnavailableOfferingsTTL, DefaultCleanupInterval),
 	}
+	u.spotKey = u.key("", "", v1beta1.CapacityTypeSpot)
```
tallaxes, line 44: If you change `key()` to non-member (which should not be a problem), all this unpleasantness would go away

pkg/cache/unavailableofferings.go
Bryce-Soghigian: sg

tallaxes (approved):
LGTM👍

tallaxes (approved):

PR #95, opened 2024-01-19
chore: karpenter-global-settings migration

comtalyst:
Fixes #  N/A

**Description**

**Context:**
- Following karpenter-core's effort, we are migrating away from `Settings` to `Options`
  - `Settings`, from pkg/api/settings = karpenter-global-settings ConfigMap
  - `Options`, from pkg/operator/options = flags OR environment variables
- At the time between "introducing `Options`", which is this PR, and the breaking API change in v0.33(?), which removed ConfigMap support entirely, we will support either options, with `Options` taking precedence

**Expected impact from this change:**
- `Options`: flags and environment variables is now supported as an alternative for all values currently utilizing the ConfigMap
  - ConfigMap can be safely removed once we create the container using flags or environment variables

**Non-impact:**
- ConfigMap is still supported, but `Options` taking precedence if both exists
- E2Es and charts are still using ConfigMap until the API update that removes `Settings` entirely

**Additional changes**
- A `networkPlugin` constant in `imagefamily` package was incorrectly set to "overlay" rather than "azure"
  - Its primary use is to determine the maximum number of pods, since Azure CNI allow more number of pods than `kubenet`
  - The change to use `Options` also reworked the default values in unit test, thus this gap/bug was exposed
  - This PR also include the fix for that constant, as well as additional cases in that unit test

* Manually test basic scale up-down for Karpenter deployed using environment variables
* E2E for both Karpenter deployed using ConfigMap and environment variables

- [ ] Yes, issue opened: # 
- [X] No

**Release Note**

```release-note
- `Options`: flags and environment variables is now supported as an alternative for all values currently utilizing the ConfigMap
  - ConfigMap can be safely removed once we create the container using flags or environment variables
```

### DIFF
--- a/pkg/apis/settings/settings.go
+++ b/pkg/apis/settings/settings.go
@@ -28,6 +28,8 @@ import (
 	"go.uber.org/multierr"
 	v1 "k8s.io/api/core/v1"
 	"knative.dev/pkg/configmap"
+
+	coresettings "github.com/aws/karpenter-core/pkg/apis/settings"
 )
 
 type settingsKeyType struct{}
@@ -73,6 +75,9 @@ func (*Settings) ConfigMap() string {
 // Inject creates a Settings from the supplied ConfigMap
 func (*Settings) Inject(ctx context.Context, cm *v1.ConfigMap) (context.Context, error) {
 	s := defaultSettings.DeepCopy()
+	if cm == nil {
+		return ToContext(ctx, s), nil
+	}
 
 	if err := configmap.Parse(cm.Data,
 		configmap.AsString("azure.clusterName", &s.ClusterName),
@@ -140,6 +145,10 @@ func (s Settings) validateEndpoint() error {
 func (s Settings) GetAPIServerName() string {
 	endpoint, _ := url.Parse(s.ClusterEndpoint) // already validated
 	return endpoint.Hostname()
+}
+
+func (*Settings) FromContext(ctx context.Context) coresettings.Injectable {
+	return FromContext(ctx)
 }
 
 func ToContext(ctx context.Context, s *Settings) context.Context {
--- a/pkg/apis/settings/settings.go
+++ b/pkg/apis/settings/settings.go
@@ -28,8 +28,6 @@ import (
 	"go.uber.org/multierr"
 	v1 "k8s.io/api/core/v1"
 	"knative.dev/pkg/configmap"
-
-	coresettings "github.com/aws/karpenter-core/pkg/apis/settings"
 )
 
 type settingsKeyType struct{}
@@ -145,10 +143,6 @@ func (s Settings) validateEndpoint() error {
 func (s Settings) GetAPIServerName() string {
 	endpoint, _ := url.Parse(s.ClusterEndpoint) // already validated
 	return endpoint.Hostname()
-}
-
-func (*Settings) FromContext(ctx context.Context) coresettings.Injectable {
-	return FromContext(ctx)
 }
 
 func ToContext(ctx context.Context, s *Settings) context.Context {
--- a/pkg/cloudprovider/suite_test.go
+++ b/pkg/cloudprovider/suite_test.go
@@ -30,7 +30,6 @@ import (
 	clock "k8s.io/utils/clock/testing"
 
 	"github.com/Azure/karpenter-provider-azure/pkg/utils"
-	coresettings "github.com/aws/karpenter-core/pkg/apis/settings"
 	corev1beta1 "github.com/aws/karpenter-core/pkg/apis/v1beta1"
 	corecloudprovider "github.com/aws/karpenter-core/pkg/cloudprovider"
 
@@ -44,8 +43,8 @@ import (
 	. "knative.dev/pkg/logging/testing"
 
 	"github.com/Azure/karpenter-provider-azure/pkg/apis"
-	"github.com/Azure/karpenter-provider-azure/pkg/apis/settings"
 	"github.com/Azure/karpenter-provider-azure/pkg/apis/v1alpha2"
+	"github.com/Azure/karpenter-provider-azure/pkg/operator/options"
 	"github.com/Azure/karpenter-provider-azure/pkg/providers/instance"
 	"github.com/Azure/karpenter-provider-azure/pkg/test"
 	. "github.com/aws/karpenter-core/pkg/test/expectations"
@@ -70,10 +69,14 @@ func TestCloudProvider(t *testing.T) {
 	RunSpecs(t, "CloudProvider")
 }
 
+func toBoolPtr(b bool) *bool {
+	return &b
+}
+
 var _ = BeforeSuite(func() {
 	env = coretest.NewEnvironment(scheme.Scheme, coretest.WithCRDs(apis.CRDs...))
-	ctx = coresettings.ToContext(ctx, coretest.Settings(coresettings.Settings{DriftEnabled: true}))
-	ctx = settings.ToContext(ctx, test.Settings())
+	ctx = coreoptions.ToContext(ctx, coretest.Options(coretest.OptionsFields{FeatureGates: coretest.FeatureGates{Drift: toBoolPtr(true)}}))
+	ctx = options.ToContext(ctx, test.Options())
 	ctx, stop = context.WithCancel(ctx)
 	azureEnv = test.NewEnvironment(ctx, env)
 
@@ -92,7 +95,7 @@ var _ = BeforeEach(func() {
 	ctx = coreoptions.ToContext(ctx, coretest.Options())
 	// TODO v1beta1 options
 	// ctx = options.ToContext(ctx, test.Options())
-	ctx = settings.ToContext(ctx, test.Settings())
+	ctx = options.ToContext(ctx, test.Options())
 	nodeClass = test.AKSNodeClass()
 	nodePool = coretest.NodePool(corev1beta1.NodePool{
 		Spec: corev1beta1.NodePoolSpec{
--- a/pkg/cloudprovider/suite_test.go
+++ b/pkg/cloudprovider/suite_test.go
@@ -23,6 +23,7 @@ import (
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	"github.com/samber/lo"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
@@ -69,13 +70,9 @@ func TestCloudProvider(t *testing.T) {
 	RunSpecs(t, "CloudProvider")
 }
 
-func toBoolPtr(b bool) *bool {
-	return &b
-}
-
 var _ = BeforeSuite(func() {
 	env = coretest.NewEnvironment(scheme.Scheme, coretest.WithCRDs(apis.CRDs...))
-	ctx = coreoptions.ToContext(ctx, coretest.Options(coretest.OptionsFields{FeatureGates: coretest.FeatureGates{Drift: toBoolPtr(true)}}))
+	ctx = coreoptions.ToContext(ctx, coretest.Options(coretest.OptionsFields{FeatureGates: coretest.FeatureGates{Drift: lo.ToPtr(true)}}))
 	ctx = options.ToContext(ctx, test.Options())
 	ctx, stop = context.WithCancel(ctx)
 	azureEnv = test.NewEnvironment(ctx, env)
--- a/pkg/operator/options/options.go
+++ b/pkg/operator/options/options.go
@@ -1,0 +1,177 @@
+/*
+Portions Copyright (c) Microsoft Corporation.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package options
+
+import (
+	"context"
+	"errors"
+	"flag"
+	"fmt"
+	"hash/fnv"
+	"math/rand"
+	"net/url"
+	"os"
+	"strings"
+
+	"github.com/Azure/karpenter-provider-azure/pkg/apis/settings"
+	coreoptions "github.com/aws/karpenter-core/pkg/operator/options"
+	"github.com/aws/karpenter-core/pkg/utils/env"
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+func init() {
+	coreoptions.Injectables = append(coreoptions.Injectables, &Options{})
+}
+
+type nodeIdentitiesValue []string
+
+func newNodeIdentitiesValue(val string, p *[]string) *nodeIdentitiesValue {
+	*p = strings.Split(val, ",")
+	return (*nodeIdentitiesValue)(p)
+}
+
+func (s *nodeIdentitiesValue) Set(val string) error {
+	*s = nodeIdentitiesValue(strings.Split(val, ","))
+	return nil
+}
+
+func (s *nodeIdentitiesValue) Get() any { return []string(*s) }
+
+func (s *nodeIdentitiesValue) String() string { return strings.Join(*s, ",") }
+
+type optionsKey struct{}
+
+type Options struct {
+	ClusterName                    string
+	ClusterEndpoint                string // => APIServerName in bootstrap, except needs to be w/o https/port
+	VMMemoryOverheadPercent        float64
+	ClusterID                      string
+	KubeletClientTLSBootstrapToken string   // => TLSBootstrapToken in bootstrap (may need to be per node/nodepool)
+	SSHPublicKey                   string   // ssh.publicKeys.keyData => VM SSH public key // TODO: move to node template?
+	NetworkPlugin                  string   // => NetworkPlugin in bootstrap
+	NetworkPolicy                  string   // => NetworkPolicy in bootstrap
+	NodeIdentities                 []string // => Applied onto each VM
+
+	setFlags map[string]bool
+}
+
+func (o *Options) AddFlags(fs *coreoptions.FlagSet) {
+	fs.StringVar(&o.ClusterName, "cluster-name", env.WithDefaultString("CLUSTER_NAME", ""), "[REQUIRED] The kubernetes cluster name for resource discovery.")
+	fs.StringVar(&o.ClusterEndpoint, "cluster-endpoint", env.WithDefaultString("CLUSTER_ENDPOINT", ""), "[REQUIRED] The external kubernetes cluster endpoint for new nodes to connect with. If not specified, will discover the cluster endpoint using DescribeCluster API.")
+	fs.Float64Var(&o.VMMemoryOverheadPercent, "vm-memory-overhead-percent", env.WithDefaultFloat64("VM_MEMORY_OVERHEAD_PERCENT", 0.075), "The VM memory overhead as a percent that will be subtracted from the total memory for all instance types.")
+	fs.StringVar(&o.ClusterID, "cluster-id", env.WithDefaultString("CLUSTER_ID", ""), "The kubernetes cluster ID. If not specified, will generated based on cluster endpoint.")
+	fs.StringVar(&o.KubeletClientTLSBootstrapToken, "kubelet-client-tls-bootstrap-token", env.WithDefaultString("KUBELET_CLIENT_TLS_BOOTSTRAP_TOKEN", ""), "[REQUIRED] The bootstrap token for new nodes to join the cluster.")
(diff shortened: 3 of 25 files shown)


pkg/test/settings.go
```diff
@@ -25,7 +25,7 @@ import (
 	azsettings "github.com/Azure/karpenter/pkg/apis/settings"
 )
 
-type SettingOptions struct {
+type SettingsFields struct {
```
Bryce-Soghigian, line 28: why change to Fields and why change Setting to Settings?

pkg/test/options.go
```diff
@@ -0,0 +1,58 @@
...
+			panic(fmt.Sprintf("Failed to merge settings: %s", err))
+		}
+	}
+	return &azoptions.Options{
+		ClusterName:                    lo.FromPtrOr(options.ClusterName, "test-cluster"),
+		ClusterEndpoint:                lo.FromPtrOr(options.ClusterEndpoint, "https://test-cluster"),
+		ClusterID:                      lo.FromPtrOr(options.ClusterID, "00000000"),
+		KubeletClientTLSBootstrapToken: lo.FromPtrOr(options.KubeletClientTLSBootstrapToken, "test-token"),
+		SSHPublicKey:                   lo.FromPtrOr(options.SSHPublicKey, "test-ssh-public-key"),
+		NetworkPlugin:                  lo.FromPtrOr(options.NetworkPlugin, "kubenet"),
```
Bryce-Soghigian, line 53: Do we want to default to kubenet? I think azure would be more realistic at least to start in a test env.

pkg/operator/options/suite_test.go
```diff
@@ -0,0 +1,274 @@
...
+
+func TestAPIs(t *testing.T) {
+	ctx = TestContextWithLogger(t)
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Options")
+}
+
+var _ = Describe("Options", func() {
+	var envState map[string]string
+	var environmentVariables = []string{
```
Bryce-Soghigian, line 46: this seems useful for other tests and other values that are set via env variables like around cilium, wonder if we should test these envs are set. 

Also would be good to validate that we do properly prepare the environment in this test: https://github.com/Azure/karpenter/blob/main/pkg/providers/instancetype/suite_test.go#L117

pkg/operator/options/options.go
```diff
@@ -0,0 +1,177 @@
...
+	mergeField(&o.SSHPublicKey, s.SSHPublicKey, o.setFlags["ssh-public-key"])
+	mergeField(&o.NetworkPlugin, s.NetworkPlugin, o.setFlags["network-plugin"])
+	mergeField(&o.NetworkPolicy, s.NetworkPolicy, o.setFlags["network-policy"])
+	mergeField(&o.NodeIdentities, s.NodeIdentities, o.setFlags["node-identities"])
+
+	if err := o.validateRequiredFields(); err != nil {
+		panic(fmt.Errorf("checking required fields, %w", err))
+	}
+
+	// if clusterID is not set, generate it from cluster endpoint
```
Bryce-Soghigian, line 144: nit ClusterID

Bryce-Soghigian (changes requested):

pkg/test/settings.go
comtalyst: Since our new module is called `Options` and will have a similar testing mechanism (look at `pkg/test/options.go`), this name is confusing. The options one used `OptionsFields`, so the parallel counterpart here would be `SettingsFields`.

pkg/operator/options/suite_test.go
comtalyst: Since those envs are from `pkg/auth/config`, maybe we could improve the test in that area.

comtalyst: Since those envs are from `pkg/auth/config`, maybe we could improve the test in that area later.

pkg/operator/options/options.go
```diff
@@ -0,0 +1,177 @@
...
+	KubeletClientTLSBootstrapToken string   // => TLSBootstrapToken in bootstrap (may need to be per node/nodepool)
+	SSHPublicKey                   string   // ssh.publicKeys.keyData => VM SSH public key // TODO: move to node template?
+	NetworkPlugin                  string   // => NetworkPlugin in bootstrap
+	NetworkPolicy                  string   // => NetworkPolicy in bootstrap
+	NodeIdentities                 []string // => Applied onto each VM
+
+	setFlags map[string]bool
+}
+
+func (o *Options) AddFlags(fs *coreoptions.FlagSet) {
```
comtalyst, line 72: This function is probably worth pay attention to:
- How's the description?
- Should we include `AZURE_` prefix to be equivalent to old `azure.` in the ConfigMap as well?

pkg/operator/options/options.go
```diff
@@ -0,0 +1,177 @@
...
+		o.setFlags[f.Name] = ok || cliFlags.Has(f.Name)
+	})
+
+	if err := o.Validate(); err != nil {
+		return fmt.Errorf("validating options, %w", err)
+	}
+
+	// if ClusterID is not set, generate it from cluster endpoint
+	// if clusterEndpoint is empty, it might awaiting merge from MergeSettings(), do not run GetAPIServerName() or it could panic
+	// TODO: chore: remove clusterEndpoint validation logic here when karpenter-global-settings (and merge logic) are completely removed
```
comtalyst, line 116: I will track this soon

pkg/test/settings.go
```diff
@@ -51,7 +51,7 @@ func Settings(overrides ...SettingOptions) *azsettings.Settings {
 		ClusterID:                      lo.FromPtrOr(options.ClusterID, "00000000"),
 		KubeletClientTLSBootstrapToken: lo.FromPtrOr(options.KubeletClientTLSBootstrapToken, "test-token"),
 		SSHPublicKey:                   lo.FromPtrOr(options.SSHPublicKey, "test-ssh-public-key"),
-		NetworkPlugin:                  lo.FromPtrOr(options.NetworkPlugin, "kubenet"),
+		NetworkPlugin:                  lo.FromPtrOr(options.NetworkPlugin, "azure"),
```
Bryce-Soghigian, line 54: this change means that you will need to change all the max-pods flags in your unit tests to "250". Currently there is a bug that will set them to 110(Default) https://github.com/Azure/karpenter-provider-azure/blob/main/pkg/providers/imagefamily/resolver.go#L37 change this value to "azure" and the max pods to 250 and you should be good

pkg/providers/imagefamily/bootstrap/aksbootstrap.go
```diff
@@ -143,7 +143,7 @@ type NodeBootstrapVariables struct {
 	ContainerRuntime                  string   // s   always containerd
 	CLITool                           string   // s   static/unnecessary
 	ContainerdDownloadURLBase         string   // -   unnecessary
-	NetworkMode                       string   // c   user input
+	NetworkMode                       string   // o   user input
```
Bryce-Soghigian, line 146: Why change these all to o? Perosonally i thought c for cluster level is was clear enough. 

I agree with changing comment to imply they come from options

pkg/operator/options/options.go
```diff
@@ -0,0 +1,177 @@
...
+	mergeField(&o.NodeIdentities, s.NodeIdentities, o.setFlags["node-identities"])
+
+	if err := o.validateRequiredFields(); err != nil {
+		panic(fmt.Errorf("checking required fields, %w", err))
+	}
+
+	// if ClusterID is not set, generate it from cluster endpoint
+	if o.ClusterID == "" {
+		o.ClusterID = getAKSClusterID(o.GetAPIServerName())
+	}
```
Bryce-Soghigian, line 147: would there ever by any difference between o.ClusterID set in env vs generating like this?

Bryce-Soghigian (changes requested):
Starting to shape up!!!!!!! 


Just FYI `make presubmit` will fix the ci issue in the pipeline.

pkg/providers/imagefamily/bootstrap/aksbootstrap.go
comtalyst: I wasn't sure if `c` stands for `Cluster` for `ConfigMap` either. I thought it was `ConfigMap` so I changed it.
But looks like it is more for `Cluster`, so let's change it back.

pkg/operator/options/options.go
comtalyst: In this way, a part of generated ClusterID is randomized. Users can choose to customize their own string if they don't want the randomization.

pkg/providers/instancetype/suite_test.go
```diff
@@ -514,7 +514,7 @@ var _ = Describe("InstanceType Provider", func() {
 			Expect(kubeletFlags).To(ContainSubstring("--eviction-hard=memory.available<750Mi"))
 			Expect(kubeletFlags).To(ContainSubstring("--eviction-soft=memory.available<1Gi"))
 			Expect(kubeletFlags).To(ContainSubstring("--eviction-soft-grace-period=memory.available=10s"))
-			Expect(kubeletFlags).To(ContainSubstring("--max-pods=100")) // kubenet
+			Expect(kubeletFlags).To(ContainSubstring("--max-pods=250")) // networkPlugin=azure
```
Bryce-Soghigian, line 517: Can we also add a case to cover kubenet and validate we are still seeing 100 as max pods for kubenet?

pkg/operator/options/options.go
Bryce-Soghigian: I don't think users should be able to provider their own string, it needs to be accurate to reference the fqdn. 

https://github.com/Azure/karpenter-provider-azure/blob/main/Makefile-az.mk#L47 it doesn't look like we pass this value in from flags or env variables. We probably should remove the if o.clusterId == "" because we need this value to be accurate always and shouldn't accept user input.

 Its used for a critical value https://github.com/Azure/karpenter-provider-azure/blob/main/pkg/providers/imagefamily/bootstrap/aksbootstrap.go#L452 so probably best to calculate it ourselves.

pkg/operator/options/options.go
```diff
@@ -0,0 +1,177 @@
...
+// getAKSClusterID returns cluster ID based on the DNS prefix of the cluster.
+// The logic comes from AgentBaker and other places, originally from aks-engine
+// with the additional assumption of DNS prefix being the first 33 chars of FQDN
+func getAKSClusterID(apiServerFQDN string) string {
+	dnsPrefix := apiServerFQDN[:33]
+	h := fnv.New64a()
+	h.Write([]byte(dnsPrefix))
+	r := rand.New(rand.NewSource(int64(h.Sum64()))) //nolint:gosec
+	return fmt.Sprintf("%08d", r.Uint32())[:8]
+}
```
Bryce-Soghigian, line 177: This seems copy pasted from settings, we have it in both places and in both places it isn't tested anywhere. Not that its easy to test the hashing but we can probably just use it from one place

pkg/operator/options/options.go
Bryce-Soghigian: @tallaxes  thoughts?

comtalyst: @tallaxes I think the consideration of changing this should be in a separate PR or issue, regardless of whether it should be removed or not. This proposed change in ClusterID is too far away from the purpose of this PR--to migrate from Settings to Options.
Unless it the consideration small enough like the unit test changes earlier...

tallaxes: Agree with both :-) We should be using calculated ClusterID exclusively; the only reason it is exposed as a configuration setting is because until certain stage I did not have enough confidence it is always calculated correctly. Since in current scenarios it is never passed externally, the configuration setting should be safe to remove. Can be handled in a separate PR.

comtalyst: @tallaxes The issue @Bryce-Soghigian brought up with the idea of separating PR is also: the separate PR will become a `TODO` that no one might pick up, and then never get fixed. 
My proposal is to create an issue to track this rather than letting it be a stray `TODO`, what do you think? I also think that that should only applied to not-so-small issues (we have to guess the size I think), otherwise it can be fixed along the way in the same PR.

tallaxes: Totally agree with creating issues for pretty much any identified follow-up work.

pkg/operator/options/options.go
comtalyst: The one in `Settings` will be removed once we drop the support, with this one only remains.

pkg/operator/options/options.go
comtalyst: https://github.com/Azure/karpenter-provider-azure/issues/130 I will get back to this after this one.

Bryce-Soghigian (dismissed):
Approved, but we need to remove settings for this feature to be usable at any stage. Looking forward to karpenter core migration

pkg/operator/options/options.go
```diff
@@ -0,0 +1,177 @@
...
+	SSHPublicKey                   string   // ssh.publicKeys.keyData => VM SSH public key // TODO: move to node template?
+	NetworkPlugin                  string   // => NetworkPlugin in bootstrap
+	NetworkPolicy                  string   // => NetworkPolicy in bootstrap
+	NodeIdentities                 []string // => Applied onto each VM
+
+	setFlags map[string]bool
+}
+
+func (o *Options) AddFlags(fs *coreoptions.FlagSet) {
+	fs.StringVar(&o.ClusterName, "cluster-name", env.WithDefaultString("CLUSTER_NAME", ""), "[REQUIRED] The kubernetes cluster name for resource discovery.")
```
tallaxes, line 73: This is not used for anything other than resource tags right now (and may actually be a good candidate for dropping). So maybe "for resource tags"?

pkg/operator/options/options.go
```diff
@@ -0,0 +1,177 @@
...
+	fs.StringVar(&o.ClusterEndpoint, "cluster-endpoint", env.WithDefaultString("CLUSTER_ENDPOINT", ""), "[REQUIRED] The external kubernetes cluster endpoint for new nodes to connect with. If not specified, will discover the cluster endpoint using DescribeCluster API.")
```
tallaxes, line 74:
```suggestion
	fs.StringVar(&o.ClusterEndpoint, "cluster-endpoint", env.WithDefaultString("CLUSTER_ENDPOINT", ""), "[REQUIRED] The external kubernetes cluster endpoint for new nodes to connect with.")
```
Has to be specified (and DescribeCluster is an EKS thing :-)

pkg/operator/options/options.go
```diff
@@ -0,0 +1,177 @@
...
+	fs.Float64Var(&o.VMMemoryOverheadPercent, "vm-memory-overhead-percent", env.WithDefaultFloat64("VM_MEMORY_OVERHEAD_PERCENT", 0.075), "The VM memory overhead as a percent that will be subtracted from the total memory for all instance types.")
+	fs.StringVar(&o.ClusterID, "cluster-id", env.WithDefaultString("CLUSTER_ID", ""), "The kubernetes cluster ID. If not specified, will generated based on cluster endpoint.")
```
tallaxes, line 76:
```suggestion
	fs.StringVar(&o.ClusterID, "cluster-id", env.WithDefaultString("CLUSTER_ID", ""), "The kubernetes cluster ID. If not specified, will be generated based on cluster endpoint.")
```

pkg/operator/options/options.go
```diff
@@ -0,0 +1,177 @@
...
+	fs.StringVar(&o.KubeletClientTLSBootstrapToken, "kubelet-client-tls-bootstrap-token", env.WithDefaultString("KUBELET_CLIENT_TLS_BOOTSTRAP_TOKEN", ""), "[REQUIRED] The bootstrap token for new nodes to join the cluster.")
```
tallaxes, line 77: Just `kubelet-boostrap-token` may be better, but not critical

pkg/operator/options/options.go
```diff
@@ -0,0 +1,177 @@
...
+	fs.StringVar(&o.SSHPublicKey, "ssh-public-key", env.WithDefaultString("SSH_PUBLIC_KEY", ""), "[REQUIRED] VM SSH public key.")
+	fs.StringVar(&o.NetworkPlugin, "network-plugin", env.WithDefaultString("NETWORK_PLUGIN", "azure"), "AKS cluster networking plugin.")
```
tallaxes, line 79:
```suggestion
	fs.StringVar(&o.NetworkPlugin, "network-plugin", env.WithDefaultString("NETWORK_PLUGIN", "azure"), "The network plugin used by the cluster.")
```

pkg/operator/options/options.go
```diff
@@ -0,0 +1,177 @@
...
+	fs.StringVar(&o.NetworkPolicy, "network-policy", env.WithDefaultString("NETWORK_POLICY", ""), "AKS cluster network policy.")
```
tallaxes, line 80:
```suggestion
	fs.StringVar(&o.NetworkPolicy, "network-policy", env.WithDefaultString("NETWORK_POLICY", ""), "The network policy used by the cluster.")
```

tallaxes (changes requested):
Some suggestions (just on option comments for now)

pkg/operator/options/options.go
tallaxes: Made some suggestions on descriptions. No need for `AZURE_` prefix, IMO; clear from context.

pkg/operator/options/options.go
comtalyst: Damn this copilot code slipped through my eyes

pkg/apis/settings/settings.go
```diff
@@ -142,6 +147,10 @@ func (s Settings) GetAPIServerName() string {
 	return endpoint.Hostname()
 }
 
+func (*Settings) FromContext(ctx context.Context) coresettings.Injectable {
+	return FromContext(ctx)
+}
```
tallaxes, line 152: Post-switch to options, this is not used anywhere?

pkg/cloudprovider/suite_test.go
```diff
@@ -70,10 +69,14 @@ func TestCloudProvider(t *testing.T) {
 	RunSpecs(t, "CloudProvider")
 }
 
+func toBoolPtr(b bool) *bool {
+	return &b
+}
```
tallaxes, line 74: Why not lo.ToPtr?

pkg/operator/options/options_validation.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+func (o Options) validateRequiredFields() error {
+	if o.ClusterEndpoint == "" {
+		return fmt.Errorf("missing field, cluster-endpoint")
+	}
+	if o.ClusterName == "" {
+		return fmt.Errorf("missing field, cluster-name")
+	}
+	return nil
```
tallaxes, line 64: `kubelet-bootstrap-token` and `ssh-public-key` are also required?

tallaxes (changes requested):
Minor nits. Please also mention in the summary the fix to network-plugin that sneaks in.

tallaxes:
We could also update the Helm chart to use options, even though ConfigMap is still supported? I presume you saw good reasons to delay / separate this.

pkg/apis/settings/settings.go
comtalyst: Correct. Some methods in `Settings` are left unused due to the references generally changed to `Options`. But still required due to we implementing the interface of `Settings` from core. This is primarily why the coverage drops.

pkg/operator/options/options_validation.go
comtalyst: Will add it. Looks like we were missing this up until this point...

comtalyst:
> We could also update the Helm chart to use options, even though ConfigMap is still supported? I presume you saw good reasons to delay / separate this.

I was planning to do so when we drop `Settings` entirely. 
The minor reason to not switch to that now is that changing `Makefile` and helm charts is a breaking interface change (if we count that as an interface, since some people might be using it). But now I think about it, if we are not having a release between this and "dropping `Settings`," I don't think this matter much. What do you think?

Testing in parallel (e.g., having another `Makefile` option) is proved to be a challenge primarily due to skaffold and the logic of "environment variables taking precedence."

pkg/apis/settings/settings.go
tallaxes: Hmm. If that's the case, it is a little unsettling that nothing obvious breaks when I remove it. Does it silently break injection? Maybe we should have a test for this then. Not critical if this goes away, just a bit surprizing.

tallaxes:
> helm charts is a breaking interface change (if we count that as an interface)

Not really breaking, as the "interface" part is represented by values, which don't change (they just flow differently into the templates), right? But I am good with doing this one step at a time.

pkg/apis/settings/settings.go
Bryce-Soghigian: Also we aren't excersising the options switch by default it wont fail it will just using settings. THis is why i was initially pushing to remove in this pr because without we aren't excercising this works for each of the other cases

comtalyst: Look at it again, looks like this **method** haven't been used at all. It is also not presented in the interface. 
Let's remove it. Should've been more careful with the public method pruning...

However, what's being used is the **function** of the same name in the same package. That is still being used in injection and merging with `Options.` This explains why removing the earlier doesn't break anything.

tallaxes (approved):

comtalyst (commented):
/test

comtalyst (commented):
/test

comtalyst (commented):
/test

Bryce-Soghigian (approved):

PR #120, opened 2024-01-28
chore: removing old crds and nodeTemplate reference

Bryce-Soghigian:
Fixes # 

**Description**
Removed old crds and nodeclass

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


Bryce-Soghigian:
Wondering if we need these old crds, I assume they were kept in for kaito since kaito uses alpha apis? Opened this pr to have that discussion 

CC: @rakechill

tallaxes:
> Wondering if we need these old crds, I assume they were kept in for kaito since kaito uses alpha apis? 

No, these CRDs are still required by the currently used version of core; so this will have to wait until we update core to v0.33+ (very soon). They are brought automatically from core within `make verify` (try `make presubmit`), so just deleting them from repo is ineffectual. (If you are wondering why this does not get caught by "file modifications detected" check - `git diff` does not consider new/untracked files ...) When these do go away, the corresponding symlinks from the chart need to go too.

`nodeTemplate` variable - good catch!

Bryce-Soghigian:
Thanks @tallaxes  will open new pr

PR #131, opened 2024-01-31
chore: create resource group before creating workload identity for local dev

rakechill:
**Description**
`make az-all` was failing at `az-create-workload-msi` with RESOURCE GROUP NOT FOUND. Added `az-mkrg` target as prereq of workload identity creation to fix this. Also updated `az- mkrg` to check if the resource group exists before trying to create it.

image

* `make az-all` succeeded

- [ ] Yes, issue opened: # 
- [X] No


Makefile-az.mk
```diff
@@ -24,7 +24,7 @@ az-login: ## Login into Azure
 az-mkrg: ## Create resource group
 	az group create --name $(AZURE_RESOURCE_GROUP) --location $(AZURE_LOCATION) -o none
 
-az-mkacr: az-mkrg ## Create test ACR
+az-mkacr: ## Create test ACR
```
charliedmcb, line 27: This breaks `make az-mkaks`. However, I think that might not be used anywhere, so could possibly be removed?

charliedmcb: Or you could add `az-mkrg` as a command to the `az-mkaks` before `az-mkacr`?

charliedmcb: Does leaving this here cause an error, if we try creating the calling `az-mkacr` twice?

If not, maybe it should be left?

charliedmcb: As it would also now break `az-mkaks-cilium` actually

charliedmcb (commented):
This would break some commands. Wondering if we can just add the rg call instead.

Makefile-az.mk
rakechill: Yeah I see that now. Running mk-rg twice will fail w/ rg already created error... can we run `az-mkacr` directly after `az-login`?

rakechill: I think this change worked bc I accidentally ran `make- az-mkrg` before this change. Let me edit this, my bad

charliedmcb: Well, `make az-all` will currently work since it will have the `az-create-workload-msi` call before, but its just the direct calls that will fail.

Makefile-az.mk
```diff
@@ -22,7 +22,9 @@ az-login: ## Login into Azure
 	az account set --subscription $(AZURE_SUBSCRIPTION_ID)
 
 az-mkrg: ## Create resource group
```
charliedmcb, line 24: Add a screenshot to the PR description with how it looks now? You mentioned that a `false`/`true` is now output?

charliedmcb (approved):
lgtm

Makefile-az.mk
rakechill:
```
make az-mkrg
if ! az group exists --name bookish-goldfish-r79r65v6x4jhggq; then \
        az group create --name bookish-goldfish-r79r65v6x4jhggq --location westus2 -o none; \
fi
false
```

rakechill: Adding an `@` before the if makes it instead look like this:
```
$ make az-mkrg
false
```

charliedmcb: I prefer having the if statement showing in the output to show what's being checked there. Gives context to the false

rakechill: Now there's no output,  * how fancy*, ty chatgpt

charliedmcb: I think, I personally would like which commands are being called, and their output to be displayed.

I think having the whole if block being output feels a little messy, but I'd possibly prefer that over no output. 

My optimal would be:
> az group exists --name $(AZURE_RESOURCE_GROUP)
> [output]
> [if the if block is executed, than output the az group create command, and its output]

charliedmcb (dismissed):
no output is fine with me, but I would generally personally prefer make commands to output what commands they are using.

charliedmcb (approved):
lgtm.

Would like a screenshot of what running the command outputs added to the PR description

PR #134, opened 2024-02-02
chore: az-mkrg fix

rakechill:
**Description**
Follow-up to #131. Currently the e2e test suite + local dev via `make az-all` is blocked by this. The if condition was always evaluating as true, because non-empty strings in bash are truthy.

* `make az-mkrg`, e2e suite (passed, except for utilization, not due to this issue)

- [ ] Yes, issue opened: # 
- [X] No


rakechill (commented):
/test

charliedmcb (approved):
I feel like there is probably something we could do more elegant than grep, but this works :)

PR #147, opened 2024-02-12
feat: add release scripts

tallaxes:
**Description**

Add `release` and `snapshot` scripts, `ksnap.azurecr.io` ACR for snapshots

* All releases (including snapshots) generate images for self-hosted and NAP (extra `-aks` in the tag), and Helm charts (referencing corresponding self-hosted images). Helm charts are packaged published to ACR as OCI artifacts.
* Snapshot releases can be generated with `make snapshot`, use commit hash as tag. Shapshot releases go to `ksnap.azurecr.io`, and are retained for 30 days. (See bicep template for ksnap.)
* Stable releases can be generated with `make release`, expected to be published to ACR onboarded to MCR. Stable releases have images locked after publishing (`--write-enabled false --delete-enabled false`).
* The corresponding https://github.com/Azure/karpenter-provider-azure/pull/148 adds and documents the process for consuming these stable or snapshot releases, deploying Karpenter via Helm charts
* Base images changed to `mcr.microsoft.com/cbl-mariner/distroless/minimal:2.0-nonroot`

* `make snapshot`

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


hack/release/common.sh
```diff
@@ -0,0 +1,163 @@
...
+  buildImages "${SNAPSHOT_REPO_ACR}"
+  updateHelmChart
+  # not locking artifacts for snapshot releases
+  cosignImage "${CONTROLLER_IMG}"
+  cosignImage "${CONTROLLER_IMG_NAP}"
+  publishHelmChart "karpenter" "${RELEASE_VERSION}" "${SNAPSHOT_REPO_ACR}"
+  publishHelmChart "karpenter-crd" "${RELEASE_VERSION}" "${SNAPSHOT_REPO_ACR}"
+}
+
+release() {
```
charliedmcb, line 49: nit: should this live in `releaes.sh`

hack/release/common.sh
```diff
@@ -0,0 +1,163 @@
...
+  local VERSION="$1"
+  local VERSION="${VERSION#[vV]}"
+  RELEASE_VERSION_MAJOR="${VERSION%%\.*}"
+  RELEASE_VERSION_MINOR="${VERSION#*.}"
+  RELEASE_VERSION_MINOR="${RELEASE_VERSION_MINOR%.*}"
+  RELEASE_VERSION_PATCH="${VERSION##*.}"
+  RELEASE_MINOR_VERSION="v${RELEASE_VERSION_MAJOR}.${RELEASE_VERSION_MINOR}"
+}
+
+snapshot() {
```
charliedmcb, line 32: nit: should this live in `snapshot.sh`?

hack/release/common.sh
```diff
@@ -0,0 +1,163 @@
...
+  lockImage "${IMG_REPOSITORY}" "${IMG_TAG}-aks"
+  cosignImage "${CONTROLLER_IMG}"
+  cosignImage "${CONTROLLER_IMG_NAP}"
+  updateHelmChart
+  publishHelmChart "karpenter" "${RELEASE_VERSION}" "${RELEASE_REPO_ACR}"
+  publishHelmChart "karpenter-crd" "${RELEASE_VERSION}" "${RELEASE_REPO_ACR}"
+}
+
+authenticatePrivateRepo() {
+  az acr login -n "${SNAPSHOT_REPO_ACR}"
```
charliedmcb, line 69: nit: don't really like the mixing of naming here, as this is used within the `release` script, and here its using the `SNAPSHOT_REPO_ACR` env var

charliedmcb: Is this actually a bug?

charliedmcb: Please leaving a comment to explain if intentional and keeping it this way.

hack/release/common.sh
```diff
@@ -0,0 +1,163 @@
...
+
+buildImages() {
+  RELEASE_REPO=$1
+  # Set the SOURCE_DATE_EPOCH and KO_DATA_DATE_EPOCH values for reproducable builds with timestamps
+  # https://ko.build/advanced/faq/
+
+  CONTROLLER_IMG=$(GOFLAGS=${GOFLAGS} \
+    SOURCE_DATE_EPOCH=$(git log -1 --format='%ct') KO_DATA_DATE_EPOCH=$(git log -1 --format='%ct') KO_DOCKER_REPO=${RELEASE_REPO} \
+    ko publish -B --sbom none -t "${RELEASE_VERSION}" "${RELEASE_PLATFORM}" ./cmd/controller)
+  CONTROLLER_IMG_NAP=$(GOFLAGS="${GOFLAGS} -tags=ccp" \
```
charliedmcb, line 80: should `ccp` be mentioned here?

hack/release/common.sh
```diff
@@ -0,0 +1,163 @@
...
+release() {
+  RELEASE_VERSION=$1
+  echo "Release Type: stable
+Release Version: ${RELEASE_VERSION}
+Commit: $(git rev-parse HEAD)
+Helm Chart Version $(helmChartVersion "$RELEASE_VERSION")"
+
+  authenticatePrivateRepo
+  buildImages "${RELEASE_REPO_ACR}"
+  updateHelmChart "${RELEASE_REPO_MAR}"
```
charliedmcb, line 58: Not quite following why we need to updateHelmChart twice here

charts/karpenter-crd/Chart.yaml
```diff
@@ -2,8 +2,8 @@ apiVersion: v2
 name: karpenter-crd
 description: A Helm chart for AKS Karpenter provider Custom Resource Definitions (CRDs)
 type: application
-version: 0.1.0
-appVersion: 0.1.0
+version: 0-41f6dcabe655c7fec84f4467e95bbd75285ae082
```
charliedmcb, line 5: Assuming this will get reverted for PR completion

charts/karpenter/Chart.yaml
```diff
@@ -2,8 +2,8 @@ apiVersion: v2
 name: karpenter
 description: A Helm chart for AKS Karpenter provider.
 type: application
-version: 0.1.0
-appVersion: 0.29.2
+version: 0-41f6dcabe655c7fec84f4467e95bbd75285ae082
+appVersion: 41f6dcabe655c7fec84f4467e95bbd75285ae082
```
charliedmcb, line 6: Assuming this will get reverted for PR completion

charts/karpenter/values.yaml
```diff
@@ -94,11 +94,13 @@ extraVolumes: []
 controller:
   image:
     # -- Repository path to the controller image.
-    repository: public.ecr.aws/karpenter/controller
+    repository: ksnap.azurecr.io/karpenter/snapshot/controller
     # -- Tag of the controller image.
-    tag: v0.32.1
+    tag: 41f6dcabe655c7fec84f4467e95bbd75285ae082
```
charliedmcb, line 99: Assuming this will get reverted for PR completion

hack/infra/ksnap-acr.bicep
charliedmcb, line 1: Is this actually referenced anywhere yet? I didn't spot it

charliedmcb (commented):
Overall, looking good, but have a few nits/questions.

hack/release/common.sh
tallaxes: It could. Though this follows some patterns established by the AWS provider that I like. The general idea is that snaphshot/release.sh are just thin entry points, most of the actual logic for both is in common.sh. Among other things, this makes it easy to see the differences and commonalities, update them in tandem, etc. I would prefer to keep it this way.

hack/release/common.sh
tallaxes: See the other comment on "snapshot".

hack/infra/ksnap-acr.bicep
tallaxes: This captures what we use for creating/maintaining the actual ksnap ACR, "infra-as-code" style. It has been applied manually.

hack/release/common.sh
charliedmcb: Sure, that's fine. Don't have a strong preference. I am mostly familiar with the common element containing only what's needed.

hack/infra/ksnap-acr.bicep
charliedmcb: Ah, ok that makes sense. Thanks for the explainer.

hack/release/common.sh
tallaxes: The second one left over by mistake, good catch, fixed

hack/release/common.sh
comtalyst: As Alex said, this part would not limited to only work with NAP, but also with having a general case of Karpenter being on a separate cluster/control plane (hence the name "ccp"). 

If it is not yet in our goals(?) to support self-hosted to be deployed like that, I am good with renaming it to "nap".

tallaxes: Took another look at the code - and the alternative operator implementation has many references to ccPlaneCtx, which make a lot of sense in that context. Feel like changing the build tag would introduce too much confusion, compared to the benefits. Leaving as is.

charts/karpenter-crd/Chart.yaml
tallaxes: Reverting. Neither value is great, sticking to the latest release probably makes the most sense for the in-repo chart. (Note that these values get automatically replaced for the published snapshot and release artifacts).

charts/karpenter/values.yaml
tallaxes: Reverting, though should be updated to proper release version (with next release)

hack/release/common.sh
tallaxes: It is a bug, fixing with proper ACR for release. (This is expected to go away for release, with auth switching to 1ES runner-based instead.)

charliedmcb (commented):
Overall, the release process for creating a snapshot was generall smooth. I do have some suggestions noted.

Additionally, the karpenter pod I created with the current commit snapshot, and the helm based install process PR did result in a `CrashLoopBackOff` which I will look into further:
https://github.com/Azure/karpenter-provider-azure/pull/148/files

image

Makefile
```diff
@@ -94,6 +100,12 @@ vulncheck: ## Verify code vulnerabilities
 codegen: ## Auto generate files based on Azure API responses
 	./hack/codegen.sh
 
+snapshot: ## Builds and publishes snapshot release
```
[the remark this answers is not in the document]
tallaxes: I don't' think shortcutting is worth the trouble. For actual releases we lock/protect the published image, but not for snapshots.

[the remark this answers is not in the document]
tallaxes: Yes, but let me see if we can avoid having to go through interactive login if alrady logged in ... Maybe something ike this:
```sh
	az account show -o none || az login
```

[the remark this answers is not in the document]
tallaxes: In most cases this is because the updates the chart version & values, so the next run will detect uncommited changes and refuse to proceed. Opting to add a message that clarifies what happened, rather than any kind automatic cleanup, which should help interactive scenarios. In GH action (which should drive this in most cases) this should not be a problem.

hack/release/release.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+GIT_TAG=$(git describe --exact-match --tags || echo "no tag")
```
charliedmcb, line 4: What if the commit has multiple tags?

hack/release/common.sh
```diff
@@ -0,0 +1,168 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+config(){
+  GITHUB_ACCOUNT="Azure"
+  RELEASE_ACR=${RELEASE_ACR:-ksnap.azurecr.io} # will always be ovreridden
+  RELEASE_REPO_ACR=${RELEASE_REPO_ACR:-${RELEASE_ACR}/public/aks/karpenter/}
```
charliedmcb, line 7:
```suggestion
  RELEASE_REPO_ACR=${RELEASE_REPO_ACR:-${RELEASE_ACR}/public/aks/karpenter}
```

Should the `/` at the end be removed?

Makefile
```diff
@@ -2,6 +2,12 @@ include Makefile-az.mk
 
 export K8S_VERSION ?= 1.27.x
 
+## Inject the app version into operator.Version
+LDFLAGS ?= -ldflags=-X=github.com/aws/karpenter-core/pkg/operator.Version=$(shell git describe --tags --always)
```
charliedmcb, line 6: This shouldn't be karpenter-core anymore, right?

hack/release/common.sh
```diff
@@ -0,0 +1,168 @@
...
+  ACR=$1
+  az acr login -n "${ACR}"
+}
+
+buildAndPublishImages() {
+  RELEASE_REPO=$1
+  # Set the SOURCE_DATE_EPOCH and KO_DATA_DATE_EPOCH values for reproducable builds with timestamps
+  # https://ko.build/advanced/faq/
+
+  CONTROLLER_IMG=$(GOFLAGS=${GOFLAGS} \
```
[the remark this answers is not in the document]
tallaxes: The wiring here just propagates the GOFLAGS into the build (originally set in the makefile); GOFLAGS in the makefile have comments

hack/release/release.sh
tallaxes: git describe "finds the most recent tag that is reachable from a commit", and I think this would actually work for annotated tags (which have dates). But let's not do that 🙂

hack/release/common.sh
```diff
@@ -0,0 +1,168 @@
...
+
+release() {
+  RELEASE_VERSION=$1
+  echo "Release Type: stable
+Release Version: ${RELEASE_VERSION}
+Commit: $(git rev-parse HEAD)
+Helm Chart Version $(helmChartVersion "$RELEASE_VERSION")"
+
+  authenticatePrivateRepo "${RELEASE_ACR}"
+  buildImages "${RELEASE_REPO_ACR}"
```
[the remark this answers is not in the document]
tallaxes: Renamed to buildAndPublish (now does more than images)

hack/release/common.sh
```diff
@@ -0,0 +1,157 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+RELEASE_ACR=${RELEASE_ACR:-ksnap.azurecr.io} # will always be ovreridden
```
uniemimu, line 4:
```suggestion
RELEASE_ACR=${RELEASE_ACR:-ksnap.azurecr.io} # will always be overridden
```

hack/release/common.sh
```diff
@@ -0,0 +1,157 @@
...
+# to the OCI registry alongside the image you're signing. For details see https://github.com/sigstore/cosign.
+cosignOciArtifact() {
+  local version commit_sha build_date artifact
+
+  version="${1}"
+  commit_sha="${2}"
+  build_date="${3}"
+  artifact="${4}"
+
+  cosign sign --yes -a version="${version}" -a commitSha="${commit_sha}" -a buildDate="${build_date}" "${artifact}"
```
uniemimu, line 144: cosign is probably worthy of mentioning in the README.md Install utilities section

hack/release/common.sh
```diff
@@ -0,0 +1,157 @@
...
+
+  cd charts
+  helm dependency update "${helm_chart}"
+  helm lint "${helm_chart}"
+  helm package "${helm_chart}" --version "${version}"
+  helm push "${helm_chart_artifact}" "oci://${oci_repo}"
+  rm "${helm_chart_artifact}"
+  cd ..
+
+  helm_chart_digest="$(crane digest "${oci_repo}/${helm_chart}:${version}")"
```
uniemimu, line 127: crane is probably worthy of mentioning in the README.md Install utilities section

charliedmcb (commented):
Looking good

Makefile-az.mk
```diff
@@ -26,7 +26,7 @@ az-all-custom-vnet: az-login az-create-workload-msi az-mkaks-custom-vnet az-crea
 az-all-savm:    az-login az-mkaks-savm az-perm-savm az-patch-skaffold-azure az-build az-run az-run-sample ## Provision the infra (ACR,AKS); build and deploy Karpenter; deploy sample Provisioner and workload - StandaloneVirtualMachines
 
 az-login: ## Login into Azure
-	az login
+	az account show -o none || az login
```
charliedmcb, line 29: Is this an intended change for this PR?

hack/release/common.sh
```diff
@@ -0,0 +1,157 @@
...
+  img="$(GOFLAGS=${GOFLAGS:-} \
+    SOURCE_DATE_EPOCH="${date_epoch}" KO_DATA_DATE_EPOCH="${date_epoch}" KO_DOCKER_REPO="${oci_repo}" \
+    ko publish -B --sbom none -t "${version}"     ./cmd/controller)"
+  img_nap="$(GOFLAGS="${GOFLAGS:-} -tags=ccp" \
+    SOURCE_DATE_EPOCH="${date_epoch}" KO_DATA_DATE_EPOCH="${date_epoch}" KO_DOCKER_REPO="${oci_repo}" \
+    ko publish -B --sbom none -t "${version}"-aks ./cmd/controller)"
+
+  img_repo="$(echo "${img}" | cut -d "@" -f 1 | cut -d ":" -f 1)"
+  img_tag="$(echo "${img}" | cut -d "@" -f 1 | cut -d ":" -f 2 -s)"
+  img_digest="$(echo "${img}" | cut -d "@" -f 2)"
```
charliedmcb, line 72: Could possibly use some minor comments on data format, and what exactly is getting extracted here. Little hard to just directly read/understand.

hack/release/common.sh
```diff
@@ -0,0 +1,157 @@
...
+
+  img_registry="$(echo "${img_repo}" | cut -d "/" -f 1)"
+  img_path="$(echo "${img_repo}" | cut -d "/" -f 2-)"
```
charliedmcb, line 75: Same comment on, "could possibly use some minor comments on data format, and what exactly is getting extracted here. Little hard to just directly read/understand."

Makefile-az.mk
tallaxes: Yes. While not strictly speaking part of the PR functionality, it addresses a long-standing problem with our makefile workflows: the fact that login is performed even if one is already logged in.

hack/release/common.sh
```diff
@@ -0,0 +1,157 @@
...
+}
+
+lockImage() {
+  local img_registry img_path img_tag
+
+  img_registry="$1"
+  img_path="$2"
+  img_tag="$3"
+
+	az acr repository update -n "${img_registry}" --image "${img_path}:${img_tag}" \
```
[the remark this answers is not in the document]
tallaxes: Not for me, but a tab instead of space snuck it; fixed

hack/release/common.sh
tallaxes: Added example, hopefully helps

hack/release/common.sh
tallaxes: Added a comment, thought the variable names tell the story ...

hack/release/common.sh
```diff
@@ -1,168 +1,157 @@
...
+  version="${2}"
+  helm_chart_version="${3}"
+  commit_sha="${4}"
+
+  date_epoch="$(dateEpoch)"
+  build_date="$(buildDate "${date_epoch}")"
+
+  img="$(GOFLAGS=${GOFLAGS:-} \
+    SOURCE_DATE_EPOCH="${date_epoch}" KO_DATA_DATE_EPOCH="${date_epoch}" KO_DOCKER_REPO="${oci_repo}" \
+    ko publish -B --sbom none -t "${version}"     ./cmd/controller)"
```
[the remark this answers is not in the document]
tallaxes: Sharp eyes! Those moved into defaults in ko.yaml

hack/release/common.sh
```diff
@@ -1,168 +1,157 @@
...
+  img_nap="$(GOFLAGS="${GOFLAGS:-} -tags=ccp" \
+    SOURCE_DATE_EPOCH="${date_epoch}" KO_DATA_DATE_EPOCH="${date_epoch}" KO_DOCKER_REPO="${oci_repo}" \
+    ko publish -B --sbom none -t "${version}"-aks ./cmd/controller)"
```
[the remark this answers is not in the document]
tallaxes: Same, ko.yaml

hack/release/common.sh
```diff
@@ -1,168 +1,157 @@
...
+
+  img_repo="$(echo "${img}" | cut -d "@" -f 1 | cut -d ":" -f 1)"
+  img_tag="$(echo "${img}" | cut -d "@" -f 1 | cut -d ":" -f 2 -s)"
+  img_digest="$(echo "${img}" | cut -d "@" -f 2)"
+
+  img_registry="$(echo "${img_repo}" | cut -d "/" -f 1)"
+  img_path="$(echo "${img_repo}" | cut -d "/" -f 2-)"
+
+  lockImage "${img_registry}" "${img_path}" "${img_tag}"
+  lockImage "${img_registry}" "${img_path}" "${img_tag}-aks"
```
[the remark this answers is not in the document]
tallaxes: Good catch, fixed

hack/release/common.sh
charliedmcb: This is great! Just having some info on the data format is perfect!

hack/release/common.sh
charliedmcb: I still think these little data examples are great!

charliedmcb (approved):
lgtm!

PR #152, opened 2024-02-16
chore: cleaning up chart from v0.33 upgrade

Bryce-Soghigian:
Fixes # 

**Description**
This PR removes rbac roles and permissions for the provisioners
- test command

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note
NONE
```


Bryce-Soghigian (commented):
/test

charts/karpenter/templates/aggregate-clusterrole.yaml
```diff
@@ -10,14 +10,10 @@ metadata:
     {{- toYaml . | nindent 4 }}
   {{- end }}
 rules:
-  # at least listing machines appears to be required for now
```
rakechill, line 13: this is just to remove v1alpha5 API?

pkg/providers/instance/instance.go
```diff
@@ -355,8 +355,8 @@ func setVMPropertiesBillingProfile(vmProperties *armcompute.VirtualMachineProper
 	}
 }
 
-// setVMTagsProvisionerName sets "karpenter.sh/provisioner-name" tag
-func setVMTagsProvisionerName(tags map[string]*string, nodeClaim *corev1beta1.NodeClaim) {
+// setVMTagsNodepoolName sets "karpenter.sh/nodepool-name" tag
```
rakechill, line 358: It's just `karpenter.sh/nodepool`, should fix the comment + func name.

> Karpenter no longer supports using the `karpenter.sh/provisioner-name` label in NodePool labels and requirements or in application node selectors, affinities, or topologySpreadConstraints. If you were previously using this label to target applications to specific Provisioners, you should update your applications to use the `karpenter.sh/nodepool` label instead before upgrading. If you upgrade without changing these labels, you may begin to see pod scheduling failures for these applications.

rakechill: Also if we want to do a full cleanup, then there are a lot more uses of Provisioner still around: https://github.com/search?q=repo%3AAzure%2Fkarpenter-provider-azure%20provisioner&type=code

rakechill (commented):
mostly lgtm, left some comments wrt other use cases of Provisioner. Also a correction for the nodepool tag name.

pkg/providers/instance/instance.go
Bryce-Soghigian: this was initially just aimed ath the chart and some code, but can do full cleanup now

rakechill (approved):

tallaxes (approved):

PR #170, opened 2024-02-28
chore(settings): cleanup final settings references

charliedmcb:
Fixes # 

**Description**
There were a few lingering settings references I found in the code, and deployment files. Cleaning up and removing those final references.

`/test` & `make presubmit`

Previously passing E2E run:
- https://github.com/Azure/karpenter-provider-azure/actions/runs/8087985217
*

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
NONE
```


designs/aks-node-bootstrap.md
```diff
@@ -9,7 +9,7 @@
 - [Karpenter configuration sources](#karpenter-configuration-sources)
   - [Hardcoded values](#hardcoded-values)
   - [Environment variables](#environment-variables)
-  - [Karpenter global settings](#karpenter-global-settings)
+  - [Karpenter global settings [OUTDATED]](#karpenter-global-settings)
```
charliedmcb, line 12: I think we should have a practice of leaving design docs as a snapshot of their state at design completion, and not take on the burden of continuously updating. If others agree I think I'll remove this change. However, ok with adding this `[OUTDATED]` tag too. 

I think if we want to update design docs as the assumptions they were built upon change that is more a symptom of needing better documentation, rather than actually a practice we want to have.

Bryce-Soghigian: I would say true, but we have zero documentation. this outdated tag is probably enough and this section isn't important. I updated the gpu doc a couple of times because we simply have no other documents covering it

charliedmcb (commented):
/test

charliedmcb (commented):
/test

charliedmcb (commented):
/test

charliedmcb (commented):
/test

charliedmcb (commented):
/test

Bryce-Soghigian (approved):

designs/aks-node-bootstrap.md
charliedmcb: Ok, I'll leave this change in for now. But yea I think we should move towards better documentation and freezing design docs after completion.

Issue #193, opened 2024-03-11
InstanceTypes Provider should cache instance types by image family


Bryce-Soghigian:
### Tell us about your request

We should be caching the instance types by image family  rather than recalcuating on the full list each time.

### Tell us about the problem you're trying to solve. What are you trying to do, and why is it hard?

WE want to get the instance types by image family as well in the cache. 

https://github.com/Azure/karpenter-provider-azure/blob/main/pkg/providers/instancetype/instancetypes.go#L73
```go
// Get all instance type options
func (p *Provider) List(
	ctx context.Context, kc *corev1beta1.KubeletConfiguration, nodeClass *v1alpha2.AKSNodeClass) ([]*cloudprovider.InstanceType, error) {
	p.Lock()
	defer p.Unlock()
	// Get SKUs from Azure
	skus, err := p.getInstanceTypes(ctx)
	if err != nil {
		return nil, err
	}

	// Get Viable offerings
	/// Azure has zones availability directly from SKU info
	var result []*cloudprovider.InstanceType
	for _, sku := range skus {
		vmsize, err := sku.GetVMSize()
		if err != nil {
			logging.FromContext(ctx).Errorf("parsing VM size %s, %v", *sku.Size, err)
			continue
		}
		architecture, err := sku.GetCPUArchitectureType()
		if err != nil {
			logging.FromContext(ctx).Errorf("parsing SKU architecture %s, %v", *sku.Size, err)
			continue
		}
		instanceTypeZones := instanceTypeZones(sku, p.region)
		instanceType := NewInstanceType(ctx, sku, vmsize, kc, p.region, p.createOfferings(sku, instanceTypeZones), nodeClass, architecture)
		if len(instanceType.Offerings) == 0 {
			continue
		}

		if !p.isInstanceTypeSupportedByImageFamily(sku.GetName(), lo.FromPtr(nodeClass.Spec.ImageFamily)) {
			continue
		}
		result = append(result, instanceType)
	}
	return result, nil
}
```

### Are you currently working around this issue?

No

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

tallaxes:
[...]

PR #194, opened 2024-03-11
perf: caching instance types in instancetypes provider by image family

Bryce-Soghigian:
Fixes # 

**Description**
This PR adds a new key for `nodeclass.Spec.ImageFamily` to the InstanceTypes Provider cache. This allows us to cache the instances available to ubuntu2204, and azurelinux separately, rather than recalculating on each iteration. 

- make presubmit. The AzureLinux GPU Instance type test should validate this works as expected. https://github.com/Azure/karpenter-provider-azure/blob/main/pkg/providers/instancetype/suite_test.go#L321

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
improving InstanceTypes provider List() performance by caching instance types in instancetypes provider by image family
```

### DIFF
--- a/pkg/providers/instancetype/instancetypes.go
+++ b/pkg/providers/instancetype/instancetypes.go
@@ -75,7 +75,7 @@ func (p *Provider) List(
 	p.Lock()
 	defer p.Unlock()
 	// Get SKUs from Azure
-	skus, err := p.getInstanceTypes(ctx)
+	skus, err := p.getInstanceTypes(ctx, lo.FromPtr(nodeClass.Spec.ImageFamily))
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


Bryce-Soghigian (commented):
/test

tallaxes (commented):
Commented on #193 [here](https://github.com/Azure/karpenter-provider-azure/issues/193#issuecomment-2016596976), since most feedback is around what needs to be done for instance type list caching in general, rather than this PR. Overall I think this has the right idea (there should indeed be caching to avoid recalculating every time; and image family, being one of the inputs affecting the resulting set of instance types for a given NodePool, should indeed be part of the overall cache key), but there is more to be done.

PR #223, opened 2024-03-20
test: output of grep command is capital letter and AZURE_ACR_NAME should be unique 

jacobbaek:
* changed to False from false. 
* can make a unique AZURE_ACR_NAME

Fixes # 

**Description**
'az group exists' printed true/false as capital letter.
So i added the ignore-case option(-i).
Also AZURE_ACR_NAME should be unique. but if the codespace is not using, only set karpenter as the AZURE_ACR_NAME 

make az-all

*

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**
NONE

```release-note

```

### DIFF
--- a/Makefile-az.mk
+++ b/Makefile-az.mk
@@ -1,5 +1,5 @@
 AZURE_LOCATION ?= westus2
-COMMON_NAME ?= karpenter2test
+COMMON_NAME ?= karpenter
 ifeq ($(CODESPACES),true)
   AZURE_RESOURCE_GROUP ?= $(CODESPACE_NAME)
   AZURE_ACR_NAME ?= $(subst -,,$(CODESPACE_NAME))
--- a/Makefile-az.mk
+++ b/Makefile-az.mk
@@ -23,7 +23,7 @@ az-login: ## Login into Azure
 	az account set --subscription $(AZURE_SUBSCRIPTION_ID)
 
 az-mkrg: ## Create resource group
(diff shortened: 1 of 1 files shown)


Makefile-az.mk
```diff
@@ -1,13 +1,14 @@
 AZURE_LOCATION ?= westus2
+COMMON_NAME ?= karpenter2test
```
Bryce-Soghigian, line 2: Love, the common name bit. can we keep it as karpenter? Then I will approve

jacobbaek: sure. I will revert this name into karpenter.

Bryce-Soghigian (approved):

tallaxes:
@jacobbaek, thank you for the contribution!
* I am curios which environment produced capitalized True/False? (Which I suspect is not valid JSON ...)
* You can (and could) set `AZURE_ACR_NAME` prior to running make (it uses conditional assignment)

jacobbaek:
Hi @tallaxes ,
In my case, the core.output setting was table. so the result of grep was printed "True".
It is not problem If use the json default setting.
To clarify, I also opened PR for updating document. there is 'az config set core.output=json'.
https://github.com/Azure/karpenter-provider-azure/pull/225
Thanks for your advice.

tallaxes:
> In my case, the core.output setting was table. so the result of grep was printed "True".

Ok, that explains it. A better fix for that command would be to not rely on the default setting and have explicit `-o json` or -`o tsv`. 

And thank you for helping document non-codespace dev environment!

PR #230, opened 2024-03-26
perf: instance types cache

### LINKED ISSUE #226: perf: cache fully initialized instance types

Fully initialized instance types (as returned by instanceTypeProvider.List) need to be cached. 

* This needs to be done with a new cache, which would cache the _fully initialized_ instance type list for a given set of inputs. (A convenient way to think about it is "per NodePool", but in reality, a set of inputs may be shared between different nodepools.)
* The existing lower-level cache (in getInstanceTypes) should stay the same, representing all (input-independent) instance types Karpenter can work with. It is there essentially to reduce skewer / resource API calls. It should also have a version (facilitated by change monitor), that input-specific / fully initialized instance type cache can use.
* All inputs affecting the resulting set of instance types need to be considered when caching. There are fields from NodePool (kubelet config) and AKSNodeClass (ImageFamily, OSDiskSize) that need to be hashed/included in the key. (There may be others, but those stand out.)
* Note that the (version of) unavailable offerings cache is yet another input, and should be included as well. (We need to add versioning support to that cache).

Here as well, AWS provider implementation can serve as a reference. Note the implementation of cache versioning (via SeqNum; facilited by ChangeMonitor where needed), composition of the synthetic key, etc.

To demonstrate improvement, it would be good to capture before and after numbers (e.g. maybe just via GetInstanceTypes method duration metric? pprof CPU and memory profile?) Adding a benchmark test (that would also catch regressions) would be great, but could be done separately.

Need to avoid having to compose / reinitialize the full list of instance types (with offerings) every time on List.

The current implementation works, at the expense of some performance hit (and likely quite a bit of extra garbage & GC)

### Additional Context

_No response_

### Attachments

_No response_

### Community Note

tallaxes:
Fixes #226 

**Description**

Introduces improved caching for fully initialized instance types. Please see #226 for details, this PR implements everything on the list there.

* make presubmit
* perf before/after - using CP method duration metric for GetInstanceTypes

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note
Improved performance and memory footprint via instance type caching
```

### DIFF
--- a/pkg/providers/instancetype/instancetypes.go
+++ b/pkg/providers/instancetype/instancetypes.go
@@ -107,7 +107,7 @@ func (p *Provider) List(
 		p.unavailableOfferings.SeqNum,
 		kcHash,
 		to.String(nodeClass.Spec.ImageFamily),
-		nodeClass.Spec.OSDiskSizeGB,
+		to.Int32(nodeClass.Spec.OSDiskSizeGB),
 		kubeReservedHash,
 		systemReservedHash,
 	)
--- a/pkg/providers/instancetype/instancetypes.go
+++ b/pkg/providers/instancetype/instancetypes.go
@@ -23,10 +23,13 @@ import (
 	"net/http"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
+	"github.com/mitchellh/hashstructure/v2"
 	"github.com/samber/lo"
 
+	"github.com/Azure/go-autorest/autorest/to"
 	"github.com/Azure/karpenter-provider-azure/pkg/apis/v1alpha2"
 	kcache "github.com/Azure/karpenter-provider-azure/pkg/cache"
 	"github.com/Azure/karpenter-provider-azure/pkg/utils"
@@ -41,6 +44,8 @@ import (
 	"github.com/alecthomas/units"
 	corev1beta1 "sigs.k8s.io/karpenter/pkg/apis/v1beta1"
 	"sigs.k8s.io/karpenter/pkg/cloudprovider"
+	"sigs.k8s.io/karpenter/pkg/utils/pretty"
+	"sigs.k8s.io/karpenter/pkg/utils/resources"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


tallaxes (commented):
/test

pkg/providers/instancetype/instancetypes.go
```diff
@@ -184,7 +226,13 @@ func (p *Provider) getInstanceTypes(ctx context.Context) (map[string]*skewer.SKU
 		}
 	}
 
-	logging.FromContext(ctx).Debugf("%d SKUs remaining after filtering", len(instanceTypes))
+	if p.cm.HasChanged("instance-types", instanceTypes) {
+		// Only update instanceTypesSeqNun with the instance types have been changed
+		// This is to not create new keys with duplicate instance types option
+		atomic.AddUint64(&p.instanceTypesSeqNum, 1)
+		logging.FromContext(ctx).With(
+			"count", len(instanceTypes)).Debugf("discovered instance types")
```
Bryce-Soghigian, line 234: can we use a name other than count? writing queries against the logs would be useful to have a term other than count in the json we can query against. 


InstanceCount maybe?

Bryce-Soghigian:
Would be good to get proper benchmarks of before and after, potentially flagging other areas that could use performance improvements as well.

pkg/providers/instancetype/instancetypes.go
tallaxes: Wouldn't it be possible to discern context from other fields? I dislike the idea of having to come up with different names in different contexts ...

Bryce-Soghigian: yeah you could do something like 

FakeEventsTable
| where msg startswith "discovered instance types" 
| project count

but the problem is that count is also a reserved keyword in KQL and other languages. So you would have do to 


FakeEventsTable
| where msg startswith "discovered instance types" 
| project ['count']

for it to actually run iirc

tallaxes:
Before:
image
After:
image

tallaxes:
https://github.com/Azure/karpenter-provider-azure/pull/232 refreshes dev monitoring tools and adds the above Provider Method Duration pane to Karpenter Performance dashboard

Bryce-Soghigian (approved):

tallaxes (commented):
/test

tallaxes (commented):
/test

PR #234, opened 2024-03-28
test(make): add taint node make command

charliedmcb:
Fixes # 

**Description**
I've found I've been wanting this make command myself, so adding one to the file.

Nothing additional
*

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
NONE
```

### DIFF
--- a/Makefile-az.mk
+++ b/Makefile-az.mk
@@ -232,6 +232,9 @@ az-rmnodeclaims-fin: ## Remove Karpenter finalizer from all nodeclaims (use with
 az-rmnodeclaims: ## kubectl delete all nodeclaims; don't wait for finalizers (use with care!)
 	kubectl delete --wait=false nodeclaims --all
 
+az-taintnodes: ## Run e2etests
+	kubectl taint nodes CriticalAddonsOnly=true:NoSchedule --all --overwrite
+
 az-e2etests: ## Run e2etests
 	kubectl taint nodes CriticalAddonsOnly=true:NoSchedule --all --overwrite
 	TEST_SUITE=Utilization make e2etests


Bryce-Soghigian (approved):

Makefile-az.mk
```diff
@@ -232,6 +232,9 @@ az-rmnodeclaims-fin: ## Remove Karpenter finalizer from all nodeclaims (use with
 az-rmnodeclaims: ## kubectl delete all nodeclaims; don't wait for finalizers (use with care!)
 	kubectl delete --wait=false nodeclaims --all
 
+az-taintnodes: ## Run e2etests
+	kubectl taint nodes CriticalAddonsOnly=true:NoSchedule --all --overwrite
```
tallaxes, line 236: Do you have use cases where this is not constrained to system nodepool? If not - should we target it explicitly with selector?

Bryce-Soghigian: I use this sometiems when running perf test, but not all of the nodes have scaled down I just want to launch a bunch of small instances to test somethign. 

So tainting all nodes is useful

Bryce-Soghigian: maybe tainting by label could be a second command?

charliedmcb: That's a fair idea. I've been running it right after cluster creation. However, it would be helpful to ensure it only targets system nodepool. I'll look at doing that.

Makefile-az.mk
```diff
@@ -232,6 +232,9 @@ az-rmnodeclaims-fin: ## Remove Karpenter finalizer from all nodeclaims (use with
 az-rmnodeclaims: ## kubectl delete all nodeclaims; don't wait for finalizers (use with care!)
 	kubectl delete --wait=false nodeclaims --all
 
+az-taintnodes: ## Run e2etests
```
tallaxes, line 235: The comment is wrong

Makefile-az.mk
charliedmcb: Will need a follow up PR though as I already completed

Makefile-az.mk
charliedmcb: Thx for the catch. Got a little eager on the completion. Another reason for a follow up

PR #235, opened 2024-03-28
test(make): minor follow up on taint make command

charliedmcb:
Fixes # 

**Description**
@tallaxes noticed a few solid changes to make to the command after completion, so addressing them here:
- https://github.com/Azure/karpenter-provider-azure/pull/234

Ran this on a newly created cluster and it tainted the nodes
*

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
NONE
```

### DIFF
--- a/Makefile-az.mk
+++ b/Makefile-az.mk
@@ -232,7 +232,7 @@ az-rmnodeclaims-fin: ## Remove Karpenter finalizer from all nodeclaims (use with
 az-rmnodeclaims: ## kubectl delete all nodeclaims; don't wait for finalizers (use with care!)
 	kubectl delete --wait=false nodeclaims --all
 
-az-taintnodes: ## Taint all system nodepool nodes
+az-taintsystemnodes: ## Taint all system nodepool nodes
 	kubectl taint nodes CriticalAddonsOnly=true:NoSchedule --selector='kubernetes.azure.com/mode=system' --overwrite
 
 az-e2etests: ## Run e2etests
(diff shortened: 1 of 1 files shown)


tallaxes (approved):
LGTM (could consider having target name mention system nodes, but not critical)

Makefile-az.mk
```diff
@@ -232,8 +232,8 @@ az-rmnodeclaims-fin: ## Remove Karpenter finalizer from all nodeclaims (use with
 az-rmnodeclaims: ## kubectl delete all nodeclaims; don't wait for finalizers (use with care!)
 	kubectl delete --wait=false nodeclaims --all
 
-az-taintnodes: ## Run e2etests
-	kubectl taint nodes CriticalAddonsOnly=true:NoSchedule --all --overwrite
+az-taintnodes: ## Taint all system nodepool nodes
+	kubectl taint nodes CriticalAddonsOnly=true:NoSchedule --selector='kubernetes.azure.com/mode=system' --overwrite
```
Bryce-Soghigian, line 236: I would prefer to taint all nodes not just system

charliedmcb: I agree with Alex that it should not taint the karpenter nodes. Could look at tainting any "aks nodepool" nodes. Is there a particular reason you want all nodes?

I know for myself I'd prefer otherwise as I ended up accidentally tainting some karpenter nodes before

charliedmcb:
> LGTM (could consider having target name mention system nodes, but not critical)

Updated. I think it makes sense to more specifically mention system.

tallaxes (approved):

Bryce-Soghigian (approved):

Makefile-az.mk
charliedmcb: Talked offline for a bit, around the fact that sometimes with stuck nodes being able to blanket taint them all and run tests on top can be useful. I think the scenario of just wanting system nodes will be more common. However, open to adding a taint all if it would be useful within the future.

PR #238, opened 2024-03-29
feat: support custom subnet ID (`--vnet-subnet-id`)

Bryce-Soghigian:
Fixes #199 

**Description**
This pr solves the issue linked above and allows karpenter to tap into a specifed custom vnet + subnet id, via `--vnet-subnet-id` global flag. Both self-hosted and NAP deployments are expected to populate this from custom subnet specified for the cluster, which can be looked up from the system nodepool.

**How was this change tested?**
- ran make az-all-custom-subnet with new custom vnet cluster create:
- ran make az-perftest100 took 3 minutes to create all 100 nodes, they all join the subnet fine
real    3m6.050s
user    0m0.304s
sys     0m0.072s

**Does this change impact docs?**
- [ ] Yes, PR includes docs updates
- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note
Support custom subnet ID via `--vnet-subnet-id` flag
```

PR #239, opened 2024-03-29
WIP: Support for VnetSubnetID as a nodeclass field enabling BYO Subnet scenarios

### LINKED ISSUE #231: BYO subnet per node pool (AKSNodeClass)
**Add support to karpenter for BYO Subnet per AKSNodeclass**

A workload may require splitting cluster nodes into separate pools for logical isolation. Separate subnets dedicated to each node pool in the cluster can help support this isolation, which can address requirements such as having noncontiguous virtual network address space to split across node pools. 

The goal of this issue is to define how unique subnets in a custom vnet will work inside of karpenter. It will be largely based on the existing [AKS Nodepool VNETSubnetID](https://learn.microsoft.com/en-us/azure/aks/create-node-pools#node-pools-with-unique-subnets) field.

## Acceptance Criteria
- BYO Subnet per AKSNodeclass 
- supported for both Node Auto Provisioning and Self Hosted Karpenter
- we will not be supporting `--pod-subnet-id` as karpenter currently only supports azure cni + overlay

## Limitations
- If you expand your VNET after creating the cluster, you must update your cluster before adding a subnet outside the original CIDR block.
- System pods must have access to all nodes and pods in the cluster to provide critical functionality, such as DNS resolution and tunneling kubectl logs/exec/port-forward proxy.
- Internal load balancers default to one of the node pool subnets.
- All Subnets assigned to nodepools must belong to the same virtual network

----

Bryce-Soghigian:
Fixes #231

**Description**
This PR Introduces the `VnetSubnetID field from the aks api into the v1alpha2 AKSNodeclass api. This will allow customers to specify individual subnets outside of the default subnet. 

This paired with https://github.com/Azure/karpenter-provider-azure/pull/238 will allow users to bring their own VNET, and pick their own node subnets. 

This change does not add support for pod subnet since overlay uses fake ips for pods on each node, and only uses real ips for Nodes. Pod subnet will need to be introduced alongside the azure cni without overlay. 

- make test 
- e2e 

**Followup Work** 
- Subnet Drift should be supported and have e2e tests 
- e2e test case for allocating to an additional subnet

**Release Note**

```release-note
Adding support for BYO subnet, allowing users to configure a subnet per nodeclass
```

### DIFF
--- a/pkg/providers/instance/instance.go
+++ b/pkg/providers/instance/instance.go
@@ -86,7 +86,7 @@ type Provider struct {
 	launchTemplateProvider *launchtemplate.Provider
 	loadBalancerProvider   *loadbalancer.Provider
 	resourceGroup          string
-	subnetID               string
+	defaultSubnetID               string
 	subscriptionID         string
 	unavailableOfferings   *cache.UnavailableOfferings
 }
@@ -99,7 +99,7 @@ func NewProvider(
 	offeringsCache *cache.UnavailableOfferings,
 	location string,
 	resourceGroup string,
-	subnetID string,
+	defaultSubnetID string,
 	subscriptionID string,
 ) *Provider {
 	listQuery = GetListQueryBuilder(resourceGroup).String()
@@ -109,8 +109,8 @@ func NewProvider(
 		launchTemplateProvider: launchTemplateProvider,
 		loadBalancerProvider:   loadBalancerProvider,
 		location:               location,
+		defaultSubnetID:        defaultSubnetID,
 		resourceGroup:          resourceGroup,
-		subnetID:               subnetID,
 		subscriptionID:         subscriptionID,
 		unavailableOfferings:   offeringsCache,
 	}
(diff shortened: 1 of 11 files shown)


Bryce-Soghigian (commented):
/test

Bryce-Soghigian (commented):
/test

Bryce-Soghigian (commented):
/test

Bryce-Soghigian (commented):
/test

pkg/providers/instancetype/suite_test.go
```diff
@@ -142,7 +139,54 @@ var _ = Describe("InstanceType Provider", func() {
...
+				ContainSubstring("kubernetes.azure.com/network-name=karpentervnet"),
+				ContainSubstring("kubernetes.azure.com/network-subnet=karpentersub"),
+				ContainSubstring("kubernetes.azure.com/network-subscription=00000000-0000-0000-0000-00000000000"),
+				ContainSubstring("kubernetes.azure.com/nodenetwork-vnetguid=test-vnet-guid"),
+				ContainSubstring("kubernetes.azure.com/podnetwork-type=overlay"),
+			))
+		})
+	})
 	Context("VM Creation Failures", func() {
+		//It("should not continue to use a nodepool with a full subnet", func() {})
```
Bryce-Soghigian, line 189: Will remove this, but thought it might be an interesting mitigation to return ice for subnet full and let karpenter move to a different nodepool with allocatable IPs.

pkg/providers/instance/instance.go
```diff
@@ -86,7 +86,7 @@ type Provider struct {
 	launchTemplateProvider *launchtemplate.Provider
 	loadBalancerProvider   *loadbalancer.Provider
 	resourceGroup          string
-	subnetID               string
+	defaultSubnetID        string
```
Bryce-Soghigian, line 89: should we remove this and just use the env variable instead? How do we feel about these global access patterns with the environment variable?

Bryce-Soghigian:
Going to close this entire pr since the 300 or so lines of refactoring has been moved over to the custom vnet pr. That means its time to throw this one away and start anew with fresh commit history

PR #245, opened 2024-04-02
chore: az-run-sample should provision some nodes

Bryce-Soghigian:
Fixes NA

**Description**
This change bumps up the inflate deployment to include a couple of replicas and taints the systempool before applying the inflate deployment. This should result in scale up from make az-run-sample. I find myself always running k scale deployment inflate --replicas 10 when creating the cluster. 

**Release Note**

```release-note

```


tallaxes:
Tainting systempool makes sense to me, but I would prefer to stick to manual scaling (or a separate target that is not executed automatically)

rakechill (approved):

tallaxes:
Discussed and decided against this

PR #249, opened 2024-04-03
perf(creation): remove polling from creation

charliedmcb:
Fixes # 

**Description**
Currently the Azure Karpenter Provider is not able to leverage budgets for doing multi-Drift, since the core logic only looks to start a second node Drift, after the first create responds back with its Provider ID. Since we wait for the actual Async operations to Azure to complete before moving on, this means that we are delaying this response, and losing out on the possible asynchronous multi-Drift functionality within budgets.

This change removes our Polling (waiting) on the create calls to allow us to respond back with the Provider ID quickly, and take advantage of multi-Drift.

We still track the vm create async errors internally, and asynchronously to track and mark unavailable offerings.

Note: for creations that fail asynchronously it will fall back on registerationTTL to clean up any failed instances, which is currently set at 15min:
- https://github.com/kubernetes-sigs/karpenter/blob/3d493b30f2464833a6482b99c80117dc20a23a44/pkg/controllers/nodeclaim/lifecycle/liveness.go#L40 

E2E run:
- https://github.com/Azure/karpenter-provider-azure/actions/runs/11245208770

Also, did multiple manual tests in code spaces.
*

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
Updates the approach for VM creation to return only after the request has been made to CRP and other resources. This unblocks processing of other nodes within Drift Disruption. However, means that async errors within the call will not be caught before returning the initial request. For creations that fail asynchronously it will fall back on registerationTTL to clean up any failed instances, which is currently set at 15min:
- https://github.com/kubernetes-sigs/karpenter/blob/3d493b30f2464833a6482b99c80117dc20a23a44/pkg/controllers/nodeclaim/lifecycle/liveness.go#L40  
```


charliedmcb (commented):
/test

charliedmcb (commented):
/test

charliedmcb (commented):
/test

charliedmcb (commented):
/test

Bryce-Soghigian:
> Note: for creations that fail asynchronously it will fall back on registerationTTL to clean up any failed instances, which is currently set at 15min:

This doesn't seem acceptable as for any async error that fails we will have an additional 14 minutes for a retry vs before it was 2s 4s 8s 16s 60s retries on nodeclaim create. 


Is there a better design for dealing with the long running error messages? Can we check async for those error messages in a separate goroutine not blocking the main Create thread and report back the status, if it fails we delete the nodeclaim or something like that?

pkg/providers/instance/instance.go
```diff
@@ -380,19 +380,18 @@ func setVMPropertiesBillingProfile(vmProperties *armcompute.VirtualMachineProper
 
 // setNodePoolNameTag sets "karpenter.sh/nodepool" tag
 func setNodePoolNameTag(tags map[string]*string, nodeClaim *corev1beta1.NodeClaim) {
-	if val, ok := nodeClaim.Labels[corev1beta1.NodePoolLabelKey]; ok {
-		tags[NodePoolTagKey] = &val
-	}
+	nodePoolLabel := nodeClaim.Labels[corev1beta1.NodePoolLabelKey]
+	tags[NodePoolTagKey] = &nodePoolLabel
```
Bryce-Soghigian, line 384: Why did we change this?

charliedmcb: Oh, good catch. This is unrelated to the change.

Pretty sure its a bug I found months back. Will separate it out into a new PR.

theunrepentantgeek:
Copilot summarizes the current state as:

>This branch is effectively obsolete. The "remove polling from creation" motivation was addressed by the Promise/Poller/MachineCache rework and the AKS Machine API migration; simply calling  Get  after BeginCreateOrUpdate  (as this branch does) would today be a regression against the more explicit sync-vs-async error contract in  VirtualMachinePromise . Safe to close/abandon; no salvage needed.

Closing as obsolete.

PR #261, opened 2024-04-11
chore: update dev tools

tallaxes:
**Description**

Bump the versions of development tools (including skaffold)

* `make presubmit`

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


hack/toolchain.sh
```diff
@@ -11,17 +11,17 @@ main() {
...
+    go install sigs.k8s.io/controller-runtime/tools/setup-envtest@v0.0.0-20240409134613-20f3f4bed925
     go install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.14.0
-    go install github.com/sigstore/cosign/v2/cmd/cosign@v2.2.3
+    go install github.com/sigstore/cosign/v2/cmd/cosign@v2.2.4
 #   go install -tags extended github.com/gohugoio/hugo@v0.110.0
-    go install golang.org/x/vuln/cmd/govulncheck@v1.0.3
-    go install github.com/onsi/ginkgo/v2/ginkgo@v2.15.0
-    go install github.com/rhysd/actionlint/cmd/actionlint@v1.6.26
+    go install golang.org/x/vuln/cmd/govulncheck@v1.0.4
+    go install github.com/onsi/ginkgo/v2/ginkgo@latest
```
Bryce-Soghigian, line 23: This means we will reinstall the ginkgo CLI whenever there is an update. Currently running `make presubmit` and having it trigger `make toolchain` works because it doesn't often change. 

We don't update it frequently: https://github.com/onsi/ginkgo/releases 

This is still something to consider.

Bryce-Soghigian (approved):

hack/toolchain.sh
tallaxes: Considered. The very reason for changing to `@latest` is because we usually bump the ginkgo packages (e.g. via dependabot) and CLI gets outdated, which generates a warning. So making an exception from pinning the versions here.

PR #274, opened 2024-04-18
feat: CEL validation for provider-specific labels/requirements

tallaxes:
**Description**

This adds additional CEL rules for validating provider-specific labels in NodePools and requirements in NodePools/NodeClaims. These are added on the provider side to core CRDs. The actual impactful changes are in the CRDs, the rest is infra for generating the patches (and maintaining them in a readable way). The infra part can be improved, this puts just enough to capture the results.

It also removes `topology.kubernetes.io/zone` from allowed labels and requirements, until the continuous drift it causes is fixed.

* `make presumbit`
* manual testing of CRs
* (current testing framework need to be updated to cover provider-specific CEL validation)

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


Bryce-Soghigian (approved):
> (current testing framework need to be updated to cover provider-specific CEL validation)

The only feedback here would be to add tests enforcing these changes. Seems you thought of that ahead of time in the pr. 

Approving given we come back and add tests when the framework supports it

Bryce-Soghigian (commented):
/test

hack/validation/requirements.sh
```diff
@@ -0,0 +1,37 @@
...
+        "karpenter.azure.com/sku-accelerator",
+        "karpenter.azure.com/sku-networking-accelerated",
+        "karpenter.azure.com/sku-storage-premium-capable",
+        "karpenter.azure.com/sku-storage-ephemeralos-maxsize",
+        "karpenter.azure.com/sku-encryptionathost-capable",
+        "karpenter.azure.com/sku-gpu-name",
+        "karpenter.azure.com/sku-gpu-manufacturer",
+        "karpenter.azure.com/sku-gpu-count"
+    ]
+    || !self.find("^([^/]+)").endsWith("karpenter.azure.com")
```
Bryce-Soghigian, line 22: > !self.find("^([^/]+)"

Can we leave a comment explaining this regex a bit? It is a bit magical to me. Rule applies for any well known labels and any labels that do not end with "karpenter.azure.com"? 


The () creates a grouping for extracting a string. then the slash means we want to extract a substring using that `/` as a back reference?

hack/validation/labels.sh
```diff
@@ -0,0 +1,33 @@
...
+        "karpenter.azure.com/sku-networking-accelerated",
+        "karpenter.azure.com/sku-storage-premium-capable",
+        "karpenter.azure.com/sku-storage-ephemeralos-maxsize",
+        "karpenter.azure.com/sku-encryptionathost-capable",
+        "karpenter.azure.com/sku-gpu-name",
+        "karpenter.azure.com/sku-gpu-manufacturer",
+        "karpenter.azure.com/sku-gpu-count"
+    ]
+    || !x.find("^([^/]+)").endsWith("karpenter.azure.com")
+)
```
Bryce-Soghigian, line 23: rule is the same for both of these right? can they just be merged into one script?

hack/validation/requirements.sh
tallaxes: No, it just means "all non-slash characters after start" = "everything from start to the first /". I can add a comment ...

hack/validation/labels.sh
tallaxes: It is close, but not quite the same, one operates on collection (of labels), another operates on a single thing (key). Yes, there are opportunities for sharing, that's what I mean by improving infra, but readability and result are more important for now.

Bryce-Soghigian (approved):

PR #275, opened 2024-04-18
perf: use vmname-nic for network interface names

tallaxes:
**Description**

Uses "vmname-nic" when naming network interfaces, as cloud-provider-azure has some optimizations for this case. Minor other cleanup.

* `make presubmit`

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/providers/instance/instance.go
+++ b/pkg/providers/instance/instance.go
@@ -178,7 +178,7 @@ func (p *Provider) List(ctx context.Context) ([]*armcompute.VirtualMachine, erro
 }
 
 func (p *Provider) Delete(ctx context.Context, resourceName string) error {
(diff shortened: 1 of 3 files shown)


pkg/providers/instance/instance.go
```diff
@@ -410,6 +410,11 @@ func (p *Provider) launchInstance(
 	return resp, instanceType, nil
 }
 
+func getNetworkInterfaceName(vmName string) string {
+	// cloud-provider-azure has certain optimizations (can avoid extra API calls) if this pattern is followed
```
Bryce-Soghigian, line 414: Can we link to the parts from cloud provider azure this is referencing?

PR #290, opened 2024-04-22
chore(make): fix create-cluster.sh deployment script to work with correct namespace

charliedmcb:
Fixes # 

**Description**
When testing `v0.4.0.rc.3` I noticed the service account federated credential not aligning correctly. The namespace within `create-cluster.sh` should be `karpenter`, so fixing.

tested this exact change with `v0.4.0.rc.3` on a different branch/code space and worked correctly.
*

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ GOFLAGS ?= $(LDFLAGS)
 WITH_GOFLAGS = GOFLAGS="$(GOFLAGS)"
 
 # # CR for local builds of Karpenter
-SYSTEM_NAMESPACE ?= karpenter
+KARPENTER_NAMESPACE ?= karpenter
 
 # Common Directories
 # TODO: revisit testing tools (temporarily excluded here, for make verify)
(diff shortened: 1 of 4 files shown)


tallaxes (commented):
`kube-system` is very likely a better place for it (at least for proper self-hosted deployment); ideally it should work with whatever namespace is specified (parameterized), but yes, there is certain value using one namespace consistently

charliedmcb:
> `kube-system` is very likely a better place for it (at least for proper self-hosted deployment); ideally it should work with whatever namespace is specified (parameterized), but yes, there is certain value using one namespace consistently

Agreed we should ideally be able to support whatever namespace is specified. 

I did a comb over the namespace usage, and with so much already using the `karpenter` namespace I think it makes sense to continue with that for now. If we think `kube-system` is better and want to try and tackle a conversion we could, but I think consistency is more important imo, than having a partial split repo within documentation and usage.

While we also shouldn't be hardcoding namespaces I found many scattered throughout, including make commands, and others that'd fail if using a different namespace.

Here's a little demo PR showing potential namespace changes. Although, would need to go through a little more comprehensively and make sure these are right/test it:
- https://github.com/Azure/karpenter-provider-azure/pull/291/files

tallaxes (approved):

PR #293, opened 2024-04-23
fix: don't use untagged on ksnap acr purge

tallaxes:
Fixes # 

**Description**

Don't use `--untagged` on ksnap acr purge, as it appears to be broken: https://github.com/Azure/acr-cli/issues/131

* (not tested yet)

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/hack/infra/ksnap-acr.bicep
+++ b/hack/infra/ksnap-acr.bicep
@@ -15,10 +15,11 @@ resource acr 'Microsoft.ContainerRegistry/registries@2023-11-01-preview' = {
 
 var schedule = '0 1 * * Tue' // 1am UTC every Tuesday
 
+// --untagged appears to be broken: https://github.com/Azure/acr-cli/issues/131
 var purgeOldArtifacts = '''
 version: v1.1.0
 steps:
-- cmd: acr purge --filter 'karpenter/snapshot/.*:.*' --ago 30d --untagged
+- cmd: acr purge --filter 'karpenter/snapshot/.*:.*' --ago 30d
   disableWorkingDirectoryOverride: true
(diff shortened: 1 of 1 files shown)


hack/infra/ksnap-acr.bicep
```diff
@@ -15,10 +15,11 @@ resource acr 'Microsoft.ContainerRegistry/registries@2023-11-01-preview' = {
 
 var schedule = '0 1 * * Tue' // 1am UTC every Tuesday
 
+// --untagged appears to be broken: https://github.com/Azure/acr-cli/issues/131
 var purgeOldArtifacts = '''
 version: v1.1.0
 steps:
-- cmd: acr purge --filter 'karpenter/snapshot/.*:.*' --ago 30d --untagged
+- cmd: acr purge --filter 'karpenter/snapshot/.*:.*' --ago 30d
```
Bryce-Soghigian, line 22: this will remove any image that is older than 30 days for snapshot images? 

I also read about dry run we could use to test some of this if we wanted to add tests for this kind of thing. 

--dry-run - Specifies that no data is deleted, but the output is the same as if the command is run without this flag. This parameter is useful for testing a purge command to make sure it does not inadvertently delete data you intend to preserve.

tallaxes: > this will remove any image that is older than 30 days for snapshot images?

Yes

> --dry-run

Yes, though I'd prefer to check the results of the actual execution instead; don't quite trust `--dry-run` at this point

Bryce-Soghigian (approved):

PR #306, opened 2024-04-28
feat: adding wiring for --network-plugin-mode and vnet labels cleanup

Bryce-Soghigian:
Fixes # 

**Description**
We need to pass in the wiring of the `network-plugin-mode` for supporting azure cni without overlay. This is the first step for that support. It also includes some general cleanup. 

*

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/operator/options/options.go
+++ b/pkg/operator/options/options.go
@@ -66,6 +66,7 @@ type Options struct {
 	SSHPublicKey                   string   // ssh.publicKeys.keyData => VM SSH public key // TODO: move to v1alpha2.AKSNodeClass?
 	NetworkPlugin                  string   // => NetworkPlugin in bootstrap
 	NetworkPolicy                  string   // => NetworkPolicy in bootstrap
+	NetworkPluginMode				string  // => Network Plugin mode is used to set AzureCNIs network plugin mode to overlay. Learn more about overlay networking here: https://learn.microsoft.com/en-us/azure/aks/azure-cni-overlay?tabs=kubectl#overview-of-overlay-networking
 	NodeIdentities                 []string // => Applied onto each VM
 
 	SubnetID string // => VnetSubnetID to use (for nodes in Azure CNI Overlay and Azure CNI + pod subnet; for for nodes and pods in Azure CNI), unless overridden via AKSNodeClass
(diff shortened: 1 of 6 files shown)


pkg/providers/instancetype/suite_test.go
```diff
@@ -507,13 +507,34 @@ var _ = Describe("InstanceType Provider", func() {
...
+			ExpectScheduled(ctx, env.Client, pod)
 
+			Expect(azureEnv.VirtualMachinesAPI.VirtualMachineCreateOrUpdateBehavior.CalledWithInput.Len()).To(Equal(1))
+			vm := azureEnv.VirtualMachinesAPI.VirtualMachineCreateOrUpdateBehavior.CalledWithInput.Pop().VM
+			customData := *vm.Properties.OSProfile.CustomData
+			Expect(customData).ToNot(BeNil())
+			decodedBytes, err := base64.StdEncoding.DecodeString(customData)
+			Expect(err).To(Succeed())
+			decodedString := string(decodedBytes[:])
+			// Since the network plugin is not "azure" and the network plugin is calico, it should not include the following kubeletLabels
```
Bryce-Soghigian, line 530:
```suggestion
			// Since the network plugin is not "azure" and the network policy is calico, it should not include the following kubeletLabels
```

pkg/operator/options/options.go
```diff
@@ -66,6 +66,7 @@ type Options struct {
 	SSHPublicKey                   string   // ssh.publicKeys.keyData => VM SSH public key // TODO: move to v1alpha2.AKSNodeClass?
 	NetworkPlugin                  string   // => NetworkPlugin in bootstrap
 	NetworkPolicy                  string   // => NetworkPolicy in bootstrap
+	NetworkPluginMode              string   // => Network Plugin mode is used to set AzureCNIs network plugin mode to overlay. Learn more about overlay networking here: https://learn.microsoft.com/en-us/azure/aks/azure-cni-overlay?tabs=kubectl#overview-of-overlay-networking
```
Bryce-Soghigian, line 69:
```suggestion
	NetworkPluginMode              string   // => Network Plugin is used to control the mode the network plugin should operate in. For example, "overlay" used with --network-plugin=azure will use an overlay network (non-VNET IPs) for pods in the cluster. Learn more about overlay networking here: https://learn.microsoft.com/en-us/azure/aks/azure-cni-overlay?tabs=kubectl#overview-of-overlay-networking
```

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -113,21 +114,24 @@ func (p *Provider) getStaticParameters(ctx context.Context, instanceType *cloudp
...
+
+	if options.FromContext(ctx).NetworkPlugin == networkPluginAzure && options.FromContext(ctx).NetworkPluginMode == networkModeOverlay {
+		// TODO: make conditional on pod subnet
+		vnetLabels, err := p.getVnetInfoLabels(ctx, nodeClass)
+		if err != nil {
+			return nil, err
+		}
+		labels = lo.Assign(labels, vnetLabels)
+	}
+	if options.FromContext(ctx).NetworkPolicy == networkDataplaneCilium {
```
tallaxes, line 126: Need to clearly distinguish between network-dataplane and network-policy settings (even though both can take "cilium" as one of the allowed values). Please introduce network-dataplane, and use it here.

tallaxes (changes requested):

pkg/providers/launchtemplate/launchtemplate.go
Bryce-Soghigian: Moved this change out to its own pr https://github.com/Azure/karpenter-provider-azure/pull/318. 

I suspect with network dataplane we will have some discussions about if it should be optional or not. Maybe same with this flag. Some clusters don't need overlay networking to be specified.

This flag was not added as a required flag and was excluded from the required validation for that reason. Cilium dataplane is the same for me.

pkg/operator/options/options_validation.go
```diff
@@ -31,11 +31,21 @@ func (o Options) Validate() error {
...
+		o.validateNetworkPluginMode(),
 		o.validateVnetSubnetID(),
 		validate.Struct(o),
 	)
 }
 
+func (o Options) validateNetworkPluginMode() error {
+	// TODO: Move overlay and none to shared constants in AgentBaker
+	// NOTE: Network Plugin Mode should be normalized on the AKS API Level, so if we are passing values in from the ManagedCluster, NetworkPluginMode will already be normalized.
+	if o.NetworkPluginMode != "overlay" && o.NetworkPluginMode != "" && o.NetworkPluginMode != "none" {
```
Bryce-Soghigian, line 43: we can remove "none" I validated we no longer support this as a possible input into the NetworkPluginMode AKS API.

PR #318, opened 2024-05-01
feat: --network-dataplane flag

Bryce-Soghigian:
**Description**
This pr gates the node labels we add to be gated alongside network dataplane only adding the cilium labels when cilium is enabled. 
- make presubmit

**Release Note**

```release-note
introducing the --network-dataplane flag
```


pkg/operator/options/options.go
```diff
@@ -81,6 +81,7 @@ func (o *Options) AddFlags(fs *coreoptions.FlagSet) {
 	fs.StringVar(&o.SSHPublicKey, "ssh-public-key", env.WithDefaultString("SSH_PUBLIC_KEY", ""), "[REQUIRED] VM SSH public key.")
 	fs.StringVar(&o.NetworkPlugin, "network-plugin", env.WithDefaultString("NETWORK_PLUGIN", "azure"), "The network plugin used by the cluster.")
 	fs.StringVar(&o.NetworkPolicy, "network-policy", env.WithDefaultString("NETWORK_POLICY", ""), "The network policy used by the cluster.")
+	fs.StringVar(&o.NetworkDataplane, "network-dataplane", env.WithDefaultString("NETWORK_DATAPLANE", "cilium"), "The network dataplane used by the cluster.")
```
Bryce-Soghigian, line 84: defaulting to cilium may be the wrong way to do things? We could default to empty.

Bryce-Soghigian (commented):
/test

pkg/operator/options/options_validation.go
```diff
@@ -44,6 +45,12 @@ func (o Options) validateVnetSubnetID() error {
 	return nil
 }
 
+func (o Options) validateNetworkDataplane() error {
+	if o.NetworkDataplane != "azure" && o.NetworkDataplane != "cilium" {
```
Bryce-Soghigian, line 49: these can move to using the constants package when that merges in

comtalyst (approved):
(discussed offline)

Approving code changes in general. 
But let's make sure we get things right on the relationship between `NetworkPolicy` and `NetworkDataplane`.

Bryce-Soghigian (commented):
/test

Bryce-Soghigian (commented):
/test

Bryce-Soghigian (commented):
/test

PR #336, opened 2024-05-10
[WIP] feat: Azure CNI Without Overlay

Bryce-Soghigian:
Fixes # 

**Description**

*

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


pkg/apis/consts/consts.go
```diff
@@ -0,0 +1,11 @@
+package consts
+
+const (
+	NetworkPluginAzure     = "azure"
+	NetworkPluginKubenet   = "kubenet"
+	
+	PodNetworkTypeOverlay  = "overlay"
+	PodNetworkTypeNone = ""
+
+	NetworkDataplaneCilium = "cilium"
```
Bryce-Soghigian, line 10: We could also potentially get these from the Azure sdk for go. IIRC they are exposed in the AKS API.

pkg/providers/instance/instance.go
```diff
@@ -212,26 +214,50 @@ func (p *Provider) newNetworkInterfaceForVM(vmName string, backendPools *loadbal
...
+						Primary: lo.ToPtr(false),
+						PrivateIPAllocationMethod: lo.ToPtr(armnetwork.IPAllocationMethodDynamic),
+						Subnet: &armnetwork.Subnet{
+							ID: &p.subnetID,
+						},
+					},
+				},
+			)
+		}
+	}
```
Bryce-Soghigian, line 259: Having kubeletconfiguration in the nodeclass would make this configuration for dynamic max pods easier. 

Otherwise we will end up adding some code to the launch template just to pass that configuration only to remove it a couple weeks later.

PR #351, opened 2024-05-14
chore: the output format of az command was specified

jacobbaek:
It is better to set output format.

Fixes # 

**Description**
someone might specify the ouput format as table or yaml.
If then, this script will fail.
So I added the output format as json.

it works well.

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note

```


jacobbaek:
@tallaxes 
Could you review this PR?

tallaxes (dismissed):
LGTM, thank you for the contribution!  (and apologies for the delay in reviewing)

tallaxes (approved):

PR #365, opened 2024-05-21
feat: Support for static AzureCNI without overlay networking via generating additional ip configurations

Bryce-Soghigian:
Fixes # 

**Description**
This PR adds support for azure cni without overlay, as well as introduces some makefile goodness for creating clusters or other cni configurations. 
- e2e tests 
- unit tests
- manual scale ups and goldpinger

**What this PR does not include** 
- E2E Suite for our Upstream E2Es here running network connectivity 

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note
support added for static AzureCNI without overlay networking
```


pkg/test/expectations/expectations.go
```diff
@@ -27,3 +28,18 @@ import (
...
+	Expect(customData).ToNot(BeNil())
+
+	decodedBytes, err := base64.StdEncoding.DecodeString(customData)
+	Expect(err).To(Succeed())
+	decodedString := string(decodedBytes[:])
+
+	for key, value := range expectedFlags {
+		Expect(decodedString).To(ContainSubstring(fmt.Sprintf("--%s=%s", key, value)))
+	}
+}
```
Bryce-Soghigian, line 45: Alternatively, we could write this to be more generic for checking any substring matches or doesnt match.  What I like about testing like this however is we can validate the flags themselves aren't malformed easily so I chose to do things this way.

Bryce-Soghigian (commented):
/test

pkg/test/expectations/expectations.go
```diff
@@ -27,3 +29,24 @@ import (
...
+	vm := env.VirtualMachinesAPI.VirtualMachineCreateOrUpdateBehavior.CalledWithInput.Pop().VM
+	customData := *vm.Properties.OSProfile.CustomData
+	Expect(customData).ToNot(BeNil())
+
+	decodedBytes, err := base64.StdEncoding.DecodeString(customData)
+	Expect(err).To(Succeed())
+	decodedString := string(decodedBytes[:])
+
+	return decodedString
+}
```
Bryce-Soghigian, line 52: Do we think its worth creating a separate package for assertions? Is assertDecodeCustomData a good name?

.github/actions/e2e/create-cluster/action.yaml
```diff
@@ -49,7 +49,7 @@ runs:
     run: AZURE_RESOURCE_GROUP=${{ inputs.resource_group }} AZURE_LOCATION=${{ inputs.location }} make az-create-workload-msi
   - name: create cluster
     shell: bash
-    run: AZURE_CLUSTER_NAME=${{ inputs.cluster_name }} AZURE_RESOURCE_GROUP=${{ inputs.resource_group }} AZURE_ACR_NAME=${{ inputs.acr_name }} AZURE_LOCATION=${{ inputs.location }} make az-mkaks-cilium
+    run: AZURE_CLUSTER_NAME=${{ inputs.cluster_name }} AZURE_RESOURCE_GROUP=${{ inputs.resource_group }} AZURE_ACR_NAME=${{ inputs.acr_name }} AZURE_LOCATION=${{ inputs.location }} make az-mkaks-cniv1
```
Bryce-Soghigian, line 52: Note this is just here so we can run the e2e tests on cniv1. I will revert this commit when we are ready to merge. For now I am leaving it so we can run our e2es against it if we need to change code in the node review

Makefile-az.mk
```diff
@@ -1,5 +1,5 @@
 AZURE_LOCATION ?= westus2
-COMMON_NAME ?= karpenter
+COMMON_NAME ?= karpenter
```
Bryce-Soghigian, line 2:
```suggestion
COMMON_NAME ?= karpenter
```

Makefile-az.mk
```diff
@@ -18,7 +18,12 @@ KARPENTER_FEDERATED_IDENTITY_CREDENTIAL_NAME ?= KARPENTER_FID
...
+az-all:             az-login az-create-workload-msi az-mkaks-cilium      az-create-federated-cred az-perm               az-perm-acr az-configure-values             az-build az-run          az-run-sample ## Provision the infra (ACR,AKS); build and deploy Karpenter; deploy sample Provisioner and workload 
+
+az-all-cniv1:             az-login az-create-workload-msi az-mkaks-cniv1      az-create-federated-cred az-perm               az-perm-acr az-configure-values             az-build az-run          az-run-sample ## Provision the infra (ACR,AKS); build and deploy Karpenter; deploy sample Provisioner and workload
+
+az-all-cni-overlay:  az-login az-create-workload-msi az-mkaks-overlay     az-create-federated-cred az-perm               az-perm-acr az-configure-values             az-build az-run          az-run-sample ## Provision the infra (ACR,AKS); build and deploy Karpenter; deploy sample Provisioner and workload
+
```
Bryce-Soghigian, line 26: seems we could improve the composability of these. 

Having two shared make steps 
make az-auth: az-login az-create-workload-msi 
make az-configure-and-build: az-perm az-perm-acr az-configure-values  az-build az-run


then we can refactor all of these to 
```yaml 
az-all:  az-auth az-mkaks-cilium   az-configure-and-build ## Provision the infra (ACR,AKS); build and deploy Karpenter; deploy sample Provisioner and workload 

az-all-cniv1:   az-auth az-mkaks-cniv1 az-configure-and-build 

az-all-cni-overlay:  az-auth az-mkaks-overlay  az-configure-and-build
```

Makefile-az.mk
```diff
@@ -43,13 +48,29 @@ az-mkaks: az-mkacr ## Create test AKS cluster (with --vm-set-type AvailabilitySe
...
 	skaffold config set default-repo $(AZURE_ACR_NAME).azurecr.io/karpenter
 
+az-mkaks-cniv1: az-mkacr 
+	az aks create          --name $(AZURE_CLUSTER_NAME) --resource-group $(AZURE_RESOURCE_GROUP) --attach-acr $(AZURE_ACR_NAME) \
+		--enable-managed-identity --node-count 3 --generate-ssh-keys -o none --network-plugin azure \
+		--enable-oidc-issuer --enable-workload-identity
+	az aks get-credentials --name $(AZURE_CLUSTER_NAME) --resource-group $(AZURE_RESOURCE_GROUP) --overwrite-existing
+	skaffold config set default-repo $(AZURE_ACR_NAME).azurecr.io/karpenter
+
+
```
Bryce-Soghigian, line 58: I realize there is a fair bit shared in these makefile commands @charliedmcb mentioned this before. 

But having composable steps with some var for network configuration bits and other flags seems like a pain. Is copy pasting the steps here sufficent?

hack/deploy/configure-values.sh
```diff
@@ -35,13 +35,18 @@ if [[ ! -v VNET_SUBNET_ID ]]; then
     VNET_SUBNET_ID=$(jq -r ".subnets[0].id" <<< "$VNET_JSON")
 fi
 
+# The // empty ensures that if the files is 'null' or not prsent jq will output nothing
+# If the value returned is none, its from jq and not the aks api in this case we return ""
+NETWORK_PLUGIN=$(jq -r ".networkProfile.networkPlugin // empty | if . == \"none\" then \"\" else . end" <<< "$AKS_JSON")
+NETWORK_PLUGIN_MODE=$(jq -r ".networkProfile.networkPluginMode // empty | if . == \"none\" then \"\" else . end" <<< "$AKS_JSON")
+NETWORK_POLICY=$(jq -r ".networkProfile.networkPolicy // empty | if . == \"none\" then \"\" else . end" <<< "$AKS_JSON")
+
```
Bryce-Soghigian, line 43: Open to suggestions on this part of the code I am particularly proud of it but it gets the job done.

pkg/operator/options/options.go
```diff
@@ -79,7 +81,8 @@ func (o *Options) AddFlags(fs *coreoptions.FlagSet) {
 	fs.Float64Var(&o.VMMemoryOverheadPercent, "vm-memory-overhead-percent", env.WithDefaultFloat64("VM_MEMORY_OVERHEAD_PERCENT", 0.075), "The VM memory overhead as a percent that will be subtracted from the total memory for all instance types.")
 	fs.StringVar(&o.KubeletClientTLSBootstrapToken, "kubelet-bootstrap-token", env.WithDefaultString("KUBELET_BOOTSTRAP_TOKEN", ""), "[REQUIRED] The bootstrap token for new nodes to join the cluster.")
 	fs.StringVar(&o.SSHPublicKey, "ssh-public-key", env.WithDefaultString("SSH_PUBLIC_KEY", ""), "[REQUIRED] VM SSH public key.")
-	fs.StringVar(&o.NetworkPlugin, "network-plugin", env.WithDefaultString("NETWORK_PLUGIN", "azure"), "The network plugin used by the cluster.")
+	fs.StringVar(&o.NetworkPlugin, "network-plugin", env.WithDefaultString("NETWORK_PLUGIN", consts.NetworkPluginAzure), "The network plugin used by the cluster.")
+	fs.StringVar(&o.NetworkPluginMode, "network-plugin-mode", env.WithDefaultString("NETWORK_PLUGIN_MODE", consts.PodNetworkTypeOverlay), "[REQUIRED] network plugin mode of the cluster")
```
Bryce-Soghigian, line 85: Should we be defaulting these values or should we force them to be required? Networking changes a lot for karpenter configuration of overlay vs non overlay. 

My thinking is they should be required because otherwise in some cases for non overlay clusters, passing in this default is a breaking change to them.

pkg/operator/options/options_validation.go
```diff
@@ -31,11 +32,21 @@ func (o Options) Validate() error {
...
 		o.validateEndpoint(),
 		o.validateVMMemoryOverheadPercent(),
+		o.validateNetworkPluginMode(),
 		o.validateVnetSubnetID(),
 		validate.Struct(o),
 	)
 }
 
+func (o Options) validateNetworkPluginMode() error {
+	// TODO: Move overlay and none to shared constants in AgentBaker
```
Bryce-Soghigian, line 42:
```suggestion
```

pkg/operator/options/options_validation.go
```diff
@@ -31,11 +32,20 @@ func (o Options) Validate() error {
...
 		o.validateVMMemoryOverheadPercent(),
+		o.validateNetworkPluginMode(),
 		o.validateVnetSubnetID(),
 		validate.Struct(o),
 	)
 }
 
+func (o Options) validateNetworkPluginMode() error {
+	// NOTE: Network Plugin Mode should be normalized on the AKS API Level, so if we are passing values in from the ManagedCluster, NetworkPluginMode will already be normalized.
+	if o.NetworkPluginMode != consts.PodNetworkTypeOverlay && o.NetworkPluginMode != "" && o.NetworkPluginMode != "none" {
```
Bryce-Soghigian, line 43: "none" isn't on most of all clusters in the datamodel I could part with not including it.

pkg/providers/instance/instance.go
```diff
@@ -244,7 +271,7 @@ func (p *Provider) createNetworkInterface(ctx context.Context, nicName string, l
 		return "", err
 	}
 
-	nic := p.newNetworkInterfaceForVM(nicName, backendPools, instanceType)
+	nic := p.newNetworkInterfaceForVM(ctx, nicName, backendPools, instanceType)
```
Bryce-Soghigian, line 274: these refactors for gocyclo for newNetworkInterfaceForVM and newVirtualMachine are kind of sloppy in the sense that we pass a lot of the same parameters down and the responsibilities are not that different. 


Should we refactor this elsewhere to combine the wrapper newNetworkInterfaceForVM with its lower level functions?

pkg/providers/instancetype/suite_test.go
```diff
@@ -1020,6 +1014,26 @@ var _ = Describe("InstanceType Provider", func() {
 					Expect(lo.FromPtr(nicDeleteOption)).To(Equal(armcompute.DeleteOptionsDelete))
 				}
 			})
+			It("should not create unneeded secondary ips for azure cni with overlay", func() {
+				ExpectApplied(ctx, env.Client, nodePool, nodeClass)
+				pod := coretest.UnschedulablePod()
+				ExpectProvisioned(ctx, env.Client, cluster, cloudProvider, coreProvisioner, pod)
+				ExpectScheduled(ctx, env.Client, pod)
```
Bryce-Soghigian, line 1021: I also copy paste this bit a fair amount. Should we add a helper that encapsulates this? 

ExpectProvisionedNodeForPod()?

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -113,21 +111,22 @@ func (p *Provider) getStaticParameters(ctx context.Context, instanceType *cloudp
...
 		arch = corev1beta1.ArchitectureArm64
 	}
-	// TODO: make conditional on either Azure CNI Overlay or pod subnet
-	vnetLabels, err := p.getVnetInfoLabels(ctx, nodeClass)
-	if err != nil {
-		return nil, err
-	}
-	labels = lo.Assign(labels, vnetLabels)
 
-	// TODO: Make conditional on epbf dataplane
```
Bryce-Soghigian, line 123: https://github.com/Azure/karpenter-provider-azure/pull/318 this is addressed here, comment shouldnt have been removed yet though

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -100,7 +98,7 @@ func (p *Provider) GetTemplate(ctx context.Context, nodeClass *v1alpha2.AKSNodeC
 	if err != nil {
 		return nil, err
 	}
-	launchTemplate, err := p.createLaunchTemplate(ctx, templateParameters)
+	launchTemplate, err := p.createLaunchTemplate(templateParameters)
```
Bryce-Soghigian, line 101: Is there a reason we dont want to store the context object on the Provider? I know go has this pattern of passing for each call so that you can cancel the context you created later, but we are really just using context for options and logging. 

Do we need to be passing it in the functions? I find myself adding and removing this frequently

Bryce-Soghigian (commented):
/test

pkg/utils/cni.go
```diff
@@ -0,0 +1,42 @@
...
+// DefaultMaxPods returns for a given network plugin the default value for pods per node
+func DefaultMaxPods(networkPlugin string) int {
+	if networkPlugin == consts.NetworkPluginAzure {
+		return defaultKubernetesMaxPodsAzure
+	}
+	if networkPlugin == consts.NetworkPluginKubenet {
+		return defaultKubernetesMaxPodsKubenet
+	}
+	return defaultKubernetesMaxPods
+}
```
Bryce-Soghigian, line 42: This package may become useless when we have our KubeletConfiguration implemented. For now I added it because we use DefaultMaxPods in two places, in the instance provider and in the image family resolver for the values we pass down to the kubelet. 


In future we will flow these values from the nodeclass and then we can remove this

Makefile-az.mk
```diff
@@ -43,13 +48,29 @@ az-mkaks: az-mkacr ## Create test AKS cluster (with --vm-set-type AvailabilitySe
 	az aks get-credentials --name $(AZURE_CLUSTER_NAME) --resource-group $(AZURE_RESOURCE_GROUP) --overwrite-existing
 	skaffold config set default-repo $(AZURE_ACR_NAME).azurecr.io/karpenter
 
+az-mkaks-cniv1: az-mkacr 
+	az aks create          --name $(AZURE_CLUSTER_NAME) --resource-group $(AZURE_RESOURCE_GROUP) --attach-acr $(AZURE_ACR_NAME) \
+		--enable-managed-identity --node-count 3 --generate-ssh-keys -o none --network-plugin azure \
+		--enable-oidc-issuer --enable-workload-identity
+	az aks get-credentials --name $(AZURE_CLUSTER_NAME) --resource-group $(AZURE_RESOURCE_GROUP) --overwrite-existing
+	skaffold config set default-repo $(AZURE_ACR_NAME).azurecr.io/karpenter
+
```
Bryce-Soghigian, line 57: These could alternatively be moved to use the create_cluster.sh bash script with the values we need for cluster create passed in as parameters, but this isn't so cumbersome to just copy paste a create command.

Makefile-az.mk
```diff
@@ -43,13 +48,29 @@ az-mkaks: az-mkacr ## Create test AKS cluster (with --vm-set-type AvailabilitySe
...
+
 az-mkaks-cilium: az-mkacr ## Create test AKS cluster (with --network-dataplane cilium, --network-plugin cilium, and --network-plugin-mode overlay)
 	az aks create          --name $(AZURE_CLUSTER_NAME) --resource-group $(AZURE_RESOURCE_GROUP) --attach-acr $(AZURE_ACR_NAME) \
 		--enable-managed-identity --node-count 3 --generate-ssh-keys -o none --network-dataplane cilium --network-plugin azure --network-plugin-mode overlay \
 		--enable-oidc-issuer --enable-workload-identity
 	az aks get-credentials --name $(AZURE_CLUSTER_NAME) --resource-group $(AZURE_RESOURCE_GROUP) --overwrite-existing
 	skaffold config set default-repo $(AZURE_ACR_NAME).azurecr.io/karpenter
 
+az-mkaks-overlay: az-mkacr ## Create test AKS cluster (with --network-dataplane cilium, --network-plugin cilium, and --network-plugin-mode overlay)
```
Bryce-Soghigian, line 66:
```suggestion
az-mkaks-overlay: az-mkacr ## Create test AKS cluster (with --network-plugin-mode overlay)
```

Makefile-az.mk
```diff
@@ -43,13 +48,29 @@ az-mkaks: az-mkacr ## Create test AKS cluster (with --vm-set-type AvailabilitySe
 	az aks get-credentials --name $(AZURE_CLUSTER_NAME) --resource-group $(AZURE_RESOURCE_GROUP) --overwrite-existing
 	skaffold config set default-repo $(AZURE_ACR_NAME).azurecr.io/karpenter
 
+az-mkaks-cniv1: az-mkacr
```
Bryce-Soghigian, line 51:
```suggestion
az-mkaks-cniv1: az-mkacr ## Create test AKS cluster (with --network-plugin azure)
```

.github/actions/e2e/create-cluster/action.yaml
Bryce-Soghigian, line 52:
```suggestion
    run: AZURE_CLUSTER_NAME=${{ inputs.cluster_name }} AZURE_RESOURCE_GROUP=${{ inputs.resource_group }} AZURE_ACR_NAME=${{ inputs.acr_name }} AZURE_LOCATION=${{ inputs.location }} make az-mkaks-cilium
```

pkg/operator/options/suite_test.go
```diff
@@ -115,6 +116,21 @@ var _ = Describe("Options", func() {
...
+				fs,
+				"--cluster-name", "my-name",
+				"--cluster-endpoint", "https://karpenter-000000000000.hcp.westus2.staging.azmk8s.io",
+				"--kubelet-bootstrap-token", "flag-bootstrap-token",
+				"--ssh-public-key", "flag-ssh-public-key",
+				"--vm-memory-overhead-percent", "-0.01",
+				"--network-plugin-mode", typo,
+			)
+			Expect(err).To(MatchError(ContainSubstring(errMsg)))
+		})
```
Bryce-Soghigian, line 133: These tests are brittle as the validation error message is dependent on the ordering of the calls and what flags we pass in. There is probably a better way with multi-error to test that we saw the error we are looking for.

pkg/apis/consts/consts.go
```diff
@@ -0,0 +1,27 @@
...
+
+const (
+	NetworkPluginAzure   = "azure"
+	NetworkPluginKubenet = "kubenet"
+
+	PodNetworkTypeOverlay = "overlay"
+	PodNetworkTypeNone    = ""
+
+	NetworkDataplaneCilium = "cilium"
+)
```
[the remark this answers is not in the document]
Bryce-Soghigian: Do we have these constants in the azure sdk for go for AKS?

Bryce-Soghigian (commented):
/test

pkg/test/expectations/expectations.go
charliedmcb: Maybe `ExpectDecodeVMCustomData`? `Expect` is the most commonly used within our testing, and VM in case there is any another source of custom data?

pkg/providers/instancetype/suite_test.go
```diff
@@ -513,7 +503,20 @@ var _ = Describe("InstanceType Provider", func() {
...
 		})
+		It("should not include cilium or azure cni vnet labels", func() {
+			ExpectApplied(ctx, env.Client, nodePool, nodeClass)
+			pod := coretest.UnschedulablePod()
+			ExpectProvisioned(ctx, env.Client, cluster, cloudProvider, coreProvisioner, pod)
+			ExpectScheduled(ctx, env.Client, pod)
 
+			customData := AssertDecodeCustomData(azureEnv)
+			// Since the network plugin is not "azure" it should not include the following kubeletLabels
+			Expect(customData).To(Not(SatisfyAny(
```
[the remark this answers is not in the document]
charliedmcb: These come from `getVnetInfoLabels`? However, this is customData, not labels?

Makefile-az.mk
```diff
@@ -18,7 +18,12 @@ KARPENTER_FEDERATED_IDENTITY_CREDENTIAL_NAME ?= KARPENTER_FID
...
+az-all:             az-login az-create-workload-msi az-mkaks-cilium      az-create-federated-cred az-perm               az-perm-acr az-configure-values             az-build az-run          az-run-sample ## Provision the infra (ACR,AKS); build and deploy Karpenter; deploy sample Provisioner and workload 
+
+az-all-cniv1:             az-login az-create-workload-msi az-mkaks-cniv1      az-create-federated-cred az-perm               az-perm-acr az-configure-values             az-build az-run          az-run-sample ## Provision the infra (ACR,AKS); build and deploy Karpenter; deploy sample Provisioner and workload
+
+az-all-cni-overlay:  az-login az-create-workload-msi az-mkaks-overlay     az-create-federated-cred az-perm               az-perm-acr az-configure-values             az-build az-run          az-run-sample ## Provision the infra (ACR,AKS); build and deploy Karpenter; deploy sample Provisioner and workload
```
charliedmcb, line 25: With the only things changing here between the different commands being the cluster creation, I wonder if it could make sense to make helper commands for the set of commands before, and after accordingly, to reduce duplication.

Makefile-az.mk
```diff
@@ -43,13 +48,29 @@ az-mkaks: az-mkacr ## Create test AKS cluster (with --vm-set-type AvailabilitySe
...
+	az aks create          --name $(AZURE_CLUSTER_NAME) --resource-group $(AZURE_RESOURCE_GROUP) --attach-acr $(AZURE_ACR_NAME) \
+		--enable-managed-identity --node-count 3 --generate-ssh-keys -o none --network-plugin azure \
+		--enable-oidc-issuer --enable-workload-identity
+	az aks get-credentials --name $(AZURE_CLUSTER_NAME) --resource-group $(AZURE_RESOURCE_GROUP) --overwrite-existing
+	skaffold config set default-repo $(AZURE_ACR_NAME).azurecr.io/karpenter
+
+
 az-mkaks-cilium: az-mkacr ## Create test AKS cluster (with --network-dataplane cilium, --network-plugin cilium, and --network-plugin-mode overlay)
 	az aks create          --name $(AZURE_CLUSTER_NAME) --resource-group $(AZURE_RESOURCE_GROUP) --attach-acr $(AZURE_ACR_NAME) \
 		--enable-managed-identity --node-count 3 --generate-ssh-keys -o none --network-dataplane cilium --network-plugin azure --network-plugin-mode overlay \
```
charliedmcb, line 61: Since everything within these cluster create commands is the same beyond the network settings I wonder if it'd be possibly to simply set an env var or similar for the NETWORK_FLAGS, and make a joint cluster create command.

Then expand the NETWORK_FLAGS as part of the cli command.

charliedmcb (commented):
Left a few minor comments.

Looking good to me, but want to learn a little more about the network pluggins, particularly the `networkPlugin == consts.NetworkPluginAzure && networkPluginMode != consts.PodNetworkTypeOverlay` case, before signing off.

pkg/providers/instancetype/suite_test.go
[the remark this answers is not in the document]
Bryce-Soghigian: the vnetInfoLabels are merged with the kubelet labels we pass into the node on the initial create request we make from the kubelet to the apiserver. We pass in kubelet node labels that eventually get patched there.

pkg/operator/options/options_validation.go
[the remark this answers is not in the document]
Bryce-Soghigian: Yeah we can do that. I hesitated at first to do apimachinery.sets here because it's a bit much for 3 values, but there is virtually no cost to defining that grouping so let's do it.

pkg/test/expectations/expectations.go
Bryce-Soghigian: The reason I didnt do Expect was it was sort of returning a value, but Expect does make more sense with the consistent patterns we have throughout the codebase. We can set it to expect

pkg/providers/instance/instance.go
```diff
@@ -212,26 +214,51 @@ func (p *Provider) newNetworkInterfaceForVM(vmName string, backendPools *loadbal
...
 	}
+	networkPlugin := options.FromContext(ctx).NetworkPlugin
+	networkPluginMode := options.FromContext(ctx).NetworkPluginMode
+	if networkPlugin == consts.NetworkPluginAzure && networkPluginMode != consts.PodNetworkTypeOverlay {
+		// NOTE: We don't need to set LoadBalancerBackendAddressPools for secondary ip configs.
+		// AzureCNI without overlay as pod networking type requires we configure additional secondary ips
+		// NOTE: There are some pods that we see on every node for networking like kube-proxy, and ip-masq-agent). We could be subtracting ips for these pods, and we do for AKSEngine. In the case of AKS However, we have a dynamic number of host networking addon pods. So we are explicitly choosing to not subtract these IPs like AKS engine and some parts of AKS do. The AKS networking team regrets adding this behavior and karpenter is a chance to do better so we shall.
+		// TODO: When MaxPods comes from the AKSNodeClass kubelet configuration, get the number of secondary
+		// ips from the nodeclass instead of using the default
+		for i := 1; i < utils.DefaultMaxPods(networkPlugin); i++ {
```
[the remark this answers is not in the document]
charliedmcb: Ok, this makes sense after chatting 1:1, and reviewing the PR description.

pkg/providers/instancetype/suite_test.go
```diff
@@ -152,13 +151,7 @@ var _ = Describe("InstanceType Provider", func() {
...
 			ExpectScheduled(ctx, env.Client, pod)
 
-			Expect(azureEnv.VirtualMachinesAPI.VirtualMachineCreateOrUpdateBehavior.CalledWithInput.Len()).To(Equal(1))
-			vm := azureEnv.VirtualMachinesAPI.VirtualMachineCreateOrUpdateBehavior.CalledWithInput.Pop().VM
-			customData := *vm.Properties.OSProfile.CustomData
-			Expect(customData).ToNot(BeNil())
-			decodedBytes, err := base64.StdEncoding.DecodeString(customData)
-			Expect(err).To(Succeed())
-			decodedString := string(decodedBytes[:])
+			decodedString := AssertDecodeCustomData(azureEnv)
```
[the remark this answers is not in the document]
Bryce-Soghigian: Yeah I have been meaning to refactor all of these tests but practicing restraint

pkg/apis/consts/consts.go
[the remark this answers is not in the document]
Bryce-Soghigian: After some more reflection i moved these to pkg/consts

pkg/providers/instance/instance.go
```diff
@@ -212,26 +214,51 @@ func (p *Provider) newNetworkInterfaceForVM(vmName string, backendPools *loadbal
...
-			EnableIPForwarding:          to.Ptr(true),
+			EnableAcceleratedNetworking: lo.ToPtr(enableAcceleratedNetworking),
+			EnableIPForwarding:          lo.ToPtr(true),
 		},
 	}
+	networkPlugin := options.FromContext(ctx).NetworkPlugin
+	networkPluginMode := options.FromContext(ctx).NetworkPluginMode
+	if networkPlugin == consts.NetworkPluginAzure && networkPluginMode != consts.PodNetworkTypeOverlay {
+		// NOTE: We don't need to set LoadBalancerBackendAddressPools for secondary ip configs.
+		// AzureCNI without overlay as pod networking type requires we configure additional secondary ips
```
tallaxes, line 241:
```suggestion
		// AzureCNI without overlay requires secondary IPs, for pods. (These IPs are not included in backend address pools.)
```

pkg/utils/cni.go
```diff
@@ -0,0 +1,42 @@
...
+
+import (
+	"github.com/Azure/karpenter-provider-azure/pkg/consts"
+)
+
+const (
+	// defaultKubernetesMaxPodsAzure is the maximum number of pods to run on a node for Azure CNI.
+	defaultKubernetesMaxPodsAzure = 250
+	// defaultKubernetesMaxPodsKubenet is the maximum number of pods to run on a node for Kubenet.
+	defaultKubernetesMaxPodsKubenet = 100
```
tallaxes, line 27: Why wouldn't we remove this altogether?

Makefile-az.mk
tallaxes: Can we please keep these vertically aligned, so that it is easy to spot the difference?

Makefile-az.mk
```diff
@@ -43,13 +48,29 @@ az-mkaks: az-mkacr ## Create test AKS cluster (with --vm-set-type AvailabilitySe
 	az aks get-credentials --name $(AZURE_CLUSTER_NAME) --resource-group $(AZURE_RESOURCE_GROUP) --overwrite-existing
 	skaffold config set default-repo $(AZURE_ACR_NAME).azurecr.io/karpenter
 
+az-mkaks-cniv1: az-mkacr ## Create test AKS cluster (with --network-plugin azure)
+	az aks create          --name $(AZURE_CLUSTER_NAME) --resource-group $(AZURE_RESOURCE_GROUP) --attach-acr $(AZURE_ACR_NAME) \
+		--enable-managed-identity --node-count 3 --generate-ssh-keys -o none --network-plugin azure \
```
tallaxes, line 53: Do we support non-cilium dataplane now?

tallaxes (commented):
Submitting initial feedback, will come back to finish the review

Makefile-az.mk
Bryce-Soghigian: I ran our e2e tests on non cilium dataplane. It seems to work ok so far, but we can do more testing if we choose(goldpinger and networking suite).

pkg/providers/instance/instance.go
```diff
@@ -212,26 +214,51 @@ func (p *Provider) newNetworkInterfaceForVM(vmName string, backendPools *loadbal
...
+		// NOTE: There are some pods that we see on every node for networking like kube-proxy, and ip-masq-agent). We could be subtracting ips for these pods, and we do for AKSEngine. In the case of AKS However, we have a dynamic number of host networking addon pods. So we are explicitly choosing to not subtract these IPs like AKS engine and some parts of AKS do. The AKS networking team regrets adding this behavior and karpenter is a chance to do better so we shall.
+		// TODO: When MaxPods comes from the AKSNodeClass kubelet configuration, get the number of secondary
+		// ips from the nodeclass instead of using the default
+		for i := 1; i < utils.DefaultMaxPods(networkPlugin); i++ {
+			nic.Properties.IPConfigurations = append(
+				nic.Properties.IPConfigurations,
+				&armnetwork.InterfaceIPConfiguration{
+					Name: lo.ToPtr(fmt.Sprintf("ipconfig-%d", i)),
```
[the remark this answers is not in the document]
Bryce-Soghigian: does rp just omit the dash? 

                            "ipConfigurations": [
                                {
                                    "name": "ipconfig1",
                                    "properties": {
                                        "primary": true,
                                        "subnet": {
                                            "id": "/.....aks-subnet"
                                        },
            
                                    }
                                }
                            ]
                            
        ```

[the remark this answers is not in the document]
Bryce-Soghigian: This is one from SAVM pool
     
        "ipConfigurations": [
            {
                "name": "ipconfig1",
                "id": ""
                "type": "Microsoft.Network/networkInterfaces/ipConfigurations",

pkg/operator/options/options_validation.go
Bryce-Soghigian: fair enough.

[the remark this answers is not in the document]
Bryce-Soghigian: good point

pkg/providers/instance/instance.go
```diff
@@ -212,26 +214,51 @@ func (p *Provider) newNetworkInterfaceForVM(vmName string, backendPools *loadbal
...
 				},
 			},
-			EnableAcceleratedNetworking: to.Ptr(enableAcceleratedNetworking),
-			EnableIPForwarding:          to.Ptr(true),
+			EnableAcceleratedNetworking: lo.ToPtr(enableAcceleratedNetworking),
+			EnableIPForwarding:          lo.ToPtr(true),
 		},
 	}
+	networkPlugin := options.FromContext(ctx).NetworkPlugin
+	networkPluginMode := options.FromContext(ctx).NetworkPluginMode
```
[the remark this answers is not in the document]
Bryce-Soghigian: Yeah, I was wondering about this too, it seems some of these values could be put in provider level values in their own structure.They could even be parameterized for cnis which parts we want to construct at a higher level.

pkg/utils/cni.go
Bryce-Soghigian: I wasn't certain if we would ever support kubelet. 

It could make some code much simpler if we didn't have to also consider network-plugin. Although maybe by just haivng this someone may get the impression that we do support it.

Makefile-az.mk
Bryce-Soghigian: Feels kind of brittle to pass in the flags as a string. I have thought of "ADDITIONAL_CLUSTER_ARGS" but then you end up just writing the flags again to type them in. Probably easiest to keep as is.

pkg/providers/instance/instance.go
```diff
@@ -212,26 +214,50 @@ func (p *Provider) newNetworkInterfaceForVM(vmName string, backendPools *loadbal
...
+	if networkPlugin == consts.NetworkPluginAzure && networkPluginMode != consts.NetworkPluginModeOverlay {
+		// AzureCNI without overlay requires secondary IPs, for pods. (These IPs are not included in backend address pools.)
+		// NOTE: Unlike AKS RP, this logic does not reduce secondary IP count by the number of expected hostNetwork pods, favoring simplicity instead
+		// TODO: When MaxPods comes from the AKSNodeClass kubelet configuration, get the number of secondary
+		// ips from the nodeclass instead of using the default
+		for i := 1; i < utils.DefaultMaxPods(networkPlugin); i++ {
+			nic.Properties.IPConfigurations = append(
+				nic.Properties.IPConfigurations,
+				&armnetwork.InterfaceIPConfiguration{
+					Name: lo.ToPtr(fmt.Sprintf("ipconfig-%d", i)),
```
Bryce-Soghigian, line 248:
```suggestion
					Name: lo.ToPtr(fmt.Sprintf("ipconfig%d", i)),
```

Bryce-Soghigian (commented):
/test

pkg/operator/options/options_validation.go
```diff
@@ -31,11 +32,20 @@ func (o Options) Validate() error {
...
 		o.validateEndpoint(),
 		o.validateVMMemoryOverheadPercent(),
+		o.validateNetworkPluginMode(),
 		o.validateVnetSubnetID(),
 		validate.Struct(o),
 	)
 }
 
+func (o Options) validateNetworkPluginMode() error {
+	// NOTE: Network Plugin Mode should be normalized on the AKS API Level, so if we are passing values in from the ManagedCluster, NetworkPluginMode will already be normalized.
```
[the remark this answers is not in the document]
Bryce-Soghigian: Removed these comments

pkg/providers/instance/instance.go
```diff
@@ -208,46 +210,77 @@ func (p *Provider) newNetworkInterfaceForVM(vmName string, backendPools *loadbal
...
 
-func (p *Provider) createNetworkInterface(ctx context.Context, nicName string, launchTemplateConfig *launchtemplate.Template, instanceType *corecloudprovider.InstanceType) (string, error) {
+type createNICOptions struct {
+	NICName           string
+	BackendPools      *loadbalancer.BackendAddressPools
+	InstanceType      *corecloudprovider.InstanceType
+	LaunchTemplate    *launchtemplate.Template
+	NetworkPlugin     string
+	NetworkPluginMode string
+}
```
Bryce-Soghigian, line 272: Ideally we could split out more options, but this is a good starting point. It does feel awkward to have this and not createOptions for the vm create calls.

.github/actions/e2e/create-cluster/action.yaml
rakechill: hm I wonder if we should just run some e2e tests using both? trying to be mindful of making e2e tests take longer tho...

would we have to recreate the aks cluster to use a different CNI or can we run a patch command to update this? if so, it might be worthwhile to have at least one test that runs the different CNIs so we know if things break

pkg/providers/instancetype/suite_test.go
rakechill: which lines are you referring to?

pkg/providers/launchtemplate/launchtemplate.go
rakechill: add the comment line back then?

pkg/utils/cni.go
rakechill: One point here -- not gone forever. We always have this code in our git history, so we can re-add and resolve any library/API conflicts quickly if they arise

rakechill (commented):
Left some initial comments, still probably want to take another look before merge.

pkg/providers/instancetype/suite_test.go
```diff
@@ -152,13 +151,7 @@ var _ = Describe("InstanceType Provider", func() {
 			ExpectProvisioned(ctx, env.Client, cluster, cloudProvider, coreProvisioner, pod)
 			ExpectScheduled(ctx, env.Client, pod)
 
-			Expect(azureEnv.VirtualMachinesAPI.VirtualMachineCreateOrUpdateBehavior.CalledWithInput.Len()).To(Equal(1))
```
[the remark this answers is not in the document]
Bryce-Soghigian: These were moved into their own helper that encapsulates all of that

.github/actions/e2e/create-cluster/action.yaml
Bryce-Soghigian: its probably better to write a specific goldpinger style suite exercising the actual node to node networking rather than just running on both. The only value of these suites running for them is that we get nodes joining apiserver and utilization tests max pods correctly.

pkg/providers/instancetype/suite_test.go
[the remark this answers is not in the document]
rakechill: I see `ExpectDecodedCustomData` now -- thanks

pkg/operator/options/suite_test.go
```diff
@@ -115,6 +116,21 @@ var _ = Describe("Options", func() {
 	})
 
 	Context("Validation", func() {
+		It("should fail when network-plugin-mode is invalid", func() {
+			typo := "overlaay"
+			errMsg := fmt.Sprintf("network-plugin-mode %v is invalid. network-plugin-mode must equal 'overlay' or ''", typo)
```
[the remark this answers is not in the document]
Bryce-Soghigian: i agree, maybe we should create a separate item as I believe this PR has more than enough structual changes and refactors.

pkg/providers/instance/instance.go
```diff
@@ -208,46 +209,76 @@ func (p *Provider) newNetworkInterfaceForVM(vmName string, backendPools *loadbal
...
 }
 
-func (p *Provider) createNetworkInterface(ctx context.Context, nicName string, launchTemplateConfig *launchtemplate.Template, instanceType *corecloudprovider.InstanceType) (string, error) {
+type createNICOptions struct {
+	NICName           string
+	BackendPools      *loadbalancer.BackendAddressPools
+	InstanceType      *corecloudprovider.InstanceType
+	LaunchTemplate    *launchtemplate.Template
+	NetworkPluginMode string
+}
```
Bryce-Soghigian, line 270: There probably exists a better pattern for this that splits the population of the creation options and the actual reading of those options. Its a bit messy here because we set some values like BackendPools after the fact rather than initially. 

Also creation options as a pattern could be more widely leveraged for VM creation as well. This was feedback placed here originally to split up some options rather than just using context as a global retrieval object. 

But seems like a lot to tackle restructuring the entire dataflow for the scope of this PR. It deserves its own proper refactoring PR.

Bryce-Soghigian (commented):
/test

Bryce-Soghigian (commented):
/test

pkg/providers/imagefamily/azlinux.go
```diff
@@ -97,7 +97,6 @@ func (u AzureLinux) UserData(kubeletConfig *corev1beta1.KubeletConfiguration, ta
 		ClusterID:                      u.Options.ClusterID,
 		APIServerName:                  u.Options.APIServerName,
 		KubeletClientTLSBootstrapToken: u.Options.KubeletClientTLSBootstrapToken,
-		NetworkPlugin:                  u.Options.NetworkPlugin,
```
tangyouzzz, line 100: Hello, after reading your PR, I think this is a great idea. However, I have a question as to why this line needs to be removed. I went to look at the provider.state.sh  of Azure VM, which determines whether to install Azure CNI based on the ` NETWORK-POLICY `. If this line is removed, will it cause Azure CNI installation to fail?

tangyouzzz: installNetworkPlugin() {
    if [[ "${NETWORK_PLUGIN}" = "azure" ]]; then
        installAzureCNI
    fi
    installCNI #reference plugins. Mostly for kubenet but loopback plugin is used by containerd until containerd 2
    rm -rf $CNI_DOWNLOADS_DIR & 
}

Bryce-Soghigian: Good catch you are correct, I discovered similar problems when testing it this morning, seems you are pointing to the exact sad part in agentbaker.

Bryce-Soghigian: :tada: aks-general-purpose-p7wg4           Ready    <none>   31s   v1.29.7

Nodes come up as ready after commit `5d88685` with fully initialized cni.

hack/deploy/configure-values.sh
```diff
@@ -35,13 +35,18 @@ if [[ ! -v VNET_SUBNET_ID ]]; then
     VNET_SUBNET_ID=$(jq -r ".subnets[0].id" <<< "$VNET_JSON")
 fi
 
+# The // empty ensures that if the files is 'null' or not prsent jq will output nothing
+# If the value returned is none, its from jq and not the aks api in this case we return ""
+NETWORK_PLUGIN=$(jq -r ".networkProfile.networkPlugin // empty | if . == \"none\" then \"\" else . end" <<< "$AKS_JSON")
```
Bryce-Soghigian: Will remove from karpenter-values.yaml after im done testing some other things.

pkg/operator/options/options.go
```diff
@@ -58,16 +59,18 @@ func (s *nodeIdentitiesValue) String() string { return strings.Join(*s, ",") }
...
-	VMMemoryOverheadPercent        float64
-	ClusterID                      string
+	ClusterName             string
+	ClusterEndpoint         string // => APIServerName in bootstrap, except needs to be w/o https/port
+	VMMemoryOverheadPercent float64
+	ClusterID               string
+
 	KubeletClientTLSBootstrapToken string // => TLSBootstrapToken in bootstrap (may need to be per node/nodepool)
 	SSHPublicKey                   string // ssh.publicKeys.keyData => VM SSH public key // TODO: move to v1alpha2.AKSNodeClass?
-	NetworkPlugin                  string // => NetworkPlugin in bootstrap
```
tallaxes, line 67: As we add support for selected CNI variants, we also want to make/keep BYO CNI possible - even if we don't fully support it right away. With this in mind (and for better alignment with AKS CNI settings in general), let's keep network plugin setting?

Bryce-Soghigian:
From what I can tell, in terms of the contract there are two places where NETWORK_PLUGIN is used

https://github.com/Azure/AgentBaker/blob/1c4f87e96b26cd8933a16cc31eaa9d18025932f1/self-contained/bootstrap_install.sh#L62
```go
installNetworkPlugin() {
    if [[ "${NETWORK_PLUGIN}" = "azure" ]]; then
        installAzureCNI
    fi
    installCNI  #reference plugins. Mostly for kubenet but loop back used by contaierd until containerd 2
    rm -rf $CNI_DOWNLOADS_DIR &
}
```

https://github.com/Azure/AgentBaker/blob/1c4f87e96b26cd8933a16cc31eaa9d18025932f1/self-contained/bootstrap_config.sh#L283
```go
configureCNIIPTables() {
    if [[ "${NETWORK_PLUGIN}" = "azure" ]]; then
        mv $CNI_BIN_DIR/10-azure.conflist $CNI_CONFIG_DIR/
        chmod 600 $CNI_CONFIG_DIR/10-azure.conflist
        if [[ "${NETWORK_POLICY}" == "calico" ]]; then
          sed -i 's#"mode":"bridge"#"mode":"transparent"#g' $CNI_CONFIG_DIR/10-azure.conflist
        elif [[ "${NETWORK_POLICY}" == "" || "${NETWORK_POLICY}" == "none" ]] && [[ "${NETWORK_MODE}" == "transparent" ]]; then
          sed -i 's#"mode":"bridge"#"mode":"transparent"#g' $CNI_CONFIG_DIR/10-azure.conflist
        fi
        /sbin/ebtables -t nat --list
    fi
}
```

but we need it to support BYO CNI it seems based on these two references in Agentbaker

pkg/operator/options/options.go
Bryce-Soghigian: Added it all back.

Bryce-Soghigian:
E2E Run: https://github.com/Azure/karpenter-provider-azure/actions/runs/10950653145/job/30407746822

tangyouzzz:
> From what I can tell, in terms of the contract there are two places where NETWORK_PLUGIN is used
> 
> https://github.com/Azure/AgentBaker/blob/1c4f87e96b26cd8933a16cc31eaa9d18025932f1/self-contained/bootstrap_install.sh#L62
> 
> ```go
> installNetworkPlugin() {
>     if [[ "${NETWORK_PLUGIN}" = "azure" ]]; then
>         installAzureCNI
>     fi
>     installCNI  #reference plugins. Mostly for kubenet but loop back used by contaierd until containerd 2
>     rm -rf $CNI_DOWNLOADS_DIR &
> }
> ```
> 
> https://github.com/Azure/AgentBaker/blob/1c4f87e96b26cd8933a16cc31eaa9d18025932f1/self-contained/bootstrap_config.sh#L283
> 
> ```go
> configureCNIIPTables() {
>     if [[ "${NETWORK_PLUGIN}" = "azure" ]]; then
>         mv $CNI_BIN_DIR/10-azure.conflist $CNI_CONFIG_DIR/
>         chmod 600 $CNI_CONFIG_DIR/10-azure.conflist
>         if [[ "${NETWORK_POLICY}" == "calico" ]]; then
>           sed -i 's#"mode":"bridge"#"mode":"transparent"#g' $CNI_CONFIG_DIR/10-azure.conflist
>         elif [[ "${NETWORK_POLICY}" == "" || "${NETWORK_POLICY}" == "none" ]] && [[ "${NETWORK_MODE}" == "transparent" ]]; then
>           sed -i 's#"mode":"bridge"#"mode":"transparent"#g' $CNI_CONFIG_DIR/10-azure.conflist
>         fi
>         /sbin/ebtables -t nat --list
>     fi
> }
> ```
> 
> but we need it to support BYO CNI it seems based on these two references in Agentbaker

After adding the NETWORK-POLICY variable, Karpenter should be able to be used in underlay network and Azure Cni AKS

pkg/providers/imagefamily/resolver.go
```diff
@@ -111,6 +90,19 @@ func (r Resolver) Resolve(ctx context.Context, nodeClass *v1alpha2.AKSNodeClass,
 	return template, nil
 }
 
+func prepareKubeletConfiguration(instanceType *cloudprovider.InstanceType, nc *corev1beta1.NodeClaim) *corev1beta1.KubeletConfiguration {
+	kubeletConfig := nc.Spec.Kubelet
+	if kubeletConfig == nil {
+		kubeletConfig = &corev1beta1.KubeletConfiguration{}
+	}
+	kubeletConfig.MaxPods = lo.ToPtr[int32](consts.DefaultKubernetesMaxPodsAzure)
```
tangyouzzz, line 98: `consts.DefaultKubernetesMaxPodsAzure' is a fixed value, and some nodes may not seem to need to allocate so many IPs. Can we first obtain it through the maxpod field in the nodepool, and only use the default value if it is not defined?

Bryce-Soghigian: kubelet configuration isnt something we are supporting for GA

tangyouzzz: Okay, thank you for your answer

pkg/test/expectations/expectations.go
```diff
@@ -27,3 +29,24 @@ import (
 func ExpectUnavailable(env *test.Environment, instanceType string, zone string, capacityType string) {
 	Expect(env.UnavailableOfferingsCache.IsUnavailable(instanceType, fmt.Sprintf("%s-%s", fake.Region, zone), capacityType)).To(BeTrue())
 }
+
+func ExpectKubeletFlags(env *test.Environment, customData string, expectedFlags map[string]string) {
+	kubeletFlags := customData[strings.Index(customData, "KUBELET_FLAGS=")+len("KUBELET_FLAGS="):]
```
tallaxes, line 34: Consider this (using "KUBELET_NODE_LABELS" is a little fagile, but avoids sucking in the rest of custom data):
```suggestion
	kubeletFlags := customData[strings.Index(customData, "KUBELET_FLAGS=")+len("KUBELET_FLAGS=") : strings.Index(customData, "KUBELET_NODE_LABELS")]
```

tallaxes: Also, custom expectations should normally have `GinkoHelper()` as first line.
(And we can consider co-locating them the suites they are using in, if they are only used in one.)

tallaxes (changes requested):
Looks good, mosly minor things

.github/actions/e2e/create-cluster/action.yaml
tallaxes: A matrix (ideally optional, but easy to engage on demand) would be great - and option-specific test could plug into that ...

tallaxes, line 52: Can we switch to passing vars in via `env` block? Would improve readability ...

pkg/consts/consts.go
```diff
@@ -0,0 +1,30 @@
...
+	NetworkPluginAzure = "azure"
+	NetworkPluginNone  = "none"
+
+	NetworkPluginModeOverlay = "overlay"
+	NetworkPluginModeNone    = ""
+
+	NetworkDataplaneCilium = "cilium"
+	NetworkDataplaneAzure  = "azure"
+
+	DefaultKubernetesMaxPodsAzure = 250
```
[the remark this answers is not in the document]
Bryce-Soghigian: Before we used to have a default for kubenet, but doesn't make sense anymore now those consts are removed. 

I will change the name.

pkg/providers/instance/instance.go
```diff
@@ -208,46 +209,77 @@ func (p *Provider) newNetworkInterfaceForVM(vmName string, backendPools *loadbal
...
 	backendPools, err := p.loadBalancerProvider.LoadBalancerBackendPools(ctx)
 	if err != nil {
 		return "", err
 	}
-
-	nic := p.newNetworkInterfaceForVM(nicName, backendPools, instanceType)
-	p.applyTemplateToNic(&nic, launchTemplateConfig)
-	logging.FromContext(ctx).Debugf("Creating network interface %s", nicName)
-	res, err := createNic(ctx, p.azClient.networkInterfacesClient, p.resourceGroup, nicName, nic)
+	opts.BackendPools = backendPools
```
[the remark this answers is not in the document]
Bryce-Soghigian: Ok fixed moved to get backendPools before we create options.

pkg/test/expectations/expectations.go
Bryce-Soghigian: Really seems like we should at some point refactor all of suite_test for instancetypes to other places rather than putting all of those tests there. 

So im going to wait on co-locating them and leave them here. 

But i did take your suggestion on KUBELET_NODE_LABELS

pkg/operator/options/options_validation.go
```diff
@@ -31,12 +32,28 @@ func (o Options) Validate() error {
...
 		o.validateNetworkDataplane(),
+		o.validateNetworkPlugin(),
 		o.validateVnetSubnetID(),
 		validate.Struct(o),
 	)
 }
 
+func (o Options) validateNetworkPluginMode() error {
+	if o.NetworkPluginMode != consts.NetworkPluginModeOverlay && o.NetworkPluginMode != consts.NetworkPluginModeNone {
+		return fmt.Errorf("network-plugin-mode %v is invalid. network-plugin-mode must equal 'overlay' or ''", o.NetworkPluginMode)
```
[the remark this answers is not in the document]
Bryce-Soghigian: addressed

.github/actions/e2e/create-cluster/action.yaml
Bryce-Soghigian: Seems out of the scope of this PR but if we want to invest in parameterizing this and maybe some sort of matrix/suite for running the two modes + making these vars in this file come from env instead of passing in we can later.

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -113,23 +111,25 @@ func (p *Provider) getStaticParameters(ctx context.Context, instanceType *cloudp
 	if err := instanceType.Requirements.Compatible(scheduling.NewRequirements(scheduling.NewRequirement(v1.LabelArchStable, v1.NodeSelectorOpIn, corev1beta1.ArchitectureArm64))); err == nil {
 		arch = corev1beta1.ArchitectureArm64
 	}
-	// TODO: make conditional on either Azure CNI Overlay or pod subnet
-	vnetLabels, err := p.getVnetInfoLabels(ctx, nodeClass)
-	if err != nil {
-		return nil, err
+
+	if options.FromContext(ctx).NetworkPlugin == consts.NetworkPluginAzure && options.FromContext(ctx).NetworkPluginMode == consts.NetworkPluginModeOverlay {
```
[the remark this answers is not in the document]
Bryce-Soghigian: Will Daly 9:36am
these labels are used for Azure CNI Overlay and podsubnet. They are not used for Azure CNI nodesubnet or BYO CNI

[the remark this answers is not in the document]
Bryce-Soghigian: Resolving.

pkg/providers/instancetype/suite_test.go
Bryce-Soghigian: 1018-1021

pkg/utils/cni.go
Bryce-Soghigian: Yeah removed all of this garbage.

.github/actions/e2e/create-cluster/action.yaml
Bryce-Soghigian: Pushed the changes here, I will try running it and see what happens

pkg/providers/imagefamily/resolver.go
Bryce-Soghigian: Yeah, that was my same rationale, we should just do the max until we support setting it. The default for azure cni without overlay is 30, but for azure cni with overlay is 250. 

That also makes me wonder if that limit comes from the pod ip scaling of overlay and maybe there is a historical reason for that limit.

.github/actions/e2e/create-cluster/action.yaml
tallaxes: Worked (https://github.com/Azure/karpenter-provider-azure/actions/runs/11111453449), resolving

pkg/providers/imagefamily/resolver.go
tallaxes: Discussed offline. There is a concern about exhausting the IP space quickly; on the other hand Karpenter can provision larger nodes more often, so underspecifying max-pods is also undesirable. Keeping it 250 for now, will revisit (hopefully just make it configurable).

pkg/operator/options/options_validation.go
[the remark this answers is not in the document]
tallaxes: May want to validate to allow just the supported combination of plugin/mode/dataplane, but can do this as follow-up. Resolving for now.

tallaxes (approved):

tallaxes (approved):

pdeva:
thank you for fixing this folks! hope a new release is cut soon. we cant wait to use this on our AKS cluster.

marchermans:
@Bryce-Soghigian Question, as it seems to confuse us: Does this PR support running without an Azure CNI network plugin (for example by running Calico Enterprise natively)?

Bryce-Soghigian:
@marchermans We do not support the calico network policy, is there anything driving you towards calico rather than cilium?

 CC: @paulgmiller, @chasewilson

marchermans:
> @marchermans We do not support the calico network policy, is there anything driving you towards calico rather than cilium?
> 
> CC: @paulgmiller, @chasewilson

Ehm, several reasons, but right now: We have an enterprise license for it, and it is an officially supported infrastructure within AKS.

Secondly, why did you close #380, as this clearly indicates that there is need/demand for this ability.

marchermans:
Also to me this seems rather a weird concept to support one CNI while not the other....... CNIs are standardized implementations of an API, are they not? Especially if the CNI it self is supported on 

Why would karpenter care infra-structurally what CNI is actually installed, once `networkPlugin: none` is chosen? Specifically I am trying to understand if there is something I can PR, and potentially how, and where existing documentation exists, on this area of Karpenters implementation.

Issue #224, opened 2024-03-20
Identities management becomes increasingly complex and error-prone; should be revisited/refactored for ease of continuing development


comtalyst:
### Tell us about your request

The area of identities management, as well as its interface, should be revisited to prevent impending, if not existing, engineering mistakes and security concerns of the codebase.

### Tell us about the problem you're trying to solve. What are you trying to do, and why is it hard?

As the project becomes more mature, the existing assumptions on the uses of identity have evolved, and will continue to do so.
For example, kubelet identity (i.e., user assigned identity) is being used on everything Karpenter needs(?), while its original expectation is just to assign it to the provisioning nodes. This fact should, at least, be clearly reflected in the code (e.g., through proper naming) to prevent unexpected misuses in the case where changes will be introduced in this area, as well as to prevent security breaches.

### Are you currently working around this issue?

No

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

comtalyst:
Working on it

PR #366, opened 2024-05-22
refactor: rework aged authentication config interface

comtalyst:
Fixes # 

**Description**
Part of #224 
This involves refactoring of auth config, removing some redundant code (e.g., in favor of `DefaultAzureCredentials`, remove some unused variables), and renaming `ARM_USER_ASSIGNED_IDENTITY_ID` to `KUBELET_IDENTITY_CLIENT_ID` for accuracy to its purpose/use. See core motivations in the issue.

* Unit tests
* E2E - Utilization (through `make az-e2etests`)
* E2E for NAP (using managed identity)

- [ ] Yes, issue opened: # 
- [X] No

**Release Note**

```release-note
- Deprecate `ARM_USE_CREDENTIAL_FROM_ENVIRONMENT` and `ARM_USE_MANAGED_IDENTITY_EXTENSION` in favor of track2-based authentication: https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azidentity/README.md#defaultazurecredential
- Rename `ARM_USER_ASSIGNED_IDENTITY_ID` to `KUBELET_IDENTITY_CLIENT_ID` for accuracy to its purpose/use
- Remove unused environment variables `AZURE_CLUSTER_NAME` and `ARM_VM_TYPE`
```


comtalyst (commented):
/test

pkg/auth/config.go
```diff
@@ -165,22 +101,35 @@ func (cfg *Config) GetAzureClientConfig(authorizer autorest.Authorizer, env *azu
...
+	cfg.ResourceGroup = strings.TrimSpace(os.Getenv("ARM_RESOURCE_GROUP"))
+	cfg.TenantID = strings.TrimSpace(os.Getenv("ARM_TENANT_ID"))
+	cfg.SubscriptionID = strings.TrimSpace(os.Getenv("ARM_SUBSCRIPTION_ID"))
+	cfg.VMType = strings.ToLower(os.Getenv("ARM_VM_TYPE"))
+	cfg.ClusterName = strings.TrimSpace(os.Getenv("AZURE_CLUSTER_NAME"))
+	cfg.NodeResourceGroup = strings.TrimSpace(os.Getenv("AZURE_NODE_RESOURCE_GROUP"))
+	cfg.AuthMethod = strings.TrimSpace(os.Getenv("ARM_AUTH_METHOD"))
+	cfg.KubeletIdentityClientID = strings.TrimSpace(os.Getenv("ARM_KUBELET_IDENTITY_CLIENT_ID"))
+
+	return nil
```
Bryce-Soghigian, line 116: if we are refactoring this right now can we also refactor them into the options struct?

comtalyst: (discussed offline) that is being considered in the next PR since it would requires more refactor apart from the interface.

comtalyst (commented):
/test

comtalyst (commented):
/test

pkg/auth/cred.go
```diff
@@ -30,26 +30,19 @@ func NewCredential(cfg *Config) (azcore.TokenCredential, error) {
 		return nil, fmt.Errorf("failed to create credential, nil config provided")
 	}
 
-	if cfg.UseCredentialFromEnvironment {
+	if cfg.ArmAuthMethod == authMethodWorkloadIdentity {
 		klog.V(2).Infoln("cred: using workload identity for new credential")
 		return azidentity.NewDefaultAzureCredential(nil)
```
tallaxes, line 35: "using workload identity" - this is not quite what `NewDefaultAzureCredential` does, it will [attempt a variaty of mechanisms](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azidentity/README.md#defaultazurecredential). If we are Ok with this - which may well be the case - let's just say it, and we don't need to special case Managed Identity below. If we want to constrain this to Workload Identity - needs to use `NewWorkloadIdentityCredential`?

tallaxes (commented):
Before commenting on the appropriate configuration surface, and corresponding implementation, I would like to see a clear description of which authentication methods we want to support - and if we have a good reason to constrain them. Is there anything that prevents us from just using [`DefaultAzureCredential`](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azidentity/README.md#defaultazurecredential), which attempts a variety of authentication mechanisms (including the ones we need to support), using its existing/documented configuration surface? Would be good to avoid custom switches (like `ARM_AUTH_METHOD`) if the "standard"/OOB works well.

pkg/auth/cred.go
comtalyst: There is a discussion related to this with @charliedmcb in another PR. I think our understanding match yours, with the new knowledge here is that `NewDefaultAzureCredential()` might cover managed identity as well.

So far, I think that we should scope it down to workload identity for simplicity. This means we will support only two auth methods for now: workload identity and system assigned MSI.
In the current implementation, there might be "shortcuts" where `ARM_AUTH_METHOD` is workload identity, but they ended up using something else that's compatible with this implementation. 

So, I think I will make changes to `NewWorkloadIdentityCredential`, if possible.

comtalyst: I'm not sure what you mean by below; could you provide more example? 
>  Would be good to avoid custom switches (like ARM_AUTH_METHOD) if the "standard"/OOB works well.

tallaxes: What I mean is that `azidentity` already provides (via environment variables) an existing configuration surface, and (via `DefaultAzureCredential`) a reasonable sequence of auth methods to try, including the two that we need to support. We should have good reasons to a) constrain the auth methods, and b) introduce additional/custom configuration surface, such as the proposed `ARM_AUTH_METHOD` setting. (Which we may need to remove later if decide to use `NewDefaultAzureCredential`? Removal is always hard ...) There may be valid reasons for both (improved security posture, the built-in sequence not meeting our order requirements, established pattern elsewhere, etc.) - but would be good to have clarity on which ones - if any - apply here.

tallaxes: One more option to consider - if we decide there is a good reason to support only workload identity and MSI - is to just try them in order (similar to how `DefaultAzureCredential` does it), eliminating the need (and capability) for the explicit choice via `ARM_AUTH_METHOD`.

comtalyst: >  b) introduce additional/custom configuration surface, such as the proposed ARM_AUTH_METHOD setting

`ARM_AUTH_METHOD` is in fact a replacement for the awkward combination of `ARM_USE_CREDENTIAL_FROM_ENVIRONMENT` and `ARM_USE_MANAGED_IDENTITY_EXTENSION` rather than an addition. 
But, if we use `DefaultAzureCredential`, maybe that can be implied from the credentials and the user don't have to explicitly specify `ARM_AUTH_METHOD`, eliminating its needs.

comtalyst: > a) constrain the auth methods

I find that we never(?) have an expectation of the user using other auth methods than these two in the first place. Even if we allow it, that allow means additional responsibility/accountability (outside of this relatively simple implementation, I mean) on our side. 
So, unless there are good demands for other auth methods that is possible through `DefaultAzureCredential`, I don't think we should give the opportunity for the user to mess up things that we not yet want to invest into.

But at least, our code framework should be able to support that quickly once we decided to, but not exposing it to the user.

comtalyst: > Removal is always hard ...
Removing ARM_AUTH_METHOD to eliminate user-side complexity, given `DefaultAzureCredential` can retrieve that implicitly, is an annoyance on our side to change the code, but I think it is ok.

On the other hand, if the user secretly rely on some auth methods that `DefaultAzureCredential` supports, but then we decide to remove it (e.g., security), then that's going to affect them.
Technically, if we stop using it and switch to `NewWorkloadIdentityCredential`, it is like that. But given we are in the early stage, I think it is better to do it now.

comtalyst: Keep in mind that whatever in `DefaultAzureCredential` today might changes in the future under our radar, given its definition is open to new methods rather than always specific to a few already.

tallaxes: My preference would be for the simplest secure approach that supports the requirements, which seems to be `DefaultAzureCredential`. (Any changes there will likely require additional configuration, and are very unlikely to break existing auth scenarios. In NAP we control the configuration tightly enough for this not to be a concern, for self-hosted breaking changes, if any, would/should surface via release notes.) Second best would be to try a more controlled sequence of auth methods we need to support - unless you see a problem with this, and we have to specify the method explicitly via configuration.

Bryce-Soghigian: Agreed DefaultAzureCredential already encapsulates what we need. I don't see them breaking it either as its wildly used in many projects for wrapping many types of auth based on some quick github code searching.

comtalyst: > Any changes there will likely require additional configuration, and are very unlikely to break existing auth scenarios

> I don't see them breaking it either

I doubt it will break too, plus the tests will catch it anyway in case we neglect release notes. 
But my point is it might have new auth methods we are not aware of, which means more liability as I mentioned. This will not happen in the more controlled approach (i.e., `NewWorkloadIdentityCredential` with `ARM_AUTH_METHOD`)

>  In NAP we control the configuration tightly enough for this not to be a concern

Yep all these concerns are for self-hosted. In NAP we have static integration interface.

comtalyst: But maybe I am overthinking again. We could try going with implicit auth method and use `DefaultAzureCredential` and see how it goes.

The only problem, if it is at all, is `autorest_auth.go` is also using it. But I just haven't take a look at that deeply yet. If we are moving in this direction then I will see the possibility.

comtalyst: Circling back to this. 
- The concern of unwanted auth method is valid, from an example in Cluster Autoscaler. "Requirement" to continue the support is an engineering burden.
- Although, this `DefaultAzureCredential` is something standardized. If we are going to rely on this standardization (cloud-provider-azure?) for a next while, the simplicity we get might worth it, and the deprecation (if necessary) might not be too difficult.

TL;DR let's just use `DefaultAzureCredential`.

pkg/operator/operator.go
```diff
@@ -169,7 +170,7 @@ func getCABundle(restConfig *rest.Config) (*string, error) {
 }
 
 func getVnetGUID(cfg *auth.Config, subnetID string) (string, error) {
-	creds, err := auth.NewCredential(cfg)
+	creds, err := azidentity.NewDefaultAzureCredential(nil)
```
comtalyst, line 173: An alternative is to keep the abstraction, but just return `azidentity.NewDefaultAzureCredential(nil)`. But there are only 2 uses (another in azure_client.go) at the moment, so I don't think it is worth it.

pkg/providers/instance/skuclient/skuclient.go
```diff
@@ -48,11 +50,14 @@ func (sc *skuClient) updateInstance() {
...
-	authorizer, err := auth.NewAuthorizer(sc.cfg, sc.env)
+	// Create a new authorizer for the sku client
+	// TODO (charliedmcb): need to get track 2 support for the skewer API
+	cred, err := azidentity.NewDefaultAzureCredential(nil)
 	if err != nil {
-		klog.V(5).Infof("Error creating authorizer for sku client: %s", err)
+		klog.V(5).Infof("Error creating authorizer for sku client: default cred: %s", err)
 		return
 	}
+	authorizer := azidext.NewTokenCredentialAdapter(cred, []string{azidext.DefaultManagementScope})
```
comtalyst, line 60: An alternative is to keep the abstraction, that returns authorizer. But this is the only use of it at the moment, so I don't think it is worth it. Could refactor later.

comtalyst (commented):
/test

comtalyst (commented):
/test

pkg/auth/config.go
```diff
@@ -165,22 +90,30 @@ func (cfg *Config) GetAzureClientConfig(authorizer autorest.Authorizer, env *azu
...
-	cfg.NodeResourceGroup = strings.TrimSpace(cfg.NodeResourceGroup)
+func (cfg *Config) Build() error {
+	// May require more than this behind the scenes: https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azidentity/README.md#defaultazurecredential
+	cfg.Cloud = strings.TrimSpace(os.Getenv("ARM_CLOUD"))
+	cfg.Location = strings.TrimSpace(os.Getenv("LOCATION"))
+	cfg.ResourceGroup = strings.TrimSpace(os.Getenv("ARM_RESOURCE_GROUP"))
+	cfg.TenantID = strings.TrimSpace(os.Getenv("ARM_TENANT_ID"))
+	cfg.SubscriptionID = strings.TrimSpace(os.Getenv("ARM_SUBSCRIPTION_ID"))
+	cfg.VMType = strings.ToLower(os.Getenv("ARM_VM_TYPE"))
+	cfg.ClusterName = strings.TrimSpace(os.Getenv("AZURE_CLUSTER_NAME"))
```
Bryce-Soghigian, line 101: How is this used?

pkg/providers/instance/azure_client.go
```diff
@@ -108,11 +109,11 @@ func CreateAZClient(ctx context.Context, cfg *auth.Config) (*AZClient, error) {
 }
 
 func NewAZClient(ctx context.Context, cfg *auth.Config, env *azure.Environment) (*AZClient, error) {
-	cred, err := auth.NewCredential(cfg)
+	defaultAzureCred, err := azidentity.NewDefaultAzureCredential(nil)
```
Bryce-Soghigian, line 112: why change the name cred should suffice

pkg/providers/instance/skuclient/skuclient.go
```diff
@@ -48,11 +50,14 @@ func (sc *skuClient) updateInstance() {
 	sc.mu.RLock()
 	defer sc.mu.RUnlock()
 
-	authorizer, err := auth.NewAuthorizer(sc.cfg, sc.env)
+	// Create a new authorizer for the sku client
+	// TODO (charliedmcb): need to get track 2 support for the skewer API
+	cred, err := azidentity.NewDefaultAzureCredential(nil)
```
Bryce-Soghigian, line 55: How did you test this?

Bryce-Soghigian (approved):
Run e2e tests please before merging

pkg/auth/config.go
comtalyst: bootstrap variables, mostly

pkg/providers/instance/skuclient/skuclient.go
comtalyst: It is the code of `auth.NewAuthorizer(sc.cfg, sc.env)`, with changes to use `azidentity.NewDefaultAzureCredential(nil)`.
E2E for each auth method should be sufficient.

tallaxes (approved):
Overall looks good, noted some potential additional cleanup, but approving in case it is best handled separately.

karpenter-values-template.yaml
```diff
@@ -34,12 +34,7 @@ controller:
       value: ${AZURE_SUBSCRIPTION_ID}
     - name: LOCATION
       value: ${AZURE_LOCATION}
-    # settings for managed workload identity
-    - name: ARM_USE_CREDENTIAL_FROM_ENVIRONMENT
-      value: "true"
-    - name: ARM_USE_MANAGED_IDENTITY_EXTENSION
-      value: "false"
-    - name: ARM_USER_ASSIGNED_IDENTITY_ID
+    - name: ARM_KUBELET_IDENTITY_CLIENT_ID
```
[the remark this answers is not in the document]
comtalyst: Not really, let's remove it now.

pkg/auth/config.go
```diff
@@ -61,97 +54,29 @@ type Config struct {
 	ResourceGroup  string `json:"resourceGroup" yaml:"resourceGroup"`
 	VMType         string `json:"vmType" yaml:"vmType"`
```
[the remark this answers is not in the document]
comtalyst: Looks like it, let's clean it up now.

pkg/auth/config.go
```diff
@@ -61,97 +54,29 @@ type Config struct {
...
-	cfg.AADClientID = os.Getenv("ARM_CLIENT_ID")
-	cfg.AADClientSecret = os.Getenv("ARM_CLIENT_SECRET")
-	cfg.VMType = strings.ToLower(os.Getenv("ARM_VM_TYPE"))
-	cfg.AADClientCertPath = os.Getenv("ARM_CLIENT_CERT_PATH")
-	cfg.AADClientCertPassword = os.Getenv("ARM_CLIENT_CERT_PASSWORD")
-	cfg.ClusterName = os.Getenv("AZURE_CLUSTER_NAME")
-	cfg.NodeResourceGroup = os.Getenv("AZURE_NODE_RESOURCE_GROUP")
-}
+	// Managed identity for Kubelet (not to be confused with Azure cloud authorization)
+	KubeletIdentityClientID string `json:"kubeletIdentityClientID" yaml:"kubeletIdentityClientID"`
```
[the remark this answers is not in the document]
comtalyst: No specific reason, but could be clean up along with other options later.

tallaxes (approved):

PR #390, opened 2024-06-07
refactor: rework options flow to have a more consistent pattern

comtalyst:
Fixes # 

**Description**

*

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


pkg/operator/operator.go
```diff
@@ -67,68 +63,62 @@ type Operator struct {
...
 		azClient.ImageVersionsClient,
-		azConfig.Location,
 	)
 	imageResolver := imagefamily.New(
 		operator.GetClient(),
 		imageProvider,
 	)
 	launchTemplateProvider := launchtemplate.NewProvider(
 		ctx,
+		opts,
```
comtalyst, line 98: This should be the highlights worth discussing.

Prior to this, in addition to passing the options (and `azConfig`) directly that you can see in the changes preview, some of the providers did call `options.FromContext(ctx)` again. I find this inconsistent and would prefer to stick to one side—either pass all required parameters here, or just pass in `ctx`/`opts`.

The current code chooses passing in `opts`. This will make the providers depends on the knowledge of options (no change), while sparing operator from having to know which provider needs what. This is particularly helpful in `launchTemplateProvider`, which is have a lot of demands for fields in options.

With that said, each provider have different demands in options fields, but ideally I want to keep the dependencies/pattern consistent for all providers.

comtalyst: Alternatively, we could stay explicit on the interface and pass all variables that we currently getting from `options.FromContext(ctx)` through here instead. The interface will be be big for high-demanding providers, but we will be able to cut their dependence on `options`.

comtalyst: Another alternative is to add some layering in `options` to categorize them by uses, appearing at "sub-structs".
But this will force `options` to contain knowledge about the exact uses of their fields, which is going to add additional complexity, given some fields are shared with multiple uses.

An extension to that is to have additional layer between `options` (which will mostly take envs in raw) and this layer to map those to its provider uses. But that would be no difference that just creating structs here, which will only be better than explicit passing by extensibility and small reusability, at the cost of additional complexity.

pkg/operator/options/options.go
```diff
@@ -57,66 +53,81 @@ func (s *nodeIdentitiesValue) String() string { return strings.Join(*s, ",") }
...
+// So, if one day we want to support dynamic configuration updates, consider creating the a new instance of Options and reinitialize the operator/controllers.
+//
+// If some fields need to be updated/refreshed and have their own (valid) cascading update procedure, consider moving it away to prevent confusion with the above assumption.
+// At that point, we should consider having a more clear distinction between user options and global variables.
 type Options struct {
-	ClusterName                    string
-	ClusterEndpoint                string // => APIServerName in bootstrap, except needs to be w/o https/port
-	VMMemoryOverheadPercent        float64
-	ClusterID                      string
+	// Target cluster information; might be use for both bootstrapping and ARM authentications
```
comtalyst, line 65: The idea behind this is I try to categorize this by user-facing behavior (e.g., target cluster for the operations to occur, what will be on provisioned nodes) rather than implementation details.

comtalyst: Some should be renamed. But let's put that in a different PR

pkg/operator/options/options_defaulting.go
```diff
@@ -0,0 +1,113 @@
...
+	"github.com/Azure/karpenter-provider-azure/pkg/utils"
+
+	armopts "github.com/Azure/karpenter-provider-azure/pkg/utils/opts"
+)
+
+// Default sets the default values for the options, but only for those too complicated to be in env default (e.g., depends on other envs)
+func (o *Options) Default(ctx context.Context) error {
+	var err error
+
+	var authManager *auth.AuthManager
```
comtalyst, line 37: This is a flaw in the implementation, as `authManager` is created in both here and `operator`. Given core, there is nothing much we can do(?) if we want to follow this pattern.

pkg/operator/options/options.go
```diff
@@ -57,66 +53,81 @@ func (s *nodeIdentitiesValue) String() string { return strings.Join(*s, ",") }
...
+	Cloud           string
+	Location        string
+	TenantID        string
+	SubscriptionID  string
+	ResourceGroup   string
+	ClusterName     string
+	ClusterEndpoint string // => APIServerName in bootstrap, except needs to be w/o https/port
+	ClusterID       string
+	APIServerName   string
```
comtalyst, line 74: This and `VnetGUID` (that I just added) are the options that are not available to be configurable directly by the customer, but directly depends on other options through defaulting. Open to suggestions if this should be elsewhere.

pkg/controllers/nodeclaim/inplaceupdate/controller.go
```diff
@@ -44,13 +44,17 @@ import (
 type Controller struct {
 	kubeClient       client.Client
 	instanceProvider *instance.Provider
+
+	// Will be used to calculate the goal state
+	opts *options.Options
```
comtalyst, line 49: Another instance of "not sure if options is overkill", see the comments in `operator.go` for similar discussions.

pkg/operator/operator.go
comtalyst: But all these options are not really making use of `options.FromContext(ctx)` pattern established by karpenter-core and AWS provider. If we want to keep that, another options is to just pass in `ctx` to all places and let `options.FromContext(ctx)` retrieve the options whenever it is needed. This will assume that all instances here are singletons or at least have no differences between each, which is true(?), for now.

karpenter-values-template.yaml
```diff
@@ -34,15 +34,12 @@ controller:
...
-      value: "true"
-    - name: ARM_USE_MANAGED_IDENTITY_EXTENSION
-      value: "false"
-    - name: ARM_USER_ASSIGNED_IDENTITY_ID
+    - name: ARM_KUBELET_IDENTITY_CLIENT_ID
       value: ""
     - name: AZURE_NODE_RESOURCE_GROUP
       value: ${AZURE_RESOURCE_GROUP_MC}
+    - name: ARM_AUTH_METHOD
+      value: "workload-identity"
```
charliedmcb, line 42: nit: the reason its called "from environment" is because its actually using a more generic/general auth option that finds the creds in the environment. This is the flow that works for workload-identity. However, I think the naming should probably keep the more generic term since that is what its actually doing. Unless you have more tied to this.

pkg/auth/authmanager.go
```diff
@@ -0,0 +1,111 @@
...
+	if am.authMethod == AuthMethodWorkloadIdentity {
+		klog.V(2).Infoln("auth: using workload identity for new authorizer")
+		cred, err := azidentity.NewDefaultAzureCredential(nil)
+		if err != nil {
+			return nil, fmt.Errorf("default cred: %w", err)
+		}
+		return azidext.NewTokenCredentialAdapter(cred, []string{azidext.DefaultManagementScope}), nil
+	}
+
+	if am.authMethod == AuthMethodWorkloadIdentity {
```
charliedmcb, line 89: Are both of these if statements supposed to be `am.authMethod == AuthMethodWorkloadIdentity`

Seems like a potential reg flag on bool logic.

charliedmcb (commented):
Just did a quick scan. I like the direction of streamlining options. However, would recommend that we keep the naming generic since the flow doesn't necessarily have to just be used for workload identities.

karpenter-values-template.yaml
comtalyst: So, by theory, there could be "from environment" auth option that is not workload identity, right?
If that's the case then I think keeping "from-environment" would be a better choice, given we doesn't seem to have ties to workload identity (right?).

pkg/auth/authmanager.go
comtalyst: That's a catch. The latter should be `AuthMethodSysMSI`.
But generally I think this should be an enum-like rather than a bool, given none of the 2 are "inferior" to others to be undermined using `false`, and it would open up the possibility of new methods too.

pkg/auth/authmanager.go
```diff
@@ -0,0 +1,111 @@
...
+		klog.V(2).Infoln("cred: using system assigned MSI for new credential")
+		msiCred, err := azidentity.NewManagedIdentityCredential(nil)
+		if err != nil {
+			return nil, err
+		}
+		return msiCred, nil
+	}
+
+	return nil, fmt.Errorf("cred: unsupported auth method: %s", am.authMethod)
+}
```
Bryce-Soghigian, line 76: Does this belong in the shared SDK? For my token fix i imagine other ppl may want the same fix. Uing azure sdk for go extensions may be a good idea for our auth so we can share that logic too. Should we move this there?

pkg/controllers/controllers.go
```diff
@@ -26,15 +26,18 @@ import (
...
 )
 
 func NewControllers(ctx context.Context, kubeClient client.Client, cloudProvider *cloudprovider.CloudProvider, instanceProvider *instance.Provider) []controller.Controller {
 	logging.FromContext(ctx).With("version", project.Version).Debugf("discovered version")
+	opts := options.FromContext(ctx)
+
 	controllers := []controller.Controller{
 		nodeclaimgarbagecollection.NewController(kubeClient, cloudProvider),
-		inplaceupdate.NewController(kubeClient, instanceProvider),
+		inplaceupdate.NewController(kubeClient, instanceProvider, opts),
```
Bryce-Soghigian, line 40: why not just options.FromContext(ctx)? or for that matter why not just pass in ctx so that future updates to the controller can access everything from the context object? Are we refactoring options out of ctx?

pkg/operator/options/options_defaulting.go
```diff
@@ -0,0 +1,108 @@
...
+
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/network/armnetwork"
+	"github.com/Azure/karpenter-provider-azure/pkg/auth"
+	"github.com/Azure/karpenter-provider-azure/pkg/utils"
+
+	armopts "github.com/Azure/karpenter-provider-azure/pkg/utils/opts"
+)
+
+// Default sets the default values for the options, but only for those too complicated to be in env default (e.g., depends on other envs)
+func (o *Options) Default(ctx context.Context) error {
```
Bryce-Soghigian, line 34: If we are going to go this far, can't we just resolve a lot of the configuration from having clustername and rg? 

Why make users pass in things like "NetworkDataplane" or "NetworkPluginMode"? Why not just use the AKS SDK to dynamically resolve that configuration for them?

pkg/operator/options/options_defaulting.go
```diff
@@ -0,0 +1,108 @@
...
+	} else if o.ArmAuthMethod == auth.AuthMethodSysMSI {
+		authManager = auth.NewAuthManagerSystemAssignedMSI(o.Location)
+	}
+
+	if o.APIServerName, err = getAPIServerName(o.ClusterEndpoint); err != nil {
+		return fmt.Errorf("failed to get APIServerName: %w", err)
+	}
+
+	if o.ClusterID, err = getAKSClusterID(o.APIServerName); err != nil {
+		return fmt.Errorf("failed to get ClusterID: %w", err)
```
Bryce-Soghigian, line 49: should we make these errors typed? Makes this part easier to test, rachel left this feedback in another of my prs but you touch this part a lot more so lmk what you think about changing all of these to typed errors

pkg/operator/options/options_defaulting.go
```diff
@@ -0,0 +1,108 @@
...
+// The logic comes from AgentBaker and other places, originally from aks-engine
+// with the additional assumption of DNS prefix being the first 33 chars of FQDN
+func getAKSClusterID(apiServerFQDN string) (string, error) {
+	dnsPrefix := apiServerFQDN[:33]
+	h := fnv.New64a()
+	h.Write([]byte(dnsPrefix))
+	r := rand.New(rand.NewSource(int64(h.Sum64()))) //nolint:gosec
+	return fmt.Sprintf("%08d", r.Uint32())[:8], nil
+}
+
```
Bryce-Soghigian, line 75: General question, does this belong in agentbaker? We stole it from agentbaker will self contained have this in the pkg/agent/common or wahtever we can just import rather than redefining it?

pkg/operator/options/options_defaulting.go
```diff
@@ -0,0 +1,108 @@
...
+	return *vnet.Properties.ResourceGUID, nil
+}
+
+func contains(slice []string, target string) bool {
+	for _, element := range slice {
+		if target == element {
+			return true
+		}
+	}
+	return false
```
Bryce-Soghigian, line 107: kubernetes helpers for slices should have this already

pkg/operator/options/options_validation.go
```diff
@@ -29,22 +29,45 @@ func (o Options) Validate() error {
...
+	}
+	if o.ClusterEndpoint == "" {
+		return fmt.Errorf("missing field, cluster-endpoint")
+	}
+	if o.ClusterName == "" {
+		return fmt.Errorf("missing field, cluster-name")
 	}
+
+	if o.NodeResourceGroup == "" {
+		return fmt.Errorf("missing field, node-resource-group")
```
Bryce-Soghigian, line 55: same comment about shared error type. we really only care that 

1. its missing 
2. what is missing

pkg/providers/imagefamily/bootstrap/aksbootstrap.go
```diff
@@ -40,7 +40,7 @@ type AKS struct {
 	Arch                           string
 	TenantID                       string
 	SubscriptionID                 string
-	UserAssignedIdentityID         string
+	KubeletIdentityClientID        string
```
Bryce-Soghigian, line 43: great name change!

Bryce-Soghigian: Do we worry about these values being different from those in azure json? So people may grep for UserAssignedIdentityID since its called that everywhere else?

pkg/providers/imagefamily/image.go
```diff
@@ -51,11 +52,11 @@ const (
 	imageIDFormat = "/CommunityGalleries/%s/images/%s/versions/%s"
 )
 
-func NewProvider(kubernetesInterface kubernetes.Interface, kubernetesVersionCache *cache.Cache, versionsClient CommunityGalleryImageVersionsAPI, location string) *Provider {
+func NewProvider(opts *options.Options, kubernetesInterface kubernetes.Interface, kubernetesVersionCache *cache.Cache, versionsClient CommunityGalleryImageVersionsAPI) *Provider {
 	return &Provider{
 		kubernetesVersionCache: kubernetesVersionCache,
 		imageCache:             cache.New(imageExpirationInterval, imageCacheCleaningInterval),
-		location:               location,
+		location:               opts.Location,
```
Bryce-Soghigian, line 59: why not just put options in the Provider struct itself and we can access location from that? I also will need some values from options in some of these providers in this pr https://github.com/Azure/karpenter-provider-azure/pull/365/files so that pattern would make access a bit easier.

pkg/providers/instancetype/instancetype.go
```diff
@@ -298,11 +295,11 @@ func memoryMiB(sku *skewer.SKU) int64 {
 	return int64(memoryGiB(sku) * 1024)
 }
 
-func memory(ctx context.Context, sku *skewer.SKU) *resource.Quantity {
+func memory(sku *skewer.SKU, vmMemoryOverheadPercent float64) *resource.Quantity {
 	memory := resources.Quantity(fmt.Sprintf("%dGi", int64(memoryGiB(sku))))
 	// Account for VM overhead in calculation
 	memory.Sub(resource.MustParse(fmt.Sprintf("%dMi", int64(math.Ceil(
-		float64(memory.Value())*options.FromContext(ctx).VMMemoryOverheadPercent/1024/1024)))))
```
Bryce-Soghigian, line 305: does this come from core?

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -178,15 +164,15 @@ func mergeTags(tags ...map[string]string) (result map[string]*string) {
 	})
 }
 
-func (p *Provider) getVnetInfoLabels(ctx context.Context, _ *v1alpha2.AKSNodeClass) (map[string]string, error) {
-	// TODO(bsoghigian): this should be refactored to lo.Ternary(nodeClass.Spec.VnetSubnetID != nil, lo.FromPtr(nodeClass.Spec.VnetSubnetID), os.Getenv("AZURE_SUBNET_ID")) when we add VnetSubnetID to the nodeclass
-	vnetSubnetComponents, err := utils.GetVnetSubnetIDComponents(options.FromContext(ctx).SubnetID)
+func (p *Provider) getVnetInfoLabels(_ *v1alpha2.AKSNodeClass) (map[string]string, error) {
+	// TODO(bsoghigian): this should be refactored to lo.Ternary(nodeClass.Spec.SubnetID != nil, lo.FromPtr(nodeClass.Spec.SubnetID), p.opts.SubnetID) when we add VnetSubnetID to the nodeclass
```
Bryce-Soghigian, line 168: are you sure we should differ from the AKS Agentpool api in what we call SubnetID? We discussed this somewhere on the other PR and i thought we agreed on calling it VnetSubnetID

Bryce-Soghigian: I named it SubnetID from the start but got feedback we should change it to be VNETSubnetID

pkg/providers/loadbalancer/loadbalancer.go
```diff
@@ -69,11 +70,11 @@ type BackendAddressPools struct {
 }
 
 // NewProvider creates a new LoadBalancer provider
-func NewProvider(loadBalancersAPI LoadBalancersAPI, cache *cache.Cache, resourceGroup string) *Provider {
+func NewProvider(opts *options.Options, loadBalancersAPI LoadBalancersAPI, cache *cache.Cache) *Provider {
 	return &Provider{
 		loadBalancersAPI: loadBalancersAPI,
 		cache:            cache,
-		resourceGroup:    resourceGroup,
+		resourceGroup:    opts.NodeResourceGroup,
```
Bryce-Soghigian, line 77: 😍

pkg/providers/pricing/pricing.go
```diff
@@ -64,7 +65,9 @@ func NewAPI() client.PricingAPI {
 	return client.New()
 }
 
-func NewProvider(ctx context.Context, pricing client.PricingAPI, region string, startAsync <-chan struct{}) *Provider {
+func NewProvider(ctx context.Context, opts *options.Options, pricing client.PricingAPI, startAsync <-chan struct{}) *Provider {
+	region := opts.Location
```
Bryce-Soghigian, line 69: why not just use opts.Location in the map lookup?

pkg/controllers/controllers.go
comtalyst: (discussed offline)
Although both share the same principle, `options.FromContext(ctx)` seems to be the pattern that is generally discouraged by all opinions we have discussed with, except karpenter-core.
For my own reasoning, this would allow us to separate `options` from `ctx`, which might wanted to be used for other purposes. We could still argue that `options` still contain things that might not be used, but it still allow extensibility and hide the knowledge of "what this component requires" from the creation interface. So I would say it is something in between `ctx` and explicitly passing parameters.

Anyway, it is not a strong opinion.

pkg/operator/options/options_defaulting.go
comtalyst: Not a bad idea. Now that we establish this framework, we could consider simplifying the integration interface and get obvious fields by ourselves. A trade-off for this is Karpenter will have more responsibility of "getting required info of a cluster" which could at least lead to having more traffic at the initialization.

I think the changes for this, if we decided to, should be in a separate PR at least. But we can start discussing here if we have strong ideas.

pkg/operator/options/options_defaulting.go
comtalyst: I think it would be a nice to have, but not a necessity at the moment.

pkg/operator/options/options_defaulting.go
comtalyst: Good point. I will dive deeper into this along with the contract interface rework. Hate

pkg/providers/imagefamily/bootstrap/aksbootstrap.go
comtalyst: (discussed offline)

It is better this way given that the `UserAssignedIdentityID` that is known by `azure.json` is in fact `KubeletIdentityClientID` that is known by the cluster. This confusion is the prime motivation for this entire work.

Currently, I choose to keep this variable name until we feed it to the bootstrap module. There could be even more opportunities around that area to make it even more clear...

comtalyst: Now that we have this knowledge, we should be aware when coming across their uses in those "everywhere else." Maybe we can catch some big bugs.

pkg/providers/imagefamily/image.go
comtalyst: (discussed offline)

I am still hesitate to make these providers depends on `options` entirely. 
Right now it does module-wise, but not the struct, as the only reference to `options` is on creation function (`NewProvider`). Maybe the better name to this is `NewProviderFromOptions`.

But again, not a strong opinion. Wondering what do others think? Should we be more extreme on one side (e.g., using `opts`/`ctx` everywhere)?

pkg/providers/instancetype/instancetype.go
comtalyst: Not core but AWS provider: https://github.com/aws/karpenter-provider-aws/blob/531d7e00ad69bed8fa74de79abdaf8d61b476910/pkg/providers/instancetype/types.go#L224.

comtalyst: But, yes, this will introduce a difference between the two.
In general, I think our changes it will be for the better (at least, intention-wise). But the trade-off is consistency.

In this case, I am not completely sure, but leaning towards keeping current change.
Current changes mean consistency with this pattern within our provider (e.g., providers).
Restoring to original means consistency with the AWS provider. But regardless whether we keep other changes or restore all, inconsistency remains within our provider, unless we become more extreme with the use of `ctx`.

pkg/providers/launchtemplate/launchtemplate.go
comtalyst: I think this change is a mistake from mass text replace. My bad.

comtalyst: But there will be similar changes (not in this PR) that I intended to rename all `SubnetID` to `VnetSubnetID`. We can discuss more later along with the rest of the renames.

pkg/providers/loadbalancer/loadbalancer.go
comtalyst: 😳

pkg/providers/pricing/pricing.go
comtalyst: It is being used in other places (at least line 76/79) as well. But I don't think any choice matters.

theunrepentantgeek:
Closing as no longer relevant given more recent changes.

PR #391, opened 2024-06-07
fix: long running setup for expired auth token track 2 bug

Bryce-Soghigian:
Fixes # 

**Description**
The track 2 sdk has a bug for long living tokens(longer than 24hrs) they will not be refreshed properly. This pr adds a thin wrapper around the GetToken() interface method to set the ExpiresOn timestamp to be earlier and force a token refresh earlier. 

- 2 long running clusters both on NAP and Self Hosted ran without seeing expired token errors
- make presubmit

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


pkg/auth/cred.go
```diff
@@ -17,24 +17,48 @@ limitations under the License.
...
+	}
+}
+
+func (w *TokenWrapper) GetToken(ctx context.Context, options policy.TokenRequestOptions) (azcore.AccessToken, error) {
+	token, err := w.cred.GetToken(ctx, options) 
+	if err != nil {
+		return azcore.AccessToken{}, err
+	}
+	logging.FromContext(ctx).Info("refreshing MDAL Token")		
+	token.ExpiresOn = time.Now().Add(1 * time.Hour)
```
andyzhangx, line 48: I think 2 hours is better since the original expiration time is 2 hours

pkg/auth/cred.go
```diff
@@ -17,19 +17,42 @@ limitations under the License.
...
+	}
+}
+
+func (w *TokenWrapper) GetToken(ctx context.Context, options policy.TokenRequestOptions) (azcore.AccessToken, error) {
+	token, err := w.cred.GetToken(ctx, options)
+	if err != nil {
+		return azcore.AccessToken{}, err
+	}
+	logging.FromContext(ctx).Info("refreshing MDAL Token")
+	token.ExpiresOn = time.Now().Add(2 * time.Hour)
```
zarvd, line 47: Do we need to check the original `ExpiresOn` in case it’s already less than 2 hours?

Bryce-Soghigian:
@andyzhangx yes it seems to be working for over a week for me

pkg/providers/instance/azure_client.go
```diff
@@ -112,7 +112,7 @@ func NewAZClient(ctx context.Context, cfg *auth.Config, env *azure.Environment)
 	if err != nil {
 		return nil, err
 	}
-
+	cred = auth.NewTokenWrapper(cred)
```
comtalyst, line 115: I suppose these are _generally_ temporary until things are fixed on the other side. Could you add some comments about that?

pkg/auth/cred.go
comtalyst, line 47: My understanding is: the bug was `ExpiresOn` being set to be later than its real TTL, right?

pkg/providers/instance/azure_client.go
Bryce-Soghigian: Ideally we could remove this code entirely so I have it separated. Also we need it to be generic so any team can pass in a token without the same azure config logic we have going on. 

The plan is for some other teams to yoink the code so coupling with azureconfig is not ideal. Instead we should integrate the wrapper with just the token so that anyone with an auth token can use the code.

pkg/auth/cred.go
```diff
@@ -17,19 +17,42 @@ limitations under the License.
...
+		cred: cred,
+	}
+}
+
+func (w *TokenWrapper) GetToken(ctx context.Context, options policy.TokenRequestOptions) (azcore.AccessToken, error) {
+	token, err := w.cred.GetToken(ctx, options)
+	if err != nil {
+		return azcore.AccessToken{}, err
+	}
+	logging.FromContext(ctx).Info("refreshing MDAL Token")
```
tallaxes, line 46: Not sure how frequently this is called, but logging as `Info` seems excessive? I would also suggest something like "adjusting token expiration", this does not actually refresh it, does it?

pkg/auth/cred.go
tallaxes: I think that would be prudent

tallaxes: A more detailed description of the issue would be useful; would also help understand (and review) the mechanism of the fix.

zarvd: IIRC the underlying token credential attempts to [fetch from the cache first](https://github.com/Azure/azure-sdk-for-go/blob/93e07a3a4ef307a418a8090595aa2fb88d227083/sdk/azidentity/confidential_client.go#L104). It seems like it could return a short-lived token?

pkg/auth/cred.go
```diff
@@ -17,19 +17,42 @@ limitations under the License.
...
+	"time"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
 	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
 	"k8s.io/klog/v2"
+	"knative.dev/pkg/logging"
 )
 
+type TokenWrapper struct {
```
bcho, line 31: nit: can we document this token wrapper usage?

pkg/auth/cred.go
bcho: agree we should remove this log line as it's a hot path for refreshing. We probably should only log on error and in trace level for non-error

pkg/auth/cred.go
bcho: yep this value should be set with the minimum value of token expires on and the 2hr we want to use here.

bcho: I think the bug is kind of related to the controller missed the refresh action due to unknown reasons (one possible cause could be clock skew). And for some reason, the controller cached the old token without really refreshing it (i guess this part of the [code](https://github.com/Azure/azure-sdk-for-go/blob/76e5495a4d1c7ed72e1c1dc973b126cc31cf88e4/sdk/internal/temporal/resource.go#L45-L114) could be potential culprit). The default refresh interval for the MI token from IMDS is around 23 hours, and expires on is set to 5 minutes before 23 hours. The idea of this fix is to reduce the token refresh interval to make it refresh often / earlier to address the potential clock skew.

Unfortunately, the original bug report doesn't come with some timestamp info about the original token expires on value and the actual system clock time etc, it's a little bit difficult to fully root cause here.

related discussion can be found from here: https://github.com/hashicorp/terraform-provider-azurerm/issues/20834

pkg/auth/cred.go
bcho: Also i would suggest rename it to a meaningful name to reflect the behavior.

bcho: and to avoid exposing this type to the public caller, i would suggest to declare this as a private type

Bryce-Soghigian: done

pkg/auth/cred.go
Bryce-Soghigian: moved to debug

pkg/auth/cred.go
Bryce-Soghigian: Yeah i have a similar level of understanding its not perfect but thats the gist

pkg/auth/cred.go
```diff
@@ -17,19 +17,47 @@ limitations under the License.
...
+
+// expireEarlyTokenCredential is a wrapper around the azcore.TokenCredential that 
+// returns an earlier ExpiresOn timestamp to avoid conditions like clockSkew, or a race 
+// condition during polling. 
+// See: https://github.com/hashicorp/terraform-provider-azurerm/issues/20834 for more details
+type expireEarlyTokenCredential struct {
+	cred azcore.TokenCredential
+}
+
+func NewTokenWrapper(cred azcore.TokenCredential) *expireEarlyTokenCredential {
```
bcho, line 40: nit: return public type for public function

```
func NewTokenWrapper(/* args */) azcore.TokenCredential {

}
```

Bryce-Soghigian, line 40:
```suggestion
func NewTokenWrapper(cred azcore.TokenCredential) azcore.TokenCredential {
```

pkg/auth/cred.go
```diff
@@ -17,19 +17,55 @@ limitations under the License.
...
+	}
+}
+
+func (w *expireEarlyTokenCredential) GetToken(ctx context.Context, options policy.TokenRequestOptions) (azcore.AccessToken, error) {
+	token, err := w.cred.GetToken(ctx, options)
+	if err != nil {
+		return azcore.AccessToken{}, err
+	}
+
+	logging.FromContext(ctx).Debug("adjusting token ExpiresOn")
```
bcho, line 51: nit: put the log after checking before?

bcho (approved):
LGTM

tallaxes (approved):

PR #409, opened 2024-06-18
chore(release): switch release workflow to have tag trigger

charliedmcb:
Fixes # 

**Description**
Working towards a more automated release process. Removing the manual step of triggering with specified tag to automatic trigger based off tagging.

This is aligned behavior with the aws provider:
- https://github.com/aws/karpenter-provider-aws/blob/9e113aaca23bce17e99cd6f14447927e7c89e3d4/.github/workflows/release.yaml

Nothing additional
*

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
NONE
```

### DIFF
--- a/.github/workflows/build-publish-mcr.yml
+++ b/.github/workflows/build-publish-mcr.yml
@@ -3,11 +3,6 @@
 
 name: Build and publish to MCR
 on:
-  workflow_dispatch:
-    inputs:
-      releaseTag:
-        description: 'Release tag to publish, defaults to the latest one'
-        type: string
   push:
     # Only release on supported semantic version tagging e.g. v0.0.1-rc.0
     tags:
--- a/.github/workflows/build-publish-mcr.yml
+++ b/.github/workflows/build-publish-mcr.yml
@@ -8,42 +8,24 @@ on:
       releaseTag:
         description: 'Release tag to publish, defaults to the latest one'
         type: string
+  push:
+    # Only release on supported semantic version tagging e.g. v0.0.1-rc.0
+    tags:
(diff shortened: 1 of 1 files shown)


.github/workflows/build-publish-mcr.yml
```diff
@@ -3,47 +3,24 @@
...
-      releaseTag:
-        description: 'Release tag to publish, defaults to the latest one'
-        type: string
+  push:
+    # Only release on supported semantic version tagging e.g. v0.0.1-rc.0
+    tags:
+      - 'v[0-9]+.[0-9]+.[0-9]+'
+      - 'v[0-9]+.[0-9]+.[0-9]+-rc.[0-9]+'
+      - 'v[0-9]+.[0-9]+.[0-9]+-alpha.[0-9]+'
+      - 'v[0-9]+.[0-9]+.[0-9]+-beta.[0-9]+'
```
Bryce-Soghigian, line 12: i like the enforcement based on tags in this format. Means we can never accidentally screw up our versioning due to alha typo or something like that.

Bryce-Soghigian (approved):
Approved, but does this mean we should lock down who can actually create tags to be scoped directly to our team? Is it already fairly locked down?

.github/workflows/build-publish-mcr.yml
charliedmcb: Agreed.

charliedmcb:
> pproved, but does this mean we should lock down who can actually create tags to be scoped directly to our team? Is it already fairly locked down?

We already have restrictions around version tags, but I will do a sanity check over current permissions. Good callout.

PR #410, opened 2024-06-19
fix: Updated instanceTypeZones logic to handle non-zonal SKUs in zonal regions

rakechill:
Fixes # 

**Description**
I'm working with the Kaito team to integrate with Karpenter Azure Provider. They noticed that they were unable to use some GPU skus due to "insufficient capacity", but were able to confirm that the capacity was available and that this wasn't happening when provisioning nodes directly using AKS APIs.  

It turns out that this was due to a small bug in retrieving availability zones for SKUs. We properly handle the case for non-zonal regions, but do not properly handle non-zonal offerings within zonal regions. I've updated the condition to handle this scenario.

I also have included the `manual-nodeclaim.yaml` that I use to test this. I think it will be helpful for future validations with Kaito, but I can remove it if needed.

* manually by applying the manual nodeclaim
* added another unit test to account for this

- [ ] Yes, issue opened: # 
- [X] No

**Release Note**

```release-note
NONE
```

### DIFF
--- a/examples/v1beta1/nodeclaim/manual-nodeclaim.yaml
+++ b/examples/v1beta1/nodeclaim/manual-nodeclaim.yaml
@@ -1,0 +1,23 @@
+apiVersion: karpenter.sh/v1beta1
+kind: NodeClaim
+metadata:
+  name: manual-kaito
+  labels:
+    karpenter.sh/nodepool: kaito
+  annotations:
+    karpenter.sh/do-not-disrupt: "true"
+spec:
+  nodeClassRef:
+    name: default
+    kind: AKSNodeClass
+  taints:
+    - key: "sku"
+      value: "gpu"
+      effect: NoSchedule
+  requirements:
+    - key: "karpenter.sh/nodepool"
+      operator: In
+      values: ["kaito"]
+    - key:  karpenter.azure.com/sku-name
+      operator: In
+      values: ["Standard_NC12s_v3"]
--- a/pkg/providers/instancetype/instancetypes.go
+++ b/pkg/providers/instancetype/instancetypes.go
@@ -145,7 +145,8 @@ func instanceTypeZones(sku *skewer.SKU, region string) sets.Set[string] {
 	// skewer returns numerical zones, like "1" (as keys in the map);
 	// prefix each zone with "<region>-", to have them match the labels placed on Node (e.g. "westus2-1")
 	// Note this data comes from LocationInfo, then skewer is used to get the SKU info
-	if hasZonalSupport(region) {
+	// If an offering is non-zonal, the availability zones will be empty.
+	if hasZonalSupport(region) && sku.AvailabilityZones(region) != nil {
 		return sets.New(lo.Map(lo.Keys(sku.AvailabilityZones(region)), func(zone string, _ int) string {
 			return fmt.Sprintf("%s-%s", region, zone)
 		})...)


pkg/providers/instancetype/suite_test.go
```diff
@@ -319,7 +319,7 @@ var _ = Describe("InstanceType Provider", func() {
 		})
 	})
 
-	Context("Filtering in InstanceType Provider List", func() {
+	FContext("Filtering in InstanceType Provider List", func() {
```
rakechill, line 322: note: remove prefix "F" for focus when you go back and update this test

pkg/providers/instancetype/suite_test.go
```diff
@@ -1182,6 +1181,18 @@ var _ = Describe("InstanceType Provider", func() {
...
 			ExpectScheduled(ctx, env.Client, pod)
 
+			Expect(azureEnvNonZonal.VirtualMachinesAPI.VirtualMachineCreateOrUpdateBehavior.CalledWithInput.Len()).To(Equal(1))
+			vm := azureEnvNonZonal.VirtualMachinesAPI.VirtualMachineCreateOrUpdateBehavior.CalledWithInput.Pop().VM
+			Expect(vm.Zones).To(BeEmpty())
+		})
+		It("should support provisioning non-zonal instance types in zonal regions", func() {
+			ExpectApplied(ctx, env.Client, nodePool, nodeClass)
+			pod := coretest.UnschedulablePod(coretest.PodOptions{
+				NodeSelector: map[string]string{v1.LabelInstanceTypeStable: "Standard_A0"},
```
rakechill, line 1191: This `Standard_A0` is the only instance type that we currently have generated [here ](https://github.com/Azure/karpenter-provider-azure/blob/main/pkg/fake/zz_generated.sku.eastus.go#L81) that has no zones for `eastus`.
```
LocationInfo: &[]compute.ResourceSkuLocationInfo{{Location: lo.ToPtr("eastus"), Zones: &[]string{}}},
```

However, this VM size is [restricted](https://github.com/Azure/karpenter-provider-azure/blob/main/pkg/providers/instancetype/restricted.go#L23). Thus, it won't be provisioned. 

I tried to add `Standard_NC12s_v3` to the codegen.sh file + run it using our typical test subscription (AKS E2E - Large Cluster is what we used to use), but it shows that it has zones available in our subscription (thus this code path is not utilized).

Do we have a standard subscription (please provide name + not ID) that we use to generate these skus? I want to see what is available there that would meet these requirements. 

However, if we think it may change in the future for a particular subscription, maybe a "true" unit test would make sense just to ensure this particular path is covered.

pkg/providers/instancetype/suite_test.go
rakechill: fyi: this is causing CI to fail
```
Detected Programmatic Focus - setting exit status to 197
```

pkg/providers/instancetype/suite_test.go
tallaxes: Wouldn't `Standard_NC6s_v3` itself work? It lists zones, but also has all of them restricted ...

rakechill: yep you're correct, that should work. I think it was end of night for me + I didn't re-check the sku.AvailabilityZones() code: https://github.com/Azure/skewer/blob/main/sku.go#L511-L563

rakechill: Updated, everything should be good now

pkg/providers/instancetype/instancetypes.go
```diff
@@ -145,10 +145,14 @@ func instanceTypeZones(sku *skewer.SKU, region string) sets.Set[string] {
 	// skewer returns numerical zones, like "1" (as keys in the map);
 	// prefix each zone with "<region>-", to have them match the labels placed on Node (e.g. "westus2-1")
 	// Note this data comes from LocationInfo, then skewer is used to get the SKU info
+	// If an offering is non-zonal, the availability zones will be empty.
```
Bryce-Soghigian, line 148: Very interesting discovery so certain offerings are not zonal and only have one location? Or are they physically hosted in all of the locations you just can't request a specific location? How does this work?

pkg/providers/instancetype/suite_test.go
```diff
@@ -1186,6 +1185,22 @@ var _ = Describe("InstanceType Provider", func() {
...
 			Expect(vm.Zones).To(BeEmpty())
 		})
+		It("should support provisioning non-zonal instance types in zonal regions", func() {
+			nonZonalInstanceType := "Standard_NC6s_v3"
+			ExpectApplied(ctx, env.Client, nodePool, nodeClass)
+
+			pod := coretest.UnschedulablePod(coretest.PodOptions{
+				NodeSelector: map[string]string{v1.LabelInstanceTypeStable: nonZonalInstanceType},
+			})
+			ExpectProvisioned(ctx, env.Client, cluster, cloudProvider, coreProvisioner, pod)
```
Bryce-Soghigian, line 1195: We seem to always repeat these steps should we combine them? In core this pattern makes more sense because we test more workload scheduling so we have a more diverse set of pods/workloads we pass in. Here almost every pod is just unschedulable pods with some node selector or no diversity at all.

pkg/providers/instancetype/suite_test.go
```diff
@@ -1186,6 +1185,22 @@ var _ = Describe("InstanceType Provider", func() {
...
+
+			node := ExpectScheduled(ctx, env.Client, pod)
+			Expect(node.Labels["node.kubernetes.io/instance-type"]).To(Equal(nonZonalInstanceType))
```
Bryce-Soghigian, line 1198: since we already are popping the vm below, could we also check the VMSize in properties.HardwareProfile.VMSize?

pkg/providers/instancetype/suite_test.go
```diff
@@ -1186,6 +1185,22 @@ var _ = Describe("InstanceType Provider", func() {
 			vm := azureEnvNonZonal.VirtualMachinesAPI.VirtualMachineCreateOrUpdateBehavior.CalledWithInput.Pop().VM
 			Expect(vm.Zones).To(BeEmpty())
 		})
+		It("should support provisioning non-zonal instance types in zonal regions", func() {
+			nonZonalInstanceType := "Standard_NC6s_v3"
```
Bryce-Soghigian, line 1189: Is it mainly GPUs that are nonzonal? Or just skus with limited capacity? I am so curious about how all of this works

examples/v1beta1/nodeclaim/manual-nodeclaim.yaml
```diff
@@ -0,0 +1,23 @@
...
+    name: default
+    kind: AKSNodeClass
+  taints:
+    - key: "sku"
+      value: "gpu"
+      effect: NoSchedule
+  requirements:
+    - key: "karpenter.sh/nodepool"
+      operator: In
+      values: ["kaito"]
```
Bryce-Soghigian, line 20: Can you explain why we need this?

examples/v1beta1/nodeclaim/manual-nodeclaim.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: karpenter.sh/v1beta1
+kind: NodeClaim
+metadata:
+  name: manual-kaito
+  labels:
+    karpenter.sh/nodepool: kaito
+  annotations:
+    karpenter.sh/do-not-disrupt: "true"
```
Bryce-Soghigian, line 8: I wonder if the [forceful Expiration RFC](https://github.com/kubernetes-sigs/karpenter/pull/1303) Upstream messes with kaito since we rely on do-not-disrupt to prevent these nodes manually scaled up from ever being removed and we don't actually have a real nodepool. 

I need to learn more about nodeclaim lifecycle in general to see how it works in this case where we have a nodeclaim but no nodepool or nodeclass

tallaxes (changes requested):

pkg/providers/instancetype/suite_test.go
tallaxes: We could (even though, all things being equal, it is nice to use provider-independent checks), but I don't think we need either check for instance type, since many other tests verify instance type constraints are respected. What we _should_ check here instead is the zone label `v1alpha2.AlternativeLabelTopologyZone`).

pkg/providers/instancetype/instancetypes.go
rakechill: > What is not clear to me, is whether non-zonal offering is always available (as long as SKU/size is in the list); would be good to check - especially since the proposed logic makes this assumption. This will also be important longer term, if we decide to expand support for non-zonal offerings alongside zonal, and not just as fallbacks.

This is a good point. I'll try to find some docs/ask around AKS to understand the context here + ensure that this is a fair assumption. I'm going to proceed with merging for now to unblock Heba.

pkg/providers/instancetype/suite_test.go
rakechill: Not sure yet. Will look further into why certain SKUs are non-zonal.

examples/v1beta1/nodeclaim/manual-nodeclaim.yaml
rakechill: My intention in leaving this would be to make testing easier in the future. I can go ahead and remove it for now. If, in the future, we decide to add testing around manual nodeclaim creation, maybe this is worthwhile using. TBD + can access again from the commit here.

examples/v1beta1/nodeclaim/manual-nodeclaim.yaml
rakechill: As far as I know, we have not tested this. cc @helayoty -- do kaito nodeclaims have any expiration associated with them?

@Bryce-Soghigian are you saying this would apply to any nodeclaim or would nodeclaims w/ corresponding nodes provisioned be excepted from that default?

I'll need to look at that RFC, as well.

pkg/providers/instancetype/instancetypes.go
```diff
@@ -145,10 +145,14 @@ func instanceTypeZones(sku *skewer.SKU, region string) sets.Set[string] {
...
-		return sets.New(lo.Map(lo.Keys(sku.AvailabilityZones(region)), func(zone string, _ int) string {
-			return fmt.Sprintf("%s-%s", region, zone)
-		})...)
+		keys := lo.Keys(sku.AvailabilityZones(region))
+		if len(keys) > 0 {
+			return sets.New(lo.Map(keys, func(zone string, _ int) string {
+				return fmt.Sprintf("%s-%s", region, zone)
+			})...)
+		}
 	}
```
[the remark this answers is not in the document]
rakechill: My micro-optimization was to only define the extra variable if it's a zonal region. I do agree that skuZones is a clearer way to name this variable, though.

pkg/providers/instancetype/suite_test.go
```diff
@@ -1186,6 +1185,22 @@ var _ = Describe("InstanceType Provider", func() {
...
+			ExpectApplied(ctx, env.Client, nodePool, nodeClass)
+
+			pod := coretest.UnschedulablePod(coretest.PodOptions{
+				NodeSelector: map[string]string{v1.LabelInstanceTypeStable: nonZonalInstanceType},
+			})
```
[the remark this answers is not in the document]
rakechill: I saw examples from both sides. Happy to restrict from the nodepool instead, though. Don't have a strong opinion either way.

pkg/providers/instancetype/suite_test.go
rakechill: I think sometimes you want different orderings/things in between applied, provisioned, scheduled. I did simplify this test a bit based on other discussions with Alex.

pkg/providers/instancetype/instancetypes.go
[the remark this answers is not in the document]
tallaxes: There is a bit more than naming here. Moving skuZones out puts both conditions in one "if" and simplifies the flow. It becomes obvious in which cases an empty string is returned; with two ifs one has to think about it a bit. IMO readability beats optimization here. Note also that the resulting set of offerings is cached, so this is not called that often.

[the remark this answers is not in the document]
tallaxes: There is a bit more than naming here. Moving `skuZones` out puts both conditions in one "if" and simplifies the flow. It becomes obvious in which cases an empty string is returned; with two ifs one has to think about it a bit. IMO readability beats optimization here. Note also that the resulting set of offerings is cached, so this is not called that often.

pkg/providers/instancetype/suite_test.go
tallaxes: The above comment was only about checking instance type being unnecessary, `Expect(vm.Zones).To(BeEmpty())` is very much appropriate, can we restore it?

tallaxes (changes requested):
Minor nits, looks good otherwise!

pkg/providers/instancetype/instancetypes.go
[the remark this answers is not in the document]
rakechill: Yes, I meant that the nested if statement + joint if statement both seem clear to me. I agreed with the better naming convention, though. I don't have strong opinions on either structure here, so I'll go ahead and update as you suggested.

tallaxes (approved):

examples/v1beta1/nodeclaim/manual-nodeclaim.yaml
tallaxes: Yes, it may be worth checking the expiration behavior of standalone NodeClaims (irrespective of Forceful Expiration, which would still respects do-not-disrupt). I strongly suspect that expiration just won't work, since all configuration for it (including defaults) comes from NodePool, but worth checking nonetheless.

PR #413, opened 2024-06-19
chore(release): add automated github releases for workflow

### LINKED ISSUE #388: Autogenerate release notes
Currently our release process requires a manual collection, and creation of any release notes. It would be good to automate this process to improve the quality of life for releases.

charliedmcb:
Fixes #388

**Description**
We currently have to create the github release, and its release notes manually. This additional change makes the workflow automatically create the github release and generate the release notes. 

This is modeled off aws:
- https://github.com/aws/karpenter-provider-aws/blob/2e907f90b82ea04ff76da13c5a486c71a421e827/.github/workflows/release.yaml

 However, since the repo `aws` is using is no longer maintained:
 
image

I looked over the recommended substitutes, and found that I think the [action-gh-release](https://github.com/softprops/action-gh-release) matches our current desires the best. Its closest to the old action. While the  [semantic-release](https://github.com/semantic-release/semantic-release) action could be useful with some of its additional feature sets, I think for our current usage it goes above and beyond to a point we aren't currently looking for. This would require us to scope it down/not use it in the fully intended way.

This said the [action-gh-release](https://github.com/softprops/action-gh-release) repo does have a lot of open issues. While its getting updates at what seem to be an ok cadence, and still maintained it seems like issues/fixes aren't being put through too frequently. That's my main reservation with it. Although, I still think its the correct choice for now.
 
This change also has some additional elements around automated pre-releases as that is something we have been currently doing, which is missing from the aws release.

Sadly there isn't a great way to test this until we are actually doing a release and running the workflow.

However, I did check the regex works correctly within bash. Although, not fully confirmed with the actual release variable.
*

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
NONE
```


tallaxes (approved):
Overall LGTM (subject to testing). One concern: While there is value in aligning with AWS provider, [marvinpinto/actions](https://github.com/marvinpinto/actions) is no longer maintained. The author recommends [semantic-release](https://github.com/semantic-release/semantic-release) or [action-gh-release](https://github.com/softprops/action-gh-release). I wonder how hard it would be to move to one of those instead (maybe together with AWS provider)?

charliedmcb:
> Overall LGTM (subject to testing). One concern: While there is value in aligning with AWS provider, [marvinpinto/actions](https://github.com/marvinpinto/actions) is no longer maintained. The author recommends [semantic-release](https://github.com/semantic-release/semantic-release) or [action-gh-release](https://github.com/softprops/action-gh-release). I wonder how hard it would be to move to one of those instead (maybe together with AWS provider)?

Good callout. I'll take a look over those other options and how they would align here. Can share over with `aws` too if I find anything.

rakechill:
> While the [semantic-release](https://github.com/semantic-release/semantic-release) action could be useful with some of its additional feature sets, I think for our current usage it goes above and beyond to a point we aren't currently looking for. This would require us to scope it down/not use it in the fully intended way.

@charliedmcb do you think it would be more complex to understand this tool? from my perspective, additional feature sets that we don't use shouldn't impact us? And could potentially be more useful in the long term. Maybe I'm misunderstanding your reservations, though.

> This said the [action-gh-release](https://github.com/softprops/action-gh-release) repo does have a lot of open issues. While its getting updates at what seem to be an ok cadence, and still maintained it seems like issues/fixes aren't being put through too frequently.

Does it seem like `semantic-release` also has some of these maintenance concerns?

charliedmcb:
> @charliedmcb Do you think it would be more complex to understand this tool? From my perspective, additional feature sets that we don't use shouldn't impact us? And could potentially be more useful in the long term. Maybe I'm misunderstanding your reservations, though.

Yea, so I could see using it within the future. However, my main reservations are that while the tool does have extra features, it also seems to be build to handle releases in a way we aren't really looking to do atm. 

So, while it contains the tools we want, it also contains things that we'd want to strip away, and feels the added complexity might be excessive.

Examples:
- It appears to release to npm, which isn't something we are currently looking to do.
- It is meant for CI releases, which I think we don't want something that frequent yet.

Things I'd like from it:
- Auto-generated release notes (we can get this from the other tool)
- Auto-generation of sematic tags (looking into still leveraging this rn)
    - I hoping if its scoped down this far it might be less complexity. Also, though this I'm hoping I could gain a better understanding and we might be able to leverage more of the tool.

I so far also wasn't able to find an official github action for it, which means we'd have to make the action usage for it a bit extra complicated too.

rakechill:
Your responses make sense to me 👍 
[action-gh-release](https://github.com/softprops/action-gh-release) seems like a reasonable first step and if we need more improvements in the future or want to make this more automated after testing it for our upcoming release, then we can return to `semantic-release` and see if it's better.

approving for now if you want to move forward with this!

rakechill (approved):

charliedmcb:
Switching to using semantic-release, so closing this PR

PR #414, opened 2024-06-20
fix: specify location when creating ACR

tallaxes:
**Description**

Explicitly specifying the location for ACR seems to be fixing "ACR not found" issues during E2E. (Would like to follow-up to see why we are trying to create ACR _twice_ ...)

It also switches the default region from `eastus` (which appears to have some capacity constraints right now) to `southcentralus`. We don't have GPU quota there, so GPU tests will fail for now.

* Manual run of E2E from this branch: https://github.com/Azure/karpenter-provider-azure/actions/runs/9592960325

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


tallaxes (commented):
/test

tallaxes (commented):
/test

Bryce-Soghigian (approved):

PR #418, opened 2024-06-27
chore(release): add production env to release workflow

charliedmcb:
Fixes # 

**Description**

[Using environments for deployment](https://docs.github.com/en/actions/deployment/targeting-different-environments/using-environments-for-deployment) will allow us to put [Required reviewers](https://docs.github.com/en/actions/deployment/targeting-different-environments/using-environments-for-deployment#required-reviewers) on the release workflows.

Note: this does require us to add a `production` env to the github repo. I think this approach is correct for us as we are moving towards a more automated release process. However, it does deviate from `aws` some.

Nothing additional
*

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
NONE
```


tallaxes:
As Charlie mentioned, likely need to actually add production environment prior to merging

PR #420, opened 2024-06-27
chore(release): add new semantic-release workflow

charliedmcb:
Fixes # 

**Description**
We currently have to create the github release, and its release notes manually.

The repo `aws` is using is no longer maintained:
 - https://github.com/aws/karpenter-provider-aws/blob/2e907f90b82ea04ff76da13c5a486c71a421e827/.github/workflows/release.yaml
image

Because of this there are two approaches I'm looking at for our more automated release process.

I looked over the recommended substitutes and found that I think the [action-gh-release](https://github.com/softprops/action-gh-release) might match our current desires the best. Its closest to the old action. However, the [semantic-release](https://github.com/semantic-release/semantic-release) action could also be useful with some of its additional feature sets, so want to explore that option as well, even though more complex. This PR is exploring the [semantic-release](https://github.com/semantic-release/semantic-release) option more. 

I'm only giving the workflow the following perms, even though its known to now be enough:
- contents: read
This is a safeguard from the workflow actually doing a release for now. I'm wanting to complete this since I want to run the workflow to test semantic-release. However, can't run the workflow without it having the file in existence in main. I'm curious if we can maybe leverage it, and if the components are correcting up correctly within the workflow.

Sadly there isn't a great way to test this until we have it in main.
*

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
NONE
```

### DIFF
--- a/.github/workflows/release-trigger.yaml
+++ b/.github/workflows/release-trigger.yaml
@@ -1,0 +1,28 @@
+# This Github action is the release trigger for Azure/karpenter-provider-azure
+# It will auto-generate the next semantic release version and tag the git ref
+
+name: Release Trigger
+on:
+    workflow_dispatch:
+
+jobs:
+  generate-sem-ver:
+    permissions:
+      contents: read
+    runs-on: ubuntu-latest
+    steps:
+      - name: Harden Runner
+        uses: step-security/harden-runner@17d0e2bd7d51742c71671bd19fa12bdc9d40a3d6 # v2.8.1
+        with:
+          egress-policy: audit
+      - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+        with:
+          fetch-depth: 0
+      - uses: actions/setup-node@v4
+        with:
+          node-version: '20.x' # semantic-release requires Node version 20.8.1 or higher
+      - name: semantic-release
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          npx semantic-release@18
--- a/.github/workflows/release-trigger.yaml
+++ b/.github/workflows/release-trigger.yaml
@@ -18,7 +18,7 @@ jobs:
       - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
         with:
           fetch-depth: 0
-      - uses: actions/setup-node@v4
+      - uses: actions/setup-node@60edb5dd545a775178f52524783378180af0d1f8 # v4.0.2
         with:
           node-version: '20.x' # semantic-release requires Node version 20.8.1 or higher
       - name: semantic-release
--- a/.github/workflows/release-trigger.yaml
+++ b/.github/workflows/release-trigger.yaml
@@ -22,7 +22,5 @@ jobs:
         with:
           node-version: '20.x' # semantic-release requires Node version 20.8.1 or higher
       - name: semantic-release
-        env:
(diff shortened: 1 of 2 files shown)


rakechill (dismissed):
makes sense to me, approving since it should be read-only/for testing

rakechill (approved):

.github/workflows/release-trigger.yaml
```diff
@@ -0,0 +1,26 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Harden Runner
+        uses: step-security/harden-runner@17d0e2bd7d51742c71671bd19fa12bdc9d40a3d6 # v2.8.1
+        with:
+          egress-policy: audit
+      - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+        with:
+          fetch-depth: 0
+      - uses: actions/setup-node@v4
```
tallaxes, line 21: This will need to be pinned, but can let dependabot take care of this

.github/workflows/release-trigger.yaml
```diff
@@ -0,0 +1,26 @@
...
+        with:
+          node-version: '20.x' # semantic-release requires Node version 20.8.1 or higher
+      - name: semantic-release
+        run: |
+          npx semantic-release@18
```
tallaxes, line 26: Curios about where this version comes from, and how we will be updating it?
(Also wondering if there is a reason this is not available as GH action?)

tallaxes (dismissed):

.github/workflows/release-trigger.yaml
charliedmcb: It needs to be pinned to let `dependabot` take care of it? Pinning to the major was the given example in the docs, so was following that.

.github/workflows/release-trigger.yaml
charliedmcb: @18 is the most recent major version for `semantic-release`, and the recommended approach they have for using it.

As for no action, I also was wondering the same thing. From my exploration thus far I still haven't found a great answer to it, but I'm thinking that it might be simply due to the complexity of it, that they've decided not to publish an actual action.

.github/workflows/release-trigger.yaml
tallaxes: [Pinned dependencies](https://github.com/ossf/scorecard/blob/main/docs/checks.md#pinned-dependencies) is best practice from security perspective. Now that I think about it, dependabot won't switch unpinned to pinned (I think), so may need to do this manually once.

charliedmcb: Pinned:
https://github.com/actions/setup-node/commit/60edb5dd545a775178f52524783378180af0d1f8

charliedmcb: Pinned:
https://github.com/actions/setup-node/commit/60edb5dd545a775178f52524783378180af0d1f8

charliedmcb: Pinned:
https://github.com/actions/setup-node/commit/60edb5dd545a775178f52524783378180af0d1f8

tallaxes (approved):

.github/workflows/release-trigger.yaml
charliedmcb: Ensuring a procedure for updating this dependency is an important problem

charliedmcb: Resolving to unblock completion, but updating both this version and node is something that will need a solution.

PR #433, opened 2024-07-11
chore(release): configure semantic-release automation

### LINKED ISSUE #388: Autogenerate release notes

charliedmcb:
Fixes #388 

**Description**
This configuration of the `semantic-release` tool:
- https://github.com/semantic-release/semantic-release
Provides us with automatic version detection for releases, tagging, and autogenerating draft release notes, which can be published if the actual release succeeds.

I tested earlier versions in dry-run mode, and pipeline failure issues disabled:
- https://github.com/Azure/karpenter-provider-azure/actions/runs/9898742036/job/27346208600

Plan to test this version with an actual release once committed to `main`
*

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
NONE
```


.github/workflows/release-trigger.yaml
```diff
@@ -22,5 +24,8 @@ jobs:
         with:
           node-version: '20.x' # semantic-release requires Node version 20.8.1 or higher
       - name: semantic-release
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
-          npx semantic-release@18
\ No newline at end of file
+          npm install conventional-changelog-conventionalcommits -D
+          npx semantic-release@24
```
tallaxes, line 31: The version proved to be important, so maybe we can add a not about why this one?

.releaserc.yaml
tallaxes, line 1: Wonder why we end up with JSON in .yaml. Not a big deal, but if YAML is actually supported may be easier to maintain?

tallaxes (approved):
LGTM

.github/workflows/release-trigger.yaml
charliedmcb: Yea, this one is just the latest major. It was the recommended approach they had. They just had the incorrect version documented.

By linking to major we won't accidentally pick up a breaking change, but otherwise we'll get patch versions etc asap.

.releaserc.yaml
charliedmcb: Yea, I was kind of wondering this too. The mixed formats seemed weird. Could try changing to yaml. The doc examples where just done in json like this, but said we could use multiple file extensions. If yaml doesn't work, switching to a json file might be best

charliedmcb: Changed to json file to keep matching format with the semantic-release docs.

tallaxes (approved):

PR #455, opened 2024-08-13
fix: removing --keep-terminated-pod-volumes from kubelet flags and adding k8s 1.31 to ci-test

Bryce-Soghigian:
Fixes # 

**Description**
For supporting 1.31 node bootstrapping, we need to remove the `--keep-terminated-pod-volumes` from the kubelet flags. This pr introduces that logic into our bootstrapping contract. 

- make presubmit 
- manual testing with kubelet version set to 1.31

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
Support 1.31 node bootstrapping via removing the --keep-terminated-pod-volumes flag that was removed in kubelet version 1.31. 
```

### DIFF
--- a/hack/toolchain.sh
+++ b/hack/toolchain.sh
@@ -15,7 +15,7 @@ tools() {
     go install github.com/google/ko@v0.15.2
     go install github.com/mikefarah/yq/v4@v4.43.1
     go install github.com/norwoodj/helm-docs/cmd/helm-docs@v1.13.1
-    go install sigs.k8s.io/controller-runtime/tools/setup-envtest@v0.0.0-20240409134613-20f3f4bed925
+    go install sigs.k8s.io/controller-runtime/tools/setup-envtest@0c7827e417acc15f29e7c4bfccede809d372676a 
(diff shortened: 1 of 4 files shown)


Bryce-Soghigian:
This can't be properly tested until upstream images are replicated to our acs upstream mirror. As we can't bootstrap nodes with 1.31 until we have those images published.

hack/toolchain.sh
```diff
@@ -15,7 +15,7 @@ tools() {
     go install github.com/google/ko@v0.15.2
     go install github.com/mikefarah/yq/v4@v4.43.1
     go install github.com/norwoodj/helm-docs/cmd/helm-docs@v1.13.1
-    go install sigs.k8s.io/controller-runtime/tools/setup-envtest@v0.0.0-20240409134613-20f3f4bed925
+    go install sigs.k8s.io/controller-runtime/tools/setup-envtest@0c7827e417acc15f29e7c4bfccede809d372676a
```
tallaxes, line 18: Would be nice to keep the previous version format, but not critical

tallaxes (approved):
LGTM

PR #466, opened 2024-08-23
feat: disabling newly added kubelet serving certificate rotation

Bryce-Soghigian:
Fixes # 

**Description**
This [PR](https://github.com/Azure/AgentBaker/pull/4796/files) in AgentBaker enables the kubelet flag `--rotate-server-certificates` kubelet flag.  This flag will auto request and rotate the kubelet serving certificates by requesting new certs from the apiservers `certificates.k8s.io` apigroup when a certificate expiration approaches. This also requires the RotateKubeletServerCertificate feature gate to be enabled. The existing model just self signs the certificates.

In terms of the karpenter impact, we are just going to stick with self signed certificates for now since we don't have everything configured to leverage this functionality.

- make presubmit

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


cameronmeissner:
we'll need to come up with a way Karpenter can determine when it's safe to enable this flag

cameronmeissner:
will probably end up deploying the required plugin whenever karpenter is enabled to ensure this flag is safe to set, for now will keep as false until the new plugin as rolled out

cameronmeissner:
one thing to also note, Karpenter will need to make to not pass --tls-cert-file and --tls-private-key-file as kubelet flags when this is enabled

Bryce-Soghigian (commented):
/test

README.md
```diff
@@ -1,4 +1,4 @@
-
+FFFFF
```
Bryce-Soghigian, line 1:
```suggestion
```

tallaxes:
Why do we need to do anything to preserve the current functionality? The support `ENABLE_KUBELET_SERVING_CERTIFICATE_ROTATION` is (correctly) added with backward compatibility: not having this flag in contract is the same as setting it to false.

PR #477, opened 2024-09-05
fix: harden harden-runner egress policy

tallaxes:
**Description**

Switch to Harden Runner `egress-policy: block`, add selected `allowed-endpoints`. Also `disable-telemetry: true` for now.

The endpoints come from StepSecurity insights for corresponding runs (such as [this one](https://app.stepsecurity.io/github/Azure/karpenter-provider-azure/actions/runs/10379526294?jobid=28737859970&tab=recommendations); "Recommendations" tab as base, "Networks Events" for further drill down), with some selected wildcarding (to accommodate variations) and removals (mostly things that are no longer used, but might still be reported in the cumulative insights). Notes and outliers:
* The largest set of endpoints is (naturally) used by E2E tests, which use az CLI, create AKS clusters, and run tests. The only one using wildcards.
* Historically observed but removed: `9236a389bd48b984df91adc1bc924620.r2.cloudflarestorage.com` - [blob storage for cgr.dev](https://edu.chainguard.dev/chainguard/administration/network-requirements/#chainguard-images-third-party-hosts), and we no longer use Chainguard for base images
* `clients3.google.com:80` and `firebaselogging-pa.googleapis.com:443` - confirmed Skaffold, likely [telemetry](https://skaffold.dev/docs/resources/telemetry/), likely can be avoided/blocked with/after `skaffold config set --global collect-metrics false`
* There are no changes to "Build and publish to MCR" workflow, as it runs on self-hosted 1ES pool. (Use of Harden Runner in that context needs to be reviewed separately.)

* Selected checks (using the updated workflows) [passed](https://github.com/Azure/karpenter-provider-azure/pull/477/checks)

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note

```

### DIFF
--- a/.github/workflows/ci-test.yml
+++ b/.github/workflows/ci-test.yml
@@ -21,12 +21,22 @@ jobs:
     - name: Harden Runner
       uses: step-security/harden-runner@5c7944e73c4c2a096b17a9cb74d65b6c2bbafbde # v2.9.1
       with:
-        egress-policy: audit
+        disable-telemetry: true
+        egress-policy: block
+        allowed-endpoints: >
+          api.github.com:443
+          coveralls.io:443
+          github.com:443
+          objects.githubusercontent.com:443
+          proxy.golang.org:443
+          raw.githubusercontent.com:443
+          storage.googleapis.com:443
+          sum.golang.org:443
 
     - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
     - uses: ./.github/actions/install-deps
       with:
-        k8sVersion: ${{ matrix.k8sVersion }}    
+        k8sVersion: ${{ matrix.k8sVersion }}
     - run: K8S_VERSION=${{ matrix.k8sVersion }} make ci-test
     - name: Send coverage
       # should only send coverage once https://docs.coveralls.io/parallel-builds
--- a/.github/workflows/e2e-matrix.yaml
+++ b/.github/workflows/e2e-matrix.yaml
@@ -34,6 +34,7 @@ jobs:
           disable-telemetry: true
           disable-sudo: true
           egress-policy: block
+          # no allowed endpoints
           allowed-endpoints: >
 
       - id: generate-e2e-run-hash
--- a/.github/workflows/e2e-matrix.yaml
+++ b/.github/workflows/e2e-matrix.yaml
@@ -21,7 +21,7 @@ on:
 
 permissions:
   contents: read
-        
+
 jobs:
   initialize-generative-params:
     runs-on: ubuntu-latest
@@ -31,7 +31,10 @@ jobs:
       - name: Harden Runner
         uses: step-security/harden-runner@5c7944e73c4c2a096b17a9cb74d65b6c2bbafbde # v2.9.1
         with:
-          egress-policy: audit
+          disable-telemetry: true
+          disable-sudo: true
+          egress-policy: block
+          allowed-endpoints: >
 
       - id: generate-e2e-run-hash
         run: |
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -49,6 +49,9 @@ jobs:
         with:
           disable-telemetry: true
           egress-policy: block
+          # - wildcards to allow for variation in targeted clusters
+          # - clients3.google.com:80 and firebaselogging-pa.googleapis.com:443 - confirmed Skaffol
+          #   likely telemetry, likely can be avoided/blocked with/after "skaffold config set --global collect-metrics false"
           allowed-endpoints: >
             *.azmk8s.io:443
             *.azurecr.io:443
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -47,7 +47,33 @@ jobs:
       - name: Harden Runner
         uses: step-security/harden-runner@5c7944e73c4c2a096b17a9cb74d65b6c2bbafbde # v2.9.1
         with:
-          egress-policy: audit
+          disable-telemetry: true
+          egress-policy: block
+          allowed-endpoints: >
+            *.azmk8s.io:443
+            *.azurecr.io:443
+            *.data.mcr.microsoft.com:443
+            aka.ms:443
+            api.github.com:443
+            app.aladdin.microsoft.com:443
+            auth.docker.io:443
+            azcliextensionsync.blob.core.windows.net:443
+            clients3.google.com:80
+            dc.services.visualstudio.com:443
+            distroless.dev:443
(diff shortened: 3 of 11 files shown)


.github/workflows/e2e.yaml
```diff
@@ -47,7 +47,33 @@ jobs:
...
+          allowed-endpoints: >
+            *.azmk8s.io:443
+            *.azurecr.io:443
+            *.data.mcr.microsoft.com:443
+            aka.ms:443
+            api.github.com:443
+            app.aladdin.microsoft.com:443
+            auth.docker.io:443
+            azcliextensionsync.blob.core.windows.net:443
+            clients3.google.com:80
```
tallaxes, line 61: Skaffold (confirmed), likely [telemetry](https://skaffold.dev/docs/resources/telemetry/), likely can be avoided/blocked with/after `skaffold config set --global collect-metrics false`

.github/workflows/e2e.yaml
```diff
@@ -47,7 +47,33 @@ jobs:
...
+            dc.services.visualstudio.com:443
+            distroless.dev:443
+            firebaselogging-pa.googleapis.com:443
```
tallaxes, line 64: Skaffold (confirmed), likely telemetry

tallaxes (commented):
/test

.github/workflows/ci-test.yml
```diff
@@ -21,12 +21,22 @@ jobs:
     - name: Harden Runner
       uses: step-security/harden-runner@5c7944e73c4c2a096b17a9cb74d65b6c2bbafbde # v2.9.1
       with:
-        egress-policy: audit
+        disable-telemetry: true
+        egress-policy: block
+        allowed-endpoints: >
```
weinong, line 26: how did you get this list?

tallaxes: > The endpoints come from StepSecurity insights for corresponding runs (such as [this one](https://app.stepsecurity.io/github/Azure/karpenter-provider-azure/actions/runs/10379526294?jobid=28737859970&tab=recommendations)), with some selected wildcarding (to accommodate variations) and removals (mostly things that are no longer used, but might still be reported in the cumulative insights).

More details in the PR description

charliedmcb:
E2E Run:
- https://github.com/Azure/karpenter-provider-azure/actions/runs/10728344943

.github/workflows/e2e-matrix.yaml
```diff
@@ -31,7 +31,10 @@ jobs:
       - name: Harden Runner
         uses: step-security/harden-runner@5c7944e73c4c2a096b17a9cb74d65b6c2bbafbde # v2.9.1
         with:
-          egress-policy: audit
+          disable-telemetry: true
+          disable-sudo: true
+          egress-policy: block
+          allowed-endpoints: >
```
charliedmcb, line 37: Does this mean no allowed endpoints, and it by default inherits the allowed endpoints from any parent workflow, or does this explicitly call out to inherits the endpoints?

charliedmcb: Assuming its doing something, but if not, can we remove?

charliedmcb: Maybe just drop a comment to explain?

.github/workflows/e2e.yaml
charliedmcb: Can we leave a comment within the actual code for any odd cases?

.github/workflows/e2e.yaml
charliedmcb: + on leaving a comment for odd cases in code

.github/workflows/e2e.yaml
```diff
@@ -47,7 +47,33 @@ jobs:
       - name: Harden Runner
         uses: step-security/harden-runner@5c7944e73c4c2a096b17a9cb74d65b6c2bbafbde # v2.9.1
         with:
-          egress-policy: audit
+          disable-telemetry: true
+          egress-policy: block
+          allowed-endpoints: >
+            *.azmk8s.io:443
+            *.azurecr.io:443
+            *.data.mcr.microsoft.com:443
```
charliedmcb, line 55: Maybe call these out in a comment too, since they are our only regex ones?

The PR description is very clarifying for all the odd cases, but that won't directly stick around with the code, and will be harder to fine in the future.

charliedmcb (commented):
Overall, looks pretty good. Mostly nits.

However, do have a bit larger concern for dev flow impact. If we update the workflows, and need to add new allowed-endpoints, will there be an easy way to determine what those endpoints we need to add are?

Also, need to confirm E2E tests pass.

.github/workflows/e2e.yaml
tallaxes: Yes (will add), but not inline due https://github.com/step-security/harden-runner/issues/199

.github/workflows/e2e-matrix.yaml
tallaxes: This means no allowed endpoints (besides some built-in ones, like GH cache) for this job (and no inheritance). Will add a comment.

tallaxes:
> However, do have a bit larger concern for dev flow impact. If we update the workflows, and need to add new allowed-endpoints, will there be an easy way to determine what those endpoints we need to add are?

Yes, likely using either the failure itself or Harden Runner post-action log (or report in the action run summary). If this fails - setting `disable-telemetry: false` temporarily, do another run, and use StepSecurity report. (We are also likely to switch telemetry back on permanently after some further internal review.)

.github/workflows/e2e.yaml
```diff
@@ -47,7 +47,36 @@ jobs:
       - name: Harden Runner
         uses: step-security/harden-runner@5c7944e73c4c2a096b17a9cb74d65b6c2bbafbde # v2.9.1
         with:
-          egress-policy: audit
+          disable-telemetry: true
+          egress-policy: block
+          # - wildcards to allow for variation in targeted clusters
+          # - clients3.google.com:80 and firebaselogging-pa.googleapis.com:443 - confirmed Skaffol
```
charliedmcb, line 53: nit: Skaffold

charliedmcb (approved):
lgtm. 1 nit

charliedmcb:
> > However, do have a bit larger concern for dev flow impact. If we update the workflows, and need to add new allowed-endpoints, will there be an easy way to determine what those endpoints we need to add are?
> 
> Yes, likely using either the failure itself or Harden Runner post-action log (or report in the action run summary). If this fails - setting `disable-telemetry: false` temporarily, do another run, and use StepSecurity report. (We are also likely to switch telemetry back on permanently after some further internal review.)

Ok, sounds good.

charliedmcb (approved):
lgtm

PR #484, opened 2024-09-16
fix(bootstrap): add defaults for required IMDS restriction settings

tallaxes:
**Description**

Fixes bootstrap failure on latest VHD images by specifying default values for required IMDS restriction settings

* `make presubmit`
* manual local testing
* E2E: https://github.com/Azure/karpenter-provider-azure/actions/runs/10890699757

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/providers/imagefamily/bootstrap/cse_cmd.sh.gtpl
+++ b/pkg/providers/imagefamily/bootstrap/cse_cmd.sh.gtpl
@@ -144,4 +144,6 @@ KUBE_CA_CRT="{{.KubeCACrt}}"
 KUBENET_TEMPLATE="{{.KubenetTemplate}}"
 CONTAINERD_CONFIG_CONTENT="{{.ContainerdConfigContent}}"
 IS_KATA="{{.IsKata}}"
+ENABLE_IMDS_RESTRICTION=false
+INSERT_IMDS_RESTRICTION_RULE_TO_MANGLE_TABLE=false
 /usr/bin/nohup /bin/bash -c "/bin/bash /opt/azure/containers/provision_start.sh"


pkg/providers/imagefamily/bootstrap/cse_cmd.sh.gtpl
```diff
@@ -144,4 +144,6 @@ KUBE_CA_CRT="{{.KubeCACrt}}"
 KUBENET_TEMPLATE="{{.KubenetTemplate}}"
 CONTAINERD_CONFIG_CONTENT="{{.ContainerdConfigContent}}"
 IS_KATA="{{.IsKata}}"
+ENABLE_IMDS_RESTRICTION=false
+INSERT_IMDS_RESTRICTION_RULE_TO_MANGLE_TABLE=false
```
Bryce-Soghigian, line 148: Why set the defaults here like this rather than in NodeBootstrappingVariables like we have previously?

Bryce-Soghigian (approved):

pkg/providers/imagefamily/bootstrap/cse_cmd.sh.gtpl
tallaxes: Overall we want to be shrinking the required contract. There is no need to propagate a temporary fix (which would like to remove as soon as the contract allows it) higher up the data flow chain than absolutely necessary. When and if we decide to actually expose this for control / customization - then would be the time to establish the correct data flow. Yes, this is a bit of an inconsistency with other settings; we may want to start taking similar approach to them as well - though maybe let's wait until the contract moves forward.

PR #486, opened 2024-09-19
feat: BYO Subnet Per AKSNodeClass + Subnet Drift

### LINKED ISSUE #231: BYO subnet per node pool (AKSNodeClass)

Bryce-Soghigian:
Fixes #231 

**Description**
This PR introduces support for bringing your own subnet per AKSNodeClass. Karpenter has decided to opt into subnet drift. This pr does not solve validation of the subnet and its compatibility with other subnets on the cluster.  This PR does also not solve any observability considerations. 

- make az-all 
- make test
- new drift scenario in our e2e validating we drift on subnet id change

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


pkg/apis/v1alpha2/aksnodeclass.go
```diff
@@ -27,6 +27,11 @@ import (
 // AKSNodeClassSpec is the top level specification for the AKS Karpenter Provider.
 // This will contain configuration necessary to launch instances in AKS.
 type AKSNodeClassSpec struct {
+	// VnetSubnetID is the subnet used by nics provisioned with this nodeclass.
+	// If not specified, we will use the default --vnet-subnet-id specified on the managed cluster
```
Bryce-Soghigian, line 31: todo: Add a aka.ms/aks/nap/subnet-id tsg for making sure they specify a correct subnet without overlapping cidr etc.

pkg/apis/v1alpha2/aksnodeclass.go
```diff
@@ -27,6 +27,11 @@ import (
...
+	// +kubebuilder:validation:Pattern=`(?i)^\/subscriptions\/[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}\/resourceGroups\/[a-zA-Z0-9][a-zA-Z0-9_.-]{0,61}[a-zA-Z0-9]\/providers\/Microsoft\.Network\/virtualNetworks\/[a-zA-Z0-9][a-zA-Z0-9_.-]{0,61}[a-zA-Z0-9]\/subnets\/[a-zA-Z0-9][a-zA-Z0-9_.-]{0,61}[a-zA-Z0-9]$`
+	// +optional
+	VnetSubnetID *string `json:"vnetSubnetID,omitempty"`
```
tallaxes, line 34:
```suggestion
	VNETSubnetID *string `json:"vnetSubnetID,omitempty"`
```

test/pkg/environment/common/expectations.go
```diff
@@ -686,3 +686,13 @@ func (env *Environment) GetDaemonSetCount(np *corev1beta1.NodePool) int {
 		return true
 	})
 }
+
+func (env *Environment) EventuallyExpectDrifted(nodeClaims ...*corev1beta1.NodeClaim) {
+	GinkgoHelper()
+	Eventually(func(g Gomega) {
+		for _, nc := range nodeClaims {
+			g.Expect(env.Client.Get(env, client.ObjectKeyFromObject(nc), nc)).To(Succeed())
```
charliedmcb, line 694: We don't need:
`g.Expect(nodeClaim.StatusConditions().GetCondition(corev1beta1.Drifted)).ToNot(BeNil())`, which was in the previous code?

test/suites/drift/suite_test.go
```diff
@@ -154,4 +146,34 @@ var _ = Describe("Drift", func() {
...
+
+		By(fmt.Sprintf("waiting for nodeclaim %s to be marked as drifted", nodeClaim.Name))
+		env.EventuallyExpectDrifted(nodeClaim)
+		By(fmt.Sprintf("waiting for pod %s to to update", pod.Name))
+		delete(pod.Annotations, corev1beta1.DoNotDisruptAnnotationKey)
+		env.ExpectUpdated(pod)
+
+		By(fmt.Sprintf("expect pod %s, nodeclaim %s, and node %s to eventually not exist", pod.Name, nodeClaim.Name, node.Name))
+		SetDefaultEventuallyTimeout(10 * time.Minute)
+		env.EventuallyExpectNotFound(pod, nodeClaim, node)
```
charliedmcb, line 175: Aren't we not going to be able to get a working replacement?

So that means we will still nuke the old pod, nodeClaim, and node that have drifted, even if we don't get a working replacement?

charliedmcb (commented):
First pass. Generally looking good, found a couple issues, and a few areas I'm wondering about minor refactoring/helpers to make the code areas cleaner.

pkg/operator/options/options.go
```diff
@@ -84,6 +86,7 @@ func (o *Options) AddFlags(fs *coreoptions.FlagSet) {
 	fs.StringVar(&o.NetworkDataplane, "network-dataplane", env.WithDefaultString("NETWORK_DATAPLANE", "cilium"), "The network dataplane used by the cluster.")
 	fs.StringVar(&o.SubnetID, "vnet-subnet-id", env.WithDefaultString("VNET_SUBNET_ID", ""), "The default subnet ID to use for new nodes. This must be a valid ARM resource ID for subnet that does not overlap with the service CIDR or the pod CIDR")
 	fs.Var(newNodeIdentitiesValue(env.WithDefaultString("NODE_IDENTITIES", ""), &o.NodeIdentities), "node-identities", "User assigned identities for nodes.")
+	fs.StringVar(&o.NodeResourceGroup, "node-resource-group", env.WithDefaultString("AZURE_NODE_RESOURCE_GROUP", ""), "[REQUIRED] the resource group created and managed by AKS where the nodes live")
```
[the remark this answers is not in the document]
Bryce-Soghigian: yeah it was passed as an environment variable in azure config, but rather than accessing it that way I figured I would pull it into options since this is the first part of the flow that required the MC_RG that didnt have access to azure config

Really all of azure config should be migrated to options but Im leaving that out of this pr

pkg/providers/instance/instance.go
```diff
@@ -81,36 +81,33 @@ type Resource = map[string]interface{}
 
 type Provider struct {
 	location               string
-	azClient               *AZClient
+	AZClient               *AZClient
```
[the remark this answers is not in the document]
Bryce-Soghigian: Juan lee when I first started working on karpenter was really against the GetNic() pattern. Seems like if we need access to the clients, we should just expose the clients. This way the helpers stay decoupled and can be used in other projects or other places. 

You create the one client you need, and pass it into the armutil helper rather than having to fully configure a AZClient when using the code elsewhere.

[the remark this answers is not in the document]
Bryce-Soghigian: LMK if you have a different opinion or feel more strongly

pkg/providers/instance/armutils.go
```diff
@@ -75,6 +75,14 @@ func createVirtualMachineExtension(ctx context.Context, client VirtualMachineExt
 	return &res.VirtualMachineExtension, nil
 }
 
+func GetNic(ctx context.Context, client NetworkInterfacesAPI, rg, nicName string) (*armnetwork.Interface, error) {
```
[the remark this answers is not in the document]
Bryce-Soghigian: for Gets() it doesnt make as much sense since there is no polling but I just followed the existing pattern. 

For gets it may just make sense to use the client itself directly.

pkg/providers/instance/instance.go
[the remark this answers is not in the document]
charliedmcb: I don't have a strong opinion. Just mostly felt anti-pattern with the other clients since we have the others private, and now just `NetworkInterfacesClient` public. However, if we want to move that direction, and slowly transition it makes sense.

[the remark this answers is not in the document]
charliedmcb: Oh, never mind, for some reason the other public ones were hidden for me. This isn't an outlier. Mb

Bryce-Soghigian:
see test result of drift e2e: https://github.com/Azure/karpenter-provider-azure/actions/runs/11111946917/job/30885575958

pkg/providers/instance/instance.go
```diff
@@ -221,9 +218,10 @@ func (p *Provider) newNetworkInterfaceForVM(vmName string, backendPools *loadbal
 					Properties: &armnetwork.InterfaceIPConfigurationPropertiesFormat{
 						Primary:                   to.Ptr(true),
 						PrivateIPAllocationMethod: to.Ptr(armnetwork.IPAllocationMethodDynamic),
-						Subnet: &armnetwork.Subnet{
-							ID: &p.subnetID,
-						},
+						// Subnet is set in p.applyTemplateToNic
```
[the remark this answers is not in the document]
Bryce-Soghigian: Really its just because this is the only bit of networking configuration defined elsewhere, I figured an explicit callout would be good. 

We can remove if you think where subnet is set is discoverable enough. This comment isn't great anyway because it doesn't communicate the logic for how its set and exactly where in the launch template provider its set maybe we can remove. 


By typing that all I convinced myself to remove it.

test/suites/drift/suite_test.go
Bryce-Soghigian: see the comment above: `		// this will fail to actually create a new nodeclaim for the drift replacement but should still test that we are marking the nodeclaim as drifted`

pkg/cloudprovider/drift.go
```diff
@@ -119,3 +124,35 @@ func (c *CloudProvider) isImageVersionDrifted(
...
+
+	// TODO: Refactor all of AzConfig to be part of options
+	nic, err := instance.GetNic(ctx, c.instanceProvider.AZClient.NetworkInterfacesClient, options.FromContext(ctx).NodeResourceGroup, nicName)
+	if err != nil {
+		if sdkerrors.IsNotFoundErr(err) {
+			return "", nil
+		}
+		return "", err
+	}
+	nicSubnet := getFirstSubnetFromNic(nic)
```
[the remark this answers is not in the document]
Bryce-Soghigian: We should probably instead check that we are checking the primary ipConfiguration rather than just grabbing the first one. 

Although from a karpenter side of things, we should have consistency between all of the subnets set in all ipconfigurations so really any ip configuration works.

pkg/apis/v1alpha2/crd_validation_cel_test.go
```diff
@@ -59,6 +59,42 @@ var _ = Describe("CEL/Validation", func() {
 			},
 		}
 	})
+	Context("VnetSubnetID", func() {
+		DescribeTable("Should only accept valid VnetSubnetID", func(vnetSubnetID string, expected bool) {
```
charliedmcb, line 63: nit:
```suggestion
		DescribeTable("Should only accept valid VnetSubnetID", func(vnetSubnetID string, expectValid bool) {
```

Or some other similar naming. Just `expected` is minorly unclear.

pkg/apis/v1alpha2/crd_validation_cel_test.go
```diff
@@ -59,6 +59,42 @@ var _ = Describe("CEL/Validation", func() {
...
+					VNETSubnetID: &vnetSubnetID,
+				},
+			}
+			if expected {
+				Expect(env.Client.Create(ctx, nodeClass)).To(Succeed())
+			} else {
+				Expect(env.Client.Create(ctx, nodeClass)).ToNot(Succeed())
+			}
+		},
+			Entry("valid VnetSubnetID", "/subscriptions/12345678-1234-1234-1234-123456789012/resourceGroups/rgname/providers/Microsoft.Network/virtualNetworks/vnet/subnets/subnet", true),
```
charliedmcb, line 76: Can add in the edge cases like we talked about for 1 character subnets, etc. 

I'd personally add in some of the non letter special characters just to ensure those work.

However, all this depends on how deep you want to go into the validation part, and like we talked about might be better to backoff and make less restrictive validation here instead.

charliedmcb: Oh, I see you did add those in which is great. My assumption here was that all the valid cases would come first, followed by the invalid ones. I think a light refactor to just block them out as the complete valid set of cases, and then invalid would be nice.

pkg/apis/v1alpha2/crd_validation_cel_test.go
```diff
@@ -59,6 +59,42 @@ var _ = Describe("CEL/Validation", func() {
...
+			Entry("valid format with mixed-case resource group and subnet name", "/subscriptions/12345678-1234-1234-1234-123456789012/resourceGroups/MyResourceGroup/providers/Microsoft.Network/virtualNetworks/MyVirtualNetwork/subnets/MySubnet", true),
+			Entry("invalid subnet with too short ID", "/subscriptions/123/resourceGroups/rgname/providers/Microsoft.Network/virtualNetworks/vnet/subnets/subnet", false),
+			Entry("missing resourceGroups in path", "/subscriptions/12345678-1234-1234-1234-123456789012/rgname/providers/Microsoft.Network/virtualNetworks/vnet/subnets/subnet", false),
+			Entry("invalid provider in path", "/subscriptions/12345678-1234-1234-1234-123456789012/resourceGroups/rgname/providers/Microsoft.Storage/virtualNetworks/vnet/subnets/subnet", false),
+			Entry("missing virtualNetworks in path", "/subscriptions/12345678-1234-1234-1234-123456789012/resourceGroups/rgname/providers/Microsoft.Network/subnets/subnet", false),
+			Entry("valid VnetSubnetID at max length", "/subscriptions/12345678-1234-1234-1234-123456789012/resourceGroups/"+strings.Repeat("a", 63)+"/providers/Microsoft.Network/virtualNetworks/"+strings.Repeat("b", 63)+"/subnets/"+strings.Repeat("c", 63), true),
```
charliedmcb, line 85: nit: resource group name can be longer than `63`.

pkg/apis/v1alpha2/crd_validation_cel_test.go
```diff
@@ -59,6 +59,42 @@ var _ = Describe("CEL/Validation", func() {
...
+			Entry("valid resource group name 'valid.name-with-multiple.characters'", "/subscriptions/12345678-1234-1234-1234-123456789012/resourceGroups/valid.name-with-multiple.characters/providers/Microsoft.Network/virtualNetworks/vnet/subnets/subnet", true),
+			Entry("invalid resource group name ending with dot 'ends.with.dot.'", "/subscriptions/12345678-1234-1234-1234-123456789012/resourceGroups/ends.with.dot./providers/Microsoft.Network/virtualNetworks/vnet/subnets/subnet", false),
+			Entry("invalid resource group name with invalid character 'invalid#character'", "/subscriptions/12345678-1234-1234-1234-123456789012/resourceGroups/invalid#character/providers/Microsoft.Network/virtualNetworks/vnet/subnets/subnet", false),
+			Entry("invalid resource group name with unsupported chars 'name@with*unsupported&chars'", "/subscriptions/12345678-1234-1234-1234-123456789012/resourceGroups/name@with*unsupported&chars/providers/Microsoft.Network/virtualNetworks/vnet/subnets/subnet", false),
```
charliedmcb, line 95: Like all these resource group name tests 💯 :)

pkg/providers/instancetype/suite_test.go
```diff
@@ -168,6 +168,16 @@ var _ = Describe("InstanceType Provider", func() {
 				ContainSubstring("kubernetes.azure.com/podnetwork-type=overlay"),
 			))
 		})
+		It("should use the subnet specified in the nodeclass", func() {
+			nodeClass.Spec.VNETSubnetID = lo.ToPtr("/subscriptions/12345678-1234-1234-1234-123456789012/resourceGroups/sillygeese/providers/Microsoft.Network/virtualNetworks/karpenter/subnets/nodeclassSubnet")
```
charliedmcb, line 172: nit: make this string a const, at least in this function just to simplify that the comparison later is directly with the same string?

pkg/cloudprovider/drift.go
```diff
@@ -38,6 +42,7 @@ import (
 const (
 	K8sVersionDrift   cloudprovider.DriftReason = "K8sVersionDrift"
 	ImageVersionDrift cloudprovider.DriftReason = "ImageVersionDrift"
+	SubnetDrift       cloudprovider.DriftReason = "SubnetDrift"
```
charliedmcb, line 45: Not to do with your PR, but wondering if we want to add an aks prefix to these to distinguish them as cloud provider drift reasons.

pkg/apis/crds/karpenter.azure.com_aksnodeclasses.yaml
```diff
@@ -67,6 +67,12 @@ spec:
                   type: string
                 description: Tags to be applied on Azure resources like instances.
                 type: object
+              vnetSubnetID:
+                description: |-
+                  VNETSubnetID is the subnet used by nics provisioned with this nodeclass.
+                  If not specified, we will use the default --vnet-subnet-id specified in karpenter's options config
+                pattern: (?i)^\/subscriptions\/[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}\/resourceGroups\/[a-zA-Z0-9_\-().]{0,89}[a-zA-Z0-9_\-()]\/providers\/Microsoft\.Network\/virtualNetworks\/[a-zA-Z0-9][a-zA-Z0-9_.\-]{0,61}[a-zA-Z0-9]\/subnets\/[a-zA-Z0-9][a-zA-Z0-9_.\-]{0,61}[a-zA-Z0-9]$
```
charliedmcb, line 74: regex has some edge cases I don't think it catches around `subnets` name, etc.

I think that loosening these constraints is best, to make sure we aren't blocking valid cases.

charliedmcb (commented):
Mostly nits around testing that I think would provide better clarity to readers. 

Also, the regex needs some fixing. Once the regex is fixed I'll approve. lmk

pkg/apis/crds/karpenter.azure.com_aksnodeclasses.yaml
```diff
@@ -67,6 +67,12 @@ spec:
                   type: string
                 description: Tags to be applied on Azure resources like instances.
                 type: object
+              vnetSubnetID:
+                description: |-
+                  VNETSubnetID is the subnet used by nics provisioned with this nodeclass.
+                  If not specified, we will use the default --vnet-subnet-id specified in karpenter's options config
+                pattern: (?i)^\/subscriptions\/[^\/]+\/resourceGroups\/[a-zA-Z0-9_\-().]{0,89}[a-zA-Z0-9_\-()]\/providers\/Microsoft\.Network\/virtualNetworks\/[^\/]+\/subnets\/[^\/]+$
```
charliedmcb, line 74: Could still just check length of terms, but think its completely fine as is too.

charliedmcb (approved):
Looks like you covered all the things I was looking for in the last review so approving.

So long as the E2E tests work, think its good!

test/suites/drift/suite_test.go
```diff
@@ -154,4 +146,23 @@ var _ = Describe("Drift", func() {
...
+
+		nodeClaim := env.EventuallyExpectCreatedNodeClaimCount("==", 1)[0]
+		By("triggering subnet drift")
+		// TODO: Introduce azure clients to the tests to get values dynamically and be able to create azure resources inside of tests rather than using a fake id.
+		// this will fail to actually create a new nodeclaim for the drift replacement but should still test that we are marking the nodeclaim as drifted.
+		nodeClass.Spec.VNETSubnetID = lo.ToPtr("/subscriptions/12345678-1234-1234-1234-123456789012/resourceGroups/sillygeese/providers/Microsoft.Network/virtualNetworks/karpenter/subnets/nodeclassSubnet2")
+		env.ExpectCreatedOrUpdated(nodeClass)
+
+		By(fmt.Sprintf("waiting for nodeclaim %s to be marked as drifted", nodeClaim.Name))
+		env.EventuallyExpectDrifted(nodeClaim)
```
Bryce-Soghigian, line 166:
```
@Bryce-Soghigian ➜ /workspaces/karpenter-provider-azure (bsoghigian/subnet-drift) $ k describe nodeclaims 
Name:         curtainplump-2-zrfk1fab8a-2w9xh
....
  Conditions:
    Last Transition Time:  2024-10-04T16:33:03Z
    Reason:                SubnetDrift
    Severity:              Warning
    Status:                True
    Type:                  Drifted
    Last Transition Time:  2024-10-04T16:28:53Z
```
The resulting e2e test here passes

Ran 1 of 1 Specs in 512.998 seconds
SUCCESS! -- 1 Passed | 0 Failed | 0 Pending | 0 Skipped
PASS | FOCUSED
FAIL    github.com/Azure/karpenter-provider-azure/test/suites/drift     513.031s
FAIL

pkg/cloudprovider/drift.go
Bryce-Soghigian: Maybe thats something that belongs inside the actual core code since eventual controller will take a list of cloudprovider reasons in the future?

tallaxes (approved):
Overall LGTM, please see comments for some potential improvements

pkg/cloudprovider/drift.go
```diff
@@ -119,3 +124,35 @@ func (c *CloudProvider) isImageVersionDrifted(
...
 	return "", nil
 }
+
+// isSubnetDrifted returns drift if the nic for this nodeclaim does not match the expected subnet
+func (c *CloudProvider) isSubnetDrifted(ctx context.Context, nodeClaim *corev1beta1.NodeClaim, nodeClass *v1alpha2.AKSNodeClass) (cloudprovider.DriftReason, error) {
+	expectedSubnet := lo.Ternary(nodeClass.Spec.VNETSubnetID == nil, options.FromContext(ctx).SubnetID, lo.FromPtr(nodeClass.Spec.VNETSubnetID))
+	nicName := instance.GenerateResourceName(nodeClaim.Name)
+
+	// TODO: Refactor all of AzConfig to be part of options
+	nic, err := instance.GetNic(ctx, c.instanceProvider.AZClient.NetworkInterfacesClient, options.FromContext(ctx).NodeResourceGroup, nicName)
```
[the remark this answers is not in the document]
Bryce-Soghigian: Will address these in a separate PR, agree we need to take a step back and refactor this.

pkg/cloudprovider/suite_test.go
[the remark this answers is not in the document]
Bryce-Soghigian: There is a bug in the NIC GET fake that was giving me a hard time, but if we rewrite to use annotations instead we shouldnt have that concern. 

Lets adress this in my followup pr. This will come in shortly after I merge this

PR #508, opened 2024-10-02
test(Cilium): add CiliumAgentNotReadyTaint to prevent overprovisioning

charliedmcb:
Fixes # 

**Description**
See the following thread for why we need the given startupTaint:
- https://karpenter.sh/docs/faq/#why-do-i-see-extra-nodes-get-launched-to-schedule-pending-pods-that-remain-empty-and-are-later-removed

As our testing, and main example cases all use Cilium, adding this to all the areas.

Tested [general-purpose.yaml](https://github.com/Azure/karpenter-provider-azure/pull/508/files#diff-d76365775af789aed6414b86f28136204ea77b35ee25349d76eb590b9d7c03b9) in self-hosted codespace setup.

Also, ran E2E suite:
- https://github.com/Azure/karpenter-provider-azure/actions/runs/11149258534
- https://github.com/Azure/karpenter-provider-azure/actions/runs/11150952645

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
NONE
```

### DIFF
--- a/examples/v1beta1/general-purpose.yaml
+++ b/examples/v1beta1/general-purpose.yaml
@@ -12,7 +12,7 @@ spec:
   template:
     spec:
       startupTaints:
-        - key: node.kubernetes.io/network-unavailable
+        - key: node.cilium.io/agent-not-ready
           effect: NoSchedule
       requirements:
         - key: kubernetes.io/arch
--- a/examples/v1beta1/general-purpose.yaml
+++ b/examples/v1beta1/general-purpose.yaml
@@ -11,6 +11,9 @@ spec:
     expireAfter: Never
   template:
     spec:
+      startupTaints:
+        - key: node.kubernetes.io/network-unavailable
+          effect: NoSchedule
       requirements:
         - key: kubernetes.io/arch
           operator: In
--- a/examples/v1beta1/multi-arch.yaml
+++ b/examples/v1beta1/multi-arch.yaml
@@ -12,7 +12,7 @@ spec:
   template:
     spec:
       startupTaints:
-        - key: node.kubernetes.io/network-unavailable
+        - key: node.cilium.io/agent-not-ready
           effect: NoSchedule
       requirements:
         - key: kubernetes.io/arch
--- a/examples/v1beta1/multi-arch.yaml
+++ b/examples/v1beta1/multi-arch.yaml
@@ -11,6 +11,9 @@ spec:
     expireAfter: Never
   template:
     spec:
+      startupTaints:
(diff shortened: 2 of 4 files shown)


tallaxes:
If this is always required when running with Cilium dataplane, can we automatically (conditional on Cilium dataplane) add this taint to the taints we pass to kubelet internally, rather than requiring explicit external configuration?

charliedmcb:
> If this is always required when running with Cilium dataplane, can we automatically (conditional on Cilium dataplane) add this taint to the taints we pass to kubelet internally, rather than requiring explicit external configuration?

I don't think that would solve things, as the main issue here from my understanding is that the nodes come up with this taint already, but karpenter-core logic for provisioning isn't aware of it. Added it to the startuptaints, makes it so the taint will be there on startup, but also expected to be temporary, so won't mess up its provisioning logic.

See documentation:
- https://karpenter.sh/docs/concepts/nodepools/#cilium-startup-taint

tallaxes:
Ok, so this is more about letting Karpenter know, than actually placing the taint. Would be nice to not have to configure this manually, but maybe it is unavoidable (at least in self-hosted).

tallaxes (approved):

examples/v1beta1/general-purpose.yaml
```diff
@@ -11,6 +11,9 @@ spec:
     expireAfter: Never
   template:
     spec:
+      startupTaints:
+        - key: node.cilium.io/agent-not-ready
```
Bryce-Soghigian, line 15:
```suggestion
        # If you are using cilium dataplane you might notice that there is some overprovisioning where we launch extra nodes to schedule pods that remain empty and are later removed. 
        # this taint will prevent that scenario, read the docs here: https://karpenter.sh/docs/faq/#why-do-i-see-extra-nodes-get-launched-to-schedule-pending-pods-that-remain-empty-and-are-later-removed
        - key: node.cilium.io/agent-not-ready
```

Bryce-Soghigian: Maybe some context here to make it clear this is just if you are using cilium. network unavailable during cni initialization should also have this same side effect.

charliedmcb:
> Ok, so this is more about letting Karpenter know, than actually placing the taint. Would be nice to not have to configure this manually, but maybe it is unavoidable (at least in self-hosted). And would need to be careful not to use this with non-cilium dataplanes.

Yea, I would like this to be something we can automatically configure at least within the e2e tests. However, isn't the simplest right now, and our general testing suggests going to cilium, so wanting to fix.

PR #527, opened 2024-10-16
feat: CSE-based bootstrapping with bootstrapping client mode

comtalyst:
Fixes # 

**Description**

Using provision servers provide opportunities for us to delegate deterministic bootstrapping details away to keep our maintenance focus on autoscaling.
For the first mode introduced here—the bootstrapping client—will also switch back to CSE-based bootstrapping for general stability. The server is expected to provide custom data and CSE based on our inputs of `ProvisionProfile`. See swagger file for specifics.

This pathway will be supported by NAP only, for now.

* NAP E2Es
* Manual inspection of retrieved CSE values
* K8s versions: 1.28.0, 1.29.9, 1.30.1, 1.31.1
* ACRPull (on the above versions)
* E2E: https://github.com/Azure/karpenter-provider-azure/actions/runs/11508985665

- [ ] Yes, issue opened: # 
- [X] No

**Release Note**

```release-note
NONE
```


pkg/providers/imagefamily/azlinux.go
```diff
@@ -68,12 +71,13 @@ func (u AzureLinux) DefaultImages() []DefaultImageOutput {
...
 				scheduling.NewRequirement(v1alpha2.LabelSKUHyperVGeneration, v1.NodeSelectorOpIn, v1alpha2.HyperVGenerationV2),
 			),
+			Distro: "aks-azurelinux-v2-arm64-gen2",
 		},
 	}
 }
 
 // UserData returns the default userdata script for the image Family
-func (u AzureLinux) UserData(kubeletConfig *corev1beta1.KubeletConfiguration, taints []v1.Taint, labels map[string]string, caBundle *string, _ *cloudprovider.InstanceType) bootstrap.Bootstrapper {
+func (u AzureLinux) SelfContainedUserData(kubeletConfig *corev1beta1.KubeletConfiguration, taints []v1.Taint, labels map[string]string, caBundle *string, _ *cloudprovider.InstanceType) bootstrap.Bootstrapper {
```
Bryce-Soghigian, line 80: If we are going to rename this and change from the AWS Paradigm we should call this CustomData instead of UserData? 

This ends up getting converted to custom data here https://github.com/Azure/karpenter-provider-azure/blob/main/pkg/providers/instance/instance.go#L336

The reason we called it user data was to align with AWS but if we are changing might as well go all the way.

comtalyst: According to self-contained feature owners, right now that is actually `CustomData` (yes, I think we've been mislabeling it). But there could be a switch soon.

pkg/providers/imagefamily/ubuntu_2204.go
```diff
@@ -60,6 +62,7 @@ func (u Ubuntu2204) DefaultImages() []DefaultImageOutput {
 				scheduling.NewRequirement(v1.LabelArchStable, v1.NodeSelectorOpIn, corev1beta1.ArchitectureAmd64),
 				scheduling.NewRequirement(v1alpha2.LabelSKUHyperVGeneration, v1.NodeSelectorOpIn, v1alpha2.HyperVGenerationV1),
 			),
+			Distro: "aks-ubuntu-containerd-22.04",
```
Bryce-Soghigian, line 65: nit: these should be constants

pkg/providers/instance/instance.go
```diff
@@ -284,11 +301,17 @@ func newVMObject(
...
+	instanceType *corecloudprovider.InstanceType,
+	provisionMode string) armcompute.VirtualMachine {
 	// Build the image reference from template
 	imageReference := armcompute.ImageReference{
 		CommunityGalleryImageID: &launchTemplate.ImageID,
 	}
+
+	if launchTemplate.IsWindows {
+		return armcompute.VirtualMachine{} // TODO(Windows)
+	}
```
Bryce-Soghigian, line 313: why add this wiring here now if you arent going to support it anywhere? Assuming users will enable windows through the AKS NodeClass and a new ImageFamily api value. 

Wouldnt using v1alpha2.AKSNodeClass.Spec.ImageFamily make more sense rather than putting this in the template?

pkg/providers/instance/instance.go
```diff
@@ -342,16 +364,21 @@ func newVMObject(
...
 	}
-	setVMPropertiesStorageProfile(vm.Properties, instanceType, nodeClass)
+	setVMPropertiesStorageProfile(vm.Properties, launchTemplate.StorageProfile)
 	setVMPropertiesBillingProfile(vm.Properties, capacityType)
 
+	if provisionMode == options.ProvisionModeBootstrappingClient {
+		vm.Properties.OSProfile.CustomData = lo.ToPtr(launchTemplate.AgentBakerCustomData)
+	} else {
+		vm.Properties.OSProfile.CustomData = lo.ToPtr(launchTemplate.SelfContainedUserData)
+	}
```
Bryce-Soghigian, line 374:
```suggestion
        vm.Properties.OSProfile.CustomData = lo.ToPtr(lo.Ternary(provisionMode == options.ProvisionModeBootstrappingClient, launchTemplate.AgentBakerCustomData, launchTemplate.SelfContainedCustomData))
```

pkg/providers/instance/instance.go
```diff
@@ -333,7 +356,6 @@ func newVMObject(
 						},
 					},
 				},
-				CustomData: lo.ToPtr(launchTemplate.UserData),
```
Bryce-Soghigian, line 336: Rather than deleting here lets just do an inline ternary assignment 
```suggestion
                                   CustomData: lo.ToPtr(lo.Ternary(provisionMode == options.ProvisionModeBootstrappingClient, launchTemplate.AgentBakerCustomData, launchTemplate.SelfContainedCustomData))
```

pkg/providers/instance/instance.go
```diff
@@ -435,6 +462,14 @@ func (p *Provider) launchInstance(
 		return nil, nil, azErr
 	}
 
+	if p.provisionMode == options.ProvisionModeBootstrappingClient {
+		err = p.createCSExtension(ctx, resourceName, launchTemplate.AgentBakerCSE, launchTemplate.IsWindows)
```
Bryce-Soghigian, line 466: We have the aksNodeClass here we dont need launchTemplate.IsWindows we can just do something like 
```
		err = p.createCSExtension(ctx, resourceName, launchTemplate.AgentBakerCSE, lo.Ternary(nodeClass.Spec.ImageFamily == Windows), true, false)
```

we can add a helper too in the AKSNodeClass api that is called iswindows doing that check so we can reuse that everywhere we have access to the nodeclass

pkg/providers/imagefamily/resolver.go
```diff
@@ -67,24 +79,40 @@ func New(_ client.Client, imageProvider *Provider) *Resolver {
...
 		metrics.ImageSelectionErrorCount.WithLabelValues(imageFamily.Name()).Inc()
 		return nil, err
 	}
 
 	logging.FromContext(ctx).Infof("Resolved image %s for instance type %s", imageID, instanceType.Name)
 
+	storageProfile := "ManagedDisks"
+	if useEphemeralDisk(instanceType, nodeClass) {
+		storageProfile = "Ephemeral"
+	}
```
Bryce-Soghigian, line 93: why move this check here?

pkg/providers/instance/instance.go
```diff
@@ -119,6 +122,7 @@ func (p *Provider) Create(ctx context.Context, nodeClass *v1alpha2.AKSNodeClass,
 	instanceTypes = orderInstanceTypesByPrice(instanceTypes, scheduling.NewNodeSelectorRequirementsWithMinValues(nodeClaim.Spec.Requirements...))
 	vm, instanceType, err := p.launchInstance(ctx, nodeClass, nodeClaim, instanceTypes)
 	if err != nil {
+		// Currently, CSE errors will lead to here
```
Bryce-Soghigian, line 125: How does this interact with @charliedmcb's proposal for parallel drift?

pkg/providers/instance/instance.go
```diff
@@ -193,6 +197,19 @@ func (p *Provider) createAKSIdentifyingExtension(ctx context.Context, vmName str
...
+	vmExtName := *vmExt.Name
+	logging.FromContext(ctx).Debugf("Creating virtual machine CSE for %s", vmName)
+	v, err := createVirtualMachineExtension(ctx, p.AZClient.virtualMachinesExtensionClient, p.resourceGroup, vmName, vmExtName, *vmExt)
+	if err != nil {
+		logging.FromContext(ctx).Errorf("Creating VM CSE for VM %q failed, %w", vmName, err)
+		return fmt.Errorf("creating VM CSE for VM %q, %w failed", vmName, err)
+	}
+	logging.FromContext(ctx).Debugf("Created virtual machine CSE for %s, with an id of %s", vmName, *v.ID)
+	return nil
+}
```
Bryce-Soghigian, line 211: Wondering if there is a cleaner way to share the vm extension creation logic between the different extensions. Thoughts?

pkg/providers/instance/instance.go
```diff
@@ -670,6 +705,54 @@ func (p *Provider) getAKSIdentifyingExtension() *armcompute.VirtualMachineExtens
...
+	const (
+		vmExtensionType     = "Microsoft.Compute/virtualMachines/extensions"
+		cseNameWindows      = "windows-cse-agent-karpenter"
+		cseTypeWindows      = "CustomScriptExtension"
+		csePublisherWindows = "Microsoft.Compute"
+		cseVersionWindows   = "1.10"
+		cseNameLinux        = "cse-agent-karpenter"
+		cseTypeLinux        = "CustomScript"
+		csePublisherLinux   = "Microsoft.Azure.Extensions"
+		cseVersionLinux     = "2.0"
```
Bryce-Soghigian, line 718: How often do these values change? what exactly is cseVersionLinux?

pkg/providers/instance/instance.go
```diff
@@ -670,6 +705,54 @@ func (p *Provider) getAKSIdentifyingExtension() *armcompute.VirtualMachineExtens
 	return vmExtension
 }
 
+func (p *Provider) getCSExtension(cse string, isWindows bool) *armcompute.VirtualMachineExtension {
+	const (
+		vmExtensionType     = "Microsoft.Compute/virtualMachines/extensions"
+		cseNameWindows      = "windows-cse-agent-karpenter"
```
Bryce-Soghigian, line 711: why is there a unique name for karpenter CSE? Would this break any assumptions made https://www.hyrumslaw.com/

pkg/providers/instance/instance.go
```diff
@@ -670,6 +705,54 @@ func (p *Provider) getAKSIdentifyingExtension() *armcompute.VirtualMachineExtens
...
+				Publisher:               lo.ToPtr(csePublisherLinux),
+				TypeHandlerVersion:      lo.ToPtr(cseVersionLinux),
+				Settings:                &map[string]interface{}{},
+				ProtectedSettings: &map[string]interface{}{
+					"commandToExecute": cse,
+				},
+			},
+		}
+	}
+}
```
Bryce-Soghigian, line 754: Something like this is more readable
```suggestion
return &armcompute.VirtualMachineExtension{
	Location: lo.ToPtr(p.location),
	Name:     lo.ToPtr(lo.Ternary(isWindows, cseNameWindows, cseNameLinux)),
	Type:     lo.ToPtr(vmExtensionType),
	Properties: &armcompute.VirtualMachineExtensionProperties{
		AutoUpgradeMinorVersion: lo.ToPtr(true),
		Type:                    lo.ToPtr(lo.Ternary(isWindows, cseTypeWindows, cseTypeLinux)),
		Publisher:               lo.ToPtr(lo.Ternary(isWindows, csePublisherWindows, csePublisherLinux)),
		TypeHandlerVersion:      lo.ToPtr(lo.Ternary(isWindows, cseVersionWindows, cseVersionLinux)),
		Settings:                &map[string]interface{}{},
		ProtectedSettings: &map[string]interface{}{
			"commandToExecute": cse,
		},
	},
}
```

pkg/providers/instance/instance.go
```diff
@@ -670,6 +705,54 @@ func (p *Provider) getAKSIdentifyingExtension() *armcompute.VirtualMachineExtens
 	return vmExtension
 }
 
+func (p *Provider) getCSExtension(cse string, isWindows bool) *armcompute.VirtualMachineExtension {
+	const (
```
Bryce-Soghigian, line 709: Can you leave some comments explaining what all of these values are and what they do?

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -156,24 +164,37 @@ func (p *Provider) getStaticParameters(ctx context.Context, instanceType *cloudp
...
-	// render user data
-	userData, err := options.UserData.Script()
-	if err != nil {
-		return nil, err
-	}
-
+func (p *Provider) createLaunchTemplate(ctx context.Context, params *parameters.Parameters) (*Template, error) {
 	// merge and convert to ARM tags
-	azureTags := mergeTags(options.Tags, map[string]string{karpenterManagedTagKey: options.ClusterName})
+	azureTags := mergeTags(params.Tags, map[string]string{karpenterManagedTagKey: params.ClusterName})
```
Bryce-Soghigian, line 173: NICE!

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -156,24 +164,37 @@ func (p *Provider) getStaticParameters(ctx context.Context, instanceType *cloudp
...
+	if p.provisionMode == options.ProvisionModeBootstrappingClient {
+		customData, cse, err := params.AgentBakerNodeBootstrapping.GetCustomDataAndCSE(ctx)
+		if err != nil {
+			return nil, err
+		}
+		template.AgentBakerCustomData = customData
+		template.AgentBakerCSE = cse
+	} else {
+		// render user data
+		userData, err := params.SelfContainedUserData.Script()
```
Bryce-Soghigian, line 191: is self contained ever using userdata? why is it called this? Maybe im misunderstanding something?

pkg/providers/imagefamily/agentbakerbootstrap/provisionclientbootstrap.go
```diff
@@ -0,0 +1,239 @@
...
+
+	resp, err := client.Operations.NodeBootstrappingGet(params)
+	if err != nil {
+		return "", "", err
+	}
+
+	if resp.Payload == nil {
+		return "", "", fmt.Errorf("no payload in response")
+	}
+	if resp.Payload.Cse == nil || *resp.Payload.Cse == "" {
```
Bryce-Soghigian, line 193: nit: CSE should be cased consistently across the code

pkg/providers/imagefamily/agentbakerbootstrap/provisionclientbootstrap.go
```diff
@@ -0,0 +1,239 @@
...
+		return "", "", fmt.Errorf("no CSE in response")
+	}
+	if resp.Payload.CustomData == nil || *resp.Payload.CustomData == "" {
+		return "", "", fmt.Errorf("no CustomData in response")
+	}
+
+	cseWithoutBootstrapToken := *resp.Payload.Cse
+	customDataWithoutBootstrapToken := *resp.Payload.CustomData
```
Bryce-Soghigian, line 201: can we use lo instead for deferencing? Given if pointer is nil for CSE it doesnt matter lol bootstrapping is cooked

pkg/operator/options/options_validation.go
```diff
@@ -83,6 +84,13 @@ func (o Options) validateVMMemoryOverheadPercent() error {
 	return nil
 }
 
+func (o Options) validateProvisionMode() error {
+	if o.ProvisionMode != "" && o.ProvisionMode != "bootstrappingclient" {
+		return fmt.Errorf("provision-mode is invalid: %s", o.ProvisionMode)
+	}
+	return nil
+}
```
Bryce-Soghigian, line 92: again think that just putting this behind the managed karpenter flag makes the most sense, unless you plan on having more modes.

Bryce-Soghigian (dismissed):
First round of comments, still need to dig into core logic a bit more and think through the new launch template patterns and the new resolver patterns also I need to evaluate if we are passing in all the right KubeletConfiguration.

pkg/provisionclients/swagger/nodebootstrapping.json
```diff
@@ -0,0 +1,671 @@
...
+    "title": "NodeBootstrapping",
+    "description": "A swagger spec that defines the get NodeBootstrapping operation.",
+    "version": "2024-10-01"
+  },
+  "schemes": [
+    "https"
+  ],
+  "consumes": [
+    "application/json"
+  ],
```
Bryce-Soghigian, line 13: Can we add makefile helpers for regenerating the swagger clients?

pkg/providers/imagefamily/agentbakerbootstrap/provisionclientbootstrap.go
```diff
@@ -0,0 +1,239 @@
...
+		provisionProfile.OsSku = to.Ptr(models.OSSKU_Ubuntu)
+	}
+
+	if p.KubeletConfig != nil {
+		provisionProfile.CustomKubeletConfig = &models.CustomKubeletConfig{
+			// AllowedUnsafeSysctls: ..., // Unsupported as of now
+			CPUCfsQuota: p.KubeletConfig.CPUCFSQuota,
+		}
+	}
+
```
Bryce-Soghigian, line 126: Can you leave some comments here on whats defaulted in NPS vs what is not? Its very confusing to understand what flags we are getting vs need to be passing in.

pkg/providers/imagefamily/agentbakerbootstrap/provisionclientbootstrap.go
```diff
@@ -0,0 +1,239 @@
...
+	case v1alpha2.AzureLinuxImageFamily:
+		provisionProfile.OsSku = to.Ptr(models.OSSKU_AzureLinux)
+	default:
+		provisionProfile.OsSku = to.Ptr(models.OSSKU_Ubuntu)
+	}
+
+	if p.KubeletConfig != nil {
+		provisionProfile.CustomKubeletConfig = &models.CustomKubeletConfig{
+			// AllowedUnsafeSysctls: ..., // Unsupported as of now
+			CPUCfsQuota: p.KubeletConfig.CPUCFSQuota,
```
Bryce-Soghigian, line 123: Do we plan on adding official support for customers specifying kubelet configuration for GA?

pkg/providers/imagefamily/agentbakerbootstrap/provisionclientbootstrap.go
```diff
@@ -0,0 +1,239 @@
...
+	ResourceGroup                  string
+	KubeletClientTLSBootstrapToken string
+	KubernetesVersion              string
+	ImageDistro                    string
+	IsWindows                      bool
+	InstanceType                   *cloudprovider.InstanceType
+	StorageProfile                 string
+	ImageFamily                    string
+}
+
```
Bryce-Soghigian, line 64: Most of these options are shared between self contained and agentbaker bootstrap. Can we share those values with struct embedding rather than copy pasting in both places?

pkg/providers/imagefamily/agentbakerbootstrap/provisionclientbootstrap.go
```diff
@@ -0,0 +1,239 @@
...
+		// ArtifactStreamingProfile: &models.ArtifactStreamingProfile{
+		// Enabled: lo.ToPtr(false), // Unsupported as of now
+		// },
+	}
+
+	if p.Arch == "amd64" {
+		provisionProfile.Architecture = lo.ToPtr("x64")
+	} else {
+		provisionProfile.Architecture = lo.ToPtr(p.Arch)
+	}
```
Bryce-Soghigian, line 109: Could these be defaulted based on the instance type on the NPS side?

pkg/providers/imagefamily/agentbakerbootstrap/provisionclientbootstrap.go
```diff
@@ -0,0 +1,239 @@
...
+				InstallGPUDriver: lo.ToPtr(true),
+			}
+		} else {
+			provisionProfile.GpuProfile = &models.GPUProfile{
+				DriverType:       lo.ToPtr(models.DriverType_CUDA),
+				InstallGPUDriver: lo.ToPtr(true),
+			}
+		}
+
+	}
```
Bryce-Soghigian, line 152: Same here, all of this logic will resolve to the same static values based on the sku, if we have the sku NPS can resolve these values.

pkg/providers/imagefamily/agentbakerbootstrap/provisionclientbootstrap.go
```diff
@@ -0,0 +1,239 @@
...
+
+	if p.IsWindows {
+		provisionProfile.OsType = lo.ToPtr(models.OSType_Windows)
+	} else {
+		provisionProfile.OsType = lo.ToPtr(models.OSType_Linux)
+	}
```
Bryce-Soghigian, line 158: Seems odd we need this at all since we pass in OS SKU. This alongside some of these other values could be easily infered in a more minimal AgentpoolProfile payload

pkg/providers/imagefamily/agentbakerbootstrap/provisionclientbootstrap.go
```diff
@@ -0,0 +1,239 @@
...
+
+	provisionHelperValues := &models.ProvisionHelperValues{
+		SkuCPU:    lo.ToPtr(p.InstanceType.Capacity.Cpu().AsApproximateFloat64()),
+		SkuMemory: lo.ToPtr(p.InstanceType.Capacity.Memory().AsApproximateFloat64()),
+	}
+
```
Bryce-Soghigian, line 164: Again would be nice to infer these based on the instance type.  All values that are not user input probably don't need to be in the provision profile payload in the future. We can resolve them inside of NPS

pkg/providers/imagefamily/agentbakerbootstrap/provisionclientbootstrap.go
```diff
@@ -0,0 +1,239 @@
...
+	if resp.Payload == nil {
+		return "", "", fmt.Errorf("no payload in response")
+	}
+	if resp.Payload.Cse == nil || *resp.Payload.Cse == "" {
+		return "", "", fmt.Errorf("no CSE in response")
+	}
+	if resp.Payload.CustomData == nil || *resp.Payload.CustomData == "" {
+		return "", "", fmt.Errorf("no CustomData in response")
+	}
```
Bryce-Soghigian, line 198: If we get errors here would it make sense to fall back onto SelfContained bootstrap? Could we fall back on our bootstrapping logic if this fails?

pkg/providers/imagefamily/agentbakerbootstrap/provisionclientbootstrap.go
```diff
@@ -0,0 +1,239 @@
...
+	customDataWithBootstrapToken := base64.StdEncoding.EncodeToString([]byte(decodedCustomDataWithBootstrapToken))
+
+	return customDataWithBootstrapToken, cseWithBootstrapToken, nil
+}
+
+func getAgentbakerGeneratedLabels(nodeResourceGroup string, nodeLabels map[string]string) {
+	// Delegatable defaulting?
+	nodeLabels["kubernetes.azure.com/role"] = "agent"
+	nodeLabels["kubernetes.azure.com/cluster"] = normalizeResourceGroupNameForLabel(nodeResourceGroup)
+}
```
Bryce-Soghigian, line 219: Do we need to make the karpenter nodepool aware of these defaulted labels?

pkg/providers/imagefamily/azlinux.go
```diff
@@ -60,6 +62,7 @@ func (u AzureLinux) DefaultImages() []DefaultImageOutput {
 				scheduling.NewRequirement(v1.LabelArchStable, v1.NodeSelectorOpIn, corev1beta1.ArchitectureAmd64),
 				scheduling.NewRequirement(v1alpha2.LabelSKUHyperVGeneration, v1.NodeSelectorOpIn, v1alpha2.HyperVGenerationV1),
 			),
+			Distro: "aks-azurelinux-v2",
```
Bryce-Soghigian, line 65: Can we make these consts, and can we import them from agentbaker?

pkg/providers/instance/instance.go
```diff
@@ -435,6 +462,14 @@ func (p *Provider) launchInstance(
 		return nil, nil, azErr
 	}
 
+	if p.provisionMode == options.ProvisionModeBootstrappingClient {
+		err = p.createCSExtension(ctx, resourceName, launchTemplate.AgentBakerCSE, launchTemplate.IsWindows)
+		if err != nil {
+			// This should fall back to cleanupAzureResources
```
Bryce-Soghigian, line 468: Should we be logging the CSE Failure at least? Or creating k8s events for CSE Failure?

tallaxes (commented):
Overall looks great for the first (and working) cut! I presume fixing CI & tests is coming, and further refactoring/cleanup can be expected later.

pkg/operator/operator.go
```diff
@@ -103,10 +103,12 @@ func NewOperator(ctx context.Context, operator *operator.Operator) (context.Cont
 		options.FromContext(ctx).ClusterEndpoint,
 		azConfig.TenantID,
 		azConfig.SubscriptionID,
+		azConfig.ResourceGroup,
 		azConfig.KubeletIdentityClientID,
 		azConfig.NodeResourceGroup,
 		azConfig.Location,
 		vnetGUID,
+		options.FromContext(ctx).ProvisionMode,
```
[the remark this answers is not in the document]
tallaxes: There could be more than one provisioning mode (including within NAP)

pkg/providers/imagefamily/agentbakerbootstrap/provisionclientbootstrap.go
tallaxes: Agreed, though some of this may be coming from Swagger etc.

pkg/providers/imagefamily/agentbakerbootstrap/provisionclientbootstrap.go
tallaxes: Good question, but I think the answer is No

pkg/providers/imagefamily/agentbakerbootstrap/provisionclientbootstrap.go
tallaxes: Maybe, though this would only be consequential if there are workloads targeting these, which is very unlikely. (More generally, it may be useful to develop a mechanism for letting Karpenter know Nodes will have certain labels - without having to specify them in NodePool.)

pkg/providers/instance/instance.go
tallaxes: Logged from within createCSExtension

pkg/operator/operator.go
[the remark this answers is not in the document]
Bryce-Soghigian: https://github.com/Azure/karpenter-provider-azure/pull/527#discussion_r1802325589 yeah I responded down here with similar arguement. We could have multiple provisioning modes, also we could have mulitple "types" of karpenter 

self-hosted aks, managed aks, self-hosted azure, its arguable that even this boolean flag I am introducing in my PR would make more sense as a string with multiple options.

pkg/utils/gpu.go
```diff
@@ -33,7 +33,7 @@ const (
 )
 
 func GetAKSGPUImageSHA(size string) string {
-	if useGridDrivers(size) {
+	if UseGridDrivers(size) {
 		return AKSGPUGridSHA
 	}
```
Bryce-Soghigian, line 38: Do these values have to come from us? if we use the ones from agentbaker and default them in NPS we can have consistency and won't have to bump the shas for cves

pkg/providers/imagefamily/agentbakerbootstrap/provisionclientbootstrap.go
```diff
@@ -0,0 +1,239 @@
...
+		// MessageOfTheDay:         lo.ToPtr(""),                                    // Unsupported as of now
+		// AgentPoolWindowsProfile: &models.AgentPoolWindowsProfile{},               // Unsupported as of now; TODO(Windows)
+		// KubeletDiskType:         lo.ToPtr(models.KubeletDiskType_Unspecified),    // Unsupported as of now
+		// CustomLinuxOSConfig:     &models.CustomLinuxOSConfig{},                   // Unsupported as of now (sysctl)
+		// EnableFIPS:              lo.ToPtr(false),                                 // Unsupported as of now
+		// GpuInstanceProfile:      lo.ToPtr(models.GPUInstanceProfile_Unspecified), // Unsupported as of now (MIG)
+		// WorkloadRuntime:         lo.ToPtr(models.WorkloadRuntime_Unspecified),    // Unsupported as of now (Kata)
+		// ArtifactStreamingProfile: &models.ArtifactStreamingProfile{
+		// Enabled: lo.ToPtr(false), // Unsupported as of now
+		// },
```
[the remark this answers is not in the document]
comtalyst: This will be handled separately. There are additional implications.

pkg/providers/imagefamily/resolver.go
```diff
@@ -113,3 +141,22 @@ func getImageFamily(familyName *string, parameters *template.StaticParameters) I
...
+	// decimal places are truncated, so we round down
+	return int32(maxSize)
+}
+
+// setVMPropertiesStorageProfile enables ephemeral os disk for instance types that support it
+func useEphemeralDisk(instanceType *corecloudprovider.InstanceType, nodeClass *v1alpha2.AKSNodeClass) bool {
+	// use ephemeral disk if it is large enough
+	return *nodeClass.Spec.OSDiskSizeGB <= getEphemeralMaxSizeGB(instanceType)
+}
```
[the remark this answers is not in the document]
comtalyst: It is being shared between population of `ProvisionProfile` and the creation of VM.
Do you have better suggestions on where this code should be, on top of your mind?

pkg/providers/instance/instance.go
comtalyst: It was already half-done so I wouldn't want to remove it if we might add them again. Doubt it is too distracting in practice. But could still be moved elsewhere by then.

pkg/providers/instance/instance.go
comtalyst: Given we can treat CSE error as other errors that can arrive there, then I don't think it changes the flow.

pkg/providers/instance/instance.go
comtalyst: Barely changes.

pkg/providers/instance/instance.go
comtalyst: For consistency.

pkg/providers/imagefamily/agentbakerbootstrap/provisionclientbootstrap.go
comtalyst: All this come un-golang namings from the auto-generated client.

pkg/providers/imagefamily/agentbakerbootstrap/provisionclientbootstrap.go
comtalyst: The nil check is right above these.
But, would `lo` helps on anything? It better not panic.

pkg/operator/options/options_validation.go
comtalyst: (discussed offline + in a different thread) This is intentional and we keep this.

pkg/providers/imagefamily/agentbakerbootstrap/provisionclientbootstrap.go
comtalyst: Required fields are marked in the swagger spec.

pkg/providers/imagefamily/agentbakerbootstrap/provisionclientbootstrap.go
comtalyst: This is an architectural gap that I am going to look into later. Basically, it is not as simple to be resolved as other values, for now.

pkg/providers/imagefamily/agentbakerbootstrap/provisionclientbootstrap.go
comtalyst: Could look into general retriability (e.g., from network/reliability issues) later.
But I don't think we should mix provision modes, for consistency.

pkg/providers/imagefamily/agentbakerbootstrap/provisionclientbootstrap.go
comtalyst: FYI those are already there in self-contained.
The new one we will get from the provision server (and AgentBaker) will be the below. 
The emptiness is intentional to prevent workload targeting them, for now.
```
agentpool=
kubernetes.azure.com/agentpool=
```

pkg/providers/imagefamily/azlinux.go
comtalyst: Will consider that along with potential delegation of this field to the provision server.

pkg/providers/imagefamily/agentbakerbootstrap/provisionclientbootstrap.go
```diff
@@ -0,0 +1,239 @@
...
+
+	if p.IsWindows {
+		provisionProfile.OsType = lo.ToPtr(models.OSType_Windows)
+	} else {
+		provisionProfile.OsType = lo.ToPtr(models.OSType_Linux)
+	}
+
+	provisionHelperValues := &models.ProvisionHelperValues{
+		SkuCPU:    lo.ToPtr(p.InstanceType.Capacity.Cpu().AsApproximateFloat64()),
+		SkuMemory: lo.ToPtr(p.InstanceType.Capacity.Memory().AsApproximateFloat64()),
```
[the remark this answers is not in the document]
comtalyst: (see the update, there is an issue I found and a workaround)

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -45,10 +45,14 @@ const (
...
 type Template struct {
-	UserData string
-	ImageID  string
-	SubnetID string
-	Tags     map[string]*string
+	SelfContainedUserData string
+	ImageID               string
+	SubnetID              string
+	Tags                  map[string]*string
+	AgentBakerCustomData  string
```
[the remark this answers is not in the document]
comtalyst: Apart from just that intention, I don't think we truly need to separate it. 
But I would prefer to keep it separated this before we are more confident in that.

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -45,10 +45,14 @@ const (
...
+	AgentBakerCSE         string
+	IsWindows             bool
```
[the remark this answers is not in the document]
comtalyst: I plan to spend more time on this category later. Will require a more thorough understanding of the abstraction intentions.

pkg/utils/gpu.go
comtalyst: We are not using SHAs. Just that `UseGridDrivers()`. But could still be a candidate for that consideration.

pkg/operator/options/options.go
```diff
@@ -57,6 +57,7 @@ func (s *nodeIdentitiesValue) Get() any { return []string(*s) }
 func (s *nodeIdentitiesValue) String() string { return strings.Join(*s, ",") }
 
 const (
+	ProvisionModeAKSSelfContained    = "aksselfcontained"
```
tallaxes, line 60: The current terminology is "scriptless", I believe

pkg/provisionclients/swagger/nodebootstrapping.json
```diff
@@ -0,0 +1,653 @@
...
+          "x-nullable": true
+        },
+        "containerLogMaxFiles": {
+          "type": "integer",
+          "format": "int32",
+          "x-nullable": true
+        },
+        "podMaxPids": {
+          "type": "integer",
+          "format": "int32",
```
tallaxes, line 272: Note that actual kubelet `podPidsLimit` is int64. Something we likely need to fix on the AKS side, might as well type it correctly here.

Makefile-az.mk
```diff
@@ -358,3 +358,11 @@ az-helm-install-snapshot: az-configure-values ## Install Karpenter snapshot rele
 
 az-rmcrds: ## Delete Karpenter CRDs
 	kubectl delete crd nodepools.karpenter.sh nodeclaims.karpenter.sh aksnodeclasses.karpenter.azure.com
+
+az-swagger-generate-clients-raw:
+	cd pkg/provisionclients && swagger generate client -f swagger/*.json
```
tallaxes, line 363: A couple of thoughts:
* You should be able to remove make targets / fold them into just "go generate" by creating e.g. provisionclients/generate.go with something like `//go:generate swagger generate client ...` etc.
* Take a look at generate client options, consider maybe `--keep-spec-order`?
* Could we maybe keep the spec in yaml instead of json?

pkg/providers/imagefamily/agentbakerbootstrap/provisionclientbootstrap_test.go
```diff
@@ -0,0 +1,55 @@
...
+
+import (
+	"fmt"
+	"math"
+	"testing"
+
+	"github.com/Azure/karpenter-provider-azure/pkg/providers/instancetype"
+)
+
+func TestReverseVMMemoryOverhead(t *testing.T) {
```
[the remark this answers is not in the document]
comtalyst: Better than not having it?
In case someone change something and it matters. And it is not too burdening resource-wise IMO.

pkg/provisionclients/swagger/nodebootstrapping.json
comtalyst: I think we should change ours after to not misalign with AKS requirements...

pkg/consts/consts.go
```diff
@@ -28,4 +28,7 @@ const (
 	NetworkDataplaneAzure  = "azure"
```
comtalyst, line 29: IMO, most of these should be in `options` to not have knowledge coupling.
But for now, let's keep it in one place for consistency.

pkg/operator/options/options.go
```diff
@@ -96,7 +97,7 @@ func (o *Options) AddFlags(fs *coreoptions.FlagSet) {
 	fs.StringVar(&o.SubnetID, "vnet-subnet-id", env.WithDefaultString("VNET_SUBNET_ID", ""), "The default subnet ID to use for new nodes. This must be a valid ARM resource ID for subnet that does not overlap with the service CIDR or the pod CIDR")
 	fs.Var(newNodeIdentitiesValue(env.WithDefaultString("NODE_IDENTITIES", ""), &o.NodeIdentities), "node-identities", "User assigned identities for nodes.")
 	fs.StringVar(&o.NodeResourceGroup, "node-resource-group", env.WithDefaultString("AZURE_NODE_RESOURCE_GROUP", ""), "[REQUIRED] the resource group created and managed by AKS where the nodes live")
-	fs.StringVar(&o.ProvisionMode, "provision-mode", env.WithDefaultString("PROVISION_MODE", ""), "[UNSUPPORTED] The provision mode for the cluster.")
+	fs.StringVar(&o.ProvisionMode, "provision-mode", env.WithDefaultString("PROVISION_MODE", "aksselfcontained"), "[UNSUPPORTED] The provision mode for the cluster.")
 	fs.StringVar(&o.NodeBootstrappingServerURL, "nodebootstrapping-server-url", env.WithDefaultString("NODEBOOTSTRAPPING_SERVER_URL", ""), "[UNSUPPORTED] The url for the node bootstrapping provider server.")
```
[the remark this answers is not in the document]
comtalyst: I think most of our options have the full stop in their descriptions.
Is not having it a right way?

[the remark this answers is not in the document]
tallaxes: No, full stop seems correct. Just wanted it to be consistent, and looked at surrounding ones and not the full set.

Makefile-az.mk
comtalyst: (will circle back to this later, along with the rest of the suggestions)

tallaxes (approved):

PR #528, opened 2024-10-16
feat: v1 migration and adaptation

tallaxes:
**Description**

Update to Karpenter v1 API, with many follow-ups. Summary of changes (mostly grouped by commits):

* **Karpenter core update**: update to v1.0.4, refresh and relink core CRDs, update main and alt operator; migrate cloud provider, operator, providers (convert to interfaces), controllers, test package and E2E tests to v1 API
* **AKSNodeClass**: introduce kubelet config; add nodeclass status, termination and hash controllers
* **Development**: bump Go version to 1.23, bump dependencies, refresh toolchain, fix linting (golang-ci was silently failing; move codegen into subfolders), update lint configuration (enable most of govet, exclude alt operator logging), restore linting on verify, refresh pre-commit components
* **Deployment**: deploy to `kube-system` by default, update Helm charts (many changes), update karpenter-values-template (remove drift feature gate); enable webhooks (including webhook injection into CRDs); includes enablement of webhooks in CCP context (thank you, @matthchr!)
* **Testing**: improve selection of karpenter pods for debug logs on failure, add events and simplify; fix utilization suite; add cilium label and taint; fix labels and disruption for daemonset test

* `make presubmit`
* E2E: 
  - https://github.com/Azure/karpenter-provider-azure/actions/runs/11454938746
  - https://github.com/Azure/karpenter-provider-azure/actions/runs/11506976735

- [x] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

**Next**

AKSNodeClass
* Propagate kubelet config (and add tests)
* Add more CEL constraints
* Add CEL tests
* Review conditions (including properly computing readiness)

Testing
* Test nodeclass hash
* Test drift on nodeclass static fields
* Review and fix integration tests (DS & emptiness?)
* Review nodeclaim GC tests
* Tests for setting nodeClaim.Status.ImageID

Other
* Review instance provider CreateTags
* NodeClaim tagging controller


charts/karpenter/templates/deployment.yaml
```diff
@@ -99,6 +102,32 @@ spec:
                   containerName: controller
                   divisor: "0"
                   resource: limits.memory
+            - name: FEATURE_GATES
```
Bryce-Soghigian, line 105: What's the behavior if we don't specify any feature gates?

charts/karpenter/values.yaml
```diff
@@ -173,29 +156,21 @@ settings:
...
+  # TODO: not propagated yet ...
+  tags:
+
   # -- Feature Gate configuration values. Feature Gates will follow the same graduation process and requirements as feature gates
   # in Kubernetes. More information here https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/#feature-gates-for-alpha-or-beta-features
   featureGates:
-    # -- drift is in ALPHA and is disabled by default.
-    # Setting drift to true enables the drift disruption method to watch for drift between currently deployed nodes
-    # and the desired state of nodes set in provisioners and node templates
-    drift: true
```
Bryce-Soghigian, line 201: It was promoted to stable for v1 right?

tallaxes: Yes, and that's why it is not behind a feature gate anymore and why this block is removed

charts/karpenter/templates/deployment.yaml
tallaxes: It is an option with default, which will be used if the value is not specified. In v1.0.x [the default is](https://github.com/kubernetes-sigs/karpenter/blob/7b24b44084d10c028345bb0b2acbd6633af822c6/pkg/operator/options/options.go#L102) "SpotToSpotConsolidation=false", but it is likely to change in v1.1.x. In our provider we would want to keep it false until properly supported and tested.

comtalyst:
Haven't look at specifics of motivations, but overall LGTM.

comtalyst (dismissed):

comtalyst (dismissed):

Makefile
```diff
@@ -80,9 +80,12 @@ verify: toolchain tidy download ## Verify code. Includes dependencies, linting,
 	cp $(KARPENTER_CORE_DIR)/pkg/apis/crds/* pkg/apis/crds
 	yq -i '(.spec.versions[0].additionalPrinterColumns[] | select (.name=="Zone")) .jsonPath=".metadata.labels.karpenter\.azure\.com/zone"' \
 		pkg/apis/crds/karpenter.sh_nodeclaims.yaml
+	hack/validation/kubelet.sh
 	hack/validation/labels.sh
 	hack/validation/requirements.sh
 	hack/validation/common.sh
+	cp pkg/apis/crds/* charts/karpenter-crd/templates
```
charliedmcb, line 87: Curious why this changed?

Wasn't the mirroring of info already handled?

charts/karpenter/crds/karpenter.azure.com_aksnodeclasses.yaml
charliedmcb, line 1: Curious why these are still using the ref form, while the other file using the ref form now got changed to copy over in verify?

pkg/cloudprovider/drift.go
```diff
@@ -40,11 +40,57 @@ import (
 )
 
 const (
+	NodeClassDrift    cloudprovider.DriftReason = "NodeClassDrift"
```
charliedmcb, line 43: While you're at it, and adding a new one, do you want to add a prefix to these?

charliedmcb: Could use the same Group you are using elsewhere?
https://github.com/Azure/karpenter-provider-azure/blob/07052f3cbdcd6944ee5d9dde8e9b7b089aa3ffec/pkg/apis/apis.go#L31

pkg/cloudprovider/drift.go
```diff
@@ -40,11 +40,57 @@ import (
...
 	K8sVersionDrift   cloudprovider.DriftReason = "K8sVersionDrift"
 	ImageVersionDrift cloudprovider.DriftReason = "ImageVersionDrift"
 	SubnetDrift       cloudprovider.DriftReason = "SubnetDrift"
 )
 
+func (c *CloudProvider) isNodeClassDrifted(ctx context.Context, nodeClaim *karpv1.NodeClaim, nodeClass *v1alpha2.AKSNodeClass) (cloudprovider.DriftReason, error) {
+	// First check if the node class is statically drifted to save on API calls.
+	if drifted := c.areStaticFieldsDrifted(nodeClaim, nodeClass); drifted != "" {
```
charliedmcb, line 51: nit: prefer naming to be more detailed than `drifted`.

maybe `staticFieldsDrifted`

pkg/cloudprovider/drift.go
```diff
@@ -40,11 +40,57 @@ import (
...
+}
+
+func (c *CloudProvider) areStaticFieldsDrifted(nodeClaim *karpv1.NodeClaim, nodeClass *v1alpha2.AKSNodeClass) cloudprovider.DriftReason {
+	nodeClassHash, foundNodeClassHash := nodeClass.Annotations[v1alpha2.AnnotationAKSNodeClassHash]
+	nodeClassHashVersion, foundNodeClassHashVersion := nodeClass.Annotations[v1alpha2.AnnotationAKSNodeClassHashVersion]
+	nodeClaimHash, foundNodeClaimHash := nodeClaim.Annotations[v1alpha2.AnnotationAKSNodeClassHash]
+	nodeClaimHashVersion, foundNodeClaimHashVersion := nodeClaim.Annotations[v1alpha2.AnnotationAKSNodeClassHashVersion]
+
+	if !foundNodeClassHash || !foundNodeClaimHash || !foundNodeClassHashVersion || !foundNodeClaimHashVersion {
+		return ""
```
charliedmcb, line 85: Not for your PR, but realizing that we should probably have a const of the form:
`const NotDrifted = ""`

We can then use that for all the instances of returning "" signaling non-drift for clarity.

tallaxes: Ack, for later

pkg/cloudprovider/drift.go
tallaxes: I do like the idea of prefix, but don't want to do it in a rush (e.g. would be good to check if there are limits on length somewhere, etc.); let's circle back to this later

Makefile
tallaxes: The change likely has to do with webhook configuration injection, see hack/mutation/conversion_webhooks_injection.sh

charts/karpenter/crds/karpenter.azure.com_aksnodeclasses.yaml
tallaxes: Will review, but likely the same reason: Other's need webhook injection, this one does not.

Bryce-Soghigian (dismissed):

hack/validation/kubelet.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+set -euo pipefail
+
+# Kubelet Validation (placeholder)
```
charliedmcb, line 4: Add this to the follow ups? Or is this meant as part of the "and tests" for kubelet part?

pkg/apis/v1alpha2/aksnodeclass.go
```diff
@@ -46,11 +46,109 @@ type AKSNodeClassSpec struct {
...
+	// Default: "100ms"
+	// +optional
+	// +kubebuilder:default="100ms"
+	// TODO: validation
+	CPUCFSQuotaPeriod metav1.Duration `json:"cpuCFSQuotaPeriod,omitempty"`
+	// ImageGCHighThresholdPercent is the percent of disk usage after which image
+	// garbage collection is always run. The percent is calculated by dividing this
+	// field value by 100, so this field must be between 0 and 100, inclusive.
+	// When specified, the value must be greater than ImageGCLowThresholdPercent.
+	// Note: AKS AKS CustomKubeletConfig does not have "Percent" in the field name
```
charliedmcb, line 91: typo? `AKS` twice

pkg/apis/v1alpha2/aksnodeclass.go
```diff
@@ -46,11 +46,109 @@ type AKSNodeClassSpec struct {
...
 
+// KubeletConfiguration defines args to be used when configuring kubelet on provisioned nodes.
+// They are a subset of the upstream types, recognizing not all options may be supported.
+// Wherever possible, the types and names should reflect the upstream kubelet types.
+// https://pkg.go.dev/k8s.io/kubelet/config/v1beta1#KubeletConfiguration
+// https://github.com/kubernetes/kubernetes/blob/9f82d81e55cafdedab619ea25cabf5d42736dacf/cmd/kubelet/app/options/options.go#L53
+//
+// AKS CustomKubeletConfig w/o CPUReserved,MemoryReserved,SeccompDefault
+// https://learn.microsoft.com/en-us/azure/aks/custom-node-configuration?tabs=linux-node-pools
+type KubeletConfiguration struct {
```
charliedmcb, line 69: Just out of curiousity, where are all the fields for Kubelet Config we are supporting coming from?

K8s docs, AKS, Upstream, AWS refs?

pkg/apis/v1alpha2/aksnodeclass.go
```diff
@@ -46,11 +46,109 @@ type AKSNodeClassSpec struct {
...
+	// +optional
+	ContainerLogMaxFiles *int32 `json:"containerLogMaxFiles,omitempty"`
+	// podPidsLimit is the maximum number of PIDs in any pod.
+	// AKS CustomKubeletConfig uses PodMaxPids, int32 (!)
+	// Default: -1
+	// +optional
+	PodPidsLimit *int64 `json:"podPidsLimit,omitempty"`
+}
+
+// TODO: add hashing support
```
charliedmcb, line 147: Can be removed now?

pkg/apis/v1alpha2/aksnodeclass.go
tallaxes: Primarily from AKS CustomKubeletConfig (as the comment says)

pkg/apis/v1alpha2/aksnodeclass.go
```diff
@@ -60,6 +158,13 @@ type AKSNodeClass struct {
 	Status AKSNodeClassStatus `json:"status,omitempty"`
 }
 
+// We need to bump the AKSNodeClassHashVersion when we make an update to the AKSNodeClass CRD under these conditions:
```
charliedmcb, line 161: This significantly concerns me. I could easily see changing this being missed. However, I don't have a great other solution.

Need good unit tests for this to try and ensure we catch all the cases where this would need to be updated, so folks are aware, inspect the failures, and update this.

pkg/apis/v1alpha2/aksnodeclass.go
```diff
@@ -60,6 +158,13 @@ type AKSNodeClass struct {
...
+// 1. A field changes its default value for an existing field that is already hashed
+// 2. A field is added to the hash calculation with an already-set value
+// 3. A field is removed from the hash calculations
+const AKSNodeClassHashVersion = "v1"
+
+// TODO: add hash tests
```
charliedmcb, line 167: Yes, please!

hack/validation/kubelet.sh
tallaxes: This is for special cases of CEL validation, which is already on the list

pkg/apis/v1alpha2/aksnodeclass.go
charliedmcb: Thx! Somehow I missed that.

pkg/apis/v1alpha2/aksnodeclass.go
tallaxes: Ack. There is a design doc around this that I will link to

pkg/apis/v1alpha2/aksnodeclass.go
tallaxes: In the list of followups 🙂

comtalyst (approved):

charts/karpenter/templates/clusterrole-core.yaml
```diff
@@ -36,16 +36,16 @@ rules:
...
   - apiGroups: ["storage.k8s.io"]
-    resources: ["storageclasses", "csinodes"]
+    resources: ["storageclasses", "csinodes", "volumeattachments"]
     verbs: ["get", "watch", "list"]
   - apiGroups: ["apps"]
     resources: ["daemonsets", "deployments", "replicasets", "statefulsets"]
     verbs: ["list", "watch"]
-{{- if .Values.webhook.enabled }}
-  - apiGroups: ["admissionregistration.k8s.io"]
-    resources: ["validatingwebhookconfigurations", "mutatingwebhookconfigurations"]
```
charliedmcb, line 46: We aren't doing any validation, or mutating webhooks, right? Its just conversion webhooks?

comtalyst: Yes, just conversion for now.

cmd/controller/main.go
```diff
@@ -47,18 +46,22 @@ func main() {
...
 		WithControllers(ctx, controllers.NewControllers(
 			ctx,
+			op.Manager,
 			op.GetClient(),
+			op.EventRecorder,
 			aksCloudProvider,
 			op.InstanceProvider,
 		)...).
-		Start(ctx)
+		//WithWebhooks(ctx, webhooks.NewWebhooks()...).
```
charliedmcb, line 65: Commented out? Should this be removed, or added?

tallaxes: Just is a placeholder for where we would wire provider-specific webhooks (e.g. conversion for AKSNodeClass), would like to keep it as comment

pkg/alt/karpenter-core/pkg/operator/operator.go
```diff
@@ -66,14 +76,26 @@ const (
 	component = "controller"
 )
 
+type ccpClient struct{}
```
charliedmcb, line 79: nit: `ccpCleintKey`

Don't worry about this for release

pkg/alt/karpenter-core/pkg/operator/operator.go
```diff
@@ -66,14 +76,26 @@ const (
...
+
+func withCCPClient(ctx context.Context, client kubernetes.Interface) context.Context {
+	return context.WithValue(ctx, ccpClient{}, client)
```
charliedmcb, line 82: Do you not need the key to actually be the same object?

I thought you would define a "const" instance of the struct as the ccpCleintKey, and then use it for Set, and Get, no?

pkg/alt/karpenter-core/pkg/operator/operator.go
charliedmcb: Actually maybe a bug?

See other comment:
https://github.com/Azure/karpenter-provider-azure/pull/528/files#r1815782494

pkg/alt/karpenter-core/pkg/operator/operator.go
tallaxes: No, from what I remember it is only the type that matters. @matthchr

charliedmcb: Ok, saw a few more examples of this, so tried running a test, and doesn't appear to be a difference:
https://go.dev/play/p/vQY7n8VY4C9

charliedmcb: Did look around online and saw the patterning of:
> type contextKey string
> 
> const ccpCleintKey contextKey = "ccpCleintKey"

However, seems also like a fine usage, and not a bug the way it is. No needed action here, but thought I'd leave what I fount

pkg/alt/karpenter-core/pkg/operator/operator.go
charliedmcb: Ok, not a bug. See the other comment for detailed test, and further comments on this patterning

pkg/alt/knative/pkg/webhook/resourcesemantics/conversion/reconciler.go
```diff
@@ -0,0 +1,123 @@
...
+	if err != nil {
+		return fmt.Errorf("error retrieving crd: %w", err)
+	}
+
+	crd := configuredCRD.DeepCopy()
+
+	if crd.Spec.Conversion == nil ||
+		crd.Spec.Conversion.Strategy != apixv1.WebhookConverter ||
+		crd.Spec.Conversion.Webhook.ClientConfig == nil ||
+		(crd.Spec.Conversion.Webhook.ClientConfig.Service == nil && crd.Spec.Conversion.Webhook.ClientConfig.URL == nil) { // AKS customized
```
charliedmcb, line 98: Is this `AKS customized` block the if, or the return?

tallaxes: The only customization is allowing the URL (in addition to the Service)

pkg/alt/karpenter-core/pkg/operator/operator.go
matthchr: I copied this pattern from knative which does it this way (key is basically a bare struct which I think all compare the same regardless of if it's the same instance or not).

You can see that referenced in the context.WithValue docs too:
https://pkg.go.dev/context#WithValue

> The provided key must be comparable and should not be of type string or any other built-in type to avoid collisions between packages using context. Users of WithValue should define their own types for keys. To avoid allocating when assigning to an interface{}, **context keys often have concrete type struct{}**

pkg/alt/knative/pkg/webhook/resourcesemantics/conversion/table_test.go
```diff
@@ -0,0 +1,290 @@
...
+		Key:  key,
+		Objects: []runtime.Object{&corev1.Secret{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      secretName,
+				Namespace: system.Namespace(),
+			},
+			Data: map[string][]byte{
+				certresources.ServerKey:  []byte("present"),
+				certresources.ServerCert: []byte("present"),
+				// certresources.CACert:     []byte("missing"),
```
charliedmcb, line 71: commented, can be removed?

pkg/apis/apis.go
```diff
@@ -20,28 +20,18 @@ package apis
...
-		v1alpha2.SchemeBuilder.AddToScheme,
-	)
-	// AddToScheme may be used to add all resources defined in the project to a Scheme
-	AddToScheme = Builder.AddToScheme
 )
 
 //go:generate controller-gen crd object:headerFile="../../hack/boilerplate.go.txt" paths="./..." output:crd:artifacts:config=crds
 var (
+	Group = "karpenter.azure.com"
+	//CompatibilityGroup = "compatibility." + Group
```
charliedmcb, line 32: commented out, can be removed?

pkg/apis/crds/karpenter.azure.com_aksnodeclasses.yaml
```diff
@@ -73,6 +184,64 @@ spec:
...
+                    reason:
+                      description: |-
+                        reason contains a programmatic identifier indicating the reason for the condition's last transition.
+                        Producers of specific condition types may define expected values and meanings for this field,
+                        and whether the values are considered a guaranteed API.
+                        The value should be a CamelCase string.
+                        This field may not be empty.
+                      maxLength: 1024
+                      minLength: 1
+                      pattern: ^[A-Za-z]([A-Za-z0-9_,:]*[A-Za-z0-9_])?$
```
charliedmcb, line 223: If it should be camelCase, should we enforce the first letter is lowercase?

pkg/apis/crds/karpenter.azure.com_aksnodeclasses.yaml
```diff
@@ -73,6 +184,64 @@ spec:
...
+                      type: string
+                    status:
+                      description: status of the condition, one of True, False, Unknown.
+                      enum:
+                      - "True"
+                      - "False"
+                      - Unknown
```
charliedmcb, line 230: Should `Unknown` also be quoted?

pkg/apis/crds/karpenter.azure.com_aksnodeclasses.yaml
charliedmcb, line 223: Should we enforce starting with a lowercase letter if it should be camel case?

pkg/apis/crds/karpenter.azure.com_aksnodeclasses.yaml
charliedmcb, line 230: Should `Unknown` be quoted?

PR #530, opened 2024-10-17
Replace VNET_SUBNET_ID source in configure-values.sh.

CorrenSoft:
Fixes #

**Description**
In scenarios where the VNET is not managed for AKS (meaning it is not present in the node resource group), this script will not correctly populate that value. It is true that it is possible to assign that value using environment variables, but that behavior is a little obfuscated and can be confusing at first.

So, instead of retrieving the VNets from the node resource group, I propose getting the subnet directly from the cluster itself (which will work in both scenarios), making it easier for beginners and removing a call to Azure (too much of a performance fan?).

Notes:
- No sure what to use as a prefix in the PR name.
- Let me know if I need to also create an issue for this.

* Execute the command and check for the value in karpenter-values.yaml
`chmod +x ./configure-values.sh && ./configure-values.sh ${CLUSTER_NAME} ${RG} karpenter-sa karpentermsi `

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**
```release-note
NONE
```


tallaxes (approved):
Looks good, and appreciate the contribution!

tallaxes:
Not sure this works in cases where VNET _is_ managed by AKS

CorrenSoft:
> This does not work in cases where VNET _is_ managed by AKS, reverting for now.

What a rookie mistake... sorry for that.
I will re-do it, considering both scenarios.

PR #526, opened 2024-10-16
feat: ListNodeImageVersions + shared image gallery support

Bryce-Soghigian:
Fixes # 

**Description**
This pr introduces 
1. USE_SIG option that is used to control if we should or should not use shared image galleries
2. Shared Image Gallery ID Resolution
3. New ListNodeImageVersions logic for getting the latest approved agentbaker node image version rather than using galleries as a source of truth 

**Open Qs**: 
    Are we ok with all Nodes with Community Image gallery node images being drifted over to use SIG in managed karpenter with this change? 
    Should we implement the windows side changes too for the NodeImageVersions api? 

**How was this change tested?**
- manual testing on MSFT Tenant SIG Galleries for Ubuntu and Azure Linux
- make az-all 
- make test 

**Does this change impact docs?**
- [ ] Yes, PR includes docs updates
- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

PR #531, opened 2024-10-17
BREAKING CHANGE(aksnodeclass): removing imageVersion from CRD

charliedmcb:
Fixes # 

**Description**
We are working to improve the system of node image upgrades, and one of the first steps is removing the option to perminently lock the image version to any arbirary value with no given validation. Currently this is possible through the imageVersion field of the aksnodeclass. We are removing it from the CRD as a first step. 

Note: this will be a breaking change with the following effects:
- Existing nodes will drift to latest if currently specifying imageVersion
- Trying to apply any saved spec with imageVersion will fail, and need to be updated.

Tested impact in a self-hosted environment by preforming the following steps (equivalent to what the first commit of this PR is):
- Created an AKSNodeClass specifying imageVersion
- Created a deployment that made karpenter provision 5 nodes
- Updated the CRDs to remove imageVersion. (note: keeping the same api version in the crd)
- The existing nodeclaims for the 5 nodes were marked as drifted, and brought up to the latest image version karpenter could find
- Tried applying the old AKSNodeClass with imageVersion, and got an error.

I also ran 2 sets of the E2E tests, which both had a failure, but in different suites. I don't think the failures are a cause of this PR:
- https://github.com/Azure/karpenter-provider-azure/actions/runs/11393658202
- https://github.com/Azure/karpenter-provider-azure/actions/runs/11393395394

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
BREAKING CHANGE: removing imageVersion from AKSNodeClass. Impact: existing nodes will drift to latest if currently specifying imageVersion. action required: trying to apply any saved spec with imageVersion will fail, and need to be updated.
```


pkg/apis/crds/karpenter.azure.com_aksnodeclasses.yaml
```diff
@@ -53,9 +53,6 @@ spec:
                 - Ubuntu2204
                 - AzureLinux
                 type: string
-              imageVersion:
-                description: ImageVersion is the image version that instances use.
```
Bryce-Soghigian, line 57: what is the plan for replacing it?

Bryce-Soghigian: https://github.com/Azure/karpenter-provider-azure/pull/526 relies on this field so just removing it feels wrong? WIll we no longer drift on imageVersion? whats the new upgrade plan

tallaxes (approved):

PR #540, opened 2024-10-24
feat: Add auxiliary token policy and update VMsClient to use in client options

rakechill:
**Description**
This PR adds:
* `GetAuxiliaryToken()` which will return a token that can authenticate with shared node image galleries for managed Karpenter users.
* `AuxiliaryTokenPolicy` which provides a policy to add an auxiliary token to a request
* Updates `VirtualMachinesClient` to use this policy in its opts

* [WIP] e2e testing environment with aux token dependency
* unit tests w/ options + image provider list

- [ ] Yes, issue opened: # 
- [X] No

**Release Note**

```release-note

```


pkg/auth/cred.go
```diff
@@ -33,6 +37,38 @@ type expireEarlyTokenCredential struct {
...
 
+func GetAuxiliaryToken(ctx context.Context, scope string) (azcore.AccessToken, error) {
+	client := &http.Client{}
+	var token azcore.AccessToken
+	req, err := http.NewRequest("GET", "msi-connector.msi-connector.svc.cluster.local/karpenter", nil)
+	if err != nil {
+		return token, err
+	}
+	q := req.URL.Query()
+	q.Add("scope", scope)
```
rakechill, line 48: Make this a list of scopes instead. "scopes": []string{}

pkg/auth/cred.go
```diff
@@ -33,6 +37,38 @@ type expireEarlyTokenCredential struct {
 	cred azcore.TokenCredential
 }
 
+func GetAuxiliaryToken(ctx context.Context, scope string) (azcore.AccessToken, error) {
+	client := &http.Client{}
+	var token azcore.AccessToken
+	req, err := http.NewRequest("GET", "msi-connector.msi-connector.svc.cluster.local/karpenter", nil)
+	if err != nil {
+		return token, err
```
rakechill, line 45: return nil, err

pkg/auth/cred.go
```diff
@@ -33,6 +37,48 @@ type expireEarlyTokenCredential struct {
 	cred azcore.TokenCredential
 }
 
+func GetAuxiliaryToken(ctx context.Context) (azcore.AccessToken, error) {
```
tallaxes, line 40: This code is not covered by any unit tests. Ideally we would have an acceptance unit test that checks that during provisioning calls to create VM carry additional token in SIG scenario, but unfortunately our current fakes are not flexible enough for this. More focused unit testing of just GetAuxilaryToken with something like a mock server would be second best. If we have E2E tests that cover these scenarios - this becomes less critical, and can be added later.

tallaxes (commented):
Overall LGTM. Left minor suggestions. Could also additional tests (see comment), but IMO could wait if covered by E2E tests in SIG environment.

pkg/auth/cred.go
```diff
@@ -33,6 +37,48 @@ type expireEarlyTokenCredential struct {
...
+	client := &http.Client{Timeout: 10 * time.Second}
+	url := options.FromContext(ctx).AuxiliaryTokenServerURL
+	if url == "" {
+		return azcore.AccessToken{}, fmt.Errorf("auxiliary token server URL is not set")
+	}
+
+	// TODO: this env var needs to be customized based on cloud
+	scope := options.FromContext(ctx).SIGScope
```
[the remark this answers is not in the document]
rakechill: maybe pass sig scope + url then?

pkg/operator/options/options_validation.go
```diff
@@ -37,6 +37,7 @@ func (o Options) Validate() error {
 		o.validateVMMemoryOverheadPercent(),
 		o.validateVnetSubnetID(),
 		o.validateProvisionMode(),
+		o.validateUseSIG(),
```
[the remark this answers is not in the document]
rakechill: I think since we're setting on NAP + run e2e, it _should_ be fine, but happy to go ahead and add similar validation that we use for other sub IDs

pkg/operator/options/suite_test.go
```diff
@@ -56,6 +56,10 @@ var _ = Describe("Options", func() {
 		"PROVISION_MODE",
 		"NODEBOOTSTRAPPING_SERVER_URL",
 		"VNET_GUID",
+		"USE_SIG",
+		"AUXILIARY_TOKEN_SERVER_URL",
```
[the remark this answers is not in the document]
rakechill: I kind of think it's fine since this is how we are using it. We could instead call it "SIG_ACCESS_TOKEN_URL" which would also be accurate + more obviously connect it with SIG

pkg/providers/imagefamily/nodeimage_test.go
```diff
@@ -38,7 +38,8 @@ import (
 )
 
 const (
-	subscription = "12345678-1234-1234-1234-123456789012"
+	customerSubscription = "12345678-1234-1234-1234-123456789012"
+	SIGSubscription      = "10945678-1234-1234-1234-123456789012"
```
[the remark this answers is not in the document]
rakechill: ah yes, ack

pkg/operator/options/suite_test.go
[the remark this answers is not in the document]
tallaxes: If there is any chance there may be additional uses for it - let's keep the name

pkg/operator/options/options.go
```diff
@@ -102,6 +103,8 @@ func (o *Options) AddFlags(fs *coreoptions.FlagSet) {
...
 	fs.StringVar(&o.NodeResourceGroup, "node-resource-group", env.WithDefaultString("AZURE_NODE_RESOURCE_GROUP", ""), "[REQUIRED] the resource group created and managed by AKS where the nodes live")
 	fs.BoolVar(&o.UseSIG, "use-sig", env.WithDefaultBool("USE_SIG", false), "If set to true karpenter will use the AKS managed shared image galleries and the node image versions api. If set to false karpenter will use community image galleries. Only a subset of image features will be available in the community image galleries and this flag is only for the managed node provisioning addon.")
+	fs.StringVar(&o.AuxiliaryTokenServerURL, "auxiliary-token-server-url", env.WithDefaultString("AUXILIARY_TOKEN_SERVER_URL", ""), "The url for the auxiliary token server. Only used for AKS managed karpenter. UseSIG must be set tot true for this to take effect.")
+	fs.StringVar(&o.SIGScope, "sig-scope", env.WithDefaultString("SIG_SCOPE", ""), "The scope for the auxiliary token. Only used for AKS managed karpenter. UseSIG must be set to true for this to take effect.")
```
tallaxes, line 107: Feels like this could use a better name too; maybe Auxiliary Token Scope? And I presume you have good reasons for making this configurable.

rakechill: I think we could easily call it `SIG_ACCESS_TOKEN_SCOPE`. I actually like the idea of it explicitly being called `SIG-***`. Could we use this token for smth else along down the line? _maybe_, but we don't have any use cases atm.

The scope is needed, because it differs based on cloud. If we support sov cloud w/ this approach, then we'll have to provide a different scope.

pkg/operator/options/suite_test.go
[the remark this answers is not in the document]
rakechill: yep making that change now ^^
final variables will be:
```
USE_SIG
SIG_ACCESS_TOKEN_SERVER_URL
SIG_ACCESS_TOKEN_SCOPE
SIG_SUBSCRIPTION_ID
```

pkg/operator/options/suite_test.go
```diff
@@ -323,7 +335,7 @@ var _ = Describe("Options", func() {
 				"--ssh-public-key", "flag-ssh-public-key",
 				"--provision-mode", "ekeselfexposed",
 			)
-			Expect(err).To(MatchError(ContainSubstring("provision-mode")))
```
rakechill, line 326: this was edited, bc both error pathways contain `provision-mode` in this validator func: https://github.com/Azure/karpenter-provider-azure/blob/7f007f15a9039ee9dd0714fa6adb188e9679915d/pkg/operator/options/options_validation.go#L96-L105

pkg/providers/instance/azure_client.go
```diff
@@ -129,7 +134,17 @@ func NewAZClient(ctx context.Context, cfg *auth.Config, env *azure.Environment)
 	}
 	klog.V(5).Infof("Created network interface client %v using token credential", interfacesClient)
 
-	virtualMachinesClient, err := armcompute.NewVirtualMachinesClient(cfg.SubscriptionID, cred, opts)
+	var vmClientOptions *armpolicy.ClientOptions
+	o := options.FromContext(ctx)
+	if o.UseSIG {
+		klog.V(1).Info("Using SIG for image versions")
+		url, scope := o.SIGAccessTokenServerURL, o.SIGAccessTokenScope
```
rakechill, line 141: nit: can move into line 142

pkg/providers/instance/skuclient/skuclient.go
```diff
@@ -60,7 +60,6 @@ func (sc *skuClient) updateInstance() {
 	authorizer := azidext.NewTokenCredentialAdapter(cred, []string{azidext.DefaultManagementScope})
 
 	azClientConfig := sc.cfg.GetAzureClientConfig(authorizer, sc.env)
-	azClientConfig.UserAgent = auth.GetUserAgentExtension()
```
rakechill, line 63: moved to directly set in client config for all clients.

pkg/operator/options/options_validation.go
```diff
@@ -123,3 +121,31 @@ func (o Options) validateRequiredFields() error {
...
+		if o.SIGAccessTokenScope == "" {
+			return fmt.Errorf("sig-access-token-scope is required when use-sig is true")
+		}
+		if o.SIGSubscriptionID == "" {
+			return fmt.Errorf("sig-subscription-id is required when use-sig is true")
+		}
+		if !isValidURL(o.SIGAccessTokenServerURL) {
+			return fmt.Errorf("sig-access-token-server-url is not a valid URL")
+		}
+		if !isValidURL(o.SIGAccessTokenScope) {
```
tallaxes, line 139: Not positive scopes are actually valid URLs in all cases, would be good to check

pkg/operator/options/suite_test.go
```diff
@@ -100,6 +104,10 @@ var _ = Describe("Options", func() {
 			os.Setenv("VNET_SUBNET_ID", "/subscriptions/12345678-1234-1234-1234-123456789012/resourceGroups/sillygeese/providers/Microsoft.Network/virtualNetworks/karpentervnet/subnets/karpentersub")
 			os.Setenv("PROVISION_MODE", "bootstrappingclient")
 			os.Setenv("NODEBOOTSTRAPPING_SERVER_URL", "https://nodebootstrapping-server-url")
+			os.Setenv("USE_SIG", "true")
+			os.Setenv("SIG_ACCESS_TOKEN_SERVER_URL", "http://valid-server.com")
+			os.Setenv("SIG_ACCESS_TOKEN_SCOPE", "http://valid-scope.com")
```
tallaxes, line 109: Actual scopes might have something after? Like "/.default"?

pkg/operator/options/options.go
```diff
@@ -102,6 +103,8 @@ func (o *Options) AddFlags(fs *coreoptions.FlagSet) {
 	fs.StringVar(&o.NodeBootstrappingServerURL, "nodebootstrapping-server-url", env.WithDefaultString("NODEBOOTSTRAPPING_SERVER_URL", ""), "[UNSUPPORTED] The url for the node bootstrapping provider server.")
 	fs.StringVar(&o.NodeResourceGroup, "node-resource-group", env.WithDefaultString("AZURE_NODE_RESOURCE_GROUP", ""), "[REQUIRED] the resource group created and managed by AKS where the nodes live")
 	fs.BoolVar(&o.UseSIG, "use-sig", env.WithDefaultBool("USE_SIG", false), "If set to true karpenter will use the AKS managed shared image galleries and the node image versions api. If set to false karpenter will use community image galleries. Only a subset of image features will be available in the community image galleries and this flag is only for the managed node provisioning addon.")
+	fs.StringVar(&o.SIGAccessTokenServerURL, "sig-access-token-server-url", env.WithDefaultString("SIG_ACCESS_TOKEN_SERVER_URL", ""), "The url for the SIG access token server. Only used for AKS managed karpenter. UseSIG must be set tot true for this to take effect.")
```
tallaxes, line 106:
```suggestion
	fs.StringVar(&o.SIGAccessTokenServerURL, "sig-access-token-server-url", env.WithDefaultString("SIG_ACCESS_TOKEN_SERVER_URL", ""), "The URL for the SIG access token server. Only used for AKS managed karpenter. UseSIG must be set tot true for this to take effect.")
```

tallaxes (approved):
LGTM, left minor suggestions

pkg/operator/options/options_validation.go
rakechill: yeah I'll update the test URLs to have `./default` just to be extra sure

pkg/operator/options/suite_test.go
```diff
@@ -336,6 +348,73 @@ var _ = Describe("Options", func() {
...
 		})
+		It("should fail if use-sig is enabled, but sig-access-token-server-url is not set", func() {
+			err := opts.Parse(
+				fs,
+				"--cluster-name", "my-name",
+				"--cluster-endpoint", "https://karpenter-000000000000.hcp.westus2.staging.azmk8s.io",
+				"--kubelet-bootstrap-token", "flag-bootstrap-token",
+				"--ssh-public-key", "flag-ssh-public-key",
+				"--sig-access-token-scope", "http://valid-scope.com",
+				"--sig-subscription-id", "my-subscription-id",
```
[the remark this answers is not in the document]
rakechill: it matters basically as much as the URL + scope matter (we'd hope that we properly set it for NAP users)

Bryce-Soghigian (approved):

PR #542, opened 2024-10-25
feat: Artifact Streaming integration for bootstrappingclient mode

comtalyst:
Fixes # 

**Description**
This will enable [Artifact Streaming](https://learn.microsoft.com/en-us/azure/aks/artifact-streaming) by default for nodes provisioned with `bootstrappingclient` mode.

*

- [ ] Yes, issue opened: # 
- [X] No

**Release Note**

```release-note
NONE
```


Bryce-Soghigian (approved):
Should we add an integration test validating that image pull is done faster? Maybe with one of the cocalc larger images? We can have a timeout on image pull to validate that streaming is working since the latency is so significant without artifact streaming.

image

comtalyst:
This one haven't incorporate [preview limitations](https://learn.microsoft.com/en-us/azure/container-registry/container-registry-artifact-streaming?pivots=development-environment-azure-portal#preview-limitations) yet. Will evaluate later if we decide to pursue this during its preview.
For now, this PR stays as draft and should not be merged.

PR #544, opened 2024-10-26
fix: follow-up fixes on bootstrappingclient mode after integration testing with v1

comtalyst:
Fixes # 

**Description**
KubeletConfig retrieved from nodeClaim doesn't seem to have a proper defaulting(?), which could returns invalid values (e.g., out of range) when populating ProvisionProfile. The logic added here should isolate non-pointer values (which cannot be "unpopulated" by being nil) and have an empty check.

Also, new unregistered startup taints defaulting method went under the radar during the merge, making it ineffective for CustomScripts bootstrapping. This should fix it, and have similar spots scanned.

* Manual provisioning test
* Inspecting ProvisionProfile values

- [ ] Yes, issue opened: # 
- [X] No

**Release Note**

```release-note
NONE
```


tallaxes (approved):
LGTM. Would want to review the defaulting logic, but ok with this going in for now.

PR #551, opened 2024-10-28
Adding subnet-id retrieval from Cluster.

CorrenSoft:
Fixes #

**Description**
In scenarios where the VNET is not managed for AKS (meaning it is not present in the node resource group), this script will not correctly populate that value. It is possible to assign that value using environment variables, but that behavior is a little obfuscated and can be confusing at first.

This PR updates the SUBNET_ID logic to also look for the information in the agent profiles, in the following order:
1. Check if the variable already has a value (from Environment).
2. If empty, Check for the value on agentProfiles, using the already loaded `AKS_JSON` variable.
3. If null, search for the aks-managed subnet in the Nodes Resource Group.

* Execute the command and check for the value in karpenter-values.yaml
`chmod +x ./configure-values.sh && ./configure-values.sh ${CLUSTER_NAME} ${RG} karpenter-sa karpentermsi `

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**
```release-note
NONE
```

### DIFF
--- a/hack/deploy/configure-values.sh
+++ b/hack/deploy/configure-values.sh
@@ -31,12 +31,17 @@ BOOTSTRAP_TOKEN=$TOKEN_ID.$TOKEN_SECRET
 SSH_PUBLIC_KEY="$(cat ~/.ssh/id_rsa.pub) azureuser"
 
 if [[ ! -v VNET_SUBNET_ID ]]; then
-    # first subnet of first VNet found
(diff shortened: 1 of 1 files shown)


Bryce-Soghigian:
Closing this in favor of https://github.com/Azure/karpenter-provider-azure/pull/676 which we ended up touching this same code path for populating vnet guid

Issue #579, opened 2024-11-18
[BUG] GPU nodes fail to join the cluster


HakjunMIN:
### Version

Self hosted karpneter 0.7.0 with the custom version supporting H100 (https://github.com/Azure/karpenter-provider-azure/pull/553)

### Expected Behavior

node should join the cluster. Even after `az aks update with empty`

### Actual Behavior

nodeclaim is repeatedly created but failed to join the cluster

### Steps to Reproduce the Problem

generated nodeclass and node pool for A100 
Suspicious the version of 2204gen2containerd/versions/202411.12.0 in nodeclass.

### Resource Specs and Logs

* nodeclaim status
```bash
k get nodeclaim -oyaml
apiVersion: v1
items:
- apiVersion: karpenter.sh/v1
  kind: NodeClaim
  metadata:
    annotations:
      karpenter.azure.com/aksnodeclass-hash: "17390774718828605676"
      karpenter.azure.com/aksnodeclass-hash-version: v1
      karpenter.azure.com/in-place-update-hash: "383245467"
      karpenter.sh/nodepool-hash: "1090957544758137294"
      karpenter.sh/nodepool-hash-version: v3
      karpenter.sh/stored-version-migrated: "true"
    creationTimestamp: "2024-11-18T06:58:23Z"
    finalizers:
    - karpenter.sh/termination
    generateName: a100-
    generation: 1
    labels:
      karpenter.azure.com/sku-accelerator: A100
      karpenter.azure.com/sku-cpu: "48"
      karpenter.azure.com/sku-encryptionathost-capable: "true"
      karpenter.azure.com/sku-family: "N"
      karpenter.azure.com/sku-gpu-count: "2"
      karpenter.azure.com/sku-gpu-manufacturer: nvidia
      karpenter.azure.com/sku-gpu-name: A100
      karpenter.azure.com/sku-hyperv-generation: "2"
      karpenter.azure.com/sku-memory: "450560"
      karpenter.azure.com/sku-name: Standard_NC48ads_A100_v4
      karpenter.azure.com/sku-networking-accelerated: "true"
      karpenter.azure.com/sku-storage-ephemeralos-maxsize: "549.755813888"
      karpenter.azure.com/sku-storage-premium-capable: "true"
      karpenter.azure.com/sku-version: "4"
      karpenter.azure.com/zone: koreacentral-1
      karpenter.sh/capacity-type: on-demand
      karpenter.sh/nodepool: a100
[...]

Issue #517, opened 2024-10-10
H100 support


HakjunMIN:
### Tell us about your request

H100 VM can't be used for node pool. Registered UAT of MSFT internal as well 

### Tell us about the problem you're trying to solve. What are you trying to do, and why is it hard?

H100 is definitely supported by Karpenter. 

### Are you currently working around this issue?

No workaround for now

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

PR #587, opened 2024-11-28
fix: GPU bootstrap, refresh driver versions and list of supported GPU VM SKUs

tallaxes:
Fixes https://github.com/Azure/karpenter-provider-azure/issues/579, https://github.com/Azure/karpenter-provider-azure/issues/517

**Description**

* Accommodate required bootstrap changes in recent VHD images (set GPU_DRIVER_TYPE)
* Update NVIDIA driver versions
* Update (and externalize) list of supported GPU VM SKUs
* Add (and fix) tests for bootstrap GPU settings

* `make presubmit`
* E2E: https://github.com/Azure/karpenter-provider-azure/actions/runs/12092116668/job/33721452273
* manual testing

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


pkg/utils/gpu.go
```diff
@@ -143,12 +143,20 @@ func IsMarinerEnabledGPUSKU(vmSize string) bool {
 // NVv3 is untested on AKS, NVv4 is AMD so n/a, and NVv2 no longer seems to exist (?).
 func GetGPUDriverVersion(size string) string {
 	if UseGridDrivers(size) {
-		return Nvidia535GridDriverVersion
+		return NvidiaGridDriverVersion
 	}
 	if isStandardNCv1(size) {
 		return Nvidia470CudaDriverVersion
```
ganeshkumarashok, line 149: The 470 driver is not being updated as it is only required for this NCv1 size, which is not supported and has not ben supported for a while. I don't think there are any karpenter/NAP users of it.

pkg/providers/imagefamily/azlinux.go
```diff
@@ -80,17 +80,18 @@ func (u AzureLinux) DefaultImages() []DefaultImageOutput {
...
 		Options: bootstrap.Options{
-			ClusterName:      u.Options.ClusterName,
-			ClusterEndpoint:  u.Options.ClusterEndpoint,
-			KubeletConfig:    kubeletConfig,
-			Taints:           taints,
-			Labels:           labels,
-			CABundle:         caBundle,
-			GPUNode:          u.Options.GPUNode,
-			GPUDriverVersion: u.Options.GPUDriverVersion,
-			// GPUImageSHA: u.Options.GPUImageSHA - GPU image SHA only applies to Ubuntu
```
ganeshkumarashok, line 91: More context on it applying only for Ubuntu: this refers to the container images published via aks-gpu (https://github.com/Azure/aks-gpu) that have the version and SHA suffix. 

Az Linux doesn't use those container images and instead installs the drivers during CSE. The version is not pinned per VHD/CSE - it just picks the latest. That is a major gap and needs to be fixed. 

Windows handles it separately.

ganeshkumarashok: Also Az Linux does not support GRID GPUs today.

Bryce-Soghigian (dismissed):
LGTM, lets wait on removing NC1 etc. or do it in a separate PR.

Bryce-Soghigian:
> GPUNeedsFabricManager and related changes can be split into a separate PR since karpenter doesn't have MIG support today right? https://learn.microsoft.com/en-us/azure/aks/gpu-multi-instance?tabs=azure-cli

Yeah seems weird to add this support now.

pkg/providers/imagefamily/azlinux.go
tallaxes: The way I read it, _all_ CSE settings related to GPU driver image (type, version, sha) are currently ignored for Azure Linux, but it is should / is expected to start using them in the future. One option would be to remove setting all of them for Azure Linux (not just sha), but it seems we could just keep them all for symmetry with Ubuntu?

Bryce-Soghigian (approved):

pkg/providers/imagefamily/azlinux.go
```diff
@@ -88,9 +88,9 @@ func (u AzureLinux) ScriptlessCustomData(kubeletConfig *bootstrap.KubeletConfigu
 			CABundle:         caBundle,
 			GPUNode:          u.Options.GPUNode,
 			GPUDriverVersion: u.Options.GPUDriverVersion,
-			// GPUImageSHA: u.Options.GPUImageSHA - GPU image SHA only applies to Ubuntu
-			// See: https://github.com/Azure/AgentBaker/blob/f393d6e4d689d9204d6000c85623ad9b764e2a29/vhdbuilder/packer/install-dependencies.sh#L201
-			SubnetID: u.Options.SubnetID,
+			GPUDriverType:    u.Options.GPUDriverType,
+			GPUImageSHA:      u.Options.GPUImageSHA,
```
Bryce-Soghigian, line 92: are we ok passing in a value that is unused by the contract here? Or have things changed and this value is used now?

pkg/utils/gpu.go
```diff
@@ -17,124 +17,81 @@ limitations under the License.
...
+		for _, sku := range skus {
+			nvidiaEnabledSKUs[sku] = true
+		}
+	}
+	for _, skus := range nvidiaSKUConfig.MarinerNvidiaEnabledSKUFamilies {
+		for _, sku := range skus {
+			marinerNvidiaEnabledSKUs[sku] = true
+		}
+	}
+}
```
Bryce-Soghigian, line 72: why not just have the type we marshal into be our actual structure rather than iterating through and creating a duplicate structure here? Its less than 100 items so not an expensive operation just wondering if there is a need to iterate through the list at all? Can we not just use the nvidiaSKUConfig type directly?

pkg/utils/gpu.go
```diff
@@ -17,124 +17,81 @@ limitations under the License.
...
+	AKSGPUGridVersionSuffix = "20241021235607"
 )
 
-func GetAKSGPUImageSHA(size string) string {
-	if UseGridDrivers(size) {
-		return AKSGPUGridSHA
-	}
-	return AKSGPUCudaSHA
+type NvidiaSKUConfig struct {
+	NvidiaEnabledSKUFamilies        map[string][]string `yaml:"nvidiaEnabledSKUs"`
```
Bryce-Soghigian, line 39: this is a structure of 

[family][]string{sku_sizea, sku_sizeb}, but we don't use family directly. can we write a marshaler to take in the final structure we want? 

map[sku_sizea] = supported_boolean?

pkg/providers/imagefamily/azlinux.go
tallaxes: I think we are ok passing something that is not yet used, but should/will be in the future. See https://github.com/Azure/karpenter-provider-azure/pull/587#discussion_r1863996492

pkg/utils/gpu.go
tallaxes: I presume it is because the layout convenient for authoring and maintenance is different from what is required at runtime for lookup. But, whatever the case, the list (and the code) is essentially copied from elsewhere that is considered the master source in this context, the main value for this codebase to close alignment to its upstream.

pkg/utils/gpu.go
tallaxes: See elsewhere on the value of alignment with upstream

PR #590, opened 2024-12-01
chore: add aks-node-viewer

tallaxes:
**Description**

Adds aks-node-viewer to tooling

* `make presubmit` followed by `aks-node-viewer` (and `make az-node-viewer`)

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note

```


hack/toolchain.sh
```diff
@@ -25,6 +25,7 @@ tools() {
     go install github.com/mattn/goveralls@v0.0.12
     go install github.com/google/go-containerregistry/cmd/crane@v0.20.2
     go install github.com/go-swagger/go-swagger/cmd/swagger@v0.31.0
+    go install github.com/Azure/aks-node-viewer/cmd/aks-node-viewer@latest
```
Bryce-Soghigian, line 28: should we update the az-node-viewer command too?

Bryce-Soghigian (dismissed):

hack/toolchain.sh
charliedmcb, line 28: Should we be pinning to a hash here for dependency bot? Or does it not matter since it isn't directly in an action?

charliedmcb (approved):
One question, but looks good.

hack/toolchain.sh
tallaxes: Dependabot does not process these. It is up to us to decide, for just a dev tool that does not contribute to built code using latest makes sense to me.

charliedmcb: Intuitively I feel like latest is best, so we don't have to continuously update this as we release changes, but curious the interplay between dependency bot and versioning within the repo.

charliedmcb: Responded at the same time! Perfect. 100% agree since dependabot doesn't apply here.

PR #594, opened 2024-12-02
test(examples): QoL, add optional commented options to examples

charliedmcb:
Fixes # 

**Description**
Quality of Life change to add useful options to the examples, but optional with comments. 

These are ones I use a lot for testing. However, don't want them by default as they might lead to confusing behavior if unaware they are enabled.

I use these a lot. Just added them directly while commented out with a note on each.
*

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note
NONE
```


tallaxes (approved):
Maybe also add to v1 examples?

PR #598, opened 2024-12-02
test(make): update `az-run-sample` to v1

charliedmcb:
Fixes # 

**Description**
I noticed that the `az-run-sample` make command was still using, 1vbeta, and updating it here, along with updating the `v1` file in accordance with recent optional settings.

`make az-run-sample`
*

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
NONE
```

### DIFF
--- a/examples/v1/general-purpose.yaml
+++ b/examples/v1/general-purpose.yaml
@@ -9,8 +9,12 @@ metadata:
 spec:
   disruption:
     consolidateAfter: 0s
+    expireAfter: Never
     budgets:
     - nodes: 30%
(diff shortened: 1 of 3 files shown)


examples/v1/general-purpose.yaml
```diff
@@ -9,8 +9,12 @@ metadata:
 spec:
   disruption:
     consolidateAfter: 0s
+    expireAfter: Never
```
tallaxes, line 12: In v1 `expireAfter` is in the template (and already there)

charliedmcb: Yea, just removed

tallaxes (approved):

Issue #616, opened 2024-12-11
SKU-image/disk incompatibility


CCOLLOT:
### Version

Starting this morning in Europe West, I started getting these errors:

The VM size 'Standard_D4als_v6' cannot boot with OS image or disk. Please check that disk controller types supported by the OS image or disk is one of the supported disk controller types for the VM size 'Standard_D4als_v6'. Please query sku api at https://aka.ms/azure-compute-skus to determine supported disk controller types for the VM size.

nodepool config:

```yaml
apiVersion: karpenter.sh/v1beta1
kind: NodePool
metadata:
  name: app
spec:
  disruption:
    budgets:
    - nodes: 20%
    - duration: 8h
      nodes: "0"
      schedule: 0 8 * * mon-fri
    consolidationPolicy: WhenUnderutilized
    expireAfter: Never
  template:
    metadata:
      labels:
        foo.bar/workload-type: app
    spec:
      nodeClassRef:
        name: default
      requirements:
      - key: kubernetes.io/arch
        operator: In
        values:
        - amd64
      - key: kubernetes.io/os
        operator: In
        values:
        - linux
      - key: karpenter.sh/capacity-type
        operator: In
        values:
        - on-demand
      - key: karpenter.azure.com/sku-family
        operator: In
        values:
        - E
        - D
      - key: karpenter.azure.com/sku-cpu
        operator: In
        values:
        - "4"
        - "8"
  weight: 100
```

### Expected Behavior

I expect NAP to check such compatibilities issues or at least to retry with a different node type / disk type /os image if it gets this kind of errors

### Actual Behavior

No node is provisioned

### Steps to Reproduce the Problem

Not 100% certain but I assume trying to launch a Standard_D4als_v6 node with the latest azure linux image

### Resource Specs and Logs

The VM size 'Standard_D4als_v6' cannot boot with OS image or disk. Please check that disk controller types supported by the OS image or disk is one of the supported disk controller types for the VM size 'Standard_D4als_v6'. Please query sku api at https://aka.ms/azure-compute-skus to determine supported disk controller types for the VM size.

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
[...]

PR #617, opened 2024-12-13
fix: temp exclude SKUs without compatible VM image

tallaxes:
Fixes https://github.com/Azure/karpenter-provider-azure/issues/616

**Description**

Exclude from consideration VM SKUs that don't have a compatible (= tagged for NVMe) VM image in community image galleries - for now, until the VM images are properly tagged. The implementation only allows VM SKUs that support SCSI (defined as either declaring SCSI support in their DiskControllerType, or not declaring anything at all).

Possible alternative implementations (which we may consider in the future):
* Internally assign corresponding capabilities (e.g. `karpenter.azure.com/controller-disk-type`) to both instance types (derived from VM SKUs) and images (either statically or derived from image definition) and rely on requirements compatibility, and dynamically take VM SKUs without compatible image support out of consideration. This would have a side benefit of being able to select `controller-disk-type`.
* Detect persistent / non-retriable failure to provision certain VM SKUs and take them out of consideration.
* Use explicit VM SKU allowlist

* `make presubmit`
* local testing with Fv6
* E2E test: https://github.com/Azure/karpenter-provider-azure/actions/runs/12307523172 (passed)

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
Exclude from consideration VM SKUs that don't have a compatible (= tagged for NVMe) VM image
```


pkg/providers/instancetype/suite_test.go
```diff
@@ -348,6 +348,9 @@ var _ = Describe("InstanceType Provider", func() {
 		It("should not include confidential SKUs", func() {
 			Expect(instanceTypes).ShouldNot(ContainElement(WithTransform(getName, Equal("Standard_DC8s_v3"))))
 		})
+		It("should not include SKUs without compatible image", func() {
+			Expect(instanceTypes).ShouldNot(ContainElement(WithTransform(getName, Equal("Standard_D2as_v6"))))
+		})
```
Bryce-Soghigian, line 353: A followup would be to have a unit test where we set the options to useSIG and then these skus are included, but we can add additional tests then.

Bryce-Soghigian (approved):

Issue #525, opened 2024-10-15
NodeClaims stuck deleting due to finalizer


flickers:
### Version

__Karpenter Version:__ v0.5.4

__Kubernetes Version:__ v1.30.3

### Expected Behavior

When a node is deleted that the Nodeclaim is also eventually deleted

### Actual Behavior

After running karpenter for a while in 3 separate clusters in 3 separate subscriptions we have seen karpenter fail to delete nodeclaims

- The kubernetes node has been deleted 
- The Azure resources have been deleted (nic, disk and vm)
- Karpenter nodeclaim still exists for the node
- Karpenter nodeclaim has .metadata.deletionTimestamp like it has been deleted but finalizer is preventing deletion

We have been running the following command that removes the finalizer for the nodeclaims that have .metadata.deletionTimestamp
```
kubectl get nodeclaim -o json | jq -r '.items[] | select(.metadata.deletionTimestamp != null) | .metadata.name' | xargs -I {} kubectl patch nodeclaim {} --type=json -p='[{"op": "remove", "path": "/metadata/finalizers"}]'
```

### Steps to Reproduce the Problem

Not sure how to reproduce, we are also running Karpenter in AWS and have done that from the early beginning and never seen something like this.
Saw a PR for something similar in karpenter-sigs repo but that might be related to Karpenter v1
https://github.com/kubernetes-sigs/karpenter/pull/1705

### Resource Specs and Logs

```json
{
    "level": "ERROR",
    "time": "2024-10-15T12:49:39.023Z",
    "logger": "controller.nodeclaim.termination",
    "message": "virtualMachine.Delete for aks-aks-prd-nodepool-kgcpq failed: GET https://management.azure.com/subscriptions/xxxxxxx-5535-4267-8839-xxxxxxxxx/resourceGroups/rg-aks-prd-nodes/providers/Microsoft.Compute/virtualMachines/aks-aks-prd-nodepool-kgcpq\n--------------------------------------------------------------------------------\nRESPONSE 404: 404 Not Found\nERROR CODE: NotFound\n--------------------------------------------------------------------------------\n{\n  \"error\": {\n    \"code\": \"NotFound\",\n    \"message\": \"The entity was not found in this Azure location.\"\n  }\n}\n--------------------------------------------------------------------------------\n",
    "commit": "846ef96",
    "nodeclaim": "aks-prd-nodepool-kgcpq",
    "node": "aks-aks-prd-nodepool-kgcpq",
[...]

PR #633, opened 2024-12-24
fix: Resolve errors for vm resource not found

tangyouzzz:
Fixes https://github.com/Azure/karpenter-provider-azure/issues/525

**Description**
https://learn.microsoft.com/en-us/azure/azure-resource-manager/troubleshooting/error-not-found?tabs=bicep#symptoms 
When using the GET method to obtain a VM, the Error Code returned when the VM does not exist can be either NotFound or ResourceNotFound.

```bash
{"level":"ERROR","time":"2024-12-24T08:08:51.172Z","logger":"controller","message":"Reconciler error","commit":"56aa2e0-dirty","controller":"node.termination","controllerGroup":"","controllerKind":"Node","Node":{"name":"aks-cloud-fip-mcu-arm64-mkr2f"},"namespace":"","name":"aks-cloud-fip-mcu-arm64-mkr2f","reconcileID":"2ee3a64a-083f-4743-84fa-661820165f2c","error":"ensuring instance termination, getting cloudprovider instance, getting instance, failed to get VM instance, GET https://management.azure.com/subscriptions/xxxx/resourceGroups/xxxx/providers/Microsoft.Compute/virtualMachines/aks-cloud-fip-mcu-arm64-mkr2f\n--------------------------------------------------------------------------------\nRESPONSE 404: 404 Not Found\nERROR CODE: NotFound\n--------------------------------------------------------------------------------\n{\n  \"error\": {\n    \"code\": \"NotFound\",\n    \"message\": \"The entity was not found in this Azure location.\"\n  }\n}\n--------------------------------------------------------------------------------\n"}
```
In IsNotFoundErr, only the situation of ResourceNotFound will be judged, which will cause the node to remain in the notready state and cannot be deleted.
```go
func IsNotFoundErr(err error) bool {
	azErr := IsResponseError(err)
	return azErr != nil && azErr.ErrorCode == ResourceNotFound
}

```

- Using `az vm simulate-eviction` to simulate spot node release

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note

```


tangyouzzz:

Bryce-Soghigian (approved):
LGTM @tallaxes take a look for second approval?

Bryce-Soghigian:
https://github.com/Azure/azure-sdk-for-go-extensions/blob/main/pkg/errors/armerrors.go#L34 can we also add this change here too?

tallaxes (approved):
LGTM. Some possible improvements:
* Checking for `azErr.StatusCode == http.StatusNotFound` instead should catch both
* Same logic should be applied deleting other resources (like NIC); other resource providers are very likely to exhibit the same pattern
* A helper method would be better to avoid code duplication (and yes, would be good to update azure-sdk-for-go-extensions)

tallaxes:
Merging since this fixes an actual problem, but let's follow-up on [feedback](https://github.com/Azure/karpenter-provider-azure/pull/633#pullrequestreview-2535514721)

PR #634, opened 2024-12-30
test(2a): removing settings and using isNAP flag to gate the karpenter controller pod calls

Bryce-Soghigian:
**Description**
This pr introduces a cleaner set of changes for running the karpenter e2e tests on a cluster with the node auto provisioning addon. 

-  [x] E2E Tests

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note

```


Bryce-Soghigian:
Testing this code doesn't break the existing e2e tests: https://github.com/Azure/karpenter-provider-azure/actions/runs/12540459266

tallaxes (commented):
So, there are several things in tests that expect to be able to get to Karpenter controller itself (expectations around settings and being able to modify them, monitoring restarts, leases, dumping logs, etc.). I would suggest adding a field to the common `Environment` (something like `InClusterController`), initializing it either from environment, or maybe just by attempting to lookup the controller in `kube-system`. This would make this setting uniformly accessible in testing, including for gating calls to `printControllerLogs`, `ExpectSettings`, etc. (One could consider introducing an optional alternative kubeclient instead, but I would keep it simple for now ...)

I would also suggest keeping the existing settings-related expectations, they may come handy later.

Bryce-Soghigian:
Closing this in favor of a cleaner branch https://github.com/Azure/karpenter-provider-azure/pull/634/files#diff-5b1126cd8ee92bf9d7bdc16f382bb5c4ccd17f3ca6fab669e63a6f5a7c5bfe6c

PR #642, opened 2025-01-08
fix: properly garbage collecting orphaned network interfaces

### LINKED ISSUE #92: NicReservedForAnotherVM Error leaves nics in MC Resource Group
### Version

v0.2.0

### Expected Behavior

When karpenter sends a VM POST Request to create a vm, and that post request fails, karpenter should always clean up the leftover resources from that failed attempt. 

### Actual Behavior

When karpenter sends a VM Post Request to create a vm, and that post request fails(Due to quota or various other issues), there is a race conditon between the arm representation of the vm and nic deletion. 

When we attempt to issue a delete call for the network interface, we will get an error  `NicReservedForAnotherVM`.  This often occurs because the arm representation of the vm we tried to create isn't yet deleted when we issue the network interface deletion call. 

This results in the end delete call failing saying that the nic is reserved for another vm. 

One can attempt to fix this by 
1. Disassociating the nic and retrying deletion(Preferred)
2. Having controller retries for network interface deletion(This blocks the nodeclaim from failing fast so not prefered) 
3. having another garbage collection controller for dealing with ghost nics.

make az-perftest-300, essentially you just need to scale up to a high volume. 

### Resource Specs and Logs

1. VM Create Fails
{"level":"ERROR","time":"2024-01-10T23:27:06.884Z","logger":"controller.nodeclaim.lifecycle","message":"Creating virtual machine \"aks-default-bzzjc\" failed: PUT https://management.azure.com/subscriptions/<redacted>/resourceGroups/MC_blah/providers/Microsoft.Compute/virtualMachines/aks-default-bzzjc\n--------------------------------------------------------------------------------\nRESPONSE 409: 409 Conflict\nERROR CODE: OperationNotAllowed\n--------------------------------------------------------------------------------\n{\n  \"error\": {\n    \"code\": \"OperationNotAllowed\",\n    \"message\": \"Operation could not be completed as it results in exceeding approved Total Regional Cores quota. Additional details - Deployment Model: Resource Manager, Location: uksouth, Current Limit: 100, Current Usage: 60, Additional Required: 48, (Minimum) New Limit Required: 108. Submit a request for Quota increase at <redacted> by specifying parameters listed in the ‘Details’ section for deployment to succeed. Please read more about quota limits at https://docs.microsoft.com/en-us/azure/azure-supportability/regional-quota-requests\"\n  }\n}\n--------------------------------------------------------------------------------\n","commit":"832597b-dirty","nodeclaim":"default-bzzjc","nodepool":"default"}
2. Deletion of network interface fails due to nic reserved 
3. {"level":"ERROR","time":"2024-01-10T23:28:07.848Z","logger":"controller.nodeclaim.lifecycle","message":"networkInterface.Delete for aks-default-bzzjc failed: DELETE https://management.azure.com/subscriptions/redacted/resourceGroups/redacted/providers/Microsoft.Network/networkInterfaces/aks-default-bzzjc\n--------------------------------------------------------------------------------\nRESPONSE 400: 400 Bad Request\nERROR CODE: NicReservedForAnotherVm\n--------------------------------------------------------------------------------\n{\n  \"error\": {\n    \"code\": \"NicReservedForAnotherVm\",\n    \"message\": \"Nic(s) in request is reserved for another Virtual Machine for 180 seconds. Please provide another nic(s) or retry after 180 seconds. Reserved VM: /subscriptions/redacted/resourceGroups/M/providers/Microsoft.Compute/virtualMachines/aks-default-bzzjc\",\n    \"details\": []\n  }\n}\n--------------------------------------------------------------------------------\n","commit":"832597b-dirty","nodeclaim":"default-bzzjc","nodepool":"default"}

### Community Note

Bryce-Soghigian:
Fixes #92 

**Description**
This introduces a garbage collection controller that will remove network interfaces that are orphaned.

- make test 
- make az-all + scale up 

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
Properly garbage collect network interfaces.
```


pkg/cloudprovider/cloudprovider.go
```diff
@@ -350,15 +350,15 @@ func (c *CloudProvider) instanceToNodeClaim(ctx context.Context, vm *armcompute.
 	}
 	annotations[v1alpha2.AnnotationInPlaceUpdateHash] = inPlaceUpdateHash
 
-	nodeClaim.Name = GenerateNodeClaimName(*vm.Name)
+	nodeClaim.Name = GenerateNodeClaimName(lo.FromPtr(vm.Name))
 	nodeClaim.Labels = labels
 	nodeClaim.Annotations = annotations
-	nodeClaim.CreationTimestamp = metav1.Time{Time: *vm.Properties.TimeCreated}
+	nodeClaim.CreationTimestamp = metav1.Time{Time: lo.FromPtr(vm.Properties.TimeCreated)}
```
Bryce-Soghigian, line 356: this path was executed when my List query returned an empty result. 

In the case where ARG has an outage, garbage collection will break, but there is no need to panic here in that scenario. So switched to lo.FromPtr() instead of normal dereference.

Bryce-Soghigian:
Testing this locally. 



Initially after our first scaleup, we were at 810 network interfaces. 


After garbage collection ran we can see that we deleted 435 network interfaces, and were left with 375 for our 375 nodes. 

~/dev/focus/karpenter-provider-azure (bsoghigian/network-interface-gc*) » cat logs.txt| grep "garbage collected NIC" | wc -l                                                                                                             sillygoose@Bryces-MacBook-Pro
     435
     
image

tallaxes (commented):
Realize this is still draft to be cleaned up, so apologies if feedback is too early. Leaving some comments, overall I _think_ I would be tempted to have a separate "networkinterface.garbagecollection" controller - unless there is a good reason not to.

pkg/controllers/nodeclaim/garbagecollection/nic_garbagecollection.go
```diff
@@ -0,0 +1,148 @@
...
+	NodeclaimReason = "nc"
+)
+
+type NetworkInterfaceController struct {
+	kubeClient       client.Client
+	instanceProvider instance.Provider
+	// A network interface is considered unremovable if it meets the following 3 criteria
+	// 1: Reserved by NRP: When creating a nic and attempting to assign it to a vm, the nic will be reserved for that vm arm_resource_id for 180 seconds
+	// 2: Belongs to a Nodeclaim: If a nodeclaim exists in the cluster we shouldn't attempt removing it
+	// 3: Belongs to VM: If the VM Garbage Collection controller is removing a vm, we should not attempt removing it in this controller, and delegate that responsibility to the vm gc controller since deleting a successfully provisioned vm has delete options to also clean up the associated nic
```
Bryce-Soghigian, line 58:
```suggestion
    // 3: Belongs to VM: If the VM Garbage Collection controller is removing a vm, we should not attempt removing it in this controller, and delegate that responsibility to the vm gc controller since deleting a successfully provisioned vm has delete options to also clean up the associated nic
```

pkg/test/networkinterfaces.go
```diff
@@ -0,0 +1,78 @@
...
+// RandomName returns a pseudo-random resource name with a given prefix.
+func RandomName(prefix string) string {
+	// You could make this more robust by including additional random characters.
+	return prefix + "-" + k8srand.String(10)
+}
+
+func ManagedTags(nodepoolName string) map[string]*string {
+	return map[string]*string{
+		"karpenter.sh_cluster":  lo.ToPtr("test-cluster"),
+		"karpenter.sh_nodepool": lo.ToPtr(nodepoolName),
```
Bryce-Soghigian, line 76: Should use constants from instance provider? But then cyclic dependency?

pkg/controllers/nodeclaim/garbagecollection/controller.go
```diff
@@ -42,55 +44,131 @@ import (
...
+	nodeClaimList := &karpv1.NodeClaimList{}
+	if err := c.kubeClient.List(ctx, nodeClaimList); err != nil {
+		return fmt.Errorf("listing NodeClaims for NIC GC: %w", err)
+	}
+
+	// Normalize NodeClaim names to match NIC naming conventions
+	nodeClaimNames := sets.New[string]()
+	for _, nodeClaim := range nodeClaimList.Items {
+		// Adjust the prefix as per the aks naming convention
+		nodeClaimNames.Insert(fmt.Sprintf("aks-%s", nodeClaim.Name))
```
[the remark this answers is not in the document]
Bryce-Soghigian: lets leave this for a separate pr

pkg/providers/instance/instance.go
```diff
@@ -87,6 +89,8 @@ type Provider interface {
 	// CreateTags(context.Context, string, map[string]string) error
 	Update(context.Context, string, armcompute.VirtualMachineUpdate) error
 	GetNic(context.Context, string, string) (*armnetwork.Interface, error)
+	DeleteNic(context.Context, string) error
+	ListNics(context.Context) ([]*armnetwork.Interface, error)
```
Bryce-Soghigian, line 93: Future refactor here would be to factor out these calls into a network interface provider.

pkg/controllers/nodeclaim/garbagecollection/nic_garbagecollection.go
```diff
@@ -0,0 +1,149 @@
...
+)
+
+type NetworkInterface struct {
+	kubeClient       client.Client
+	instanceProvider instance.Provider
+	// A network interface is considered unremovable if it meets the following 3 criteria
+	// 1: Reserved by NRP: When creating a nic and attempting to assign it to a vm, the nic will be reserved for that vm arm_resource_id for 180 seconds
+	// 2: Belongs to a Nodeclaim: If a nodeclaim exists in the cluster we shouldn't attempt removing it
+	// 3: Belongs to VM: If the VM Garbage Collection controller is removing a vm, we should not attempt removing it in this controller, and delegate that responsibility to the vm gc controller since deleting a successfully provisioned vm has delete options to also clean up the associated nic
+	unremovableNics *cache.Cache
```
tallaxes, line 59: Are you sure it makes sense to track all "nics that should not be deleted in this loop" - VM-associated, NodeClaim-associated and failed to delete - in the same cache?

- It appears to mix network interfaces we just want to keep with those that we want to remove, but cannot.
- It is not clear to me that VM-associated and NodeClaim-associated need cahcing with expiration capabilities (do they?)
- NodeClaim-associated, specifically, are not really maintained in cache, they are added in the beginning and removed in the end.
- Couldn't this add/remove "mask" something that was cached (with the same key) for failed-to-delete reason?
- Actually, wouldn't failed-to-delete get evicted before the next try? (Asked elsewhere, for here assuming it won't.)

Maybe separating "in-use" (rebult) from "recently failed to delete" (cached) would make sense? And then the logic is "(try) delete unless in use or recently failed to delete"?

pkg/controllers/nodeclaim/garbagecollection/nic_garbagecollection.go
```diff
@@ -0,0 +1,149 @@
...
+	sdkerrors "github.com/Azure/azure-sdk-for-go-extensions/pkg/errors"
+	"github.com/Azure/karpenter-provider-azure/pkg/providers/instance"
+)
+
+const (
+	NICGCControllerName    = "networkinterface.garbagecollection"
+	NicReservationDuration = time.Second * 180
+	// We set this interval at 5 minutes, as thats how often our NRP limits are reset.
+	// See: https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/request-limits-and-throttling#network-throttling
+	NicGarbageCollectionInterval = time.Minute * 5
```
tallaxes, line 46: So this caches (with eviction) for 3 min, but the next attempt is in 5 min? Wouldn't it find the cache always empty (Have you observed a cache hit for "failed-to-delete" reason during testing?) If the next try is going to be after 5 min, maybe "failed-to-delete" state is not needed at all?

pkg/test/networkinterfaces.go
```diff
@@ -0,0 +1,94 @@
...
+
+// RandomName returns a pseudo-random resource name with a given prefix.
+func RandomName(prefix string) string {
+	// You could make this more robust by including additional random characters.
+	return prefix + "-" + k8srand.String(10)
+}
+
+func ManagedTags(nodepoolName string) map[string]*string {
+	return map[string]*string{
+		"karpenter.sh_cluster":  lo.ToPtr("test-cluster"),
```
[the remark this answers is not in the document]
Bryce-Soghigian: just for consistency for what the actual karpenter tags are. Would be good to have a proper api package that outlines the keys so that we never have inconsistency between test and the real deal.

pkg/controllers/nodeclaim/garbagecollection/nic_garbagecollection.go
Bryce-Soghigian: After thinking more, moved to use a set, and just included the vms and nics there.

pkg/controllers/nodeclaim/garbagecollection/nic_garbagecollection.go
Bryce-Soghigian: no you are correct, I removed the sdkerrors 3 minute retry. I am being conservative with retries as the price of not being able to bootstrap outweighs latent deletion attempts. 

We don't need the IsReserved check at all or the expiration cache after further reflection.

johnthompson-ybor:
I'd really love to see this merged, this is causing regular outages for us.

pkg/controllers/nodeclaim/garbagecollection/instance_garbagecollection.go
```diff
@@ -41,21 +40,21 @@ import (
 	corecloudprovider "sigs.k8s.io/karpenter/pkg/cloudprovider"
 )
 
-type Controller struct {
+type VirtualMachine struct {
 	kubeClient      client.Client
 	cloudProvider   corecloudprovider.CloudProvider
-	successfulCount uint64 // keeps track of successful reconciles for more aggressive requeueing near the start of the controller
+	successfulCount uint64 // keeps track of successful reconciles for more aggressive requeuing near the start of the controller
```
Bryce-Soghigian, line 46:
```suggestion
	successfulCount uint64 // keeps track of successful reconciles for more aggressive requeueing near the start of the controller
```

Bryce-Soghigian:
Ci here is failing due to vulncheck: https://github.com/Azure/karpenter-provider-azure/pull/665 should fix it

pkg/cloudprovider/cloudprovider.go
Bryce-Soghigian: just going to revert for now.

pkg/controllers/nodeclaim/garbagecollection/nic_garbagecollection.go
```diff
@@ -0,0 +1,149 @@
...
+		if !unremovableNic {
+			err := c.instanceProvider.DeleteNic(ctx, nicName)
+			if sdkerrors.IsNicReservedForAnotherVM(err) {
+				// cache the network interface as unremovable for 180 seconds
+				c.unremovableNics.SetDefault(nicName, NicReason)
+			}
+			if err != nil {
+				logging.FromContext(ctx).Error(err)
+				return
+			}
```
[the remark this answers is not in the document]
Bryce-Soghigian: I tried looking into the original pr to see if we got any rationality behind that: https://github.com/aws/karpenter-provider-aws/pull/3405 

Didn't see much.


The pattern as I understand it is if we see errors we retry two minutes later. We should probably change that, if VM Deletion fails due to a TooManyRequests error, we can't retry until the quota window resets(After 5 minutes?)

2m might make some sense for vms as the volume of vms that need garbage collection is in most cases significantly lower than the volume of interfaces that need garbage collection. 

We could try requeuing depending on the kind of error at different intervals. 


If we see TooManyRequests Error wait 5 minutes to retry, if we see IsNicReservedForAnotherVM retry 3 minutes later?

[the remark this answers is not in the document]
Bryce-Soghigian: Overall it seems simpler to just not do all of that requeuing and come and optimize it later if we are optimizing for performance. For now just requeuing every 5 minutes seems to be fine.

pkg/controllers/nodeclaim/garbagecollection/nic_garbagecollection.go
```diff
@@ -0,0 +1,149 @@
...
+
+			logging.FromContext(ctx).With("nic", nicName).Infof("garbage collected NIC")
+		}
+	})
+	c.removeNodeclaimsFromUnremovableNics()
+	// requeue every 5 minutes, adjust for throttling?
```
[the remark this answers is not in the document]
Bryce-Soghigian: going to resolve as I am dicussing same point up here: https://github.com/Azure/karpenter-provider-azure/pull/642#discussion_r1921907318

pkg/test/networkinterfaces.go
```diff
@@ -0,0 +1,94 @@
...
+				{
+					Name: lo.ToPtr("ipConfig"),
+					Properties: &armnetwork.InterfaceIPConfigurationPropertiesFormat{
+						PrivateIPAllocationMethod: lo.ToPtr(armnetwork.IPAllocationMethodDynamic),
+						Subnet:                    &armnetwork.Subnet{ID: lo.ToPtr("/subscriptions/.../resourceGroups/.../providers/Microsoft.Network/virtualNetworks/.../subnets/default")},
+					},
+				},
+			},
+		},
+		Tags: options.Tags,
```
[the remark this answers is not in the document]
Bryce-Soghigian: Changed the approach lmk what you think.

pkg/fake/azureresourcegraphapi_test.go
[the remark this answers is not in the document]
Bryce-Soghigian: Added these tests

pkg/providers/instance/arglist.go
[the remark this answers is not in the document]
Bryce-Soghigian: renamed

pkg/providers/instance/instance.go
```diff
@@ -642,11 +670,11 @@ func (p *DefaultProvider) cleanupAzureResources(ctx context.Context, resourceNam
 	// The order here is intentional, if the VM was created successfully, then we attempt to delete the vm, the
 	// nic, disk and all associated resources will be removed. If the VM was not created successfully and a nic was found,
 	// then we attempt to delete the nic.
-	nicErr := deleteNicIfExists(ctx, p.azClient.networkInterfacesClient, p.resourceGroup, resourceName)
+
+	nicErr := p.DeleteNic(ctx, resourceName)
```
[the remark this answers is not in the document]
Bryce-Soghigian: Sure, shouldnt we just modify the armerrors NotFound in sdkforgoextensions then both will get the fix? Or is there another fix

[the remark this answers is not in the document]
tallaxes: Yes, but doing this properly in sdk-for-go-extensions - which is also used by other projects - is more involved. (For example: It is not clear that other RPs follow the same pattern, I am even not sure about NRP - thus suggestion to check; The current `IsNotFound` is actually checking for `ResourceNotFound` - and there may be cases where client needs to distinguish between the two; changing this behavior changes the behavior subtly - probably justifies minor version bump and a warning in release notes; checking for 404 might just be the right thing to do, etc.) Figuring out the proper generic logic is not trivial, so let's not make it a pre-requisite for this PR. @matthchr

pkg/controllers/nodeclaim/garbagecollection/nic_garbagecollection.go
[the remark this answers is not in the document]
tallaxes: Appreciate the explanation. Proper treatment of retries and rate limiting is not trivial, let's not tackle it here. I am Ok with keeping constant retry period here for now.

pkg/test/virtualmachines.go
```diff
@@ -0,0 +1,78 @@
...
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/compute/armcompute"
+	"github.com/Azure/karpenter-provider-azure/pkg/fake"
+	"github.com/imdario/mergo"
+	"github.com/samber/lo"
+)
+
+// VirtualMachineOptions customizes an Azure Virtual Machine for testing.
+type VirtualMachineOptions struct {
+	Name         string
+	NodepoolName string
```
tallaxes, line 32: Having VM be aware of the Nodepool is a smell

pkg/test/expectations/expectations.go
```diff
@@ -54,3 +53,8 @@ func ExpectDecodedCustomData(env *test.Environment) string {
 
 	return decodedString
 }
+
+func ExpectNoError(err error) {
+	GinkgoHelper()
+	Expect(err).To(BeNil())
+}
```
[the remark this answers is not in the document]
tallaxes: Marked resolved with no comment or action. You feel strongly about keeping this? Or plan to address later?

Bryce-Soghigian:
> Marked resolved with no comment or action. You feel strongly about keeping this? Or plan to address later?
I removed the uses but seem to have forgotten to remove the error itself. Will get that in.

pkg/providers/instance/instance.go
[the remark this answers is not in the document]
matthchr: Yeah my 2c at least is, just check 404 because trying to check the `code` across Azure services is a fools errand, there are _tons_ of possible `code`'s that RPs set so having a complete list just won't work. HTTP StatusCode 404 will work consistently across RPs.

pkg/controllers/nodeclaim/garbagecollection/suite_test.go
```diff
@@ -332,3 +316,77 @@ var _ = Describe("GarbageCollection", func() {
...
 })
+
+var _ = Describe("NetworkInterface Garbage Collection", func() {
+	It("should not delete a network interface if a nodeclaim exists for it", func() {
+		// Create and apply a NodeClaim that references this NIC
+		nodeClaim := coretest.NodeClaim()
+		ExpectApplied(ctx, env.Client, nodeClaim)
+
+		// Create a managed NIC
+		nic := test.Interface(test.InterfaceOptions{Name: instance.GenerateResourceName(nodeClaim.Name), NodepoolName: nodePool.Name})
```
tallaxes, line 327: Yes I suggested elsewhere to collapse these for redability, but I meant (and included in the example) only `test.Interface(test.InterfaceOptions{` part, like this:

```suggestion
		nic := test.Interface(test.InterfaceOptions{
			Name: instance.GenerateResourceName(nodeClaim.Name),
			NodepoolName: nodePool.Name,
		})
```

This is the pattern that we use in many other places for readability. The rationale is that the type names are usually boilerplate, while the field values are usually of interest.

pkg/controllers/nodeclaim/garbagecollection/suite_test.go
```diff
@@ -332,3 +316,77 @@ var _ = Describe("GarbageCollection", func() {
...
+		managedVM := test.VirtualMachine(test.VirtualMachineOptions{Name: lo.FromPtr(managedNic.Name), NodepoolName: nodePool.Name})
+		azureEnv.VirtualMachinesAPI.VirtualMachinesBehavior.Instances.Store(lo.FromPtr(managedVM.ID), *managedVM)
+		ExpectSingletonReconciled(ctx, networkInterfaceGCController)
+		// We should still have a network interface here
+		nicsAfterGC, err := azureEnv.InstanceProvider.ListNics(ctx)
+		ExpectNoError(err)
+		Expect(len(nicsAfterGC)).To(Equal(1))
+
+	})
+	It("the vm gc controller should remove the nic if there is an associated vm", func() {
```
tallaxes, line 369: Had to read the code to understand what it does; maybe belongs under VM GC test section? Or just a better description:

```suggestion
	It("the VM GC controller should also delete the associated NIC", func() {
```

pkg/providers/instance/instance.go
```diff
@@ -55,9 +53,13 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/network/armnetwork"
 )
 
+var (
+	vmListQuery  string
+	nicListQuery string
+)
```
tallaxes, line 59: This is probably a leftover from VM GC implementation, but not clear to me why these need to be global rather than DefaultProvider fields

pkg/providers/instance/suite_test.go
```diff
@@ -215,4 +216,24 @@ var _ = Describe("InstanceProvider", func() {
...
+		pod := coretest.UnschedulablePod(coretest.PodOptions{})
+		ExpectProvisioned(ctx, env.Client, cluster, cloudProvider, coreProvisioner, pod)
+		ExpectScheduled(ctx, env.Client, pod)
+		interfaces, err := azureEnv.InstanceProvider.ListNics(ctx)
+		Expect(err).To(BeNil())
+		Expect(len(interfaces)).To(Equal(1))
+	})
+	It("should only list nics that belong to karpenter", func() {
+		managedNic := test.Interface(test.InterfaceOptions{NodepoolName: nodePool.Name})
+		unmanagedNic := test.Interface(test.InterfaceOptions{Tags: map[string]*string{"kubernetes.io/cluster/test-cluster": lo.ToPtr("random-aks-vm")}})
```
tallaxes, line 230: Same comment about the degree of collapsing as elsewhere ...

pkg/test/virtualmachines.go
tallaxes: Ok, I see now this was maybe in response to my question
> What happens if options.Tags are not passed?

I was just asking about whether it would be tolerated / won't crash. Did not mean to suggest they should be calculated automatically (and definetely not at the expense of having to pass nodepool name). Will be more clear next time.

tallaxes (approved):
Approving with suggestions

PR #650, opened 2025-01-14
test(e2e): update drift suite

tallaxes:
**Description**

* Update E2E drift suite (among other things, adds budget and AKSNodeClass tests). Mostly aligned with AWS provider.
* Fix AKSNodeClass hashing (add KubeletConfig) and add tests

Other E2E improvements:
* Fix test AKSNodeClass overrides
* Remove drift feature gate override (no longer needed)
* Add check for leaked leases
* Override `coretest.DefaultImage` (pause) with MCR version, once
* Introduce `AdaptToClusterConfig` wrapper for test NodePools
* Have `make az-e2etests` also run `az-cleanenv`

* `make presubmit`
* local E2E testing
* E2E CI: (note that the full Drift suite now takes ~3h to run)
  * https://github.com/Azure/karpenter-provider-azure/actions/runs/12804083976 (https://github.com/Azure/karpenter-provider-azure/commit/dd26f5e0c959594b9eee34b7fa823f5748ee0e67) ✅
  * https://github.com/Azure/karpenter-provider-azure/actions/runs/14208846100 (https://github.com/Azure/karpenter-provider-azure/pull/650/commits/20ec089f7e384a33f71577011360385314e1845c) 🔴
  * https://github.com/Azure/karpenter-provider-azure/actions/runs/14272136072 (https://github.com/Azure/karpenter-provider-azure/pull/650/commits/6ea09c6e8387d078d7272e7b2ae6c48b0bd00296) ✅

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note

```


test/suites/drift/suite_test.go
```diff
@@ -14,26 +14,35 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package drift
+package drift_test
```
charliedmcb, line 17: why the package name change?

test/suites/drift/suite_test.go
```diff
@@ -57,73 +69,653 @@ var _ = AfterEach(func() { env.Cleanup() })
...
+			Spec: karpv1.NodeClaimTemplateSpec{
+				// since this will overwrite the default requirements, add instance category and family selectors back into requirements
+				Requirements: []karpv1.NodeSelectorRequirementWithMinValues{
+					{NodeSelectorRequirement: corev1.NodeSelectorRequirement{Key: karpv1.CapacityTypeLabelKey, Operator: corev1.NodeSelectorOpIn, Values: []string{karpv1.CapacityTypeSpot}}},
+					{NodeSelectorRequirement: corev1.NodeSelectorRequirement{Key: corev1.LabelInstanceTypeStable, Operator: corev1.NodeSelectorOpIn, Values: []string{"Standard_DS4_v2"}}},
+				},
+			},
+		}),
+	)
+	FDescribeTable("AKSNodeClass", func(nodeClassSpec v1alpha2.AKSNodeClassSpec) {
```
charliedmcb, line 439: Make sure to remove `F` focus before completing the PR

test/suites/drift/suite_test.go
```diff
@@ -57,73 +69,653 @@ var _ = AfterEach(func() { env.Cleanup() })
...
+					Labels: map[string]string{
+						"app": "my-app",
+					},
+					Annotations: map[string]string{
+						karpv1.DoNotDisruptAnnotationKey: "true",
+					},
 				},
+				TerminationGracePeriodSeconds: lo.ToPtr[int64](0),
 			},
-			ResourceRequirements: corev1.ResourceRequirements{Requests: corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("0.5")}},
```
charliedmcb, line 80: Don't want base resource requirements, or am I missing something?

test/suites/drift/suite_test.go
```diff
@@ -57,73 +69,653 @@ var _ = AfterEach(func() { env.Cleanup() })
...
-			Image:                "mcr.microsoft.com/oss/kubernetes/pause:3.6",
```
charliedmcb, line 81: I'm pretty sure if we don't supply this it will use an aws image. Feel we should be using ours?

Might be good to indirectly route through a wrapper package that sets this.

test/suites/drift/suite_test.go
tallaxes: Correct, those are now specified in each test (if/as needed)

tallaxes:
Failure scenarios have been failing occasionally, so added more By (https://github.com/Azure/karpenter-provider-azure/pull/650/commits/f6d20e0e572ee797f15c7bc4424835e568671be0) and attempted to deflake / make it more deterministic (https://github.com/Azure/karpenter-provider-azure/pull/650/commits/deaa7ae03c7e39491ddba1ca4932f6389cc0d30d). The last run [passed](https://github.com/Azure/karpenter-provider-azure/actions/runs/12804083976/job/35730105524), but may need to revisit if it fails again.

charliedmcb (commented):
Dropping first batch of review comments. Looking good overall, and will complete remainder today.

pkg/apis/v1alpha2/aksnodeclass.go
```diff
@@ -51,7 +51,7 @@ type AKSNodeClassSpec struct {
 	// Wherever possible, the types and names should reflect the upstream kubelet types.
 	// +kubebuilder:validation:XValidation:message="imageGCHighThresholdPercent must be greater than imageGCLowThresholdPercent",rule="has(self.imageGCHighThresholdPercent) && has(self.imageGCLowThresholdPercent) ?  self.imageGCHighThresholdPercent > self.imageGCLowThresholdPercent  : true"
 	// +optional
-	Kubelet *KubeletConfiguration `json:"kubelet,omitempty" hash:"ignore"`
+	Kubelet *KubeletConfiguration `json:"kubelet,omitempty"`
```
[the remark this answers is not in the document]
tallaxes: Kubelet config needs to be included in hash-based drift detection

Makefile-az.mk
```diff
@@ -257,7 +257,7 @@ az-taintsystemnodes: ## Taint all system nodepool nodes
 az-taintnodes:
 	kubectl taint nodes CriticalAddonsOnly=true:NoSchedule --all --overwrite
 
-az-e2etests: ## Run e2etests
+az-e2etests: az-cleanenv ## Run e2etests
```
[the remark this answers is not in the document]
tallaxes: Went away with merging of main

test/suites/drift/suite_test.go
```diff
@@ -57,71 +69,658 @@ var _ = AfterEach(func() { env.Cleanup() })
...
+			nodePool.Labels = appLabels
+			// We're expecting to create 5 nodes, so we'll expect to see at most 3 nodes deleting at one time.
+			nodePool.Spec.Disruption.Budgets = []karpv1.Budget{{
+				Nodes: "3",
+			}}
+
+			// Create a 5 pod deployment with hostname inter-pod anti-affinity to ensure each pod is placed on a unique node
+			numPods = 5
+			selector = labels.SelectorFromSet(appLabels)
+			deployment := coretest.Deployment(coretest.DeploymentOptions{
```
charliedmcb, line 242: nit: any idea why this one test case uses `deployment` instead of `dep`? I see AWS is doing the same, but don't see a reason for it.

pkg/apis/v1alpha2/aksnodeclass.go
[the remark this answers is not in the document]
charliedmcb: Than we should make a bump on the hash version?
- https://github.com/Azure/karpenter-provider-azure/blob/373616cb59b0636a3ec5d30204eaccdb74de77df/pkg/apis/v1alpha2/aksnodeclass.go#L172

pkg/apis/v1alpha2/aksnodeclass_hash_test.go
```diff
@@ -0,0 +1,106 @@
...
+					AllowedUnsafeSysctls:        []string{"net.core.somaxconn"},
+					ContainerLogMaxSize:         "10Mi",
+					ContainerLogMaxFiles:        lo.ToPtr(int32(10)),
+				},
+				MaxPods: lo.ToPtr(int32(100)),
+			},
+		}
+	})
+	DescribeTable(
+		"should match static hash on field value change",
```
charliedmcb, line 62: All of these hashes will have to be updated, if we add more fields we hash on, correct?

Please as a comment here noting this, so when expected failures occur in the future, we know these needs to be updated. Something like:
```
// NOTE: When the hashing algorithm updates, these tests are expected to fail and need to have their hashes updated.
```

pkg/apis/v1alpha2/aksnodeclass_hash_test.go
```diff
@@ -0,0 +1,106 @@
...
+
+		// Static fields, expect changed hash from base
+		Entry("VNETSubnetID", "1592960122675270014", v1alpha2.AKSNodeClass{Spec: v1alpha2.AKSNodeClassSpec{VNETSubnetID: lo.ToPtr("subnet-id-2")}}),
+		Entry("OSDiskSizeGB", "8480605960768312946", v1alpha2.AKSNodeClass{Spec: v1alpha2.AKSNodeClassSpec{OSDiskSizeGB: lo.ToPtr(int32(40))}}),
+		Entry("ImageFamily", "1518600367124538723", v1alpha2.AKSNodeClass{Spec: v1alpha2.AKSNodeClassSpec{ImageFamily: lo.ToPtr("AzureLinux")}}),
+		Entry("Tags", "13342466710512152270", v1alpha2.AKSNodeClass{Spec: v1alpha2.AKSNodeClassSpec{Tags: map[string]string{"keyTag-test-3": "valueTag-test-3"}}}),
+		Entry("Kubelet", "13352743309220827045", v1alpha2.AKSNodeClass{Spec: v1alpha2.AKSNodeClassSpec{Kubelet: &v1alpha2.KubeletConfiguration{CPUManagerPolicy: "none"}}}),
+		Entry("MaxPods", "17237389697604178241", v1alpha2.AKSNodeClass{Spec: v1alpha2.AKSNodeClassSpec{MaxPods: lo.ToPtr(int32(200))}}),
+	)
+	It("should match static hash when reordering tags", func() {
```
charliedmcb, line 77: Add a similar comment to this test, and/or the `staticHash` const?

charliedmcb (commented):
Overall, pretty much lgtm. 

Main two points to highlight:
- Called out one test which I think we should update the timeout for while async dift isn't working: `should respect budgets for non-empty replace drift`
- I think we need to bump the hash version, since the Kubelet configuration is being added to it?

Also, I do have some serious concerns about the direction of our E2E tests with Drift taking ~3hrs to run. I'd still like to get this PR checked in since its reviewed, but I'd almost want to disable some of the tests while we address speed improvements, making things more async. Testing approaching those times is a serious concern of mine.

pkg/apis/v1alpha2/aksnodeclass.go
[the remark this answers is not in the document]
tallaxes: Yes, I noticed that too when responding. Originally I thought we don't have to, but if we don't it will consider existring NodeClaims - if any - drifted, possibly continuously? Good catch.

pkg/apis/v1alpha2/aksnodeclass_hash_test.go
charliedmcb, line 1: Do we actually maybe want a test like the following?
```
        // This test is a sanity check to update the hashing version if the algorithm has been updated.
        // - Note: this will only catch a missing version update, if the currentStaticHash hasn't been updated yet.
	It("when hashing algorithm updates, we should update the hash version", func() {
                // If you update the currentStaticHash value, you need to update v1alpha2.AKSNodeClassHashVersion, and set currentHashVersion to the same value
	        currentStaticHash := "3322684356700793451"
	        currentHashVersion := "v1"
	        if nodeClass.Hash() != currentStaticHash {
	            Expect(v1alpha2.AKSNodeClassHashVersion).ToNot(Equal(currentHashVersion))
	        } else {
                    // Note: this failure case is to ensure you have updated currentHashVersion, not AKSNodeClassHashVersion
	            Expect(currentHashVersion).To(Equal(v1alpha2.AKSNodeClassHashVersion))
	        }
	        Expect(nodeClass.Hash()).To(Equal(currentStaticHash))
	})
```

test/suites/drift/suite_test.go
```diff
@@ -57,71 +69,658 @@ var _ = AfterEach(func() { env.Cleanup() })
...
+			env.ExpectUpdated(nodePool)
 
-		By("expect created node count to be 1")
-		env.ExpectCreatedNodeCount("==", 1)
+			env.EventuallyExpectDrifted(nodeClaims...)
+
+			env.ConsistentlyExpectDisruptionsUntilNoneLeft(3, 2, 5*time.Minute)
+		})
+		It("should respect budgets for non-empty delete drift", func() {
+			nodePool = coretest.ReplaceRequirements(nodePool,
```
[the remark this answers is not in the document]
tallaxes: Agree, for now leaving as is, but would be good to do a sweep for common patterns.

test/suites/drift/suite_test.go
```diff
@@ -57,71 +69,658 @@ var _ = AfterEach(func() { env.Cleanup() })
...
+			for _, node := range originalNodes {
+				Expect(env.Client.Get(env.Context, client.ObjectKeyFromObject(node), node)).To(Succeed())
+				node.Finalizers = append(node.Finalizers, common.TestingFinalizer)
+				env.ExpectUpdated(node)
+			}
+
+			By("drifting the nodepool")
+			nodePool.Spec.Template.Annotations = lo.Assign(nodePool.Spec.Template.Annotations, map[string]string{"test-annotation": "drift"})
+			env.ExpectUpdated(nodePool)
+			env.ConsistentlyExpectDisruptionsUntilNoneLeft(5, 3, 10*time.Minute)
```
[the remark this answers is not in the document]
tallaxes: This latest version was passing rather consistently in my testing, but ok with bumping the timeout to 15 min (which I would prefer to changing the logic for now)

test/suites/drift/suite_test.go
```diff
@@ -57,71 +69,658 @@ var _ = AfterEach(func() { env.Cleanup() })
...
+			nodePool = coretest.ReplaceRequirements(nodePool,
+				karpv1.NodeSelectorRequirementWithMinValues{
+					NodeSelectorRequirement: corev1.NodeSelectorRequirement{
+						Key:      corev1.LabelInstanceTypeStable,
+						Operator: corev1.NodeSelectorOpIn,
+						Values:   []string{"Standard_DS4_v2"},
+					},
+				},
+			)
+			// We're expecting to create 3 nodes, so we'll expect to see 2 nodes deleting at one time.
```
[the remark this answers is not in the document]
charliedmcb: Adding a little more clarity here. Technically we can have multiple deleting/disrupting at a time. However, the logic stalls and is blocked for a long time, making that window of actually having multiple nodes disrupted extremely narrow, and considered non-existent in my mind. This is because core won't start disrupting another node, until the Create call returns from our CloudProvider. And we stall at that point because our Create does Polling on all the created resources, which means it waits for the full VM Create to finish, and we'll have a registered node by that point, leaving the actual time of overlap between Create returning and node cleanup extremely slim (which is the window for Async Drift).

The difference with AWS is that their Create is a sync op, which doesn't go any LRO Polling, meaning they get the long-expected overlap time before the Node is registered, and thus have multiple instances disrupting at once.

I'm pretty sure this applies to the `DisruptedNoScheduleTaint` which is what's actually being checked, since I expect it would only be applied when the node is actually selected to be disrupted, and would be blocked until our Create returns. However, will double check.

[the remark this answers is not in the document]
charliedmcb: By looking at the core code, I believe my above expectations and understanding is correct, but could confirm with a real world test.

From the previous issue I opened on this we know it gets stuck on the log `waiting on cluster sync`:
https://github.com/kubernetes-sigs/karpenter/issues/1861

Looking at the code, this log is at the following location:
https://github.com/kubernetes-sigs/karpenter/blob/008f4653825c237b9a4c84c3435f3583b25ab83e/pkg/controllers/disruption/controller.go#L117

Further along in the call stack we get the taint added:
https://github.com/kubernetes-sigs/karpenter/blob/008f4653825c237b9a4c84c3435f3583b25ab83e/pkg/controllers/disruption/controller.go#L201
\- Note: `RequireNoScheduleTaint` applies the taint when the 3rd param is true, which it is here.

pkg/apis/v1alpha2/aksnodeclass_hash_test.go
tallaxes: Added, with minor modifications, https://github.com/Azure/karpenter-provider-azure/pull/650/commits/d083d3a658b21a813971ccb8e5fa81aad80006d8

test/suites/drift/suite_test.go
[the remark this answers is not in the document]
charliedmcb: Interesting. I just did some local testing, and got some initial nodeclaims taking up to just over 2 and a half minutes to fully fill out, but then testing Drift in this way they were coming up in 1.5-2 minutes. Not sure why the Drift seemed a little faster, but very small sample size.

Either way, I think the increase to 15 minutes is best for now.

test/suites/drift/suite_test.go
tallaxes: The core test Deployment or Pods does not have any defaults, so it will end up with no resource requirements or limits. The rest of the behaviour is up to standard Kuberneter scheduling etc. - it will respect other constraints (like taints, selectors, affinity, max pods, etc.), but from resources perspective will treat it as requiring nothing (including zero CPU and memory) and can place it anywhere (even on node with no available resources).

test/suites/drift/suite_test.go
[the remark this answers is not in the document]
tallaxes: Updated the comment

pkg/apis/v1alpha2/aksnodeclass_hash_test.go
```diff
@@ -103,4 +104,18 @@ var _ = Describe("Hash", func() {
...
 	})
+	// This test is a sanity check to update the hashing version if the algorithm has been updated.
+	// Note: this will only catch a missing version update, if the staticHash hasn't been updated yet.
+	It("when hashing algorithm updates, we should update the hash version", func() {
+		// If you update the staticHash value, you need to update v1alpha2.AKSNodeClassHashVersion, and set currentHashVersion to the same value
+		currentHashVersion := "v2"
+		if nodeClass.Hash() != staticHash {
+			Expect(v1alpha2.AKSNodeClassHashVersion).ToNot(Equal(currentHashVersion))
+		} else {
+			// Note: this failure case is to ensure you have updated staticHash, not AKSNodeClassHashVersion
```
charliedmcb, line 115:
```suggestion
			// Note: this failure case is to ensure you have updated currentHashVersion, not AKSNodeClassHashVersion
```

pkg/apis/v1alpha2/aksnodeclass_hash_test.go
```diff
@@ -103,4 +104,18 @@ var _ = Describe("Hash", func() {
 		}
 		Expect(nodeClass.Hash()).To(Equal(otherNodeClass.Hash()))
 	})
+	// This test is a sanity check to update the hashing version if the algorithm has been updated.
+	// Note: this will only catch a missing version update, if the staticHash hasn't been updated yet.
+	It("when hashing algorithm updates, we should update the hash version", func() {
+		// If you update the staticHash value, you need to update v1alpha2.AKSNodeClassHashVersion, and set currentHashVersion to the same value
```
charliedmcb, line 110: Wondering if this comment makes more sense next to `staticHash`, or rephrased

test/suites/drift/suite_test.go
```diff
@@ -57,71 +69,660 @@ var _ = AfterEach(func() { env.Cleanup() })
...
+			nodePool = coretest.ReplaceRequirements(nodePool,
+				karpv1.NodeSelectorRequirementWithMinValues{
+					NodeSelectorRequirement: corev1.NodeSelectorRequirement{
+						Key:      corev1.LabelInstanceTypeStable,
+						Operator: corev1.NodeSelectorOpIn,
+						Values:   []string{"Standard_DS4_v2"},
+					},
+				},
+			)
+			// We're expecting to create 3 nodes, so we'll expect to see at most 2 nodes deleting at one time.
```
charliedmcb, line 169: Duplicate of the same comment around async drift not being supported. However it does say `at most`, so maybe just ignore here?

I mean the plan is the support will eventually be there.

test/suites/drift/suite_test.go
```diff
@@ -57,71 +69,660 @@ var _ = AfterEach(func() { env.Cleanup() })
...
+				delete(pod.Annotations, karpv1.DoNotDisruptAnnotationKey)
+				env.ExpectUpdated(pod)
+			}
+
+			env.ConsistentlyExpectDisruptionsUntilNoneLeft(3, 2, 5*time.Minute)
+		})
+		It("should respect budgets for non-empty replace drift", func() {
+			appLabels := map[string]string{"app": "large-app"}
+			nodePool.Labels = appLabels
+			// We're expecting to create 5 nodes, so we'll expect to see at most 3 nodes deleting at one time.
```
charliedmcb, line 236: Another comment around async drift not working.

charliedmcb (dismissed):
Left a few suggestions/thoughts around comments, but overall lgtm.

charliedmcb:
Would probably be good to do one final run of the E2E pipeline before checking in, since this PR is pretty old?

test/suites/drift/suite_test.go
tallaxes: Updated comment

test/suites/drift/suite_test.go
tallaxes: Updated comment

pkg/apis/v1alpha2/aksnodeclass_hash_test.go
tallaxes: Updated there, removed here

test/suites/drift/suite_test.go
```diff
@@ -57,71 +69,666 @@ var _ = AfterEach(func() { env.Cleanup() })
...
+			env.ExpectCreated(dep, nodeClass, nodePool)
+
+			By("deploying multiple replicas, pod per node")
+			startingNodeClaimState := env.EventuallyExpectCreatedNodeClaimCount("==", int(numPods))
+			env.EventuallyExpectCreatedNodeCount("==", int(numPods))
+			env.EventuallyExpectHealthyPodCount(selector, int(numPods))
+
+			By("drifting the nodeClaim with bad configuration that never initializes")
+			nodePool.Spec.Template.Spec.StartupTaints = append(nodePool.Spec.Template.Spec.StartupTaints,
+				corev1.Taint{Key: "example.com/taint", Effect: corev1.TaintEffectPreferNoSchedule})
```
tallaxes, line 648: Assignment (rather than append), losing required Cilium startup taints, was the source of recent issues (like unexpected overprovisioning) in this test

test/suites/drift/suite_test.go
```diff
@@ -57,71 +69,666 @@ var _ = AfterEach(func() { env.Cleanup() })
...
+				},
+			},
+			karpv1.NodePool{
+				Spec: karpv1.NodePoolSpec{
+					Template: nodeClaimTemplate,
+				},
+			},
+		)
+		updatedNodePool.ObjectMeta = nodePool.ObjectMeta
+		updatedNodePool = env.AdaptToClusterConfig(updatedNodePool)
```
tallaxes, line 377: Example of using `AdaptToClusterConfig` on custom coretest.NodePool

test/pkg/environment/common/expectations.go
```diff
@@ -680,13 +680,13 @@ func (env *Environment) EventuallyExpectNodesUntaintedWithTimeout(timeout time.D
...
 
-func (env *Environment) EventuallyExpectNodeClaimCount(comparator string, count int) []*karpv1.NodeClaim {
+func (env *Environment) EventuallyExpectLaunchedNodeClaimCount(comparator string, count int) []*karpv1.NodeClaim {
 	GinkgoHelper()
 	By(fmt.Sprintf("waiting for nodes to be %s to %d", comparator, count))
 	nodeClaimList := &karpv1.NodeClaimList{}
 	Eventually(func(g Gomega) {
 		g.Expect(env.Client.List(env, nodeClaimList, client.HasLabels{test.DiscoveryLabel})).To(Succeed())
-		g.Expect(len(nodeClaimList.Items)).To(BeNumerically(comparator, count),
+		g.Expect(lo.CountBy(nodeClaimList.Items, func(nc karpv1.NodeClaim) bool { return nc.StatusConditions().IsTrue(karpv1.ConditionTypeLaunched) })).To(BeNumerically(comparator, count),
```
charliedmcb, line 689: Why is it necessary to be checking Launched instead?

charliedmcb (approved):
few comments, but overall lgtm.

test/pkg/environment/common/expectations.go
tallaxes: It is not strictly necessary, but a better checkpoint, waiting for NodeClaim to be actually launched (creating VM) rather then just appearing as a resource. A bit more precise state expectation is useful in a test where the failing sequence was hard to troubleshoot.

pkg/apis/v1alpha2/aksnodeclass_hash_test.go
```diff
@@ -0,0 +1,120 @@
...
+	"github.com/samber/lo"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/karpenter/pkg/test"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Hash", func() {
+	// NOTE: When the hashing algorithm is updated, these tests are expected to fail, and the test hash constants here along with the v1alpha2.AKSNodeClassHashVersion need to be updated.
```
tallaxes, line 33:
```suggestion
	// NOTE: When the hashing algorithm is updated, these tests are expected to fail; test hash constants here would have to be updated, and currentHashVersion would have to be updated to the new version matching v1alpha2.AKSNodeClassHashVersion
```

pkg/apis/v1alpha2/aksnodeclass_hash_test.go
tallaxes: Updated the comment for clarity, should be good now

PR #672, opened 2025-02-02
dev: adding a command for creating a dev environment in codespaces from cli

Bryce-Soghigian:
**Description**
I find that I end up creating a codespace after doing some code changes as it requires me to jump into the browser, so I run CI etc after creating a PR. 

This change adds a basic command to launch a codespace inside of vscode with codespace branch="branch-name" so that I have one setup and initialized from the moment I start working. 

make codespace BRANCH=feature-name

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -115,7 +115,20 @@ tidy: ## Recursively "go mod tidy" on all directories where go.mod exists
 download: ## Recursively "go mod download" on all directories where go.mod exists
 	$(foreach dir,$(MOD_DIRS),cd $(dir) && go mod download $(newline))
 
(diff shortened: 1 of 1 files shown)


tallaxes:
> I find that I end up creating a codespace after doing some code changes as it requires me to jump into the browser, so I run CI etc after creating a PR.

I am not sure I understand your workflow. You know you can switch branch in codespace, right? While there are some use case for creating more than one codespace, it is not expected to be a very common workflow needing automation.

PR #675, opened 2025-02-05
feat: validation enforcing vnet guid is set for overlay cni clusters 

Bryce-Soghigian:
**Description**
After we roll out the v0.7.2 release, we can remove the fallback to the vnet client. This pr also introduces stricter validation to vnet guid to make it clear its required. 
- make test

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note

```


pkg/operator/options/options_validation.go
```diff
@@ -39,6 +41,20 @@ func (o Options) Validate() error {
 	)
 }
 
+func (o Options) validateVNETGUID() error {
+	if !o.isAzureCNIWithOverlay() {
```
matthchr, line 45: This seems suspicious to me.
If it's not overlay, but they pass an invalid vnet GUID, we allow it?

Should we parse the vnetGUID if it's supplied always, and then also enforce it's set when overlay is set?

pkg/operator/operator.go
```diff
@@ -235,28 +223,3 @@ func getCABundle(restConfig *rest.Config) (*string, error) {
 	}
 	return ptr.String(base64.StdEncoding.EncodeToString(transportConfig.TLS.CAData)), nil
 }
-
-func getVnetGUID(cfg *auth.Config, subnetID string) (string, error) {
```
matthchr, line 239: Do we document our options anywhere?

Helm chart?
Website?
Go struct?

All of the above?

We should document what this is and how users should set it somewhere. I am also not 100% sure what the experience should be for selfhosted Karpenter. It might be OK to support getting the VNETGUID automatically if the user doesn't set it, but allowing them to set it to avoid the need to assign vnet/read permissions to standalone.

Obviously in the NAP case, VNET Guid will always be supplied.

pkg/operator/options/options_validation.go
Bryce-Soghigian: We only use the vnet guid if overlay is set, so if it's malformed, I wouldn't want to crash the customers' application if they don't need the value. 

Thoughts?

pkg/operator/operator.go
Bryce-Soghigian: We don't have a great set of documented requirements here.

CC: @wdarko1 fyi

Bryce-Soghigian: I would prefer that we make it a fully managed field. VNET GUID unlike bootstrap token for example is rather static and only needed once and easy enough to resolve in our configure_values.sh script. 

There are other fields that may require special treatment for self hosted like bootstrap token, but I don't think vnet guid needs to be one of them. 

I would prefer to rid ourselves of that dependency.

tallaxes: The options should be reflected in the chart values files - and documented at least there. Right now the list of values needs an update.

For self-hosted, the `configure-values.sh` serves as a bridge for generating configuration appropriate for a given cluster; the values template it is targeting is currently passing the options via environment variables, bypassing even chart values that exist. Ideally values serve as the chart "contract" - and the script/template should be updated to target values instead of environment variables.

pkg/operator/options/options_validation.go
Bryce-Soghigian: Ok, I agree with this approach. If we are going to do it this way, lets use uuid.Validate() instead as it just returns a bool along with an error. 

```go
	if o.VnetGUID != "" && uuid.Validate(o.VnetGUID) != nil {
		return fmt.Errorf("vnet-guid %s is malformed", o.VnetGUID)
	}
	if o.isAzureCNIWithOverlay() && o.VnetGUID == "" {
		return fmt.Errorf("vnet-guid cannot be empty for AzureCNI clusters with networkPluginMode overlay")
	}
```

Bryce-Soghigian: Ok modified to use this approach, so I will be resolving this comment.

pkg/operator/operator.go
matthchr: Yeah as far as OSS is concerned, parameters should flow from Helm values to the pod somehow. Everything needs to be configurable via values.yaml (for the chart) and we should at least put documentation into values.yaml that describes each option and what it does.

If that's not what we're doing now, should we either fix it here or create a workitem tracking it? I see that our values.yaml already does some of this. Another example is [ASO](https://github.com/Azure/azure-service-operator/blob/main/v2/charts/azure-service-operator/values.yaml)

matthchr: I see what we're doing now is just using the broad values.yaml env overrides, which yeah is fine but ideally we'd use specific parameters with specific documentation (IMO).

matthchr (dismissed):

tallaxes (approved):

tallaxes (approved):

PR #677, opened 2025-02-06
fix(API): add missing validation rules for v1beta1

tallaxes:
**Description**

Duplicate missing validation rules from v1 into v1beta1 API

* `make presubmit`

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note

```

### DIFF
--- a/hack/validation/labels.sh
+++ b/hack/validation/labels.sh
@@ -29,6 +29,9 @@ rule=${rule//\"/\\\"}            # escape double quotes
 rule=${rule//$'\n'/}             # remove newlines
 rule=$(echo "$rule" | tr -s ' ') # remove extra spaces
 
+# check that .spec.versions has 2 entries
+[[ $(yq e '.spec.versions | length' pkg/apis/crds/karpenter.sh_nodepools.yaml) -eq 2 ]] || { echo "expected two versions"; exit 1; }
+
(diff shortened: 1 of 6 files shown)


hack/validation/labels.sh
```diff
@@ -29,6 +29,13 @@ rule=${rule//\"/\\\"}            # escape double quotes
 rule=${rule//$'\n'/}             # remove newlines
 rule=$(echo "$rule" | tr -s ' ') # remove extra spaces
 
+# nodepool
+# v1beta1
+printf -v expr '.spec.versions[1].schema.openAPIV3Schema.properties.spec.properties.template.properties.metadata.properties.labels.x-kubernetes-validations +=
```
matthchr, line 34: Should we have antoher check that checks that `.spec.versions` is length 2?

So if there's another new version (or we remove a version?) we get a failure?

matthchr (approved):
One small comment

hack/validation/labels.sh
tallaxes: Added

tallaxes:
#678 to fix vulncheck

matthchr (approved):

Bryce-Soghigian (approved):

PR #680, opened 2025-02-07
feat: support failSwapOn in kubelet config

tallaxes:
Fixes # 

**Description**

*

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/charts/karpenter-crd/templates/karpenter.azure.com_aksnodeclasses.yaml
+++ b/charts/karpenter-crd/templates/karpenter.azure.com_aksnodeclasses.yaml
@@ -105,6 +105,14 @@ spec:
                     - none
                     - static
                     type: string
+                  failSwapOn:
+                    description: |-
+                      failSwapOn tells the Kubelet to fail to start if swap is enabled on the node.
+                      Note: this is currently not propagated in self-hosted scenarios,
+                      as those currently configure the nodes without swap.
+                      Default: true
+                      kubebuilder:default:=true
+                    type: boolean
                   imageGCHighThresholdPercent:
                     description: |-
                       ImageGCHighThresholdPercent is the percent of disk usage after which image
--- a/pkg/providers/imagefamily/bootstrap/aksbootstrap.go
+++ b/pkg/providers/imagefamily/bootstrap/aksbootstrap.go
@@ -583,6 +583,10 @@ func KubeletConfigToMap(kubeletConfig *KubeletConfiguration) map[string]string {
 	if kubeletConfig.ImageGCLowThresholdPercent != nil {
 		args["--image-gc-low-threshold"] = fmt.Sprintf("%d", ptr.Int32Value(kubeletConfig.ImageGCLowThresholdPercent))
 	}
+	// ShouldConfigureSwapFile is currently hardcoded to false, so can't support propagating this yet
+	// if kubeletConfig.FailSwapOn != nil {
+	//	args["--fail-swap-on"] = fmt.Sprintf("%t", lo.FromPtr(kubeletConfig.FailSwapOn))
+	//}
 	if kubeletConfig.CPUCFSQuota != nil {
 		args["--cpu-cfs-quota"] = fmt.Sprintf("%t", lo.FromPtr(kubeletConfig.CPUCFSQuota))
 	}
--- a/pkg/providers/instancetype/suite_test.go
+++ b/pkg/providers/instancetype/suite_test.go
@@ -466,6 +466,7 @@ var _ = Describe("InstanceType Provider", func() {
 			nodeClass.Spec.Kubelet = &v1alpha2.KubeletConfiguration{
 				ImageGCHighThresholdPercent: lo.ToPtr(int32(30)),
 				ImageGCLowThresholdPercent:  lo.ToPtr(int32(20)),
+				FailSwapOn:                  lo.ToPtr(false), // opposite of default
 				CPUCFSQuota:                 lo.ToPtr(true),
 			}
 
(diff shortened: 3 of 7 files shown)


pkg/providers/instancetype/suite_test.go
```diff
@@ -466,6 +466,7 @@ var _ = Describe("InstanceType Provider", func() {
 			nodeClass.Spec.Kubelet = &v1alpha2.KubeletConfiguration{
 				ImageGCHighThresholdPercent: lo.ToPtr(int32(30)),
 				ImageGCLowThresholdPercent:  lo.ToPtr(int32(20)),
+				FailSwapOn:                  lo.ToPtr(false), // opposite of default
```
Bryce-Soghigian, line 469: Should we add comments for this in case of other boolean defaults to take it obvious that we are testing something other than the default?

pkg/providers/instancetype/suite_test.go
```diff
@@ -466,6 +466,7 @@ var _ = Describe("InstanceType Provider", func() {
 			nodeClass.Spec.Kubelet = &v1alpha2.KubeletConfiguration{
 				ImageGCHighThresholdPercent: lo.ToPtr(int32(30)),
 				ImageGCLowThresholdPercent:  lo.ToPtr(int32(20)),
```
Bryce-Soghigian, line 468: Do we also want to add a test for the default/expected setting?

charts/karpenter-crd/templates/karpenter.azure.com_aksnodeclasses.yaml
```diff
@@ -105,6 +105,14 @@ spec:
                     - none
                     - static
                     type: string
+                  failSwapOn:
```
Bryce-Soghigian, line 108: For all of our kubelet configuration settings, should we also include the upstream kubelet documentation? In this case it seems your comment also parallels well with the upstream documentation, and our default is the same as the upstream default.

Seems there are a couple of defaults, Karpenter, AKS, AWS and upstream. Is there any of these flags that differ between the aks default and the karpenter default? 

Also, if users are migrating from AWS, it might make sense for them to be aware of the aws defaults, ie what kubelet behavior is changing when moving to AKS from EKS karpenter. 

Not needed in this PR, but maybe a consideration? 
----
Reference Kubelet Upstream Configuration for FailSwapOn
--fail-swap-on     Default: true
--
  | Makes the kubelet fail to start if swap is enabled on the node. (DEPRECATED: This parameter should be set via the config file specified by the kubelet's --config flag. See kubelet-config-file for more information.)

--fail-swap-on     Default: true
Makes the kubelet fail to start if swap is enabled on the node. (DEPRECATED: This parameter should be set via the config file specified by the kubelet's --config flag. See [kubelet-config-file](https://kubernetes.io/docs/tasks/administer-cluster/kubelet-config-file/) for more information.)

Bryce-Soghigian (approved):
Overall it looks good, but this is yet another feature where we support it in managed but don't support it in self hosted. 

We should make it very clear in our readme that self hosted is turning into more of a toy/testing tool, and users should be using the managed addon. 

This feature is more minor, but some other major features are now only going to be supported on managed.

pkg/providers/instancetype/suite_test.go
tallaxes: Yes, opportunistically (I had this fixed for CPUCFSQuota)

pkg/providers/instancetype/suite_test.go
tallaxes: Maybe - as part of general KubeletConfiguration tests

charts/karpenter-crd/templates/karpenter.azure.com_aksnodeclasses.yaml
tallaxes: In all custom kubelet fields we already strive to use existing kubelet field naming, documentation and defaults - and I expect AWS is the same.

pkg/providers/imagefamily/bootstrap/aksbootstrap.go
```diff
@@ -583,6 +583,10 @@ func KubeletConfigToMap(kubeletConfig *KubeletConfiguration) map[string]string {
 	if kubeletConfig.ImageGCLowThresholdPercent != nil {
 		args["--image-gc-low-threshold"] = fmt.Sprintf("%d", ptr.Int32Value(kubeletConfig.ImageGCLowThresholdPercent))
 	}
+	// ShouldConfigureSwapFile is currently hardcoded to false, so can't support propagating this yet
```
matthchr, line 586: This comment doesn't seem to match the actual field it's talking about?
`ShouldConfigureSwapFile` vs `FailSwapOn`

pkg/providers/instancetype/suite_test.go
```diff
@@ -480,8 +481,10 @@ var _ = Describe("InstanceType Provider", func() {
 				"eviction-hard":           "memory.available<750Mi",
 				"image-gc-high-threshold": "30",
 				"image-gc-low-threshold":  "20",
-				"cpu-cfs-quota":           "true",
+				"cpu-cfs-quota":           "false",
 				"max-pods":                "250",
+				// we don't currently propagate fail-swap-on
```
matthchr, line 486: Is this a TODO for this PR to fix?

pkg/providers/imagefamily/bootstrap/aksbootstrap.go
tallaxes: `ShouldConfigSwapFile` is a different internal setting in bootstrap. It - and possibly other settings like `SwapSizeMB` - would need to be coordinated to support proper control over swap file. For self-hosted, probably worth waiting for scriptless bootstrap to implement.

pkg/providers/instancetype/suite_test.go
tallaxes: It may be worth waiting for scriptless support in bootstrap to implement this for self-hosted - unless somebody really needs it right now.

Issue #518, opened 2024-10-12
Spot instances ephemeral disk size limit is not taken into consideration


krukowskid:
### Version

__Karpenter Version:__ v0.5.4

__Kubernetes Version:__ v1.29.8

### Expected Behavior

Karpenter should create spot instance with additional disk, or use smaller size to fir into VM limits.

### Actual Behavior

It seems that karpenter is trying to create default 128Gb Ephemeral OS disk for spot instance and ends-up in create loop

### Steps to Reproduce the Problem

1. Create node pool with capacity-type `spot` and instance with max allowed ephemeral disk size less than 128GB like `Standard_D4ds_v4`
2. Deploy workload to let karpenter provision new VM

### Resource Specs and Logs

```
{
  "level": "ERROR",
  "time": "2024-10-12T14:39:39.784Z",
  "logger": "controller",
  "message": "Reconciler error",
  "commit": "846ef96",
  "controller": "nodeclaim.lifecycle",
  "controllerGroup": "karpenter.sh",
  "controllerKind": "NodeClaim",
  "NodeClaim": {
    "name": "spot-v7m6r"
  },
  "namespace": "",
  "name": "spot-v7m6r",
  "reconcileID": "8c31e447-0fda-4f74-8b96-b8acc11435e2",
  "error": "launching nodeclaim, creating instance, virtualMachine.BeginCreateOrUpdate for VM \"aks-spot-v7m6r\" failed: PUT https://management.azure.com/subscriptions/2dcb8032-974b-414f-ae22-f2f4aa2ffea6/resourceGroups/MC_akskarpenter_karpenter_swedencentral/providers/Microsoft.Compute/virtualMachines/aks-spot-v7m6r\n--------------------------------------------------------------------------------\nRESPONSE 400: 400 Bad Request\nERROR CODE: NotSupported\n--------------------------------------------------------------------------------\n{\n  \"error\": {\n    \"code\": \"NotSupported\",\n    \"message\": \"OS disk of Ephemeral VM with size greater than 100 GB is not allowed for VM size Standard_D4ds_v4 when the DiffDiskPlacement is CacheDisk. Please refer to https://aka.ms/Ephemeral for more details.\"\n  }\n}\n--------------------------------------------------------------------------------\n"
}
```

### Community Note

[...]

PR #688, opened 2025-02-10
feat: decrease minimum allowed size for OsDiskSizeGB from 100 to 30

bongi23:
Fixes # 

**Description**
Karpenter supports deploying linux only instances and the minimum disk size allowed for these is 30GB. The changes proposed allowed to request VMs with a minimum OSDiskSize of 30GB instead of 100GB

Manually modified AKSNodeClass CRD on my company production cluster and let karpenter deploy successfully a VM with OSDiskSize less than 100
*

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note

```


bongi23:

Bryce-Soghigian:
We experience many performance incidents regarding disk sizes smaller than 100gb, can you provide more justification as to why you want to reduce the limit?

alexmaurizio:
> We experience many performance incidents regarding disk sizes smaller than 100gb, can you provide more justification as to why you want to reduce the limit?

One example is running a staging environment (like we do, we run two mirror kubernetes for staging and production) where availability is not crucial and we can tolerate destroying a node if any problem arises.

A 100GB disk will allocate a Premium SSD class P10 (128GB) by default on certain Azure tiers, which is 4 times as expensive as a P4 (32GB) disk. There are a lot of instances which do not have (and we don't plan to use) ephemereal disk, or as everyone calls it, local NVME storage. Managed disks are enough and having 5 inexpensive disks running 128GB P10 disks is actually costing more in storage than in machines, which is frankly stupid. 

The default should stay the same, just allow a configuration for users.

bongi23:
> We experience many performance incidents regarding disk sizes smaller than 100gb, can you provide more justification as to why you want to reduce the limit?

What kind of performance incidents? Are they related to disk performance?
In addition to what said by @alexmaurizio, our microservices are stateless (and I think this is a very common scenario) and disk performance are negligible. In any case, I should be free to save money and accept bad performance on staging environments at least 😄

lmoretto:
+1 for this. The client should have the freedom (and at the same time the responsibility) to define their infrastructure requirements, without facing such opinionated defaults and constraints

sgtraverso:
Agree with this. Let the user define infrastructure requirements, pls.

bongi23:
Maybe related to https://github.com/Azure/karpenter-provider-azure/issues/518

bongi23:
@Bryce-Soghigian hei Bryce, sorry to bother you. Can you give us a feedback please?

Bryce-Soghigian:
tangentially related to this change, 

CC: @tallaxes 

image

Perhaps we should be doing better defaulting based on vcpu unless the user specifed the disk size in their nodeclass?

tallaxes:
Let's match what AKS is currently doing - at least for minimum, but possibly (and maybe separately) for default as well. @Bryce-Soghigian, please also check GB vs GiB just in case.

tallaxes (approved):

PR #694, opened 2025-02-12
fix: make presubmit

Bryce-Soghigian:
**Description**
Seems make presubmit somehow was bypassed, this updates that. If you run make presubmit on main, it generates content. 

*

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


charliedmcb (approved):
Still curious why this started failing on main, but good to unblock.

tallaxes:
> Still curious why this started failing on main, but good to unblock.

That's because we currently don't pin the version of controller-gen (use latest)

PR #697, opened 2025-02-16
feat: setting network plugin to none in the agentbaker contract if using overlay

Bryce-Soghigian:
**Description**
This will delegate the installation of the cni to daemonsets (azure-cns) etc. This also introduces a label azure-cni-overlay which adds better consistency between karpenter provided nodes and nodes provided from aks. We also introduce tests to explicitly validate that if the network plugin mode on the cluster is azure and uses azure cni overlay, that we properly delegate that installation

- make test 
- make az-all-cni-overlay

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note
adding support for azure-cni overlay without cilium 
```


pkg/providers/imagefamily/bootstrap/aksbootstrap.go
```diff
@@ -453,7 +455,10 @@ func (a AKS) applyOptions(nbv *NodeBootstrapVariables) {
 	nbv.ResourceGroup = a.ResourceGroup
 	nbv.UserAssignedIdentityID = a.KubeletIdentityClientID
 
-	nbv.NetworkPlugin = a.NetworkPlugin
+	// if we are using overlay, we delegate the installation of the cni to the cns daemonset and will put none as to not drop the
+	// 10-azure.conflist that will use nodesubnet
+	nbv.NetworkPlugin = lo.Ternary(a.NetworkPluginMode == consts.NetworkPluginModeOverlay, consts.NetworkPluginNone, a.NetworkPlugin)
+
```
Bryce-Soghigian, line 461: Follow-up here is if we are using 1.32 we will also set network plugin to None, for now just fixing for all existing karpenter clusters.

Bryce-Soghigian:
Passing E2E Run: https://github.com/Azure/karpenter-provider-azure/actions/runs/13350057733

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -38,10 +39,11 @@ import (
 const (
 	karpenterManagedTagKey = "karpenter.azure.com/cluster"
 
-	vnetDataPlaneLabel      = "kubernetes.azure.com/ebpf-dataplane"
-	vnetSubnetNameLabel     = "kubernetes.azure.com/network-subnet"
-	vnetGUIDLabel           = "kubernetes.azure.com/nodenetwork-vnetguid"
-	vnetPodNetworkTypeLabel = "kubernetes.azure.com/podnetwork-type"
+	vnetDataPlaneLabel       = "kubernetes.azure.com/ebpf-dataplane"
+	vnetAzureCNIOverlayLabel = "kubernetes.azure.com/azure-cni-overlay"
+	vnetSubnetNameLabel      = "kubernetes.azure.com/network-subnet"
```
tallaxes, line 44: Why do these variables have 'vnet' prefix?

pkg/providers/imagefamily/bootstrap/aksbootstrap.go
```diff
@@ -453,7 +455,10 @@ func (a AKS) applyOptions(nbv *NodeBootstrapVariables) {
 	nbv.ResourceGroup = a.ResourceGroup
 	nbv.UserAssignedIdentityID = a.KubeletIdentityClientID
 
-	nbv.NetworkPlugin = a.NetworkPlugin
+	// if we are using overlay, we delegate the installation of the cni to the cns daemonset and will put none as to not drop the
+	// 10-azure.conflist that will use nodesubnet
```
tallaxes, line 459:
```suggestion
    // In some cases we don't need the node to do anything with CNI so tell it NetworkPlugin is 'none'
```

pkg/providers/imagefamily/bootstrap/aksbootstrap.go
```diff
@@ -453,7 +455,10 @@ func (a AKS) applyOptions(nbv *NodeBootstrapVariables) {
...
+	nbv.NetworkPlugin = lo.Ternary(a.NetworkPluginMode == consts.NetworkPluginModeOverlay, consts.NetworkPluginNone, a.NetworkPlugin)
```
tallaxes, line 460: Please explicitely test for (Plugin == Azure && PluginMode = Overlay)
And let's do the same for Dataplane == Cilium

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -38,10 +39,11 @@ import (
 const (
 	karpenterManagedTagKey = "karpenter.azure.com/cluster"
 
-	vnetDataPlaneLabel      = "kubernetes.azure.com/ebpf-dataplane"
-	vnetSubnetNameLabel     = "kubernetes.azure.com/network-subnet"
-	vnetGUIDLabel           = "kubernetes.azure.com/nodenetwork-vnetguid"
-	vnetPodNetworkTypeLabel = "kubernetes.azure.com/podnetwork-type"
+	vnetDataPlaneLabel       = "kubernetes.azure.com/ebpf-dataplane"
+	vnetAzureCNIOverlayLabel = "kubernetes.azure.com/azure-cni-overlay"
```
tallaxes, line 43: Why do these constants have 'vnet' prefix?

Bryce-Soghigian: Good question, followed the naming pattern but was wondering the same thing. I'm going to remove the prefix.

pkg/providers/instancetype/suite_test.go
```diff
@@ -1222,6 +1223,53 @@ var _ = Describe("InstanceType Provider", func() {
...
+				"kubernetes.azure.com/podnetwork-type=overlay",
+			)),
+		Entry("Azure CNI w Overlay w Cilium",
+			"azure", "overlay", "cilium",
+			"none",
+			sets.New(
+				"kubernetes.azure.com/azure-cni-overlay=true",
+				"kubernetes.azure.com/network-subnet=karpentersub",
+				"kubernetes.azure.com/nodenetwork-vnetguid=test-vnet-guid",
+				"kubernetes.azure.com/podnetwork-type=overlay",
```
Bryce-Soghigian, line 1268: We could have a separate set as well with these strings that are used in multiple places, and reuse them. Not trying to tackle in this PR, but worthy refactor PR to follow-up.

pkg/providers/imagefamily/bootstrap/aksbootstrap.go
Bryce-Soghigian: Added tests

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -123,7 +125,7 @@ func (p *Provider) getStaticParameters(ctx context.Context, instanceType *cloudp
 
 	subnetID := lo.Ternary(nodeClass.Spec.VNETSubnetID != nil, lo.FromPtr(nodeClass.Spec.VNETSubnetID), options.FromContext(ctx).SubnetID)
 
-	if options.FromContext(ctx).NetworkPlugin == consts.NetworkPluginAzure && options.FromContext(ctx).NetworkPluginMode == consts.NetworkPluginModeOverlay {
+	if isAzureCNIOverlay(ctx) {
```
Bryce-Soghigian, line 128: this could go into a shared bit of utils, because this isn't the only place we check if we are using azure cni overlay

(operator.go, somewhere in instance.go)

pkg/providers/imagefamily/bootstrap/aksbootstrap.go
tallaxes: Sorry, I did not make myself clear. I think
* _the check here_ needs to include `a.NetworkPlugin == const.NetworkPluginAzure` condition
* there should be another separate check for Dataplane == Cilium - which should also result in NetworkPlugin none

pkg/providers/instancetype/suite_test.go
```diff
@@ -1223,15 +1222,54 @@ var _ = Describe("InstanceType Provider", func() {
...
+			"azure", "overlay", "",
+			"none",
+			sets.New(
+				"kubernetes.azure.com/azure-cni-overlay=true",
+				"kubernetes.azure.com/network-subnet=karpentersub",
+				"kubernetes.azure.com/nodenetwork-vnetguid=test-vnet-guid",
+				"kubernetes.azure.com/podnetwork-type=overlay",
+			)),
+		Entry("Azure CNI w Overlay w Cilium",
+			"azure", "overlay", "cilium",
```
tallaxes, line 1262: My understanding is Cilium dataplane should result in "none" even if overlay is not used

Bryce-Soghigian: Looks like you are correct, I had no idea this was possible, after looking at the validation, it does look like you can create a non overlay cluster if you have a feature flag enabled for cilium node subnet. 


```
az aks create          --name kwera --resource-group kwera --attach-acr kwera \
                --enable-managed-identity --node-count 3 --generate-ssh-keys -o none --network-dataplane cilium --network-plugin azure \
The behavior of this command has been altered by the following extension: aks-preview
The new node pool will enable SSH access, recommended to use '--ssh-access disabled' option to disable SSH access for the node pool to make it more secure.
(BadRequest) Cilium dataplane in node subnet mode requires feature flag Microsoft.ContainerService/EnableCiliumNodeSubnet to be registered.
Code: BadRequest
Message: Cilium dataplane in node subnet mode requires feature flag Microsoft.ContainerService/EnableCiliumNodeSubnet to be registered.
Target: networkProfile.networkDataplane
```

pkg/providers/imagefamily/bootstrap/aksbootstrap.go
wedaly: yes to what Alex said.

The plan is that by k8s 1.32+ the logic will simplify to: if Azure CNI -> CNS configures the CNI (= the node has networkPlugin=none), for all variations of Azure CNI. We're not quite there yet though.

wedaly (dismissed):

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -162,13 +164,28 @@ func (p *Provider) getStaticParameters(ctx context.Context, instanceType *cloudp
 		ClusterID:                      options.FromContext(ctx).ClusterID,
 		APIServerName:                  options.FromContext(ctx).GetAPIServerName(),
 		KubeletClientTLSBootstrapToken: options.FromContext(ctx).KubeletClientTLSBootstrapToken,
-		NetworkPlugin:                  options.FromContext(ctx).NetworkPlugin,
+		NetworkPlugin:                  getAgentbakerNetworkPlugin(ctx),
```
Bryce-Soghigian, line 167: This is cleaner imo than the previous approach, lmk if you disagree.

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -213,9 +230,10 @@ func (p *Provider) getVnetInfoLabels(subnetID string) (map[string]string, error)
 		return nil, err
 	}
 	vnetLabels := map[string]string{
-		vnetSubnetNameLabel:     vnetSubnetComponents.SubnetName,
-		vnetGUIDLabel:           p.vnetGUID,
-		vnetPodNetworkTypeLabel: consts.NetworkPluginModeOverlay,
+		subnetNameLabel:      vnetSubnetComponents.SubnetName,
+		vnetGUIDLabel:        p.vnetGUID,
+		azureCNIOverlayLabel: strconv.FormatBool(true),
```
charliedmcb, line 235: We only support overlay atm, so this should already be true?

tallaxes: We support non-overlay as well

tallaxes: This is ripe for refactoring, sometime very soon. Yes, getVnetInfoLabels is currently only called in the context of Azure CNI Overlay, but it is not obvious within the function.

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -162,13 +164,28 @@ func (p *Provider) getStaticParameters(ctx context.Context, instanceType *cloudp
...
-		NetworkPlugin:                  options.FromContext(ctx).NetworkPlugin,
+		NetworkPlugin:                  getAgentbakerNetworkPlugin(ctx),
 		NetworkPolicy:                  options.FromContext(ctx).NetworkPolicy,
 		SubnetID:                       subnetID,
 		ClusterResourceGroup:           p.clusterResourceGroup,
 	}, nil
 }
 
+func getAgentbakerNetworkPlugin(ctx context.Context) string {
+	if isAzureCNIOverlay(ctx) || isCiliumNodeSubnet(ctx) {
```
tallaxes, line 175: What happened to the comment we established was useful?

```
 // In some cases we don't need the node to do anything with CNI so tell it NetworkPlugin is 'none'
```

pkg/providers/imagefamily/bootstrap/aksbootstrap.go
```diff
@@ -454,6 +454,7 @@ func (a AKS) applyOptions(nbv *NodeBootstrapVariables) {
 	nbv.UserAssignedIdentityID = a.KubeletIdentityClientID
 
 	nbv.NetworkPlugin = a.NetworkPlugin
+
```
tallaxes, line 457: Intentional new line?

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -162,13 +164,28 @@ func (p *Provider) getStaticParameters(ctx context.Context, instanceType *cloudp
...
+	return consts.NetworkPluginAzure
+}
+
+func isCiliumNodeSubnet(ctx context.Context) bool {
+	return options.FromContext(ctx).NetworkPlugin == consts.NetworkPluginAzure && options.FromContext(ctx).NetworkPluginMode == consts.NetworkPluginModeNone && options.FromContext(ctx).NetworkDataplane == consts.NetworkDataplaneCilium
+}
+
+func isAzureCNIOverlay(ctx context.Context) bool {
+	return options.FromContext(ctx).NetworkPlugin == consts.NetworkPluginAzure && options.FromContext(ctx).NetworkPluginMode == consts.NetworkPluginModeOverlay
+}
```
tallaxes, line 187: Should be extracting settings from context and passing into helpers, rather than making them depend on context. Generally, the earlier settings are extracted from context the better.

tallaxes (approved):
LGTM, with suggestions

tallaxes:
Merging, follow-ups (if any) can be separate

PR #707, opened 2025-02-22
fix: karpenter should generate a number of ipconfigurations matching max pods

Bryce-Soghigian:
**Description**
When we added max pods to the AKSNodeClass, we did not update the ip configurations to match the setting. Instead, it generates the default number of IP configurations. 

MaxPods is now configurable, allowing users to override the maximum number of pods that can run on the karpenter nodes. The default behavior varies based on the network plugin:
	•	Azure CNI (v1 or NodeSubnet) → Default: 30
	•	Azure CNI (overlay) → Default: 250
	•	No Network Plugin (None) → Default: 250
	•	Other Network Plugins → Default: 110 (Kubernetes standard)

This PR also introduces a minimum and maximum value for max pods based on the AKS Defaults. 

- make presubmit
- new test coverage
- E2E: https://github.com/Azure/karpenter-provider-azure/actions/runs/13889224621

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note
MaxPods is now configurable, allowing users to override the maximum number of pods on a worker node instance, following AKS defaults. The value must be between the minimum of 10 and the maximum of 250, with Azure CNI (v1 or NodeSubnet) defaulting to 30, Azure CNI (overlay) and None defaulting to 250, and others defaulting to 110.
```


tallaxes (changes requested):
Yes, let's fix the max pods logic, but the default should depend on CNI (can't be specified in CRD), and the same value determined should be used consistently - computing the number of IPs, passed to kublet, instance type capacity calculation (currently wrong), etc.

pkg/apis/v1alpha2/aksnodeclass.go
```diff
@@ -53,8 +53,8 @@ type AKSNodeClassSpec struct {
 	// +optional
 	Kubelet *KubeletConfiguration `json:"kubelet,omitempty" hash:"ignore"`
 	// MaxPods is an override for the maximum number of pods that can run on a worker node instance.
-	// +kubebuilder:validation:Minimum:=0
-	// +kubebuilder:default=250
+	// +kubebuilder:validation:Minimum:=10
```
tallaxes, line 56: Please add a comment on reasoning (I presume this matches AKS-enforced minimum?)

(This is also a breaking change, but an acceptable one I think)

pkg/apis/v1alpha2/aksnodeclass.go
```diff
@@ -53,8 +53,8 @@ type AKSNodeClassSpec struct {
...
+	// +kubebuilder:validation:Maximum:=250
```
tallaxes, line 57: Where does this limitation come from? Please add a comment (if it is real)

pkg/consts/consts.go
```diff
@@ -27,7 +27,9 @@ const (
 	NetworkDataplaneCilium = "cilium"
 	NetworkDataplaneAzure  = "azure"
 
-	DefaultKubernetesMaxPods = 250
+	DefaultOverlayMaxPods    = 250
+	DefaultNodeSubnetMaxPods = 30
+	DefaultKubernetesMaxPods = 110
```
tallaxes, line 32: Please add a comment on source (I presume this matches AKS defaults)

pkg/apis/v1alpha2/aksnodeclass.go
```diff
@@ -54,7 +54,7 @@ type AKSNodeClassSpec struct {
 	Kubelet *KubeletConfiguration `json:"kubelet,omitempty" hash:"ignore"`
 	// MaxPods is an override for the maximum number of pods that can run on a worker node instance.
 	// +kubebuilder:validation:Minimum:=0
-	// +optional
```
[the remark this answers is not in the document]
tallaxes: Please address this older feedback

pkg/providers/instance/instance.go
```diff
@@ -295,7 +295,7 @@ func (p *DefaultProvider) newNetworkInterfaceForVM(opts *createNICOptions) armne
 		// NOTE: Unlike AKS RP, this logic does not reduce secondary IP count by the number of expected hostNetwork pods, favoring simplicity instead
 		// TODO: When MaxPods comes from the AKSNodeClass kubelet configuration, get the number of secondary
 		// ips from the nodeclass instead of using the default
```
tallaxes, line 297: This comment no longer applies

tallaxes (changes requested):
Nice! Left some minor requests for code comments and questions around min/max

(Please don't forget to leave evidence of passing E2E tests)

pkg/apis/v1alpha2/aksnodeclass.go
Bryce-Soghigian: 10 is the aks minimum, and 250 is the aks maximum.

pkg/consts/consts.go
Bryce-Soghigian: Yes this comes directly from the AKS defaults.

Bryce-Soghigian:
> Nice! One more place that needs update is instance type capacity calculation:
We should add some testing to ensure this isn't broken in the future, as all tests pass here.

pkg/providers/instance/instance.go
Bryce-Soghigian: removed

pkg/apis/v1alpha2/aksnodeclass.go
Bryce-Soghigian: Added much better comments on 

A: Min 
B: Max 
C: Default

pkg/apis/v1alpha2/aksnodeclass.go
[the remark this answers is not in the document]
Bryce-Soghigian: addressed

pkg/apis/v1alpha2/aksnodeclass.go
```diff
@@ -53,7 +53,18 @@ type AKSNodeClassSpec struct {
...
+	// See minimum + maximum pods per node documentation: https://learn.microsoft.com/en-us/azure/aks/concepts-network-ip-address-planning#maximum-pods-per-node
+	// Defaults if unspecified in the nodeclass are:
+	// _______________________________________________________________
+	// | Network Plugin | Network Plugin Mode     | Default Max Pods |
+	// |----------------|-------------------------|------------------|
+	// | Azure		    | "" aka v1 or NodeSubnet | 30               |
+	// | Azure 		    | "overlay" 			  | 250 			 |
+	// | None			| *						  | 250				 |
+	// | Anything Else  | Anything Else 		  | 110 			 | <--- Being the kubernetes default
+	// |-------------------------------------------------------------|
```
Bryce-Soghigian, line 65: image
Looks like this locally, wonder why it looks like this here? Regardless should probably document this in a cleaner way then.

tallaxes: Because tabs

pkg/utils/utils.go
```diff
@@ -138,3 +143,23 @@ func PrettySlice[T any](s []T, maxItems int) string {
...
+		return consts.DefaultNetPluginNoneMaxPods
+	}
+	if networkPlugin == consts.NetworkPluginAzure && networkPluginMode == consts.NetworkPluginModeOverlay {
+		return consts.DefaultOverlayMaxPods
+	}
+	if networkPlugin == consts.NetworkPluginAzure && networkPluginMode == consts.NetworkPluginModeNone {
+		return consts.DefaultNodeSubnetMaxPods
+	}
+	// if not an overlay or nodesubnet network-plugin-mode, return the default
+	return consts.DefaultKubernetesMaxPods
```
tallaxes, line 164: Maybe switch would be more readable? (Left checking nodeClass.Spec.MaxPods out on purpose, seems special)

```suggestion
	switch {
	case networkPlugin == consts.NetworkPluginNone:
		return consts.DefaultNetPluginNoneMaxPods
	case networkPlugin == consts.NetworkPluginAzure && networkPluginMode == consts.NetworkPluginModeOverlay:
		return consts.DefaultOverlayMaxPods
	case networkPlugin == consts.NetworkPluginAzure && networkPluginMode == consts.NetworkPluginModeNone:
		return consts.DefaultNodeSubnetMaxPods
	default:
		return consts.DefaultKubernetesMaxPods
	}
```

pkg/providers/instancetype/instancetype.go
```diff
@@ -312,16 +311,9 @@ func ephemeralStorage(nodeClass *v1alpha2.AKSNodeClass) *resource.Quantity {
...
-	switch {
-	case nc.Spec.MaxPods != nil:
-		count = int64(ptr.Int32Value(nc.Spec.MaxPods))
-	default:
-		count = 110
-	}
-	return resources.Quantity(fmt.Sprint(count))
+func pods(ctx context.Context, nc *v1alpha2.AKSNodeClass) *resource.Quantity {
+	networkPlugin, networkPluginMode := options.FromContext(ctx).NetworkPlugin, options.FromContext(ctx).NetworkPluginMode
+	return resources.Quantity(fmt.Sprint(int64(utils.GetMaxPods(nc, networkPlugin, networkPluginMode))))
```
tallaxes, line 316: Best avoid parsing if possible
```suggestion
	return resource.NewQuantity(int64(utils.GetMaxPods(nc, networkPlugin, networkPluginMode)), resource.DecimalSI)
```

tallaxes (commented):
LGTM. Minor suggestions

> Nice! One more place that needs update is instance type capacity calculation:
> We should add some testing to ensure this isn't broken in the future, as all tests pass here.

Yes, please. This PR would be a good place as it updates the capacity logic, and it is impactful for scheduling / provisioning.

pkg/providers/instancetype/suite_test.go
```diff
@@ -767,10 +766,70 @@ var _ = Describe("InstanceType Provider", func() {
...
+		BeforeEach(func() {
+			ctx = options.ToContext(ctx, test.Options())
+		})
+		It("should set pods equal to MaxPods in the AKSNodeClass when specified", func() {
+			maxPods := int32(150)
+			nodeClass.Spec.MaxPods = lo.ToPtr(maxPods)
+
+			instanceTypes, err := azureEnv.InstanceTypesProvider.List(ctx, nodeClass)
+			Expect(err).NotTo(HaveOccurred())
+			ExpectCapacityPodsToMatchMaxPods(instanceTypes, maxPods)
```
tallaxes, line 779: Thinking about it some more, can you add another NodeClass with a different MaxPods in the same test? I suspect it will fail, the synthetic key for instance type cache does not include MaxPods, would likely get incorrect cache hit

https://github.com/Azure/karpenter-provider-azure/blob/53f7670982be737ea00e7a5d308f0e696c3c0214/pkg/providers/instancetype/instancetypes.go#L114-L119

tallaxes (changes requested):

pkg/providers/instancetype/suite_test.go
Bryce-Soghigian: This is a great catch!! Updated the code now.

pkg/test/aksnodeclass.go
```diff
@@ -46,6 +46,7 @@ func AKSNodeClass(overrides ...v1alpha2.AKSNodeClass) *v1alpha2.AKSNodeClass {
 	// we sometimes reference a test.AKSNodeClass without applying it, and in that case we need to set the default values ourselves
 	nc.Spec.OSDiskSizeGB = lo.ToPtr[int32](128)
 	nc.Spec.ImageFamily = lo.ToPtr(v1alpha2.Ubuntu2204ImageFamily)
+
```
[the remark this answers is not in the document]
Bryce-Soghigian: I thought it improved readability a bit but maybe not. Reverted the change.

tallaxes (approved):
LGTM 👍

> (Please don't forget to leave evidence of passing E2E tests and update the description / release note)

👆

Bryce-Soghigian:
Running E2E Tests Here: https://github.com/Azure/karpenter-provider-azure/actions/runs/13889224621

PR #714, opened 2025-02-28
feat: add azure-cni-overlay label and set network plugin to none in some cases

matthchr:
* fix: setting network plugin to none in the agentbaker contract if we are using overlay net plugin mode. This will delegate the installation of the cni to daemonsets (azure-cns) etc. This also introduces a label azure-cni-overlay which adds better consistency between karpenter provided nodes and nodes provided from aks. We also introduce tests to explicitly validate that if the network plugin mode on the cluster is azure and uses azure cni overlay, that we properly delegate that installation

* fix: ci

* refactor: renaming cni label var

* modifying comment

* style: renaming vars to not have vnet prefix

* test: adding tests for the different cni modes on labels

* refactor: renaming test

* feat: supporting cilium + nodesubnet case


Bryce-Soghigian (dismissed):
Amazing code, great work!

Bryce-Soghigian (approved):

PR #729, opened 2025-03-10
feat(upgrade): adding new NodeClass status controller logic for upgrades of k8s, and node image versions

charliedmcb:
Fixes # 

**Description**
Adding two new AKSNodeClass Status reconcilers:

- K8sVersionReconciler, which is in charge of the K8sVersionStatus, and upgrades as follows:
    - 1. Newly created AKSNodeClass, will select the version discovered from the API server
    - 2. If a later k8s version is discovered from the API server, we will upgrade to it. [don't currently support rollback]
          - Note: We will indirectly trigger an upgrade to latest image version as well, by resetting the Image readiness.

- NodeImageReconciler, which is in charge of detecting node image version upgrades, and shifts in the customers selectors for node images, as follows:
     - Scenario A: Update all image versions to latest
        - 1. Initializes the images versions for a newly created AKSNodeClass, based on customer configuration.
        - 2. Indirectly handle image bump for k8s upgrade
        - 3. Can indirectly handle bumps for any images unsupported by node features, if required to in the future
           - Note: Currently there are no node features to be handled in this way.
       - 4. TODO: Update Images to latest if in a MW (retrieved from ConfigMap)
    - Scenario B: Calculate images to be updated based on delta of available images
        - 5. Handles update cases when customer changes image family, SIG usage, or other means of image selectors
        - 6. Handles softly adding newest image version of any newly supported SKUs by Karpenter

Note: this logic is only in charge of setting the status on the AKSNodeClasses at the moment and impacts readiness, but doesn't have any integration into logic beyond that. I will have a follow up PR for actually leveraging this status information during provisioning. 
- Follow up PRs still in progress for actual usage of status values within provisioning and drift:
    - https://github.com/Azure/karpenter-provider-azure/pull/753
    - https://github.com/Azure/karpenter-provider-azure/pull/804
- There are also some TODOs left in the code for MW support, and further design around image selectors and releasing newly supported SKUs for Karpenter.

Codespace testing:
1. Created cluster, and deployed two AKSNodeClasses, of AzureLinux, and Ubuntu2204
2. Verified the Readiness and status was updated correctly
3. Preformed a k8s upgrade
4. Verified the k8s status was updated correctly, and all readiness was good.

E2E tests:
- https://github.com/Azure/karpenter-provider-azure/actions/runs/14048182949
- https://github.com/Azure/karpenter-provider-azure/actions/runs/14067089021
- https://github.com/Azure/karpenter-provider-azure/actions/runs/14502896421
- https://github.com/Azure/karpenter-provider-azure/actions/runs/14526425281
    - Failed at karpenter install time. Appears to be a bug within the E2E pipeline surrounding the small git hash containing only numbers and being interpreted as an int. 
- https://github.com/Azure/karpenter-provider-azure/actions/runs/14539923538 (last functional change)
- https://github.com/Azure/karpenter-provider-azure/actions/runs/14541656859 (latest)

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
New status fields added to the AKSNodeClass:
- KubernetesVersion, which contains the current kubernetes version to be used for nodes provisioned for the NodeClass
    - Populated, but not actually leveraged within provisioning yet.
- Images, which contains the current set of available node images to be used by Karpenter when provisioning nodes for the NodeClass.
    - Populated, but not actually leveraged within provisioning yet
- Two new readiness conditions have been added to the AKSNodeClass for KubernetesVersionReady and ImagesReady
```


pkg/controllers/nodeclass/termination/suite_test.go
```diff
@@ -60,7 +60,7 @@ var _ = BeforeSuite(func() {
 	env = coretest.NewEnvironment(coretest.WithCRDs(apis.CRDs...), coretest.WithCRDs(v1alpha1.CRDs...), coretest.WithFieldIndexers(test.AKSNodeClassFieldIndexer(ctx)))
 	ctx = coreoptions.ToContext(ctx, coretest.Options())
 	ctx = options.ToContext(ctx, test.Options())
-	awsEnv = test.NewEnvironment(ctx, env)
+	azureEnv = test.NewEnvironment(ctx, env)
```
Bryce-Soghigian, line 63: lol

charts/karpenter-crd/templates/karpenter.azure.com_aksnodeclasses.yaml
```diff
@@ -249,6 +249,59 @@ spec:
                   - type
                   type: object
                 type: array
+              k8sVersion:
```
matthchr, line 252: IMO, avoid abbreviations in user-facing names.

`kubernetesVersion`, not `k8s`.

charts/karpenter-crd/templates/karpenter.azure.com_aksnodeclasses.yaml
```diff
@@ -249,6 +249,59 @@ spec:
...
+                description: |-
+                  K8sVersion contains the current k8s version which should be
```
matthchr, line 254: Fix in these docs too.

charts/karpenter-crd/templates/karpenter.azure.com_aksnodeclasses.yaml
```diff
@@ -249,6 +249,59 @@ spec:
...
+              nodeImages:
+                description: |-
+                  NodeImages contains the current set of images available to use
+                  for the NodeClass
+                items:
+                  description: Image contains resolved image selector values utilized
+                    for node launch
+                  properties:
+                    id:
+                      description: ID of the image
```
matthchr, line 266:
```suggestion
                      description: The ID of the image. For example ...
```

pkg/apis/crds/karpenter.azure.com_aksnodeclasses.yaml
```diff
@@ -249,6 +249,59 @@ spec:
                   - type
                   type: object
                 type: array
+              k8sVersion:
```
matthchr, line 252: Same comments as above.

pkg/apis/v1alpha2/aksnodeclass_status.go
```diff
@@ -21,8 +21,13 @@ import (
 	corev1 "k8s.io/api/core/v1"
 )
 
+const (
+	ConditionTypeNodeImageReady  = "NodeImageReady"
+	ConditionTypeK8sVersionReady = "K8sVersionReady"
```
matthchr, line 26: Avoid `K8sVersionReady` here too IMO.
```suggestion
	ConditionTypeKubernetesVersionReady = "KubernetesVersionReady"
```

pkg/controllers/controllers.go
```diff
@@ -35,14 +35,15 @@ import (
 	nodeclasstermination "github.com/Azure/karpenter-provider-azure/pkg/controllers/nodeclass/termination"
 
 	"github.com/Azure/karpenter-provider-azure/pkg/controllers/nodeclaim/inplaceupdate"
+	"github.com/Azure/karpenter-provider-azure/pkg/providers/imagefamily"
 	"github.com/Azure/karpenter-provider-azure/pkg/providers/instance"
 )
 
 func NewControllers(ctx context.Context, mgr manager.Manager, kubeClient client.Client, recorder events.Recorder,
-	cloudProvider cloudprovider.CloudProvider, instanceProvider instance.Provider) []controller.Controller {
+	cloudProvider cloudprovider.CloudProvider, instanceProvider instance.Provider, imageProvider *imagefamily.Provider) []controller.Controller {
```
matthchr, line 43: minor, style, feel free to ignore:

I think go style guide wants either all args 1 line, or all args on their own line. I know you don't add this split in this PR but you could fix it =)

pkg/controllers/nodeclass/status/k8sversion.go
```diff
@@ -0,0 +1,102 @@
...
+	"github.com/blang/semver/v4"
+
+	controllerruntime "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/karpenter/pkg/utils/pretty"
+)
+
+type K8sVersionReconciler struct {
```
matthchr, line 38: Does this need to be exported?

pkg/controllers/controllers.go
matthchr, line 43: `imagefamily.Provider` should be an interface. Generally speaking accept interfaces, return structs (that's the Go way).

This function is currently follow that rule, IMO you should maintain that.

pkg/controllers/nodeclass/status/controller.go
```diff
@@ -32,24 +32,29 @@ import (
...
-type nodeClassStatusReconciler interface {
+type Reconciler interface {
 	Reconcile(context.Context, *v1alpha2.AKSNodeClass) (reconcile.Result, error)
 }
 
 type Controller struct {
 	kubeClient client.Client
 
-	readiness *Readiness //TODO : Remove this when we have sub status conditions
+	k8sVersion *K8sVersionReconciler
```
matthchr, line 46: minor, style, feel free to ignore:

Even here I'd drop the `k8s` in favor of `Kubernetes`

pkg/controllers/nodeclass/status/controller.go
```diff
@@ -32,24 +32,29 @@ import (
 	"sigs.k8s.io/karpenter/pkg/utils/result"
 
 	"github.com/Azure/karpenter-provider-azure/pkg/apis/v1alpha2"
+	"github.com/Azure/karpenter-provider-azure/pkg/providers/imagefamily"
 	"github.com/awslabs/operatorpkg/reasonable"
 )
 
-type nodeClassStatusReconciler interface {
+type Reconciler interface {
```
matthchr, line 39: Does this need to be exported?

pkg/controllers/nodeclass/status/k8sversion.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+}
+
+func (r *K8sVersionReconciler) Register(_ context.Context, m manager.Manager) error {
+	return controllerruntime.NewControllerManagedBy(m).
+		Named("nodeclass.k8sversion").
+		For(&v1alpha2.AKSNodeClass{}).
+		WithOptions(controller.Options{
+			RateLimiter:             reasonable.RateLimiter(),
+			MaxConcurrentReconciles: 10,
```
matthchr, line 56: Looks like we do this elsewhere too but, feels like this should be either a const or configurable (per-controller? Or a single global config?)

pkg/controllers/nodeclass/status/k8sversion.go
```diff
@@ -0,0 +1,102 @@
...
+		}).
+		Complete(reconcile.AsReconciler(m.GetClient(), r))
+}
+
+// The k8s version reconciler will detect reasons to bump the k8s version:
+//  1. Newly created AKSNodeClass, will select the version discovered from the API server
+//  2. If a later k8s version is discovered from the API server, we will upgrade to it. [don't currently support rollback]
+//     - Note: We will indirectly trigger an upgrade to latest image version as well, by resetting the NodeImage readiness.
+func (r *K8sVersionReconciler) Reconcile(ctx context.Context, nodeClass *v1alpha2.AKSNodeClass) (reconcile.Result, error) {
+	logger := logging.FromContext(ctx)
```
matthchr, line 66: You can (and probalby should?) just use `logger.Named()` here?

It should save you needing to manually repeat the nodeclass.k8sversion prefix everywhere

pkg/controllers/nodeclass/status/k8sversion.go
```diff
@@ -0,0 +1,102 @@
...
+func NewK8sVersionReconciler(provider imagefamily.K8sVersionProvider) *K8sVersionReconciler {
+	return &K8sVersionReconciler{
+		k8sVersionProvider: provider,
+		cm:                 pretty.NewChangeMonitor(),
+	}
+}
+
+func (r *K8sVersionReconciler) Register(_ context.Context, m manager.Manager) error {
+	return controllerruntime.NewControllerManagedBy(m).
+		Named("nodeclass.k8sversion").
```
matthchr, line 52: Define this as a const so you can use it in the logger name too?

pkg/controllers/nodeclass/status/k8sversion.go
```diff
@@ -0,0 +1,102 @@
...
+//  1. Newly created AKSNodeClass, will select the version discovered from the API server
+//  2. If a later k8s version is discovered from the API server, we will upgrade to it. [don't currently support rollback]
+//     - Note: We will indirectly trigger an upgrade to latest image version as well, by resetting the NodeImage readiness.
+func (r *K8sVersionReconciler) Reconcile(ctx context.Context, nodeClass *v1alpha2.AKSNodeClass) (reconcile.Result, error) {
+	logger := logging.FromContext(ctx)
+	logger.Debug("nodeclass.k8sversion: starting reconcile")
+
+	k8sVersion, err := r.k8sVersionProvider.KubeServerVersion(ctx)
+	if err != nil {
+		logger.Debug("nodeclass.k8sversion: err getting k8s version")
```
matthchr, line 71: IMO logging and returning an error is an anti-pattern.

With that said, if we don't have controller-runtime logs on by default, it won't log errors returned from controllers so I see why we might do this?

There are many other places I see where we are not logging + returning error though... so possibly you don't need to here either?

pkg/controllers/nodeclass/status/k8sversion.go
```diff
@@ -0,0 +1,102 @@
...
+		Complete(reconcile.AsReconciler(m.GetClient(), r))
+}
+
+// The k8s version reconciler will detect reasons to bump the k8s version:
+//  1. Newly created AKSNodeClass, will select the version discovered from the API server
+//  2. If a later k8s version is discovered from the API server, we will upgrade to it. [don't currently support rollback]
+//     - Note: We will indirectly trigger an upgrade to latest image version as well, by resetting the NodeImage readiness.
+func (r *K8sVersionReconciler) Reconcile(ctx context.Context, nodeClass *v1alpha2.AKSNodeClass) (reconcile.Result, error) {
+	logger := logging.FromContext(ctx)
+	logger.Debug("nodeclass.k8sversion: starting reconcile")
```
matthchr, line 67: IMO this sort of log is (probably) overly verbose and doesn't add a lot of value.

pkg/controllers/nodeclass/status/k8sversion.go
matthchr: This comment applies below too.

pkg/controllers/nodeclass/status/nodeimage.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+}
+
+func (r *NodeImageReconciler) Register(_ context.Context, m manager.Manager) error {
+	return controllerruntime.NewControllerManagedBy(m).
+		Named("nodeclass.nodeimage").
+		For(&v1alpha2.AKSNodeClass{}).
+		WithOptions(controller.Options{
+			RateLimiter:             reasonable.RateLimiter(),
+			MaxConcurrentReconciles: 10,
```
matthchr, line 60: Same comment as other, though not necessarily expecting you to solve it here.

pkg/controllers/nodeclass/status/nodeimage.go
```diff
@@ -0,0 +1,179 @@
...
+func NewNodeImageReconciler(provider imagefamily.NodeImageProvider) *NodeImageReconciler {
+	return &NodeImageReconciler{
+		nodeImageProvider: provider,
+		cm:                pretty.NewChangeMonitor(),
+	}
+}
+
+func (r *NodeImageReconciler) Register(_ context.Context, m manager.Manager) error {
+	return controllerruntime.NewControllerManagedBy(m).
+		Named("nodeclass.nodeimage").
```
matthchr, line 56: const + reuse w/ named logger (same as other controller)

pkg/providers/imagefamily/nodeimage.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/Azure/karpenter-provider-azure/pkg/operator/options"
+	"sigs.k8s.io/karpenter/pkg/scheduling"
+)
+
+type NodeImage struct {
+	ID           string
+	Requirements scheduling.Requirements
+}
+
+type NodeImages []NodeImage
```
matthchr, line 33: minor, style: Not totally sold that defining an alias for `[]NodeImage` is all that helpful here?

Why not just use `[]NodeImage` directly?

pkg/controllers/nodeclass/status/nodeimage.go
```diff
@@ -0,0 +1,179 @@
...
+// Scenario B: Calculate images to be updated based on delta of available images
+//   - 5. Handles update cases when customer changes image family, SIG usage, or other means of image selectors
+//   - 6. Handles softly adding newest image version of any newly supported SKUs by Karpenter
+func (r *NodeImageReconciler) Reconcile(ctx context.Context, nodeClass *v1alpha2.AKSNodeClass) (reconcile.Result, error) {
+	logger := logging.FromContext(ctx)
+	logger.Debug("nodeclass.nodeimage: starting reconcile")
+
+	nodeImages, err := r.nodeImageProvider.List(ctx, nodeClass)
+	if err != nil {
+		logger.Debug("nodeclass.nodeimage: err listing node images")
```
matthchr, line 83: Same comment as other file

pkg/controllers/nodeclass/status/nodeimage.go
```diff
@@ -0,0 +1,179 @@
...
+		return reconcile.Result{}, fmt.Errorf("getting nodeimages, %w", err)
+	}
+	images := lo.Map(nodeImages, func(nodeImage imagefamily.NodeImage, _ int) v1alpha2.NodeImage {
+		reqs := lo.Map(nodeImage.Requirements.NodeSelectorRequirements(), func(item v1.NodeSelectorRequirementWithMinValues, _ int) corev1.NodeSelectorRequirement {
+			return item.NodeSelectorRequirement
+		})
+
+		sort.Slice(reqs, func(i, j int) bool {
```
matthchr, line 91: minor: Consider leaving a comment here explaining why things must be sorted.

matthchr: note to self, stopped here

pkg/controllers/nodeclass/status/nodeimage.go
matthchr: Also you don't even log the error here so this is doubly not useful as it is now?

pkg/controllers/nodeclass/status/nodeimage.go
```diff
@@ -0,0 +1,179 @@
...
+			ID:           nodeImage.ID,
+			Requirements: reqs,
+		}
+	})
+
+	// Scenario A: Check if we should do a full update to latest before processing any partial update
+	shouldUpdate := imageVersionsUnready(nodeClass) || isOpenMW()
+	logger.Debugf("nodeclass.nodeimage: should complete a full update to latest: %t", shouldUpdate)
+	if !shouldUpdate {
+		// Scenario B: Check if we should do any partial update based on image selectors, or newly supports SKUs
```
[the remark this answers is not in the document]
matthchr: Oh a partial update is just determining if things are different from last time, due to user changes for example?

I _think_ you don't need to do any of this logic? The actual write-back of `NodeClass` is controlled here and has an equality guard:
```
	if !equality.Semantic.DeepEqual(stored, nodeClass) {
		if err := c.kubeClient.Status().Patch(ctx, nodeClass, client.MergeFrom(stored)); err != nil {
			errs = multierr.Append(errs, client.IgnoreNotFound(err))
		}
	}
```

In theory at least, you don't need to do fine-grained diffing here, you could just recompute the whole state every time (assuming it's not more perf-expensive to do that, which I don't think it is in this case?), blindly write it to status, and if it's the same no write will happen?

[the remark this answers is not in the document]
matthchr: Calling it partial is sorta weird though because you do rewrite the whole set of images based on what was returned from the API, right? For example if there's a new version for a given base image, and MW is closed... don't we still take the `processPartialUpdate` path and update the image version?

Isn't that wrong?

It seems like you may need to differentiate between the "type" of update?
- if MW is closed and there's an existing base image w/ a new image version, you _dont_ want to take the new image version, I think?
- If the MW closed and there's a net-new base image, you either want to take the new base image or you want to somehow use the old one (using the old one would require us to store that even though at the time the image selector didn't select for it, which increases the size of images in the status, but might be desirable...)
- If MW is open, you take both kinds of updates above.

pkg/controllers/nodeclass/status/nodeimage.go
```diff
@@ -0,0 +1,179 @@
...
+		images, shouldUpdate = processPartialUpdate(nodeClass, images)
+	}
+
+	logger.Debugf("nodeclass.nodeimage: should update overall: %t", shouldUpdate)
+	if shouldUpdate {
+		if len(images) == 0 {
+			nodeClass.Status.NodeImages = nil
+			nodeClass.StatusConditions().SetFalse(v1alpha2.ConditionTypeNodeImageReady, "NodeImagesNotFound", "NodeImageSelectors did not match any NodeImages")
+			logger.Info("nodeclass.nodeimage: no images")
+			return reconcile.Result{RequeueAfter: time.Minute}, nil
```
matthchr, line 117: minor: Why a different interval than below?

pkg/providers/imagefamily/nodeimage.go
```diff
@@ -0,0 +1,84 @@
...
+
+type NodeImageProvider interface {
+	List(ctx context.Context, nodeClass *v1alpha2.AKSNodeClass) (NodeImages, error)
+}
+
+func (p *Provider) List(ctx context.Context, nodeClass *v1alpha2.AKSNodeClass) (NodeImages, error) {
+	supportedImages := getSupportedImages(nodeClass.Spec.ImageFamily)
+	if options.FromContext(ctx).UseSIG {
+		nodeImages := NodeImages{}
+		retrievedLatestImages, err := p.NodeImageVersions.List(ctx, p.location, p.subscription)
```
matthchr, line 43: Do we intend `NodeImageVersions` to be exported here?

I see most provider fields are not.

pkg/controllers/nodeclass/status/nodeimage.go
```diff
@@ -0,0 +1,179 @@
...
+//   - 4. TODO: Update NodeImages to latest if in a MW (maintenance windows) [retrieved from ConfigMap]
+//
+// Scenario B: Calculate images to be updated based on delta of available images
+//   - 5. Handles update cases when customer changes image family, SIG usage, or other means of image selectors
+//   - 6. Handles softly adding newest image version of any newly supported SKUs by Karpenter
+func (r *NodeImageReconciler) Reconcile(ctx context.Context, nodeClass *v1alpha2.AKSNodeClass) (reconcile.Result, error) {
+	logger := logging.FromContext(ctx)
+	logger.Debug("nodeclass.nodeimage: starting reconcile")
+
+	nodeImages, err := r.nodeImageProvider.List(ctx, nodeClass)
```
matthchr, line 81: This call looks uncached to me - is that what we want?

It means if we have N `AKSNodeClass`'s, each one will produce a call to this API at least every 5m. I wonder if at the provider level, the results of this call should be cached?

I see we have a `cache.Cache` we could configure.

pkg/controllers/nodeclass/status/nodeimage_test.go
```diff
@@ -0,0 +1,126 @@
...
+		ExpectApplied(ctx, env.Client, nodeClass)
+		ExpectObjectReconciled(ctx, env.Client, controller, nodeClass)
+		nodeClass = ExpectExists(ctx, env.Client, nodeClass)
+
+		Expect(len(nodeClass.Status.NodeImages)).To(Equal(3))
+		Expect(nodeClass.Status.NodeImages).To(ContainElements(getExpectedTestCommunityImages(newCIGImageVersion)))
+		Expect(nodeClass.StatusConditions().IsTrue(v1alpha2.ConditionTypeNodeImageReady)).To(BeTrue())
+	})
+
+	It("should update NodeImages and its readiness on AKSNodeClass when in an open MW", func() {
```
matthchr, line 106:
```suggestion
	It("should update NodeImages and its readiness on AKSNodeClass when in an open maintenance window", func() {
```

pkg/controllers/nodeclass/status/suite_test.go
```diff
@@ -0,0 +1,75 @@
...
+
+package status_test
+
+import (
+	"context"
+	"testing"
+
+	"sigs.k8s.io/karpenter/pkg/test/v1alpha1"
+
+	//used for launch template tests until they are migrated
```
matthchr, line 25: Does this apply here?

pkg/providers/imagefamily/k8sversion.go
```diff
@@ -0,0 +1,25 @@
...
+limitations under the License.
+*/
+
+package imagefamily
+
+import (
+	"context"
+)
+
+type K8sVersionProvider interface {
```
matthchr, line 23: minor: Same comment, `Kubernetes` rather than k8s IMO, and same with the function (`Kubernetes` rather than `Kube`)

pkg/providers/imagefamily/nodeimage.go
```diff
@@ -0,0 +1,84 @@
...
+		if err != nil {
+			return nil, err
+		}
+
+		for _, nextSupportedImage := range supportedImages {
```
matthchr, line 48: minor, style: 
```suggestion
		for _, supportedImage := range supportedImages {
```

Calling it `nextSupportedImage` is sorta weird?
Applies below too.

pkg/providers/imagefamily/nodeimage.go
```diff
@@ -0,0 +1,84 @@
...
+
+type NodeImages []NodeImage
+
+type NodeImageProvider interface {
+	List(ctx context.Context, nodeClass *v1alpha2.AKSNodeClass) (NodeImages, error)
+}
+
+func (p *Provider) List(ctx context.Context, nodeClass *v1alpha2.AKSNodeClass) (NodeImages, error) {
+	supportedImages := getSupportedImages(nodeClass.Spec.ImageFamily)
+	if options.FromContext(ctx).UseSIG {
```
matthchr, line 41: Possibly a question for @tallaxes -- is this the preferred way to pass "parameters"?

I'm asking because I am not a huge fan - I would rather see `List(ctx context.Context, nodeClass *v1alpha2.AKSNodeClass, options *SomeOptions)` (or some similar pattern) which at least makes it clear that this is an input to this function.

matthchr, line 41: minor: consider 2 internal helper functions, `listBySIG` and `listByCIG`?

charts/karpenter-crd/templates/karpenter.azure.com_aksnodeclasses.yaml
charliedmcb: Yea, I was wondering about the naming here. Makes sense to go with full name, rather than k8s. Will update.

charts/karpenter-crd/templates/karpenter.azure.com_aksnodeclasses.yaml
charliedmcb: You just meaning the description? Not contained in any docs atm.

charliedmcb: Updated description.

pkg/apis/v1alpha2/aksnodeclass_status.go
charliedmcb: Updated to `Kubernetes` as well

pkg/controllers/controllers.go
charliedmcb: Yea, this is mostly an artifct of waiting on a folllow up PR where I was planning on doing some of this refactoring, but this change can go in here. Just need to break this input param into the two seperate interfaces for k8s, and image.

charliedmcb: Updated

pkg/controllers/nodeclass/status/controller.go
charliedmcb: Agreed, since I updated the Reconciler name.

pkg/controllers/nodeclass/status/k8sversion.go
charliedmcb: I've found that Karpenter is under-logging imo. I was trying to introduce more in this PR for the newly written code, at least at the debug level. Might have overdone it though.

pkg/controllers/nodeclass/status/nodeimage.go
```diff
@@ -0,0 +1,179 @@
...
+}
+
+// Case 1: This is a new AKSNodeClass, where node images haven't been populated yet
+// Case 2: This is indirectly handling k8s version image bump, since k8s version sets this status to false
+// Case 3: Note: like k8s we would also indirectly handle node features that required an image version bump, but none required atm.
+func imageVersionsUnready(nodeClass *v1alpha2.AKSNodeClass) bool {
+	return !nodeClass.StatusConditions().Get(v1alpha2.ConditionTypeNodeImageReady).IsTrue()
+}
+
+func isOpenMW() bool {
```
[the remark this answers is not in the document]
charliedmcb: Yea, agreed like k8s would be good to spell out.

pkg/controllers/nodeclass/status/nodeimage.go
```diff
@@ -0,0 +1,179 @@
...
+	imagesBaseMapping := map[string]*v1alpha2.NodeImage{}
+	for _, image := range images {
+		baseID := TrimVersionSuffix(image.ID)
+		imagesBaseMapping[baseID] = &image
+	}
+	return imagesBaseMapping
+}
+
+// Trims off the version suffix, and leaves just the image base id
+func TrimVersionSuffix(imageID string) string {
```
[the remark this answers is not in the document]
charliedmcb: Was used in testing before, but no need to be exported anymore. Made private.

pkg/providers/imagefamily/k8sversion.go
charliedmcb: I updated the provider, but the func is a remnant from before with other existing code deps. I can address the func name in the follow up PR that will be doing some refactoring for the usage of these statuses.

pkg/controllers/nodeclass/status/suite_test.go
charliedmcb: I don't think so. Removed.

pkg/providers/imagefamily/nodeimage.go
charliedmcb: I agree that I'd like that pattern more. This was simply following the pattern I saw for retrieving UseSIG from elsewhere.

This actually did cause a panic during my testing, since I missed this input wasn't set for one of my tests.

pkg/controllers/nodeclass/status/k8sversion.go
charliedmcb: Its used in tests, which are black-box with a separate package, but could use the `Reconciler` interface instead. Though you have a question if that needs to be exported too :)

charliedmcb: Could also make that one test clear-box instead.

pkg/controllers/nodeclass/status/controller.go
charliedmcb: Nope. Made private.

pkg/providers/imagefamily/nodeimage.go
charliedmcb: Changed in both locations

pkg/providers/imagefamily/nodeimage.go
charliedmcb: Done

pkg/providers/imagefamily/nodeimage.go
charliedmcb: I don't see it needing to be, but wasn't changed in this PR. I'm not going to touch it right now, but will be doing some refactoring in this area when it comes to using the status during provisioning, so can look at addressment then.

pkg/providers/imagefamily/nodeimage.go
tallaxes: Yes, we would strongly prefer to extract these parameters from context at the earliest reasonable opportunity. Lower-level functions ideally should not depend on where the setting is coming from.

charliedmcb (commented):
Pushing review comment

charts/karpenter-crd/templates/karpenter.azure.com_aksnodeclasses.yaml
charliedmcb: Added examples of CIG, and SIG.

pkg/controllers/nodeclass/status/k8sversion.go
charliedmcb: Ginkgo started complaining about putting both clearbox, and blackbox tests in the same area while using `BeforeSuite`. I'm not opposed to breaking apart our testing in this way. However, I think the added complexity to just make this struct private is not worth it here, so leaving this for the moment as is.

pkg/controllers/nodeclass/status/k8sversion.go
charliedmcb: Adding a comment to determine why this magic number, and maybe consolidate all into one const, if we have good reason for using this value. @tallaxes do you have more context on choosing this number?

pkg/controllers/nodeclass/status/nodeimage.go
charliedmcb: Added same TODO tomment to all status Reconcilers.

pkg/controllers/nodeclass/status/k8sversion.go
charliedmcb: Added a TODO comment to all usages of `MaxConcurrentReconciles: 10`.

pkg/controllers/nodeclass/status/k8sversion.go
charliedmcb: Good callout. Updated, plus image version logger.

pkg/controllers/nodeclass/status/k8sversion.go
charliedmcb: Updated to const for both reconcilers.

pkg/controllers/nodeclass/status/nodeimage.go
```diff
@@ -0,0 +1,179 @@
...
+func mapImageBasesToImages(images []v1alpha2.NodeImage) map[string]*v1alpha2.NodeImage {
+	imagesBaseMapping := map[string]*v1alpha2.NodeImage{}
+	for _, image := range images {
+		baseID := TrimVersionSuffix(image.ID)
+		imagesBaseMapping[baseID] = &image
+	}
+	return imagesBaseMapping
+}
+
+// Trims off the version suffix, and leaves just the image base id
```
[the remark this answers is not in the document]
charliedmcb: Added for CIG and SIG

pkg/controllers/nodeclass/status/kubernetesversion_test.go
```diff
@@ -0,0 +1,92 @@
...
+		Expect(nodeClass.StatusConditions().IsTrue(v1alpha2.ConditionTypeKubernetesVersionReady)).To(BeTrue())
+	})
+
+	Context("KubernetesVersionReconciler direct tests", func() {
+		var (
+			k8sReconciler *status.KubernetesVersionReconciler
+		)
+
+		BeforeEach(func() {
+			k8sReconciler = status.NeKubernetesVersionReconciler(azureEnv.ImageProvider)
```
charliedmcb, line 76: typo: `NeKubernetesVersionReconciler`

pkg/controllers/nodeclass/status/k8sversion.go
charliedmcb: Add tech debt issue for tracking?

pkg/providers/imagefamily/nodeimage.go
charliedmcb: Will update

pkg/controllers/nodeclass/status/nodeimage.go
```diff
@@ -0,0 +1,179 @@
...
+	return !nodeClass.StatusConditions().Get(v1alpha2.ConditionTypeNodeImageReady).IsTrue()
+}
+
+func isOpenMW() bool {
+	// Case 4: check if MW is open
+	// TODO: need to add the actual logic for handling MWs once it is in the ConfigMap.
+	return true
+}
+
+// Case 5: Handles the case of users updating image selectors.
```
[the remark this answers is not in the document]
charliedmcb: Add `Handles case`

pkg/controllers/nodeclass/status/nodeimage.go
```diff
@@ -0,0 +1,193 @@
...
+//   - Currently, this is just image family, and/or usage of SIG, which means that we should just be looking at the baseID of the images
+//
+// Case 6: We will softly add newly supported SKUs by Karpenter on their latest version
+//   - Note: I think this should be assess, if this is the exact behavior we want to give users, before any actual new SKU support is released.
+//
+// TODO: Need longer term design for handling newly supported versions, and other image selectors.
+func processPartialUpdate(nodeClass *v1alpha2.AKSNodeClass, discoveredImages []v1alpha2.NodeImage) ([]v1alpha2.NodeImage, bool) {
```
matthchr, line 153: defining what a partial update is at the top of this function comment, would I think be helpful

pkg/controllers/nodeclass/status/nodeimage.go
```diff
@@ -0,0 +1,193 @@
...
+			Requirements: reqs,
+		}
+	})
+
+	// Scenario A: Check if we should do a full update to latest before processing any partial update
+	shouldUpdate := imageVersionsUnready(nodeClass) || isMaintenanceWindowOpen()
+	logger.Debugf("should complete a full update to latest: %t", shouldUpdate)
+	if !shouldUpdate {
+		// Scenario B: Check if we should do any partial update based on image selectors, or newly supports SKUs
+		images, shouldUpdate = processPartialUpdate(nodeClass, images)
```
charliedmcb, line 113: rename to `goalImages`, and rename func to detecting goal images in some way, while dropping `shouldUpdate`.

pkg/controllers/nodeclass/status/nodeimage.go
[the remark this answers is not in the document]
charliedmcb: WI-1: rename to goalImages, and rename func to detecting goal images in some way, while dropping shouldUpdate.

pkg/controllers/nodeclass/status/nodeimage.go
```diff
@@ -0,0 +1,193 @@
...
+			return reqs[i].Key < reqs[j].Key
+		})
+		return v1alpha2.NodeImage{
+			ID:           nodeImage.ID,
+			Requirements: reqs,
+		}
+	})
+
+	// Scenario A: Check if we should do a full update to latest before processing any partial update
+	shouldUpdate := imageVersionsUnready(nodeClass) || isMaintenanceWindowOpen()
```
charliedmcb, line 109: Add note on wanting to do cases 1-3 in MW or not.

pkg/controllers/nodeclass/status/nodeimage.go
charliedmcb: Will update to match

pkg/providers/imagefamily/resolver.go
```diff
@@ -148,6 +148,11 @@ func prepareKubeletConfiguration(ctx context.Context, instanceType *cloudprovide
 	return kubeletConfig
 }
 
+func getSupportedImages(familyName *string) []DefaultImageOutput {
+	imageFamily := getImageFamily(familyName, nil)
```
charliedmcb, line 152: Drop TODO comment on nil because not being used.

pkg/controllers/nodeclass/status/k8sversion.go
charliedmcb: Removed err logging, and cleaned/updated logging in general.

pkg/controllers/nodeclass/status/k8sversion.go
charliedmcb: Found this helpful when debugging things earlier, so leaving in at Debug level, since base Karp is at Info.

pkg/providers/imagefamily/nodeimage.go
charliedmcb: Updated

pkg/controllers/nodeclass/status/nodeimage.go
charliedmcb: Removed, and cleaned up logging

pkg/controllers/nodeclass/status/nodeimage.go
```diff
@@ -0,0 +1,179 @@
...
+		}
+	}
+	return updatedImages, foundUpdate
+}
+
+func mapImageBasesToImages(images []v1alpha2.NodeImage) map[string]*v1alpha2.NodeImage {
+	imagesBaseMapping := map[string]*v1alpha2.NodeImage{}
+	for _, image := range images {
+		baseID := TrimVersionSuffix(image.ID)
+		imagesBaseMapping[baseID] = &image
```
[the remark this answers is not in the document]
charliedmcb: Added minor comment on sorting for consistency.

pkg/controllers/nodeclass/status/nodeimage.go
charliedmcb: Added minor comment on sorting for consistency.

pkg/controllers/nodeclass/status/nodeimage.go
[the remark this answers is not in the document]
charliedmcb: Always latest. Might have to reassess this logic if/once we start supporting a mix of version selectors.

pkg/controllers/nodeclass/status/nodeimage.go
charliedmcb: Renamed method, and left a comment explaining it's logic

pkg/controllers/nodeclass/status/nodeimage.go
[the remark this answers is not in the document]
charliedmcb: Added sorting, but wondering if we actually want these sorted by priority instead of alphabetical order. I think this will somewhat be determined on how the next PR shakes out, so will circle back to this then.

pkg/controllers/nodeclass/status/nodeimage.go
charliedmcb: Added caching to my new node image provider as a whole.

charliedmcb: Added caching for my new node image provider as a whole

pkg/providers/imagefamily/nodeimage.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+	return nodeImages, nil
+}
+
+func (p *Provider) listCIG(_ context.Context, supportedImages []DefaultImageOutput) ([]NodeImage, error) {
+	nodeImages := []NodeImage{}
+	for _, supportedImage := range supportedImages {
+		cigImageID, err := p.getCIGImageID(supportedImage.PublicGalleryURL, supportedImage.ImageDefinition)
+		if err != nil {
+			return nil, err
```
tallaxes, line 111: Can we consider continuing on error? This returns from within `for` loop. If one image has a problem, others might still work ...

pkg/providers/imagefamily/image.go
```diff
@@ -39,6 +39,7 @@ type Provider struct {
 	location               string
 	kubernetesInterface    kubernetes.Interface
 	imageCache             *cache.Cache
+	nodeImagesCache        *cache.Cache
 	imageVersionsClient    CommunityGalleryImageVersionsAPI
 	subscription           string
 	NodeImageVersions      NodeImageVersionsAPI
```
tallaxes, line 45: Likely does not need to be exported?

tallaxes:
Unit tests are failing, possibly a race condition?

pkg/providers/imagefamily/image.go
charliedmcb: Agreed, but not changing in this PR

pkg/providers/imagefamily/nodeimage.go
charliedmcb: We could possibly. However, then we'd have to have some way to still determine if the change in returned images was from a failure, or selectors changing. Maybe we could only use the partially found case, ignoring errors if the NodeImages were empty on the NodeClass. Interesting idea. However, I'm going to leave this as a hard failure for this PR.

charliedmcb: I also was hoping to refactor this to get a List of the images from one query. However, it looks like how things are setup for the CIGs, that doesn't work within the API

pkg/providers/imagefamily/nodeimage.go
```diff
@@ -0,0 +1,132 @@
...
+		})
+	}
+	return nodeImages, nil
+}
+
+func (p *Provider) cacheKey(supportedImages []DefaultImageOutput, k8sVersion string, useSIG bool) (string, error) {
+	hash, err := hashstructure.Hash([]interface{}{
+		supportedImages,
+		k8sVersion,
+		useSIG,
```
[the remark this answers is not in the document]
charliedmcb: Reasonable to drop since it shouldn't be changing over the lifetime of the process. However, since it is a reason images can update, I think I still slightly prefer having it in the key. Leaving as is.

pkg/providers/imagefamily/nodeimage.go
```diff
@@ -0,0 +1,132 @@
...
+			Requirements: supportedImage.Requirements,
+		})
+	}
+	return nodeImages, nil
+}
+
+func (p *Provider) cacheKey(supportedImages []DefaultImageOutput, k8sVersion string, useSIG bool) (string, error) {
+	hash, err := hashstructure.Hash([]interface{}{
+		supportedImages,
+		k8sVersion,
```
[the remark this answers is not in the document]
charliedmcb: Because we bump images on k8s upgrade, we want to ensure if there is a k8s change we'll get fresh images if there are any.

pkg/controllers/nodeclass/status/kubernetesversion.go
```diff
@@ -0,0 +1,104 @@
...
+		newK8sVersion, err := semver.Parse(k8sVersion)
+		if err != nil {
+			return reconcile.Result{}, fmt.Errorf("parsing discovered kubernetes version, %w", err)
+		}
+		currentK8sVersion, err := semver.Parse(nodeClass.Status.KubernetesVersion)
+		if err != nil {
+			return reconcile.Result{}, fmt.Errorf("parsing current kubernetes version, %w", err)
+		}
+		// Handles case 2: Upgrade kubernetes version [Note: we set node image to not ready, since we upgrade node image when there is a kubernetes upgrade]
+		if newK8sVersion.GT(currentK8sVersion) {
```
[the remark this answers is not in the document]
charliedmcb: Added log.

pkg/controllers/nodeclass/status/nodeimage_test.go
```diff
@@ -0,0 +1,126 @@
...
+					Key:      "karpenter.azure.com/sku-hyperv-generation",
+					Operator: "In",
+					Values:   []string{"2"},
+				},
+			},
+		},
+	}
+}
+
+var _ = Describe("NodeClass NodeImage Status Controller", func() {
```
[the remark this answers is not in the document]
charliedmcb: Yea, we could add an E2E test that the status is being populated correctly. However, I verified this with a manual test, and have the unit tests covering things too. I think an E2E test is beyond what's necessary for this PR, but once it comes to the follow up PR which changes to provisioning logic to use the status, I think it's worth adding one. Plus, we could do an actual node image upgrade test, just with extreme recommendations against doing anything like it beyond testing, since it would have to muck with the status itself.

pkg/controllers/nodeclass/status/readiness.go
```diff
@@ -29,6 +28,5 @@ type Readiness struct {
 }
 
 func (n Readiness) Reconcile(ctx context.Context, nodeClass *v1alpha2.AKSNodeClass) (reconcile.Result, error) {
-	nodeClass.StatusConditions().SetTrue(status.ConditionReady)
```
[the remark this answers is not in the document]
charliedmcb: Removed.

pkg/apis/v1alpha2/aksnodeclass_status.go
```diff
@@ -21,9 +21,16 @@ import (
 	corev1 "k8s.io/api/core/v1"
 )
 
-// Image contains resolved image selector values utilized for node launch
-type Image struct {
-	// ID of the image
+const (
+	ConditionTypeNodeImageReady         = "NodeImageReady"
```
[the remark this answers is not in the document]
charliedmcb: Yea, good catch. I think it probably should. Updated, plus the var naming to be plural too.

pkg/controllers/nodeclass/status/k8sversion.go
charliedmcb: Sounds good. Leaving as is for this PR.

pkg/controllers/nodeclass/status/kubernetesversion.go
```diff
@@ -0,0 +1,104 @@
...
+	if !nodeClass.StatusConditions().Get(v1alpha2.ConditionTypeKubernetesVersionReady).IsTrue() || nodeClass.Status.KubernetesVersion == "" {
+		logger.Infof("init kubernetes version: %s", k8sVersion)
+		nodeClass.Status.KubernetesVersion = k8sVersion
+		nodeClass.StatusConditions().SetTrue(v1alpha2.ConditionTypeKubernetesVersionReady)
+	} else {
+		// Check if there is an upgrade
+		newK8sVersion, err := semver.Parse(k8sVersion)
+		if err != nil {
+			return reconcile.Result{}, fmt.Errorf("parsing discovered kubernetes version, %w", err)
+		}
```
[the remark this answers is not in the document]
charliedmcb: Discussed offline. Resolving as I think I already fixed this concern and don't see any others it could be pointing towards.

pkg/controllers/nodeclass/status/nodeimage_test.go
```diff
@@ -0,0 +1,126 @@
...
+		nodeClass = test.AKSNodeClass()
+	})
+
+	It("should init NodeImages and its readiness on AKSNodeClass", func() {
+		ExpectApplied(ctx, env.Client, nodeClass)
+		ExpectObjectReconciled(ctx, env.Client, controller, nodeClass)
+		nodeClass = ExpectExists(ctx, env.Client, nodeClass)
+
+		Expect(len(nodeClass.Status.NodeImages)).To(Equal(3))
+		Expect(nodeClass.Status.NodeImages).To(ContainElements(getExpectedTestCommunityImages(newCIGImageVersion)))
```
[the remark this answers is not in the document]
charliedmcb: Done!

pkg/controllers/nodeclass/status/nodeimage.go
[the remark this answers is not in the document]
charliedmcb: Switched back to previous ordering, which was already consistently sorting based on priority, and is handled for both SIG, and CIG.

pkg/providers/imagefamily/nodeimage_test.go
```diff
@@ -0,0 +1,143 @@
...
+		communityImageVersionsAPI.ImageVersions.Append(&armcompute.CommunityGalleryImageVersion{Name: &cigImageVersionTest})
+		nodeImageVersionsAPI := &fake.NodeImageVersionsAPI{}
+		kubernetesVersionCache := cache.New(azurecache.KubernetesVersionTTL, azurecache.DefaultCleanupInterval)
+		nodeImageProvider = imagefamily.NewProvider(env.KubernetesInterface, kubernetesVersionCache, communityImageVersionsAPI, fake.Region, subscription, nodeImageVersionsAPI)
+
+		nodeClass = test.AKSNodeClass()
+	})
+
+	Context("CIG Images", func() {
+		It("Ubuntu2204 successful list case", func() {
```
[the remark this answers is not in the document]
charliedmcb: Updated files naming, with similar styling.

pkg/apis/v1alpha2/aksnodeclass_status.go
```diff
@@ -21,9 +21,16 @@ import (
...
+	ConditionTypeKubernetesVersionReady = "KubernetesVersionReady"
+)
+
+// NodeImage contains resolved image selector values utilized for node launch
```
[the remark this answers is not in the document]
charliedmcb: Added comment to the nodeimage controller. I was debating adding it within this file instead. However, I wouldn't want it in the CRDs, and might not make sense just being in the API file here itself.

pkg/controllers/nodeclass/status/controller.go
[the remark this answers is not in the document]
charliedmcb: Since we don't have the instance profile, I don't think the race is a concern for now, but only did a quick scan/read.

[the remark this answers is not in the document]
charliedmcb: Added `MergeFromWithOptimisticLock` to status controller.

pkg/providers/imagefamily/nodeimage.go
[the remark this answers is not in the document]
charliedmcb: Actually dropping `useSIG` from the cache key. Will have a refactor PR to have this be part of construction instead.

pkg/providers/imagefamily/nodeimage.go
charliedmcb: Added a TODO comment to address this.

I think it makes the most sense to be setting this for the provider at construction time. However, will do this as a refactor to constructor in a follow up PR since it touches some of the existing code paths too.

tallaxes (dismissed):
LGTM. Please update the release notes (and maybe consider re-running E2E - the latest captured is not on the very last commit)

pkg/controllers/nodeclass/status/nodeimage.go
```diff
@@ -0,0 +1,200 @@
...
+
+	// Scenario A: Check if we should do a full update to latest before processing any partial update
+	//
+	// Note: We want to handle cases 1-3 regardless of maintenance window state, since they are either
+	// for initialization, based off an underlying customer operation, or a different update we're
+	// dependant upon which would have already been preformed within its required maintenance Window.
+	shouldUpdate := imageVersionsUnready(nodeClass) || isMaintenanceWindowOpen()
+	if !shouldUpdate {
+		// Scenario B: Calculate any partial update based on image selectors, or newly supports SKUs
+		goalImages = overrideAnyGoalStateVersionsWithExisting(nodeClass, goalImages)
```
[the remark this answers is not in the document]
tallaxes: Looks like I had a pending comment that did not get posted earlier, please see if still relevant, but could be addressed in a follow up PR.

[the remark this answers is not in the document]
charliedmcb: Right now this case can't be hit since maintenance window open is always set to true atm. Once the maintenance window config is supported, than we'll be able to branch out the logic here, and will add test coverage. Have TODOs in code for this to come back to, once supported.

charliedmcb:
> LGTM. Please update the release notes (and maybe consider re-running E2E - the latest captured is not on the very last commit)

The last E2E run should be the last functional change. Only `_test.go`, and unused `isSIG` param getting dropped, but doesn't hurt to have an extra E2E run done.

pkg/providers/imagefamily/image.go
```diff
@@ -33,12 +33,19 @@ import (
 	"sigs.k8s.io/karpenter/pkg/utils/pretty"
 )
 
+// TODO: Remove this provider, by refactoring it into the nodeimage.go provider, and resolver.go as needed
+// This is part of a shift to have the new controllers for k8s and node image version which populate status.
+// At the current point the provisioning logic isn't update itself, just the status being populated.
+// As part of the change to the actual provisioning, this provider will be refafctored as mentioned above.
```
matthchr, line 39:
```suggestion
// As part of the change to the actual provisioning, this provider will be refactored as mentioned above.
```

pkg/providers/imagefamily/nodeimage.go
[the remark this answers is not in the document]
matthchr: Leave a comment explaining that here?

matthchr (approved):

pkg/providers/imagefamily/nodeimage.go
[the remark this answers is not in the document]
charliedmcb: Added a comment before the `hashstructure.Hash` call, explaining this.

tallaxes (dismissed):

tallaxes (dismissed):
Re-reviewed just the latest changes https://github.com/Azure/karpenter-provider-azure/compare/1e40d29eaac44170e27e14cec880e5062f53578e...468842236ccc22001f03c0b361f8744b03bc650b

pkg/apis/v1alpha2/aksnodeclass_status.go
```diff
@@ -21,9 +21,16 @@ import (
...
-	// ID of the image
+const (
+	ConditionTypeImagesReady            = "ImagesReady"
+	ConditionTypeKubernetesVersionReady = "KubernetesVersionReady"
+)
+
+// NodeImage contains resolved image selector values utilized for node launch
+type NodeImage struct {
+	// The ID of the image. Examples:
+	// - CIG /CommunityGalleries/AKSUbuntu-38d80f77-467a-481f-a8d4-09b6d4220bd2/images/2204gen2containerd/versions/2022.10.03
```
tallaxes, line 32:
```suggestion
	// - CIG: /CommunityGalleries/AKSUbuntu-38d80f77-467a-481f-a8d4-09b6d4220bd2/images/2204gen2containerd/versions/2022.10.03
```

tallaxes (approved):

tallaxes (approved):

PR #742, opened 2025-03-21
chore(refactor): add Resolver interface in refactor of imagefamily package

charliedmcb:
Fixes # 

**Description**
Currently working on a refactor of the `imagefamily` package to leverage the new status information on node images and kubernetes version for provisioning. Felt this refactor could be broken out into a clean, simple, and smaller PR.

Nothing additional (just normal check-in tests, and running E2E manually)
- E2E test: https://github.com/Azure/karpenter-provider-azure/actions/runs/13999763078

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
NONE
```

### DIFF
--- a/pkg/providers/imagefamily/resolver.go
+++ b/pkg/providers/imagefamily/resolver.go
@@ -37,11 +37,6 @@ import (
 	"sigs.k8s.io/karpenter/pkg/cloudprovider"
 )
 
-// resolver is able to fill-in dynamic launch template parameters
-type resolver struct {
-	imageProvider *Provider
-}
-
 type Resolver interface {
 	Resolve(
 		ctx context.Context,
@@ -49,6 +44,14 @@ type Resolver interface {
 		nodeClaim *karpv1.NodeClaim,
 		instanceType *cloudprovider.InstanceType,
 		staticParameters *template.StaticParameters) (*template.Parameters, error)
+}
+
+// assert that defaultResolver implements Resolver interface
+var _ Resolver = (*defaultResolver)(nil)
+
+// defaultResolver is able to fill-in dynamic launch template parameters
+type defaultResolver struct {
+	imageProvider *Provider
 }
 
 // ImageFamily can be implemented to override the default logic for generating dynamic launch template parameters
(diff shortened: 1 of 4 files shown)


pkg/providers/imagefamily/resolver.go
```diff
@@ -67,15 +79,15 @@ type ImageFamily interface {
...
+// NewDefaultResolver constructs a new launch template Resolver
+func NewDefaultResolver(_ client.Client, imageProvider *Provider) *defaultResolver {
+	return &defaultResolver{
 		imageProvider: imageProvider,
 	}
 }
 
 // Resolve fills in dynamic launch template parameters
-func (r Resolver) Resolve(ctx context.Context, nodeClass *v1alpha2.AKSNodeClass, nodeClaim *karpv1.NodeClaim, instanceType *cloudprovider.InstanceType,
+func (r *defaultResolver) Resolve(ctx context.Context, nodeClass *v1alpha2.AKSNodeClass, nodeClaim *karpv1.NodeClaim, instanceType *cloudprovider.InstanceType,
```
charliedmcb, line 90: Little surprised this wasn't a pointer before.

pkg/providers/imagefamily/resolver.go
```diff
@@ -37,8 +37,20 @@ import (
...
+	Resolve(
+		ctx context.Context,
+		nodeClass *v1alpha2.AKSNodeClass,
+		nodeClaim *karpv1.NodeClaim,
+		instanceType *cloudprovider.InstanceType,
+		staticParameters *template.StaticParameters) (*template.Parameters, error)
+}
+
+// assert that defaultResolver implements Resolver interface
+var _ Resolver = (*defaultResolver)(nil)
```
matthchr, line 50: minor, style, feel free to ignore: I find the cleaner way to write this is:
```
var _ Resolver = &defaultResolver{}
```

Note that this _looks_ like it has an allocation, but it actually doesn't - the compiler is smart enough to realize that the `_` variable here is unused and so it typechecks this expression at compile-time but it doesn't actually allocate anything.

pkg/providers/imagefamily/resolver.go
```diff
@@ -67,15 +79,15 @@ type ImageFamily interface {
 	DefaultImages() []DefaultImageOutput
 }
 
-// New constructs a new launch template Resolver
-func New(_ client.Client, imageProvider *Provider) *Resolver {
-	return &Resolver{
+// NewDefaultResolver constructs a new launch template Resolver
+func NewDefaultResolver(_ client.Client, imageProvider *Provider) *defaultResolver {
```
matthchr, line 83: minor: Why `NewDefaultResolver` versus just `New`?

pkg/providers/imagefamily/resolver.go
charliedmcb: I was copying the style I found in a different location of the codebase. It does seem like a weird split where we are using the style you mentioned in the testing packages, but this style elsewhere. I do like the style you mentioned more, and since we are using it in the testing packages, I think I'll switch it.

pkg/providers/imagefamily/resolver.go
charliedmcb: I personally don't like using `New` for searchability, plus in a shared package like this that has the provider and resolver I think New could be extra misleading. Could do just `NewResolver`, but I think since its returning the struct even if private having it reference the base type is ok.

matthchr: ack/agree on not using `New` if this package has more stuff in it.

I do slightly prefer `NewResolver` rather than `NewDefaultResolver` - possibly as the struct name too. I mostly prefer that because of a bias against the word `Default` in type names and functions. My thinking there is that the default is the one you use by default, not a property of the object itself, if that makes sense.

i.e. if we added a second variant of this and started usign `NewSuperResolver` instead, now it's the default (but it's not _called_ default it's called something else and the old one called default isn't actually the default anymore).

`default` as a variable name (or function parameter, etc) == fine
`default` as a package-scoped method name or type name == less fine, because it's making assumptions it can't really know about usage.

Approving though because it's a nit =)

matthchr (approved):

pkg/providers/imagefamily/resolver.go
matthchr: and there are cases where Default as a function name makes sense (i.e. "make me the default one... you figure it out") so it's very subjective. i.e. in the case above where we added `SuperResolver` we could make NewDefaultResolver return that and rename the non-super one to something else, and then it's OK.

charliedmcb: Leaving as is, since it matches with Default being used across the codebase. However, insightful to have your reasoning/thinking here. :)

PR #752, opened 2025-03-27
chore(refactor): reorder fields in families to match

charliedmcb:
Fixes # 

**Description**
Was doing comparisons between the two image families, and was coming up as there was a diff between their `ScriptlessCustomData` funcs. They produce the same thing, but just have slight different ordering. Putting out a quick PR to align the two.

Nothing additional
*

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
NONE
```


pkg/providers/imagefamily/ubuntu_2204.go
```diff
@@ -94,8 +94,8 @@ func (u Ubuntu2204) ScriptlessCustomData(kubeletConfig *bootstrap.KubeletConfigu
 			CABundle:         caBundle,
 			GPUNode:          u.Options.GPUNode,
 			GPUDriverVersion: u.Options.GPUDriverVersion,
-			GPUImageSHA:      u.Options.GPUImageSHA,
 			GPUDriverType:    u.Options.GPUDriverType,
+			GPUImageSHA:      u.Options.GPUImageSHA,
```
Bryce-Soghigian, line 98: why

Bryce-Soghigian (approved):

pkg/providers/imagefamily/ubuntu_2204.go
charliedmcb: Aligning with azure linux image family. Found it a little annoying when doing the diff, so felt worth fixing, since less than 5 min change.

PR #601, opened 2024-12-03
chore: update sigs.k8s.io/karpenter to v1.4.0

tallaxes:
**Description**

Update sigs.k8s.io/karpenter to v1.4.0. Most of the changes in provider are straightforward, here are some less obvious:
* Remove support for webhooks
* Remove alternative operator - should no longer be needed
* Minimal support for Node Repair policies introduced, disabled (behind feature flag), enabled for testing only https://github.com/Azure/karpenter-provider-azure/pull/601/commits/9d265f10e57690e8d9a9a1a81412c0dd9bd1c772
* Have instance Delete call Get first (including to satisfy new sigs.k8s.io/karpenter expectation) https://github.com/Azure/karpenter-provider-azure/pull/601/commits/e283696366e1cdf27c04e2cfb9ec2c5cf32381bd
* Restore "should drain pods on a node in order" E2E test (now supported) https://github.com/Azure/karpenter-provider-azure/pull/601/commits/f8483506db48c945a20a088a8627cef52726080d
* "karpenter.azure.com/aksnodeclass" label (used internally) is allowed 

**How was this change tested?**

* `make presubmit`
* E2E:
  * https://github.com/Azure/karpenter-provider-azure/actions/runs/14214116697 (https://github.com/Azure/karpenter-provider-azure/commit/9d265f10e57690e8d9a9a1a81412c0dd9bd1c772) ✅
  * https://github.com/Azure/karpenter-provider-azure/actions/runs/14248373209 (https://github.com/Azure/karpenter-provider-azure/pull/601/commits/f8483506db48c945a20a088a8627cef52726080d) ✅
  * https://github.com/Azure/karpenter-provider-azure/actions/runs/14526432427 (https://github.com/Azure/karpenter-provider-azure/pull/601/commits/b7963334cb35768d2cfabbe637f997ef83168656) ✅ - latest

**Does this change impact docs?**
- [ ] Yes, PR includes docs updates
- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

PR #753, opened 2025-03-27
feat(upgrade): begin using KubernetesVersion from nodeclass status

charliedmcb:
Fixes # 

**Description**
Updating all Karpenter to use the KubernetesVersion from NodeClass Status within its logic, rather than fetching from API server. Cases updated:
- Provisioning
- Drift
- Listing NodeImages in Status Controller

- Follow Ups:
    - Handling `NodeNotFound` plus nodes with a `DeletionTimestamp` gracefully
    - Restructuring Drift error handling, as non-blocking of cases at the higher level
    - Reusing new NoDrift const elsewhere

- Local testing tested upgrade process from before status controllers to this PR's state:
    - Started with self-hosted install of [v0.7.4](https://github.com/Azure/karpenter-provider-azure/releases/tag/v0.7.4) [pre-status controllers]
    - Created an aksnodeclass, nodepool, and 10 pods for 10 nodeclaims
    - Upgraded to a build of karpenter based off this PR, commit [666d6640a8306afb19aca2980530e7bd596e0708](https://github.com/Azure/karpenter-provider-azure/pull/753/commits/666d6640a8306afb19aca2980530e7bd596e0708)
    - New statuses were smoothly added immediately on karpenter upgrade, with no disruption occurrences.

E2E tests:
- https://github.com/Azure/karpenter-provider-azure/actions/runs/14116761196
- https://github.com/Azure/karpenter-provider-azure/actions/runs/14321717707 
- https://github.com/Azure/karpenter-provider-azure/actions/runs/14414624541
- https://github.com/Azure/karpenter-provider-azure/actions/runs/14579583526 [ObservedGeneration check added] 
- https://github.com/Azure/karpenter-provider-azure/actions/runs/14581615365
    - Failure due to small commit hash only containing numbers. BUG in framework of E2E pipeline atm. 
- https://github.com/Azure/karpenter-provider-azure/actions/runs/14582051800
- https://github.com/Azure/karpenter-provider-azure/actions/runs/14719804444
- https://github.com/Azure/karpenter-provider-azure/actions/runs/14720417850
- https://github.com/Azure/karpenter-provider-azure/actions/runs/14737685270 (latest)
    - Failure on test case: `should delete pod with do-not-disrupt when it reaches its terminationGracePeriodSeconds`
        - I got a failure on this in a previous run too, but also spotting it had a failure on main:
        https://github.com/Azure/karpenter-provider-azure/actions/runs/14656881143/job/41133430788#step:18:10429
        I got a [successful run](https://github.com/Azure/karpenter-provider-azure/actions/runs/14719804444) for commit [44cd0ad](https://github.com/Azure/karpenter-provider-azure/pull/753/commits/44cd0ad0b655b43bbb6a27ac91e06b7411a57136). Only a minor fix to logging and sync with main has occurred since [Commits](https://github.com/Azure/karpenter-provider-azure/pull/753/files/44cd0ad0b655b43bbb6a27ac91e06b7411a57136..666d6640a8306afb19aca2980530e7bd596e0708). I don't see this being related to this change, so ignoring. 

Note: got a CI-Test failure, that looks unrelated, has to do with pricing:
- https://github.com/Azure/karpenter-provider-azure/actions/runs/14720363178/job/41312867954?pr=753#step:6:3408
    - `Log in goroutine after TestAzure has completed: 2025-04-29T00:12:22.795Z	ERROR	pricing	pricing/pricing.go:167	error fetching updated pricing for region, using existing pricing data, on-demand: 2024-12-13T01:59:24Z`

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
Karpenter's usage of kubernetes version for provisioning and drift will now be based off the value stored in each AKSNodeClass' Status, which is controlled by the NodeClass Status controllers.
```


pkg/providers/imagefamily/nodeimage.go
```diff
@@ -38,10 +38,8 @@ type NodeImageProvider interface {
 
 // Returns the list of available NodeImages for the given AKSNodeClass sorted in priority ordering
 func (p *Provider) List(ctx context.Context, nodeClass *v1alpha2.AKSNodeClass) ([]NodeImage, error) {
-	kubernetesVersion, err := p.KubeServerVersion(ctx)
-	if err != nil {
-		return []NodeImage{}, err
-	}
+	// Should we be nil checking status?
```
charliedmcb, line 41: Don't need to, as both Status, and KubernetesVersion are values not pointers. However, we should check the Status Condition for KubernetesVersion, and/or the value != "".

pkg/apis/v1alpha2/aksnodeclass_status.go
```diff
@@ -68,3 +70,27 @@ func (in *AKSNodeClass) GetConditions() []status.Condition {
...
+// - The condition's ObservedGeneration is up to date with the latest spec generation
+// - The KubernetesVersion is initialized and non-empty
+func (in *AKSNodeClass) KubernetesVersionReadyAndValid() error {
+	if in == nil {
+		return fmt.Errorf("NodeClass is nil, meaning %s is consequently unready", ConditionTypeKubernetesVersionReady)
+	}
+	kubernetesVersionStatusCondition := in.StatusConditions().Get(ConditionTypeKubernetesVersionReady)
+	if !kubernetesVersionStatusCondition.IsTrue() {
+		return fmt.Errorf("NodeClass condition %s is not ready with status %s ", ConditionTypeKubernetesVersionReady, kubernetesVersionStatusCondition.GetStatus())
+		// TODO: this needs to be uncommented as soon as we update core to 1.1.x, but until then would make tests, and code checks fail.
```
charliedmcb, line 89: DO NOT RESOLVE THIS COMMENT UNTIL THE TODO IS ADDRESSES.

This comment needs to be addressed before completing this PR, as we want to handle ObservedGeneration. The code here is commented out for testing purposes while we wait for `1.1.x` core bump.

pkg/cloudprovider/drift.go
```diff
@@ -44,14 +44,16 @@ const (
 	K8sVersionDrift   cloudprovider.DriftReason = "K8sVersionDrift"
 	ImageVersionDrift cloudprovider.DriftReason = "ImageVersionDrift"
 	SubnetDrift       cloudprovider.DriftReason = "SubnetDrift"
+
+	NoDrift cloudprovider.DriftReason = ""
```
charliedmcb, line 48: Plan to handle expanding this consts usage in a separate PR.

pkg/apis/v1alpha2/aksnodeclass_status.go
```diff
@@ -68,3 +70,27 @@ func (in *AKSNodeClass) GetConditions() []status.Condition {
...
+
+// Will return nil if the the KubernetesVersion is considered valid to use,
+// otherwise will return an error detailing the reason of failure.
+//
+// Ensures
+// - The AKSNodeClass is non-nil
+// - The AKSNodeClass' KubernetesVersionReady Condition is true
+// - The condition's ObservedGeneration is up to date with the latest spec generation
+// - The KubernetesVersion is initialized and non-empty
+func (in *AKSNodeClass) KubernetesVersionReadyAndValid() error {
```
comtalyst, line 82: Maybe "AssertKubernetesVersionReadyAndValid()" is clearer(?).

pkg/providers/imagefamily/image.go
```diff
@@ -111,6 +111,7 @@ func (p *Provider) GetLatestImageID(ctx context.Context, defaultImage DefaultIma
 	return imageID, nil
 }
 
+// TODO: refactor this into kubernetesversion.go, and split into a new kubernetes provider
 func (p *Provider) KubeServerVersion(ctx context.Context) (string, error) {
```
comtalyst, line 115: Is there still a use of this method?

pkg/test/nodeclass.go
```diff
@@ -45,12 +47,28 @@ func AKSNodeClass(overrides ...v1alpha2.AKSNodeClass) *v1alpha2.AKSNodeClass {
...
-		ObjectMeta: test.ObjectMeta(options.ObjectMeta),
+		ObjectMeta: coretest.ObjectMeta(options.ObjectMeta),
 		Spec:       options.Spec,
 		Status:     options.Status,
 	}
 }
 
+func ApplyDefaultStatus(nodeClass *v1alpha2.AKSNodeClass, env *coretest.Environment) {
+	testK8sVersion := lo.Must(semver.ParseTolerant(lo.Must(env.KubernetesInterface.Discovery().ServerVersion()).String())).String()
+	nodeClass.Status.KubernetesVersion = testK8sVersion
```
comtalyst, line 58: In real environment, how do we plan to have `nodeClass.Status.KubernetesVersion` populated on existing/new node classes by the time the image is updated with this change?

pkg/apis/v1alpha2/aksnodeclass_status.go
charliedmcb: I only think `Assert` should be used as a prefix in test code.

However, could reframe as `ValidateKubernetesVersionReadiness`

pkg/providers/imagefamily/image.go
charliedmcb: Yes, it's used in the Controller that sets the KubernetesVersion on the Status.
https://github.com/Azure/karpenter-provider-azure/blob/e2e66bf4d6b564682fb4e6d702226864d5caba53/pkg/controllers/nodeclass/status/kubernetesversion.go#L76

pkg/cloudprovider/drift.go
charliedmcb: Added TODO

pkg/cloudprovider/suite_test.go
```diff
@@ -233,5 +238,62 @@ var _ = Describe("CloudProvider", func() {
...
+
+			// TODO (charliedmcb): I'm wondering if we actually want to have these soft-error cases switch to return an error if no-drift condition was found.
+			It("shouldn't error or be drifted when NodeName is missing", func() {
+				nodeClaim.Status.NodeName = ""
+				drifted, err := cloudProvider.IsDrifted(ctx, nodeClaim)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(drifted).To(Equal(NoDrift))
+			})
+
+			// TODO (charliedmcb): wondering if we should actually ignore not found errors? Can IsDrifted be called before the node exists?
```
matthchr, line 276:
```
	// From here there are three scenarios to handle:
	// 1. If NodeClaim is not launched, remove the drift status condition
	if !nodeClaim.StatusConditions().Get(v1.ConditionTypeLaunched).IsTrue() {
		_ = nodeClaim.StatusConditions().Clear(v1.ConditionTypeDrifted)
		if hasDriftedCondition {
			log.FromContext(ctx).V(1).Info("removing drift status condition, isn't launched")
		}
		return reconcile.Result{}, nil
	}
```

Suggests no

matthchr: (from: pkg/controllers/nodeclaim/disruption/drift.go)

pkg/providers/imagefamily/nodeimage.go
```diff
@@ -38,10 +38,10 @@ type NodeImageProvider interface {
 
 // Returns the list of available NodeImages for the given AKSNodeClass sorted in priority ordering
 func (p *Provider) List(ctx context.Context, nodeClass *v1alpha2.AKSNodeClass) ([]NodeImage, error) {
-	kubernetesVersion, err := p.KubeServerVersion(ctx)
-	if err != nil {
+	if err := nodeClass.ValidateKubernetesVersionReadiness(); err != nil {
```
matthchr, line 41: minor: You do this "check then read" pattern a few places. Maybe just make a helper method for it on `nodeClass`?

`func (c *nodeClass) GetKubernetesVersion() (string, error)` that calls `ValidateKubernetesVersionReadiness` internally to produce the error, and then returns the version?

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -100,11 +100,12 @@ func (p *Provider) GetTemplate(ctx context.Context, nodeClass *v1alpha2.AKSNodeC
 		return nil, err
 	}
 
-	kubeServerVersion, err := p.imageProvider.KubeServerVersion(ctx)
-	if err != nil {
+	// Note: we check ValidateKubernetesVersionReadiness at the start of the Create call. However, to ensure no call paths slip through, also calling it
+	// where we use the KubernetesVersion
+	if err = nodeClass.ValidateKubernetesVersionReadiness(); err != nil {
```
matthchr, line 105: check-then-read

pkg/cloudprovider/drift.go
```diff
@@ -91,11 +94,19 @@ func (c *CloudProvider) areStaticFieldsDrifted(nodeClaim *karpv1.NodeClaim, node
...
 
-	nodeName := nodeClaim.Status.NodeName
+	// Only check for potential drift if kubernetes version is ready, and valid to use. Note: AWS doesn't do this.
+	if err := nodeClass.ValidateKubernetesVersionReadiness(); err != nil {
+		// Note: we don't consider this a hard failure for drift if the KubernetesVersion is invalid, so we ignore returning the error here.
+		// We simply want to ensure the stored version is valid and ready to use, if we are to calculate potential Drift based on it.
+		// TODO (charliedmcb): I'm wondering if we actually want to have these soft-error cases switch to return an error if no-drift condition was found across all of IsDrifted.
+		return "", nil //nolint:nilerr
+	}
+	k8sVersion := nodeClass.Status.KubernetesVersion
```
matthchr, line 107: check-then-read

pkg/apis/v1alpha2/aksnodeclass_status.go
```diff
@@ -68,3 +70,27 @@ func (in *AKSNodeClass) GetConditions() []status.Condition {
...
+// - The AKSNodeClass' KubernetesVersionReady Condition is true
+// - The Condition's ObservedGeneration is up to date with the latest Spec Generation
+// - The KubernetesVersion is initialized and non-empty
+func (in *AKSNodeClass) ValidateKubernetesVersionReadiness() error {
+	if in == nil {
+		return fmt.Errorf("NodeClass is nil, meaning %s is consequently unready", ConditionTypeKubernetesVersionReady)
+	}
+	kubernetesVersionStatusCondition := in.StatusConditions().Get(ConditionTypeKubernetesVersionReady)
+	if !kubernetesVersionStatusCondition.IsTrue() {
+		return fmt.Errorf("NodeClass condition %s is not ready with status %s ", ConditionTypeKubernetesVersionReady, kubernetesVersionStatusCondition.GetStatus())
```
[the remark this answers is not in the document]
charliedmcb: So, I had it this way since its grouping `Unknown`, and `False` together.

pkg/cloudprovider/drift.go
```diff
@@ -44,14 +44,17 @@ const (
 	K8sVersionDrift   cloudprovider.DriftReason = "K8sVersionDrift"
 	ImageVersionDrift cloudprovider.DriftReason = "ImageVersionDrift"
 	SubnetDrift       cloudprovider.DriftReason = "SubnetDrift"
+
+	// TODO (charliedmcb): Use this const across code and test locations which are signaling/checking for "no drift"
+	NoDrift cloudprovider.DriftReason = ""
```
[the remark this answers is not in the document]
charliedmcb: I use it here too. Was just lowering code touch, and would be a simple PR to follow up with.

pkg/providers/imagefamily/nodeimage.go
charliedmcb: Seems good. My patterning/idea was that we would always be checking for validity when we read.

pkg/cloudprovider/cloudprovider.go
```diff
@@ -110,6 +110,9 @@ func (c *CloudProvider) Create(ctx context.Context, nodeClaim *karpv1.NodeClaim)
 	if nodeClassReady.IsUnknown() {
 		return nil, fmt.Errorf("resolving NodeClass readiness, NodeClass is in Ready=Unknown, %s", nodeClassReady.Message)
 	}
+	if err = nodeClass.ValidateKubernetesVersionReadiness(); err != nil {
```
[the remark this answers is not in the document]
charliedmcb: Good question on if this should actually be fatal. 

Currently there is actually no path in the controller to set `ConditionTypeKubernetesVersionReady` to false. Only true, and will be unknown in some instances as well.

The other cases of failure here though now include the ObservedGeneration of the K8s status not equaling the NodeClass' Generation. This case should get fixed via the reconciler, so could be argued if it should be fatal or not. I think its ok for this to be fatal, but also reasonable if not.

pkg/apis/v1alpha2/aksnodeclass_status.go
```diff
@@ -68,3 +70,27 @@ func (in *AKSNodeClass) GetConditions() []status.Condition {
...
+
+// Will return nil if the the KubernetesVersion is considered valid to use,
+// otherwise will return an error detailing the reason of failure.
+//
+// Ensures
+// - The AKSNodeClass is non-nil
+// - The AKSNodeClass' KubernetesVersionReady Condition is true
+// - The Condition's ObservedGeneration is up to date with the latest Spec Generation
+// - The KubernetesVersion is initialized and non-empty
+func (in *AKSNodeClass) ValidateKubernetesVersionReadiness() error {
```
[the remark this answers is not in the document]
charliedmcb: Don't have strong opinions between the two. Slightly lean towards Validate myself.

pkg/cloudprovider/drift.go
```diff
@@ -91,11 +94,19 @@ func (c *CloudProvider) areStaticFieldsDrifted(nodeClaim *karpv1.NodeClaim, node
...
+func (c *CloudProvider) isK8sVersionDrifted(ctx context.Context, nodeClaim *karpv1.NodeClaim, nodeClass *v1alpha2.AKSNodeClass) (cloudprovider.DriftReason, error) {
 	logger := logging.FromContext(ctx)
 
-	nodeName := nodeClaim.Status.NodeName
+	// Only check for potential drift if kubernetes version is ready, and valid to use. Note: AWS doesn't do this.
+	if err := nodeClass.ValidateKubernetesVersionReadiness(); err != nil {
+		// Note: we don't consider this a hard failure for drift if the KubernetesVersion is invalid, so we ignore returning the error here.
+		// We simply want to ensure the stored version is valid and ready to use, if we are to calculate potential Drift based on it.
+		// TODO (charliedmcb): I'm wondering if we actually want to have these soft-error cases switch to return an error if no-drift condition was found across all of IsDrifted.
+		return "", nil //nolint:nilerr
```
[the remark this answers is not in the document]
charliedmcb: I hadn't considered logging here. Would probably be fine either way as the reconcile of the status should happen relatively quickly after an update.

pkg/cloudprovider/drift.go
```diff
@@ -91,11 +94,19 @@ func (c *CloudProvider) areStaticFieldsDrifted(nodeClaim *karpv1.NodeClaim, node
...
-func (c *CloudProvider) isK8sVersionDrifted(ctx context.Context, nodeClaim *karpv1.NodeClaim) (cloudprovider.DriftReason, error) {
+func (c *CloudProvider) isK8sVersionDrifted(ctx context.Context, nodeClaim *karpv1.NodeClaim, nodeClass *v1alpha2.AKSNodeClass) (cloudprovider.DriftReason, error) {
 	logger := logging.FromContext(ctx)
 
-	nodeName := nodeClaim.Status.NodeName
+	// Only check for potential drift if kubernetes version is ready, and valid to use. Note: AWS doesn't do this.
+	if err := nodeClass.ValidateKubernetesVersionReadiness(); err != nil {
+		// Note: we don't consider this a hard failure for drift if the KubernetesVersion is invalid, so we ignore returning the error here.
+		// We simply want to ensure the stored version is valid and ready to use, if we are to calculate potential Drift based on it.
+		// TODO (charliedmcb): I'm wondering if we actually want to have these soft-error cases switch to return an error if no-drift condition was found across all of IsDrifted.
```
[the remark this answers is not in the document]
charliedmcb: No, so Drift pattern is currently built on short circuiting:
- https://github.com/Azure/karpenter-provider-azure/blob/881605ecb0f8b38dcaf115ad65d477339d645330/pkg/cloudprovider/drift.go#L49
- We check all the dirft cases one by one, and if any of them produce an error, or a IsDrifted status than we exit early and return that case. This means that any of the sub-drift categories that return an error is terminal for all drift calculations and will prevent others from being considered. This TODO comment is me wondering if we should change this to still check all drift conditions on an error, and only return an error back to core if no condition is drifted, meaning that IsDrifted will take precedence over an error, as I think attempting to drift the node as a cleanup might make more sense than erroring out and not considering drift cases.

pkg/providers/imagefamily/nodeimage.go
matthchr: You could even not export the validate one, and just use the `_, err` variant in that case (avoids 2 methods). IMO either is fine.

pkg/cloudprovider/suite_test.go
charliedmcb: Actually that makes me think that we might be able to reach this point while the node doesn't exist, as the cases you pointed to us clearing the Drifted condition meaning that it would have had to get there while the it was not launched

charliedmcb: Oh, no you are right. The call for IsDrifted happens after the fix block you showed:
https://github.com/kubernetes-sigs/karpenter/blob/9877cf639e665eadcae9e46e5a702a1b30ced1d3/pkg/controllers/nodeclaim/disruption/drift.go#L58

Should be fine than.

charliedmcb: Removed TODO, and left a comment answering the question at the actual code location NotFound would occur.

pkg/providers/imagefamily/nodeimage.go
charliedmcb: Updated, and kept the validate one internal.

pkg/apis/v1alpha2/aksnodeclass_status.go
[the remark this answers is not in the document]
charliedmcb: Updated error message to match the nodeclass readiness check more closely

pkg/cloudprovider/cloudprovider.go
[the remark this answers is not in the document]
charliedmcb: So, this is actually a good question for how/if we want to return a `IsNodeClassNotReadyError` here which would be treated as a hard failure:
- https://github.com/kubernetes-sigs/karpenter/blob/9877cf639e665eadcae9e46e5a702a1b30ced1d3/pkg/controllers/nodeclaim/lifecycle/launch.go#L92

I think as we expect the KubernetesVersion controller to fix the status, it should ok to have it be a non-terminal failure, and have the nodeclaim retried by core, but if there's reason this should be changed we could switch it to a `IsNodeClassNotReadyError`

pkg/apis/v1alpha2/aksnodeclass_status.go
charliedmcb: Resolved, and uncommented with TODO removed, as core bump occurred here:
https://github.com/Azure/karpenter-provider-azure/pull/601

pkg/cloudprovider/cloudprovider.go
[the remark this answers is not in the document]
charliedmcb: So, I actually do think we can get here when the node doesn't exist yet.

In core, the Launched status will become ready after we get a successful return from the cloudprovider Create:
https://github.com/kubernetes-sigs/karpenter/blob/9877cf639e665eadcae9e46e5a702a1b30ced1d3/pkg/controllers/nodeclaim/lifecycle/launch.go#L71

Right now, this won't really be an issue we encounter, but won't we fix the LRO issue, I believe that will mean we can get here before registration happens.

pkg/cloudprovider/suite_test.go
charliedmcb: So, I actually do think we can get here when the node doesn't exist yet.

In core, the Launched status will become ready after we get a successful return from the cloudprovider Create:
https://github.com/kubernetes-sigs/karpenter/blob/9877cf639e665eadcae9e46e5a702a1b30ced1d3/pkg/controllers/nodeclaim/lifecycle/launch.go#L71

Right now, this won't really be an issue we encounter, but won't we fix the LRO issue, I believe that will mean we can get here before registration happens.

matthchr (dismissed):
Overall LGTM. I had some comments around error handling for drift, and I do think the other way you suggested (see comments) is better, but OK if it's changed in a follow up PR.

pkg/cloudprovider/drift.go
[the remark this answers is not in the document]
charliedmcb: Yea, that's my thinking.

If we update the general error handling like above, than I'd switch this to return an error, but right now it would be fatal, which I don't want.

pkg/apis/v1alpha2/aksnodeclass_status.go
```diff
@@ -68,3 +70,35 @@ func (in *AKSNodeClass) GetConditions() []status.Condition {
 func (in *AKSNodeClass) SetConditions(conditions []status.Condition) {
 	in.Status.Conditions = conditions
 }
+
+// Returns the Status.KubernetesVersion if its up to date and valid to use, otherwise returns an error.
```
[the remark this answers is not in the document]
charliedmcb: Updated

pkg/cloudprovider/suite_test.go
```diff
@@ -235,5 +241,61 @@ var _ = Describe("CloudProvider", func() {
...
+			It("shouldn't error or be drifted when KubernetesVersion is empty", func() {
+				nodeClass = ExpectExists(ctx, env.Client, nodeClass)
+				nodeClass.Status.KubernetesVersion = ""
+				ExpectApplied(ctx, env.Client, nodeClass)
+				drifted, err := cloudProvider.IsDrifted(ctx, nodeClaim)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(drifted).To(Equal(NoDrift))
+			})
+
+			// TODO (charliedmcb): I'm wondering if we actually want to have these soft-error cases switch to return an error if no-drift condition was found.
```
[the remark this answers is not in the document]
charliedmcb: Yea, that's fair. Probably don't want to error in that case.

pkg/cloudprovider/suite_test.go
```diff
@@ -235,5 +241,61 @@ var _ = Describe("CloudProvider", func() {
...
+			})
+
+			It("should error when node is not found", func() {
+				nodeClaim.Status.NodeName = "NodeWhoDoesNotExist"
+				drifted, err := cloudProvider.IsDrifted(ctx, nodeClaim)
+				Expect(err).To(HaveOccurred())
+				Expect(drifted).To(Equal(NoDrift))
+			})
+
+			It("should succeed with drift true when KubernetesVersion is new", func() {
```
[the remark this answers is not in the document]
charliedmcb: The kubernetes version status controller will block this from happening. I don't think its Drifts job, since Drift is just comparing expected and actual state and trying to match the two, so I don't plan on introducing checked for rollback in Drift.

pkg/cloudprovider/suite_test.go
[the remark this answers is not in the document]
charliedmcb: Removed TODO, and added a comment in the code location

pkg/apis/v1alpha2/aksnodeclass_status.go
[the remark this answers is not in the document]
charliedmcb: Plan on leaving as Validate.

matthchr (approved):
Please make sure to file an item so we don't forget to follow-up on the NodeNotFound and DeletionTimestamp scenarios.

pkg/cloudprovider/cloudprovider.go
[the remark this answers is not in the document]
charliedmcb: I think its ok how we are handling this here, as a retriable error, so closing this thread.

pkg/cloudprovider/suite_test.go
charliedmcb: Deferred as following up: Gracefully handle `NodeNodeFound` and `DeletionTimestamp` in K8s Drift

pkg/cloudprovider/drift.go
```diff
@@ -91,27 +94,36 @@ func (c *CloudProvider) areStaticFieldsDrifted(nodeClaim *karpv1.NodeClaim, node
...
 
+	nodeName := nodeClaim.Status.NodeName
 	if nodeName == "" {
 		return "", nil
 	}
 
 	n := &v1.Node{}
 	if err := c.kubeClient.Get(ctx, client.ObjectKey{Name: nodeName}, n); err != nil {
-		// TODO (charliedmcb): should we ignore is not found errors? Will it ever be trying to check for drift before the node/vm exist?
+		// Core's check for Launched status should currently prevent us from getting here before the node exists because of the LRO block on Create:
```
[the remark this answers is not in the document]
charliedmcb: Deferred as following up: Gracefully handle `NodeNodeFound` and `DeletionTimestamp` in K8s Drift

pkg/cloudprovider/suite_test.go
```diff
@@ -235,5 +241,61 @@ var _ = Describe("CloudProvider", func() {
...
+
+			// TODO (charliedmcb): I'm wondering if we actually want to have these soft-error cases switch to return an error if no-drift condition was found.
+			It("shouldn't error or be drifted when NodeName is missing", func() {
+				nodeClaim.Status.NodeName = ""
+				drifted, err := cloudProvider.IsDrifted(ctx, nodeClaim)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(drifted).To(Equal(NoDrift))
+			})
+
+			It("should error when node is not found", func() {
```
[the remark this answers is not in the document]
charliedmcb: Deferred as following up: Gracefully handle `NodeNodeFound` and `DeletionTimestamp` in K8s Drift

pkg/cloudprovider/drift.go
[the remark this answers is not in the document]
charliedmcb: Deferred as follow-up: Reuse NodeDrift const elsewhere

pkg/cloudprovider/drift.go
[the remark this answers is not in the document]
charliedmcb: Deferring as follow-up: Restructure Drift error handling, as non-blocking of cases at the higher level

PR #764, opened 2025-04-02
test(env): fix ACR env naming for local testing

charliedmcb:
Fixes # 

**Description**
During my local testing the `make az-e2etests` command was panicking when trying to get the ACR name from the env vars.

Tracking this down, I noticed that we are passing in `AZURE_ACR_NAME` to the command not `ACR_NAME`, so updating to match. 
- Note: this has already been fixed in `main` by also passing in `ACR_NAME`. However, I think we should align around `AZURE_ACR_NAME` naming instead as it matches the others, so keeping this change, and updating `Makefile-az.mk` too.

Note: this should still work correctly for our regular E2E testing pipeline, since it appears `ACR_NAME`, and `AZURE_ACR_NAME` are both set. However, will run a test to sanity check.

- Local `make az-e2etests`
- `make presubmit`
- E2E Pipeline run:
    - https://github.com/Azure/karpenter-provider-azure/actions/runs/14227192488
    - [post `make presubmit`] https://github.com/Azure/karpenter-provider-azure/actions/runs/14227681983

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
NONE
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,10 +1,10 @@
 repos:
 - repo: https://github.com/gitleaks/gitleaks
-  rev: v8.20.1
+  rev: v8.24.2
   hooks:
   - id: gitleaks
 - repo: https://github.com/golangci/golangci-lint
-  rev: v1.61.0
+  rev: v1.64.8
   hooks:
   - id: golangci-lint
 - repo: https://github.com/jumanjihouse/pre-commit-hooks
@@ -12,7 +12,7 @@ repos:
   hooks:
   - id: shellcheck
 - repo: https://github.com/crate-ci/typos
-  rev: v1.28.1
+  rev: v1.31.1
   hooks:
   - id: typos
     args: [--write-changes, --force-exclude, --exclude, go.mod]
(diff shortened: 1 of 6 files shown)


matthchr (dismissed):

tallaxes:
Doesn't this mapping work (already in main): https://github.com/Azure/karpenter-provider-azure/commit/ad1aeec8468a751467a6cc9fc461ff99345484d3#diff-91889fe1def48a242c98253301b5c50b947d889b021b66af4d5dffc4c39648d4R266

charliedmcb:
> Doesn't this mapping work (already in main): [ad1aeec#diff-91889fe1def48a242c98253301b5c50b947d889b021b66af4d5dffc4c39648d4R266](https://github.com/Azure/karpenter-provider-azure/commit/ad1aeec8468a751467a6cc9fc461ff99345484d3#diff-91889fe1def48a242c98253301b5c50b947d889b021b66af4d5dffc4c39648d4R266)

I thought this had been fixed in some way. Dug into it, and see my syncing with `main` due to levels of branch on branch stacking missed the change.

I do still think this var naming should be updated to align with the expected value of `AZURE_ACR_NAME`, so updated in `Makefile-az.mk` now too.

tallaxes (approved):
LGTM, but if renaming please also update other known contexts that use these tests without using make target (discussed offline).

PR #769, opened 2025-04-03
revert: allowing users to not specify vnet guid for managed vnet cluster case, it will still be optional if you want to pass in chart

Bryce-Soghigian:
**Description**
For managed vnet case it makes more sense to allow karpenter to default the value for you, but also have the option to pass in VNET GUID for custom vnet case. 

This way you dont need the cluster identity to need vnet/get in cases of passing in via helm values, but we also keep parity for users on managed vnet. 

- [Ran e2e test suites](https://github.com/Azure/karpenter-provider-azure/actions/runs/14309176441)
- make test
- make presubmit

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


pkg/operator/operator.go
```diff
@@ -223,3 +234,28 @@ func getCABundle(restConfig *rest.Config) (*string, error) {
 	}
 	return ptr.String(base64.StdEncoding.EncodeToString(transportConfig.TLS.CAData)), nil
 }
+
+func getVnetGUID(cfg *auth.Config, subnetID string) (string, error) {
```
matthchr, line 238: minor: Maybe document in a comment that this function requires the users to grant certain permissions to the identity that Karpenter is running as, which wouldn't otherwise need to be set?

matthchr (approved):

pkg/operator/operator.go
Bryce-Soghigian: Im thinking we should have a document for all of our helm values in general.  That we update as an artifact for each version?

PR #773, opened 2025-04-03
fix(bootstrapping): Refactor bootstrapping to encapsulate between provisioning

charliedmcb:
Fixes # 

**Description**
I discovered there was a bug within the provisioning where bootstrapping configuration can be leaked between different instances of provisioning. This is particularly the case with `kubeletFlagsBase`, which is stored as a `var`, causing it to keep info between provisioning requests. This surfaced as an issue when implementing a self-hosted test for upgrading to kubernetes version 1.31.6, as a kubelet flag that should have been removed for that version was still around due to nodes being on 1.30.x having stored the flag.

- `make presubmit`
- Tested deepcopying of `kubeletFlagsBase` locally fixes the k8s e2e test.
    - Also, can be seen to fix KubernetesUpgrade test suite: 
    https://github.com/Azure/karpenter-provider-azure/actions/runs/14254730540/job/39955199339 
- Latest E2E test: https://github.com/Azure/karpenter-provider-azure/actions/runs/14254553711

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
- Fix an issue in bootstrapping, which causes karpenter to fail to provision nodes until restarted when a cluster is upgraded to kubernetes version v1.31.x
```

### DIFF
--- a/pkg/providers/imagefamily/bootstrap/aksbootstrap.go
+++ b/pkg/providers/imagefamily/bootstrap/aksbootstrap.go
@@ -18,11 +18,9 @@ package bootstrap
 
 import (
 	"bytes"
-	_ "embed"
 	"encoding/base64"
 	"fmt"
 	"strings"
-	"text/template"
 
 	"github.com/Azure/karpenter-provider-azure/pkg/utils"
 	"github.com/blang/semver/v4"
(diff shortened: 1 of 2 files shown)


pkg/providers/imagefamily/bootstrap/aksbootstrap.go
```diff
@@ -222,190 +220,21 @@ type NodeBootstrapVariables struct {
...
-	}
-)
-
-const (
-	globalAKSMirror = "https://acs-mirror.azureedge.net"
-)
-
 func (a AKS) aksBootstrapScript() (string, error) {
 	// use these as the base / defaults
-	nbv := staticNodeBootstrapVars // don't need deep copy (yet)
```
tallaxes, line 396: > // don't need deep copy (yet)
😆

tallaxes (approved):
LGTM

pkg/providers/imagefamily/bootstrap/aksbootstrap.go
charliedmcb: Haha, yea I noticed that comment too :p

charliedmcb:
> LGTM. Any way to have a test that that would catch sharing?

I want to get this checked in, and as its been tested to fix things in e2e. I'm going to push it forward.

I do think we could create a unit test at least for the case I discovered around `kubeletFlags`. However, would have to dig into what potential testing could be implemented beyond that.

PR #812, opened 2025-04-19
feat: moving nic creation to be handled by Microsoft.Compute rather than direct calls to Microsoft.Network

Bryce-Soghigian:
Fixes # 

**Description**
This pr delegates network interface creation to the vm creation call. This also involves modifying our testing framework to have the nic api depend on the vm api. 

- make az-all 
- make presubmit 
- make az-e2etest

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note
BREAKING: tags are no longer set on the nic on the initial put
```


pkg/utils/utils.go
```diff
@@ -69,6 +69,11 @@ func MkVMID(resourceGroupName string, vmName string) string {
 	return fmt.Sprintf(idFormat, resourceGroupName, vmName)
 }
 
+func MkNICID(resourceGroupName string, vmName string) string {
+	const idFormat = "/subscriptions/subscriptionID/resourceGroups/%s/providers/Microsoft.Network/networkInterfaces/%s"
+	return fmt.Sprintf(idFormat, resourceGroupName, vmName)
+}
```
Bryce-Soghigian, line 75:
```suggestion
func MkNICID(resourceGroupName string, nicName string) string {
	const idFormat = "/subscriptions/subscriptionID/resourceGroups/%s/providers/Microsoft.Network/networkInterfaces/%s"
	return fmt.Sprintf(idFormat, resourceGroupName, nicName)
}
```

pkg/test/environment.go
```diff
@@ -90,10 +89,12 @@ func NewEnvironmentNonZonal(ctx context.Context, env *coretest.Environment) *Env
 func NewRegionalEnvironment(ctx context.Context, env *coretest.Environment, region string, nonZonal bool) *Environment {
 	testOptions := Options()
 
+	networkInterfacesAPI := &fake.NetworkInterfacesAPI{}
 	// API
-	virtualMachinesAPI := &fake.VirtualMachinesAPI{}
+	virtualMachinesAPI := &fake.VirtualMachinesAPI{
+		NetworkInterfacesAPI: networkInterfacesAPI,
```
Bryce-Soghigian, line 95: Should we be using NetworkInterfacesBehavior instead?

pkg/providers/instance/instance.go
```diff
@@ -782,20 +827,8 @@ func (p *DefaultProvider) pickSkuSizePriorityAndZone(
...
-	// nic, disk and all associated resources will be removed. If the VM was not created successfully and a nic was found,
-	// then we attempt to delete the nic.
-
-	nicErr := deleteNicIfExists(ctx, p.azClient.networkInterfacesClient, p.resourceGroup, resourceName)
-	if nicErr != nil {
-		log.FromContext(ctx).Error(nicErr, fmt.Sprintf("networkinterface.Delete for %s failed", resourceName))
-	}
-	return errors.Join(vmErr, nicErr)
+func (p *DefaultProvider) cleanupAzureResources(ctx context.Context, resourceName string) (err error) {
+	return deleteVirtualMachineIfExists(ctx, p.azClient.virtualMachinesClient, p.resourceGroup, resourceName)
```
Bryce-Soghigian, line 831: I was debating keeping in the garbage collection controller in case nic delete fails or NRP isn't succeeding in cleanup. Thoughts?

pkg/providers/instance/instance.go
```diff
@@ -375,6 +340,108 @@ type createVMOptions struct {
...
+// setNetworkProfile creates a network profile for the VM with the NIC configuration
+func setNetworkProfile(vmProperties *armcompute.VirtualMachineProperties, opts *createVMOptions) {
+	nicConfig := createNetworkInterfaceConfiguration(opts)
+
+	networkProfile := &armcompute.NetworkProfile{
+		NetworkAPIVersion: lo.ToPtr(armcompute.NetworkAPIVersionTwoThousandTwenty1101),
+		NetworkInterfaceConfigurations: []*armcompute.VirtualMachineNetworkInterfaceConfiguration{
+			nicConfig,
+		},
+		NetworkInterfaces: []*armcompute.NetworkInterfaceReference{
```
Bryce-Soghigian, line 434: If anyone knows how to specify tags please let me know

rakechill: yeah by checking sdk, i don't see anything:
image
image

pkg/providers/instance/instance.go
tallaxes: Yes, that makes sense to me. No harm in keeping it for a while, until we have evidence it is not needed.

pkg/test/environment.go
tallaxes: That's a great question. The use of networking API from our perspective essentially disappears altogether, handling network interfaces becomes an internal implementational detail of VM API _behavior_ - and IMO that's the right way to model the associated fakes. In that model, ideally there would not even be a fake for networking API anymore, since nothing in our process would be actually calling it. The unit tests that check for network interface properties should be checking them on/within fake VMs instead.

The only potential complication is if, for whatever reason, we need networking API elsewhere (would NIC GC be the only place left?), and thus need the corresponding fakes, and tests that might expect to observe the same network interfaces. Modeling this properly would require shared fake store of network interfaces. If figuring out the right way to do _that_ is hard, using network API as part of VM fake behavior is probably acceptable, as a shortcut. Or maybe we don't need NIC GC anymore at all, which would make fakes simple.

pkg/providers/instance/instance.go
Bryce-Soghigian: Thinking more on it, the gc controller relies on the existence of the tags 'karpenter.sh', which we no longer are able to put on the network interface with this api.

Bryce-Soghigian:
Closing for now since we are going a different direction

PR #819, opened 2025-04-22
fix: specify version in user agent rather than unspecified

Bryce-Soghigian:
Fixes # 

**Description**
This introduces the version field in project so that from now on we can track what version of karpenter requests are coming from. 

it wasn't.

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/utils/project/project.go
+++ b/pkg/utils/project/project.go
@@ -18,4 +18,4 @@ package project
 
 // Version is the karpenter app version injected during compilation
 // when using the Makefile
-var Version = "unspecified"
+var Version = "v0.7.6"


pkg/utils/project/project.go
```diff
@@ -18,4 +18,4 @@ package project
 
 // Version is the karpenter app version injected during compilation
 // when using the Makefile
-var Version = "unspecified"
+var Version = "v0.7.6"
```
matthchr, line 21: IMO this shouldnt be hardcoded it should be derived from tags.

See https://github.com/Azure/azure-service-operator/blob/main/v2/internal/version/version.go#L18
https://github.com/Azure/azure-service-operator/blob/main/Taskfile.yml#L40
https://github.com/Azure/azure-service-operator/blob/main/Taskfile.yml#L33

or
https://github.com/kro-run/kro/blob/main/Makefile#L26

tallaxes: It is already derived from tags/commits in our makefile. This Version needs to be removed, and user agent should be using the one from operator.

tallaxes (changes requested):
See comments

tallaxes:
Superceeded by https://github.com/Azure/karpenter-provider-azure/pull/948

PR #820, opened 2025-04-23
chore: basic reformatting

matthchr:
- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
NONE
```

### DIFF
--- a/pkg/providers/instance/instance.go
+++ b/pkg/providers/instance/instance.go
@@ -25,44 +25,37 @@ import (
 	"strings"
 	"time"
 
+	sdkerrors "github.com/Azure/azure-sdk-for-go-extensions/pkg/errors"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/compute/armcompute"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/network/armnetwork"
 	"github.com/samber/lo"
 	v1 "k8s.io/api/core/v1"
-
 	"k8s.io/apimachinery/pkg/util/sets"
 	"knative.dev/pkg/logging"
-
-	"github.com/Azure/karpenter-provider-azure/pkg/cache"
-	"github.com/Azure/karpenter-provider-azure/pkg/providers/instancetype"
-	"github.com/Azure/karpenter-provider-azure/pkg/providers/launchtemplate"
-	"github.com/Azure/karpenter-provider-azure/pkg/providers/loadbalancer"
-	"github.com/Azure/karpenter-provider-azure/pkg/utils"
-
+	karpv1 "sigs.k8s.io/karpenter/pkg/apis/v1"
 	corecloudprovider "sigs.k8s.io/karpenter/pkg/cloudprovider"
 	"sigs.k8s.io/karpenter/pkg/scheduling"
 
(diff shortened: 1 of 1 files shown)


pkg/providers/instance/instance.go
```diff
@@ -653,7 +683,11 @@ func GetAllSingleValuedRequirementLabels(instanceType *corecloudprovider.Instanc
 }
 
 // pick the "best" SKU, priority and zone, from InstanceType options (and their offerings) in the request
-func (p *DefaultProvider) pickSkuSizePriorityAndZone(ctx context.Context, nodeClaim *karpv1.NodeClaim, instanceTypes []*corecloudprovider.InstanceType) (*corecloudprovider.InstanceType, string, string) {
+func (p *DefaultProvider) pickSkuSizePriorityAndZone(
+	ctx context.Context,
+	nodeClaim *karpv1.NodeClaim,
+	instanceTypes []*corecloudprovider.InstanceType,
+) (*corecloudprovider.InstanceType, string, string) {
```
Bryce-Soghigian, line 690: iirc alex doesnt' like this formatting? Maybe im mistaken I feel like i tried to do it a long time ago and got told not to, if we are gonna go this route as well, maybe a linter for line length is in order? 

Thoughts?

Bryce-Soghigian (dismissed):
Personally though I use vim so I like this kind of thing, so will approve, but lets get wider teams thoughts

pkg/providers/instance/instance.go
```diff
@@ -522,7 +528,19 @@ func (p *DefaultProvider) launchInstance(
 	sshPublicKey := options.FromContext(ctx).SSHPublicKey
 	nodeIdentityIDs := options.FromContext(ctx).NodeIdentities
 	useSIG := options.FromContext(ctx).UseSIG
-	vm := newVMObject(resourceName, nicReference, zone, capacityType, p.location, sshPublicKey, nodeIdentityIDs, nodeClass, launchTemplate, instanceType, p.provisionMode, useSIG)
+	vm := newVMObject(
```
matthchr, line 531: This is maybe excessive - possibly could revert.

Really this function has too many args.

tallaxes (approved):

pkg/providers/instance/instance.go
tallaxes: It does feel a little excessive

pkg/providers/instance/instance.go
matthchr: I _think_ general guidance is that the function definition should either be:
1. On a single line
2. Like this.

I don't really want to enforce a line-length limit, I'm fine with some long lines, but at least IMO for functions like `getLaunchTemplate` above (which was already split across multiple lines) this format is objectively easier to read.

Some of the error messages above that I reformatted were long enough that they were off the right hand side of my monitor (which isn't ultra-wide, but it's not tiny either).

tallaxes: Can you point to a case?

pkg/providers/instance/instance.go
matthchr: I'll revert this, I don't love the formatting of it either.

Probably we should revisit this function and reduce its args at some point, but I am not going to do that here.

tallaxes (approved):

PR #836, opened 2025-04-25
feat: switch CloudProvider Create to be (mostly) non-blocking

### LINKED ISSUE #600: BUG: Budgets don't allow for async Drift
### Note:

For `Expected Behavior`, `Actual Behavior`, and `Steps to Reproduce the Problem` see the upstream issue:
https://github.com/kubernetes-sigs/karpenter/issues/1861

This Issue is purely setup to track the work needed in `Azure/karpenter-provider-azure` to complete the fix.

### Version

- Chart Version: Using Dev Build of Azure provider on HEAD of main with commit [d84e7c8](https://github.com/Azure/karpenter-provider-azure/commit/d84e7c840cdd73da640614f7104f4304c5f147f8)
    - `sigs.k8s.io/karpenter: v1.0.5`
- Kubernetes Version (`kubectl version`): v1.29.9

### Community Note

matthchr:
* NIC creation is still blocking, but that may change in the future if we roll NIC creation into VM creation.
 * instance provider now returns a VMFuture, which CloudProvider polls in a separate goroutine. If an error is encountered, the NodeClaim is deleted and the pending pods will trigger another NodeClaim to be created.

This improves disruption speed significantly, but there is still slowness in disruption (caused by core as far as I can tell).

Fixes #600

* Unit tests
* E2E tests (pass 1): https://github.com/Azure/karpenter-provider-azure/actions/runs/14605092382 - this had some flakes but I re-ran them locally and they passed.
* E2E tests (pass 2): https://github.com/Azure/karpenter-provider-azure/actions/runs/14673602666 - running a second pass because there were a few changes from pass 1 and want to make sure things are OK.
* E2E tests (pass 3): after rebase, passed: https://github.com/Azure/karpenter-provider-azure/actions/runs/14740596543

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
Updated provider node create to be non-blocking, which significantly increases the throughput of disruption. Disruption is now more likely to actually utilize the full disruption budget.
```

### DIFF
--- a/pkg/cloudprovider/cloudprovider.go
+++ b/pkg/cloudprovider/cloudprovider.go
@@ -25,12 +25,12 @@ import (
 	"time"
 
 	"github.com/awslabs/operatorpkg/status"
-
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/karpenter/pkg/metrics"
 
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -122,10 +122,20 @@ func (c *CloudProvider) Create(ctx context.Context, nodeClaim *karpv1.NodeClaim)
 	if len(instanceTypes) == 0 {
 		return nil, cloudprovider.NewInsufficientCapacityError(fmt.Errorf("all requested instance types were unavailable during launch"))
 	}
-	instance, err := c.instanceProvider.Create(ctx, nodeClass, nodeClaim, instanceTypes)
+	instancePromise, err := c.instanceProvider.BeginCreate(ctx, nodeClass, nodeClaim, instanceTypes)
 	if err != nil {
 		return nil, fmt.Errorf("creating instance, %w", err)
 	}
+
+	// Launch a single goroutine to poll the returned promise.
+	// Note that we could store the LRO details on the NodeClaim, but we don't bother today because Karpenter
+	// crashes should be rare, and even in the case of a crash, as long as the node comes up successfully there's
+	// no issue. If the node doesn't come up successfully in that case, the node and the linked claim will
+	// be garbage collected after the TTL, but the cause of the nodes issue will be lost, as the LRO URL was
+	// only held in memory.
+	go c.waitOnPromise(ctx, instancePromise, nodeClaim)
+
+	instance := instancePromise.VM
 	instanceType, _ := lo.Find(instanceTypes, func(i *cloudprovider.InstanceType) bool {
 		return i.Name == string(lo.FromPtr(instance.Properties.HardwareProfile.VMSize))
 	})
@@ -135,6 +145,68 @@ func (c *CloudProvider) Create(ctx context.Context, nodeClaim *karpv1.NodeClaim)
 		v1alpha2.AnnotationAKSNodeClassHashVersion: v1alpha2.AKSNodeClassHashVersion,
 	})
 	return nc, err
+}
+
+func (c *CloudProvider) waitOnPromise(ctx context.Context, promise *instance.VirtualMachinePromise, nodeClaim *karpv1.NodeClaim) {
+	defer func() {
+		if r := recover(); r != nil {
+			err := fmt.Errorf("%v", r)
+			log.FromContext(ctx).Error(err, "panic during waitOnPromise")
+		}
+	}()
+
+	err := promise.Wait()
+
+	// Wait until the claim is Launched, to avoid racing with creation.
+	// This isn't strictly required, but without this, failure test scenarios are harder
+	// to write because the nodeClaim gets deleted by error handling below before
+	// the EnsureApplied call finishes, so EnsureApplied creates it again (which is wrong/isn't how
+	// it would actually happen in production).
+	c.waitUntilLaunched(ctx, nodeClaim)
+
+	if err != nil {
+		c.recorder.Publish(cloudproviderevents.NodeClaimFailedToRegister(nodeClaim, err))
+		log.FromContext(ctx).Error(err, "failed launching nodeclaim")
+
+		// TODO: This won't clean up leaked NICs if the VM doesn't exist... intentional?
+		vmName := lo.FromPtr(promise.VM.Name)
+		err = c.instanceProvider.Delete(ctx, vmName)
+		if cloudprovider.IgnoreNodeClaimNotFoundError(err) != nil {
+			log.FromContext(ctx).Error(err, "failed to delete VM %s", vmName)
+		}
+
+		if err = c.kubeClient.Delete(ctx, nodeClaim); err != nil {
+			err = client.IgnoreNotFound(err)
+			if err != nil {
+				log.FromContext(ctx).Error(err, "failed to delete nodeclaim %s, will wait for liveness TTL", nodeClaim.Name)
+			}
+		}
+		metrics.NodeClaimsDisruptedTotal.Inc(map[string]string{
+			metrics.ReasonLabel:       "async_provisioning",
+			metrics.NodePoolLabel:     nodeClaim.Labels[karpv1.NodePoolLabelKey],
+			metrics.CapacityTypeLabel: nodeClaim.Labels[karpv1.CapacityTypeLabelKey],
+		})
+
+		return
+	}
+}
+
+func (c *CloudProvider) waitUntilLaunched(ctx context.Context, nodeClaim *karpv1.NodeClaim) {
+	freshClaim := &karpv1.NodeClaim{}
+	for {
+		err := c.kubeClient.Get(ctx, types.NamespacedName{Namespace: nodeClaim.Namespace, Name: nodeClaim.Name}, freshClaim)
+		if err != nil {
+			if client.IgnoreNotFound(err) == nil {
+				return
+			}
+			log.FromContext(ctx).Error(err, "failed getting nodeclaim to wait until launched")
+		}
+
+		if cond := freshClaim.StatusConditions().Get(karpv1.ConditionTypeLaunched); !cond.IsUnknown() {
+			return
+		}
+		time.Sleep(500 * time.Millisecond)
+	}
 }
 
 func (c *CloudProvider) List(ctx context.Context) ([]*karpv1.NodeClaim, error) {
--- a/pkg/cloudprovider/cloudprovider.go
+++ b/pkg/cloudprovider/cloudprovider.go
@@ -172,7 +172,7 @@ func (c *CloudProvider) waitOnPromise(ctx context.Context, promise *instance.Vir
 		vmName := lo.FromPtr(promise.VM.Name)
 		err = c.instanceProvider.Delete(ctx, vmName)
 		if cloudprovider.IgnoreNodeClaimNotFoundError(err) != nil {
-			log.FromContext(ctx).Error(err, "failed to delete VM %s", vmName)
+			log.FromContext(ctx).Error(err, fmt.Sprintf("failed to delete VM %s", vmName))
 		}
 
 		if err = c.kubeClient.Delete(ctx, nodeClaim); err != nil {
@@ -205,7 +205,12 @@ func (c *CloudProvider) waitUntilLaunched(ctx context.Context, nodeClaim *karpv1
 		if cond := freshClaim.StatusConditions().Get(karpv1.ConditionTypeLaunched); !cond.IsUnknown() {
 			return
 		}
-		time.Sleep(500 * time.Millisecond)
+
+		select {
+		case <-time.After(500 * time.Millisecond):
+		case <-ctx.Done():
+			return // context was canceled
+		}
 	}
 }
 
--- a/pkg/cloudprovider/events/events.go
+++ b/pkg/cloudprovider/events/events.go
@@ -26,13 +26,15 @@ import (
 )
 
 const (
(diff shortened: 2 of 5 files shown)


pkg/cloudprovider/cloudprovider.go
```diff
@@ -134,6 +144,68 @@ func (c *CloudProvider) Create(ctx context.Context, nodeClaim *karpv1.NodeClaim)
...
+
+	if err != nil {
+		c.recorder.Publish(cloudproviderevents.NodeClaimFailedToRegister(nodeClaim, err))
+		log.FromContext(ctx).Error(err, "failed launching nodeclaim")
+
+		// TODO: This won't clean up leaked NICs if the VM doesn't exist... intentional?
+		vmName := lo.FromPtr(promise.VM.Name)
+		err = c.instanceProvider.Delete(ctx, vmName)
+		if cloudprovider.IgnoreNodeClaimNotFoundError(err) != nil {
+			log.FromContext(ctx).Error(err, "failed to delete VM %s", vmName)
```
tallaxes, line 172: Does Error actually do Sprintf? (And should we do more structured logging with keys/values instead?)

Bryce-Soghigian:
```go
// Error logs an error, with the given message and key/value pairs as context.
// It functions similarly to Info, but may have unique behavior, and should be
// preferred for logging errors (see the package documentations for more
// information). The log message will always be emitted, regardless of
// verbosity level.
//
// The msg argument should be used to add context to any underlying error,
// while the err argument should be used to attach the actual error that
// triggered this log line, if present. The err parameter is optional
// and nil may be passed instead of an error instance.
func (l Logger) Error(err error, msg string, keysAndValues ...any) {
	if l.sink == nil {
		return
	}
	if withHelper, ok := l.sink.(CallStackHelperLogSink); ok {
		withHelper.GetCallStackHelper()()
	}
	l.sink.Error(err, msg, keysAndValues...)
}
```

pkg/cloudprovider/events/events.go
```diff
@@ -40,3 +46,22 @@ func NodeClaimFailedToResolveNodeClass(nodeClaim *v1.NodeClaim) events.Event {
 		DedupeValues:   []string{string(nodeClaim.UID)},
 	}
 }
+
+func NodeClaimFailedToRegister(nodeClaim *v1.NodeClaim, err error) events.Event {
+	return events.Event{
+		InvolvedObject: nodeClaim,
+		Type:           corev1.EventTypeWarning,
+		Reason:         AsyncProvisioningReason, // TODO: our other events don't set reason... is it an oversight?
```
tallaxes, line 54: No reason elsewhere: Yes, oversight

pkg/providers/instance/instance.go
```diff
@@ -533,25 +548,45 @@ func (p *DefaultProvider) launchInstance(
...
+	vm.Properties.TimeCreated = lo.ToPtr(time.Now())
 
-	err = p.createAKSIdentifyingExtension(ctx, resourceName)
-	if err != nil {
-		return nil, err
-	}
-	return resp, nil
+	return &VirtualMachinePromise{
+		Wait: func() error {
+			_, err = poller.PollUntilDone(ctx, nil)
```
tallaxes, line 566: Do we have to wait for this before initiating CSE create?

pkg/providers/instancetype/suite_test.go
```diff
@@ -663,8 +670,8 @@ var _ = Describe("InstanceType Provider", func() {
 			pod.Spec.Affinity = &v1.Affinity{NodeAffinity: &v1.NodeAffinity{PreferredDuringSchedulingIgnoredDuringExecution: []v1.PreferredSchedulingTerm{
 				{
 					Weight: 1, Preference: v1.NodeSelectorTerm{MatchExpressions: []v1.NodeSelectorRequirement{
-						{Key: v1.LabelTopologyZone, Operator: v1.NodeSelectorOpIn, Values: []string{fakeZone1}},
-					}},
+					{Key: v1.LabelTopologyZone, Operator: v1.NodeSelectorOpIn, Values: []string{fakeZone1}},
+				}},
 				},
```
tallaxes, line 675: Weird indent?

pkg/providers/instance/instance.go
matthchr: It's what we were doing before and I wanted to keep it the same just from a safety perspective.

pkg/cloudprovider/cloudprovider.go
matthchr: Yeah I think this is an oversight from my log refactor (I fixed all the existing logs but then didn't fix this in my rebase)

pkg/providers/instance/instance.go
tallaxes: @charliedmcb may have tried without wait?

pkg/cloudprovider/events/events.go
```diff
@@ -40,3 +46,22 @@ func NodeClaimFailedToResolveNodeClass(nodeClaim *v1.NodeClaim) events.Event {
...
+		Message:        fmt.Sprintf("NodeClaim %s failed to register: %s", nodeClaim.Name, truncateMessage(err.Error())),
```
tallaxes, line 55: Do we need to include nodeClaim.Name in message if it is recorded as InvolvedObject?

tallaxes (approved):
The core logic update looks great to me. Let's make sure to run E2E.

Would also be good to follow-up with a test that restarts Karpenter during scaleup (manually, or maybe extra low memory limit to induce OOMKills) to validate the behavior - but this could be done later.

pkg/cloudprovider/cloudprovider.go
matthchr: Fixed.

pkg/cloudprovider/events/events.go
matthchr: I have removed this comment. Do you want me to just add a reason to the other 2 events here?

matthchr: I've added a reason here too.

pkg/providers/instancetype/suite_test.go
matthchr: I added some newlines to this format to make it a bit more obvious what it was doing and fixed this wonky indentation mistake.

pkg/cloudprovider/events/events.go
matthchr: It doesn't make sense to include nodeClaim name in this message, agreed!

matthchr:
Updated PR description with links to E2E. Had run one pass already, but had refactored some since then, will run again.

matthchr:
Looks like I need to deal with test races and the logger issue now, as UTs are failing due to that.

tallaxes (dismissed):

pkg/providers/instance/instance.go
```diff
@@ -571,19 +591,44 @@ func (p *DefaultProvider) launchInstance(
...
+				// Poller is nil means the VM existed already and we're done
+				return nil
+			}
+
+			_, err = result.Poller.PollUntilDone(ctx, nil)
+			if err != nil {
+				azErr := p.handleResponseErrors(ctx, instanceType, zone, capacityType, err)
+				return azErr
+			}
+
```
comtalyst, line 613: IMO, `Wait()` should not do a new write each time it is called. E.g., behaving similar to short poll.
In this case it is fine given we always call Wait() once and expect restart to be rare. But wdyt about adding a warning note for the below calls?

pkg/providers/instance/instance.go
```diff
@@ -571,19 +591,44 @@ func (p *DefaultProvider) launchInstance(
...
+			if p.provisionMode == consts.ProvisionModeBootstrappingClient {
+				err = p.createCSExtension(ctx, resourceName, launchTemplate.CustomScriptsCSE, launchTemplate.IsWindows)
+				if err != nil {
+					// This should fall back to cleanupAzureResources
+					// TODO: Confirm this
```
comtalyst, line 618: I don't think so, not from `Create()` as we are returning `(promise, nil)` here. And if we delete the NodeClaim ourselves, then not through `Delete()` too?

pkg/providers/instance/instance.go
```diff
@@ -125,23 +131,25 @@ func NewDefaultProvider(
...
 	instanceTypes = orderInstanceTypesByPrice(instanceTypes, scheduling.NewNodeSelectorRequirementsWithMinValues(nodeClaim.Spec.Requirements...))
-	vm, err := p.launchInstance(ctx, nodeClass, nodeClaim, instanceTypes)
+	vmPromise, err := p.beginLaunchInstance(ctx, nodeClass, nodeClaim, instanceTypes)
 	if err != nil {
-		// Currently, CSE errors will lead to here
+		// There may be orphan NICs
 		if cleanupErr := p.cleanupAzureResources(ctx, GenerateResourceName(nodeClaim.Name)); cleanupErr != nil {
 			log.FromContext(ctx).Error(cleanupErr, fmt.Sprintf("failed to cleanup resources for node claim %s", nodeClaim.Name))
 		}
 		return nil, err
```
comtalyst, line 150: Don't think this will be hit from errors during async phase at all. Is it intentional? If it is, maybe add a warning comment here to prevent misunderstanding?

pkg/providers/instance/instance.go
```diff
@@ -125,23 +131,25 @@ func NewDefaultProvider(
...
 	}
+	vm := vmPromise.VM
 	zone, err := utils.GetZone(vm)
```
comtalyst, line 153: It feels a bit hacky here to take a sync-populated part out of async. But no quick solution in my mind...

pkg/cloudprovider/cloudprovider.go
```diff
@@ -122,10 +122,20 @@ func (c *CloudProvider) Create(ctx context.Context, nodeClaim *karpv1.NodeClaim)
 	if len(instanceTypes) == 0 {
 		return nil, cloudprovider.NewInsufficientCapacityError(fmt.Errorf("all requested instance types were unavailable during launch"))
 	}
-	instance, err := c.instanceProvider.Create(ctx, nodeClass, nodeClaim, instanceTypes)
+	instancePromise, err := c.instanceProvider.BeginCreate(ctx, nodeClass, nodeClaim, instanceTypes)
```
comtalyst, line 125: IIRC, there were some errors that are explicitly defined on core to be retriable/non-retriable on the same NodeClaim. Making this async + delete NodeClaim directly if it fails on async would stray us away from that. Has it been considered?

pkg/providers/instance/instance.go
matthchr: Wait won't ever be called more than once, AFAIK.

In the happy-path case, where Karpenter doesn't restart, the NodeClaim gets moved to `Launched` state, and then `Wait` gets called. Since the claim was launched, CloudProvider.Create won't get called again.

In the sad-path case, where Karpenter restarts, there are two sub-cases:
1. Karpenter crashes after CloudProvider Create finished, so the NodeClaim was launched. When Karpenter restarts it will just wait for the node to come up. If it comes up, great things go on as normal. If it doesn't come up, the node will be GC'ed after the liveness TTL.
2. Karpenter crashes during CloudProvider Create (before LRO is returned to us, so Wait hasn't been called yet), Bryce's recent PR will find the VM and so won't create it again, and when `Done()`  is called `poller` will be nil and we'll return. @Bryce-Soghigian this is where we have an issue because the VM extension is probably not installed yet, but I think we're punting on that for now... I should maybe put a comment though.

pkg/providers/instance/instance.go
matthchr: Correct, this is specifically for sync failures.

pkg/providers/instance/instance.go
matthchr: Not sure I understand the question.

pkg/providers/instance/instance.go
matthchr: We call `instanceProvider.Delete` if an error is returned from `Wait`, so I think this is true/correct

pkg/providers/instance/instance.go
matthchr: I added a comment here

pkg/cloudprovider/cloudprovider.go
matthchr: As discussed, sync failures (things like quota, etc) will be handled as they always have. The one case which was a retry but will now be a NodeClaim delete is ZonalProvisioning failures, but we will retry by recreating a node claim and (AFAIK) we remember the zonal restrictions so next time round we won't hit the same issue again.

cc @tallaxes to confirm that.

tallaxes: Seems correct to me. Not totally sure this covers all possible errors we can get on the async side, but whatever the case is, it seems like re-creating NodeClaim is the right option for those - even if we did not do that before. (Separately, `InsufficientCapacityError` also special treatment by core - NodeClaim delete - we already return that properly.)

pkg/cloudprovider/cloudprovider.go
```diff
@@ -122,10 +122,20 @@ func (c *CloudProvider) Create(ctx context.Context, nodeClaim *karpv1.NodeClaim)
...
 	if err != nil {
 		return nil, fmt.Errorf("creating instance, %w", err)
```
tallaxes, line 127: It looks like core now has [`cloudprovider.CreateError`](https://github.com/kubernetes-sigs/karpenter/blob/31eb6b511ee63fd2d04f382c0d82eec5cb6131d9/pkg/cloudprovider/types.go#L416-L420), which has a bit better mapping for reporting on NodeClaim conditions. May not be directly related to this PR, but possibly a good followup, for anything we return from Create?

tallaxes (approved):

pkg/cloudprovider/cloudprovider.go
matthchr: Let's tackle this later.

PR #838, opened 2025-04-27
Bsoghigian/vm extension

Bryce-Soghigian:
Fixes # 

**Description**

*

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/cloudprovider/suite_test.go
+++ b/pkg/cloudprovider/suite_test.go
@@ -22,6 +22,8 @@ import (
 	"testing"
 	"time"
 
+	"fmt"
+
 	"sigs.k8s.io/karpenter/pkg/test/v1alpha1"
 
 	"github.com/awslabs/operatorpkg/object"
@@ -134,6 +136,7 @@ var _ = AfterEach(func() {
 })
 
 var _ = Describe("CloudProvider", func() {
+	It("should delete the nodeclaim if we spot a vm without the extensions in the cache", func() {})
 	It("should list nodeclaim created by the CloudProvider", func() {
 		ExpectApplied(ctx, env.Client, nodeClass, nodePool)
 		pod := coretest.UnschedulablePod()
@@ -167,7 +170,7 @@ var _ = Describe("CloudProvider", func() {
 		Expect(corecloudprovider.IsInsufficientCapacityError(err)).To(BeTrue())
 		Expect(cloudProviderMachine).To(BeNil())
 	})
-	Context("Drift", func() {
+	FContext("Drift", func() {
 		var nodeClaim *karpv1.NodeClaim
 		var pod *v1.Pod
 		BeforeEach(func() {
@@ -183,6 +186,7 @@ var _ = Describe("CloudProvider", func() {
 			input := azureEnv.VirtualMachinesAPI.VirtualMachineCreateOrUpdateBehavior.CalledWithInput.Pop()
 			rg := input.ResourceGroupName
 			vmName := input.VMName
+			fmt.Println(input)
 			// Corresponding NodeClaim
 			nodeClaim = coretest.NodeClaim(karpv1.NodeClaim{
 				Status: karpv1.NodeClaimStatus{
--- a/pkg/providers/instance/errors.go
+++ b/pkg/providers/instance/errors.go
@@ -1,0 +1,43 @@
+/*
+Portions Copyright (c) Microsoft Corporation.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package instance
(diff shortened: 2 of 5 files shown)


pkg/cloudprovider/suite_test.go
```diff
@@ -134,6 +136,7 @@ var _ = AfterEach(func() {
 })
 
 var _ = Describe("CloudProvider", func() {
+	It("should delete the nodeclaim if we spot a vm without the extensions in the cache", func() {})
```
matthchr, line 139: Missing an actual test body?

pkg/cloudprovider/suite_test.go
```diff
@@ -183,6 +186,7 @@ var _ = Describe("CloudProvider", func() {
 			input := azureEnv.VirtualMachinesAPI.VirtualMachineCreateOrUpdateBehavior.CalledWithInput.Pop()
 			rg := input.ResourceGroupName
 			vmName := input.VMName
+			fmt.Println(input)
```
matthchr, line 189: Remove?

pkg/providers/instance/errors.go
```diff
@@ -0,0 +1,43 @@
...
+)
+
+type InvalidInstanceStateError struct {
+	err          error
+	InstanceName string
+	Reason       string
+}
+
+func (e *InvalidInstanceStateError) Error() string {
+	return fmt.Sprintf("instance %s is in an invalid state for reason: %s", e.InstanceName, e.Reason)
```
matthchr, line 30: Not printing err

pkg/providers/instance/errors.go
```diff
@@ -0,0 +1,43 @@
...
+
+func NewInvalidInstanceStateError(instanceName, failureReason string) error {
+	return &InvalidInstanceStateError{
+		InstanceName: instanceName,
+		Reason:       failureReason,
+	}
+}
+
+func IsInvalidInstanceStateError(err error) bool {
+	_, ok := err.(*InvalidInstanceStateError)
```
matthchr, line 41: GitHub actions is correct, you should use `errors.As` here

pkg/providers/instance/vmextension.go
```diff
@@ -0,0 +1,115 @@
...
+)
+
+const (
+	aksIdentifyingExtensionName      = "computeAksLinuxBilling"
+	aksIdentifyingExtensionPublisher = "Microsoft.AKS"
+	aksIdentifyingExtensionTypeLinux = "Compute.AKS.Linux.Billing"
+)
+
+// createAKSIdentifyingExtension attaches a VM extension to identify that this VM participates in an AKS cluster
+func (p *DefaultProvider) createAKSIdentifyingExtension(ctx context.Context, vmName string) (err error) {
```
matthchr, line 34: style opinion: I'm _really_ not a fan of this "split a struct across N files" pattern.

The code here is too coupled to code where the DefaultProvider is defined to really be standalone, and IMO all this serves is to hide when a struct is too big - i.e. you look at a module and see 7 nicely sized files and then realize "wait these are all just methods on a single uber-struct that does everything".

IMO file complexity/structure should match (as closely as possible) code complexity/structure. If you think this stuff deserves its own file (I don't disagree) then it also probably deserves its own structure. OTOH, if we don't want to do that, then IMO just put it in the instance.go file so we're honest about how much there is there.

pkg/providers/instance/vmextension.go
```diff
@@ -0,0 +1,115 @@
...
+	log.FromContext(ctx).V(1).Info(fmt.Sprintf("Created virtual machine CSE for %s, with an id of %s", vmName, *v.ID))
+	return nil
+}
+
+func (p *DefaultProvider) requiredVMExtensionsInstalled(vm armcompute.VirtualMachine) bool {
+	if len(vm.Resources) < 2 {
+		return false
+	}
+
+	var foundCSE bool // zero values are bool
```
matthchr, line 65: This comment doesn't really make sense. It might be cleaner for you to just do:
```
foundCSE := false
foundIdentitfying := false
```

it saves the need for what I think the comment is trying to express.

pkg/providers/instance/vmextension.go
```diff
@@ -0,0 +1,115 @@
...
+			foundCSE = true
+		}
+		if lo.FromPtr(extension.Name) == aksIdentifyingExtensionName {
+			foundIdentifying = true
+		}
+	}
+	return foundCSE && foundIdentifying
+}
+
+func (p *DefaultProvider) getAKSIdentifyingExtension() *armcompute.VirtualMachineExtension {
```
matthchr, line 81: minor, style: I find it's often cleaner to name these `new...` rather than `get...`, because `get...` can imply "go to Azure and check if it exists", whereas `new` is generally understood to be construction a local thing in memory.

pkg/cloudprovider/suite_test.go
Bryce-Soghigian: oh i wrote this test somewhere else let me remove it

pkg/providers/instance/vmextension.go
Bryce-Soghigian: For me the biggest benefit is when we have two people working on a change in the instance.go we don't have conflicts. Also you immediately know where to go to find the drift code in cloudprovider for example. 

Definitions are in vmextensions whereas invocations are in instance.go(AKA MEGA STRUCT). I was tempted to do what you are mentioning, but then I feel i would ahve to for both nic and vm.

PR #853, opened 2025-04-30
test: making startup taints and annotations for cilium conditional on the cilium ds being present

Bryce-Soghigian:
**Description**
When running our test suite against other cni configurations, it will fail. This change makes the configurations dynamic

- make az-all-cniv1 
- make az-e2etest

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/test/pkg/environment/common/expectations.go
+++ b/test/pkg/environment/common/expectations.go
@@ -1035,3 +1035,19 @@ func (env *Environment) GetDaemonSetOverhead(np *karpv1.NodePool) corev1.Resourc
 		return p, true
 	})...)
 }
+
+func (env *Environment) IsCilium() bool {
+	GinkgoHelper()
+
+	dsList := &appsv1.DaemonSetList{}
+	Expect(env.Client.List(env.Context, dsList,
+		client.InNamespace("kube-system"),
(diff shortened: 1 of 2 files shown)


test/pkg/environment/common/expectations.go
```diff
@@ -1035,3 +1035,19 @@ func (env *Environment) GetDaemonSetOverhead(np *karpv1.NodePool) corev1.Resourc
...
+		client.InNamespace("kube-system"),
+	)).To(Succeed())
+
+	for _, ds := range dsList.Items {
+		if strings.HasPrefix(ds.Name, "cilium") {
+			return true
+		}
+	}
+	return false
+}
```
Bryce-Soghigian, line 1053: We could do something more advanced, like passing in network_dataplane from cluster configuration, but i opted into this to add as little wiring as possible

comtalyst (approved):

Bryce-Soghigian:
https://github.com/Azure/karpenter-provider-azure/actions/runs/14792136186

Bryce-Soghigian:
Tests minus drift are passing now so merging, drift is also failing on master for the same case

PR #866, opened 2025-05-05
chore: update charts

tallaxes:
**Description**

Updates chart to the match the latest. Selected changes:
* Removal of webhooks

* `make az-run`

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


charts/karpenter/templates/role.yaml
```diff
@@ -14,19 +14,7 @@ rules:
   - apiGroups: ["coordination.k8s.io"]
     resources: ["leases"]
     verbs: ["get", "watch"]
-{{- if .Values.webhook.enabled }}
```
matthchr, line 17: @charliedmcb do we need to retain this permission for the MW case?

matthchr: If we arent toggling our _attempt_ to read the MW off (I dont think we are?) then we'll get an auth error if we dont have the perms.

charliedmcb: We will attempt to read the MW ConfigMap in all cases. However, its gated on one condition beforehand, which is the assumption the `SYSTEM_NAMESPACE` is set in the Env:
- [Reading SYSTEM_NAMESPACE](https://github.com/Azure/karpenter-provider-azure/blob/84e449787ec72268efb0c7af81ec87a6b3ee95fa/pkg/controllers/nodeclass/status/images.go#L68)
- [Gate on it being set](https://github.com/Azure/karpenter-provider-azure/blob/84e449787ec72268efb0c7af81ec87a6b3ee95fa/pkg/controllers/nodeclass/status/images.go#L181)

So, we wouldn't necessarily fail, unless if the `SYSTEM_NAMESPACE` is set. However, I agree keeping the perms would be best, regardless.

tallaxes: Added configmap read permission

charts/karpenter/values.yaml
```diff
@@ -88,22 +101,27 @@ tolerations:
     operator: Exists
 # -- Additional volumes for the pod.
 extraVolumes: []
-
 controller:
+  # -- Distinguishing container name (containerName: karpenter-controller).
```
charliedmcb, line 105: nit: why does the comment for `containerName` not match the value? Intended?

charts/karpenter/values.yaml
```diff
@@ -156,24 +165,30 @@ settings:
...
+  # node and pod affinities/anti-affinities and ScheduleAnyways topologySpreadConstraints. Can be one of 'Ignore' and 'Respect'
+  preferencePolicy: Respect
   # -- Cluster CA bundle for TLS configuration of provisioned nodes. If not set, this is taken from the controller's TLS configuration for the API server.
   clusterCABundle: ""
   # -- Cluster name.
   clusterName: ""
   # -- Cluster endpoint.
   clusterEndpoint: ""
-  # -- The VM memory overhead as a percent that will be subtracted from the total memory for all instance types
+  # -- The VM memory overhead as a percent that will be subtracted from the total memory for all instance types. The value of `0.075` equals to 7.5%.
```
charliedmcb, line 177: hyper-hit (matching other quotes):
```suggestion
  # -- The VM memory overhead as a percent that will be subtracted from the total memory for all instance types. The value of '0.075' equals to 7.5%.
```

charts/karpenter/templates/role.yaml
```diff
@@ -37,42 +28,3 @@ rules:
...
-    {{- toYaml . | nindent 4 }}
-  {{- end }}
-rules:
-  # Read
-  - apiGroups: ["coordination.k8s.io"]
-    resources: ["leases"]
-    verbs: ["get", "list", "watch"]
-  # Write
-  - apiGroups: ["coordination.k8s.io"]
-    resources: ["leases"]
```
charliedmcb, line 77: Why is this section being dropped? Webhook removal?

charliedmcb (approved):
Approving, as I don't see anything that looks blocking to me, though, few changes I don't fully understand. Would recommend getting another review if desired.

Left a few minor questions.

charts/karpenter/templates/_helpers.tpl
```diff
@@ -69,12 +69,18 @@ Karpenter image to use
 */}}
 {{- define "karpenter.controller.image" -}}
 {{- if .Values.controller.image.digest }}
-{{- printf "%s:%s@%s" .Values.controller.image.repository  (default (printf "v%s" .Chart.AppVersion) .Values.controller.image.tag) .Values.controller.image.digest }}
+{{- printf "%s:%s@%s" .Values.controller.image.repository  (default (printf "%s" .Chart.AppVersion) .Values.controller.image.tag) .Values.controller.image.digest }}
```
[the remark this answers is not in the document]
tallaxes: To make it semver compliant

PR #867, opened 2025-05-05
chore: add InClusterClient in operator.go

matthchr:
Can be used to access the kube-apiserver in the cluster where the Karpenter pod is running, for cases where the Karpenter pod is running in a different cluster than it is managing.

* UT and manually ran utilization E2E locally.

- [ ] Yes, issue opened: # 
- [x No

**Release Note**

```release-note
NONE
```

### DIFF
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -24,10 +24,13 @@ import (
 	"github.com/patrickmn/go-cache"
 	"github.com/samber/lo"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/transport"
+	"k8s.io/client-go/util/flowcontrol"
 	karpv1 "sigs.k8s.io/karpenter/pkg/apis/v1"
 	"sigs.k8s.io/karpenter/pkg/operator"
+	coreoptions "sigs.k8s.io/karpenter/pkg/operator/options"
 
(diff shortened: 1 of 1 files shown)


pkg/operator/operator.go
```diff
@@ -77,6 +85,12 @@ func NewOperator(ctx context.Context, operator *operator.Operator) (context.Cont
 		options.FromContext(ctx).VnetGUID = vnetGUID
 	}
 
+	// These options are set similarly to those used by operator.KubernetesInterface
+	inClusterConfig := lo.Must(rest.InClusterConfig())
+	inClusterConfig.RateLimiter = flowcontrol.NewTokenBucketRateLimiter(float32(coreoptions.FromContext(ctx).KubeClientQPS), coreoptions.FromContext(ctx).KubeClientBurst)
+	inClusterConfig.UserAgent = "karpenter"
```
tallaxes, line 91: Could use `GetUserAgentExtension` (which should probably be called GetUserAgent, and is being fixed in #819)

matthchr: Is that what we want for the kube-client useragent too? It's not what upstream Karpenter is currently using for its kube agent, it just uses `karpenter`

tallaxes: I think so. I can't think of a very good reason to omit the version. I can think of some (maybe ease of querying if one can't do partial matches, or to minimize request payload, etc.) - but they don't seem to be very convincing. I say we use the same - until we find out why this is wrong :-)

tallaxes (approved):

PR #873, opened 2025-05-07
feat: wait for CRDs

tallaxes:
**Description**

Add wait for CRDs on start, prior to starting core operator with timeout (2m) - though note that liveness probe, if not tuned, is likely to kill the process earlier.

##### Background

We had removed the alternative operator that waited for CRDs. The current core operator implementation ["fails open"](https://github.com/kubernetes-sigs/karpenter/pull/1683) on missing CRDs, proceeding without configuring indexers. Unfortunately, for now there are controller (including in core) that use caching client with `client.MatchingFields` option, which require indexes. These controllers will stay broken after CRDs are deployed, since there is nothing that would (re-)create the indexes then. On top of this, after CRDs are in place, it will stop exiting on CacheSyncTimeout (2m by default), so it will stay broken. Consequently (in the absence of something that would restart the process, or rebuild indexes), waiting for CRDs on startup is appropriate.

The implementation is similar to what core uses for readyz probe (probably introduced as a gate for webhooks).

* Starting w/CRDs in place
* Starting w/o CRDs in place
* Starting w/o CRDs in place and deploying them later

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
Wait for CRDs on startup
```

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -28,8 +28,6 @@ import (
 	"github.com/go-logr/zapr"
 	"github.com/samber/lo"
 
-	"k8s.io/client-go/kubernetes"
-
 	ctrl "sigs.k8s.io/controller-runtime"
 
 	"sigs.k8s.io/karpenter/pkg/cloudprovider/metrics"
@@ -44,8 +42,7 @@ import (
 func main() {
 	ctx := injection.WithOptionsOrDie(context.Background(), options.Injectables...)
 	logger := zapr.NewLogger(logging.NewLogger(ctx, "controller"))
-	client := kubernetes.NewForConfigOrDie(ctrl.GetConfigOrDie())
-	lo.Must0(operator.WaitForCRDs(ctx, client.DiscoveryClient, 2*time.Minute, logger), "failed waiting for CRDs")
+	lo.Must0(operator.WaitForCRDs(ctx, 2*time.Minute, ctrl.GetConfigOrDie(), logger), "failed waiting for CRDs")
 
 	ctx, op := operator.NewOperator(coreoperator.NewOperator())
 	aksCloudProvider := cloudprovider.New(
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -56,6 +56,8 @@ func main() {
 		op.ImageProvider,
 	)
 
+	lo.Must0(op.AddHealthzCheck("cloud-provider", aksCloudProvider.LivenessProbe))
+
 	cloudProvider := metrics.Decorate(aksCloudProvider)
 	clusterState := state.NewCluster(op.Clock, op.GetClient(), cloudProvider)
 
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -19,19 +19,34 @@ limitations under the License.
 package main
 
 import (
-	"github.com/samber/lo"
+	"context"
+	"time"
 
 	"github.com/Azure/karpenter-provider-azure/pkg/cloudprovider"
 	"github.com/Azure/karpenter-provider-azure/pkg/controllers"
 	"github.com/Azure/karpenter-provider-azure/pkg/operator"
+	"github.com/go-logr/zapr"
+	"github.com/samber/lo"
+
+	"k8s.io/client-go/kubernetes"
+
+	ctrl "sigs.k8s.io/controller-runtime"
 
 	"sigs.k8s.io/karpenter/pkg/cloudprovider/metrics"
 	corecontrollers "sigs.k8s.io/karpenter/pkg/controllers"
 	"sigs.k8s.io/karpenter/pkg/controllers/state"
 	coreoperator "sigs.k8s.io/karpenter/pkg/operator"
+	"sigs.k8s.io/karpenter/pkg/operator/injection"
+	"sigs.k8s.io/karpenter/pkg/operator/logging"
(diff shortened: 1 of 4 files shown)


pkg/operator/operator.go
```diff
@@ -213,3 +219,41 @@ func getVnetGUID(cfg *auth.Config, subnetID string) (string, error) {
 	}
 	return *vnet.Properties.ResourceGUID, nil
 }
+
+// WaitForCRDs waits for the required CRDs to be available with a timeout
+func WaitForCRDs(ctx context.Context, client *discovery.DiscoveryClient, timeout time.Duration, log logr.Logger) error {
+	var requiredCRDs = []schema.GroupVersionResource{
```
matthchr, line 225: I think you could avoid hardcoding these and instead just reference the scheme?

https://github.com/kubernetes-sigs/karpenter/blob/93c0ef3a4d02bfc1124533a5661d19b9373dfb89/pkg/apis/v1/doc.go#L32

and 

for us at: pkg/apis/v1alpha2/doc.go

There's likely other stuff in scheme too, but you could filter to the `karpenter.sh` and `karpenter.azure.com` groups via `scheme.Scheme.ObjectKinds()` (though note that returns `Kind` rather than `Resource` so the format changes a bit, but it accomplishes the same thing I think.

matthchr: Alternatively you could possibly leave it hardcoded here but write a test (possibly using the scheme above) that ensures we don't have new versions and/or new resources sneak in but not wait for them?

pkg/operator/operator.go
```diff
@@ -213,3 +219,41 @@ func getVnetGUID(cfg *auth.Config, subnetID string) (string, error) {
...
+	}
+
+	log.WithValues("timeout", timeout).Info("waiting for required CRDs to be available")
+	ctx, cancel := context.WithTimeout(ctx, timeout)
+	defer cancel()
+
+	for _, crd := range requiredCRDs {
+		crdName := fmt.Sprintf("%s.%s/%s", crd.Resource, crd.GroupVersion().Group, crd.Version)
+		if err := wait.PollUntilContextCancel(ctx, 10*time.Second, true, func(ctx context.Context) (bool, error) {
+			_, err := client.RESTClient().Get().AbsPath("/apis", crd.Group, crd.Version, crd.Resource).Do(ctx).Raw()
```
matthchr, line 238: I think this deserves a comment for why you're avoiding the full `/apis` and using the RESTClient. (I assume to avoid the full /apis list which can be really expensive)

In fact you could probably use `ServerResourcesForGroupVersion` which would still avoid the full /apis query without needing to resort to the raw client. I don't know that this is way clearer though (it might even be more code?) so I'm fine not using it, but think a comment should be here.

pkg/operator/operator.go
```diff
@@ -213,3 +219,41 @@ func getVnetGUID(cfg *auth.Config, subnetID string) (string, error) {
...
+		{Group: "karpenter.azure.com", Version: "v1alpha2", Resource: "aksnodeclasses"},
+	}
+
+	log.WithValues("timeout", timeout).Info("waiting for required CRDs to be available")
+	ctx, cancel := context.WithTimeout(ctx, timeout)
+	defer cancel()
+
+	for _, crd := range requiredCRDs {
+		crdName := fmt.Sprintf("%s.%s/%s", crd.Resource, crd.GroupVersion().Group, crd.Version)
+		if err := wait.PollUntilContextCancel(ctx, 10*time.Second, true, func(ctx context.Context) (bool, error) {
```
matthchr, line 237: minor, style: maybe just assign `err` here and check it on the subsequent line to avoid this:
`}); err != nil {`

not a big deal though.

cmd/controller/main.go
```diff
@@ -41,7 +56,6 @@ func main() {
 		op.ImageProvider,
 	)
 
-	lo.Must0(op.AddHealthzCheck("cloud-provider", aksCloudProvider.LivenessProbe))
```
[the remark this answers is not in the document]
tallaxes: No, will restore (I don't think we need it, but let's separate this)

pkg/operator/operator.go
tallaxes: Fixed

pkg/operator/operator.go
tallaxes: There was nothing built-in that works nicely, but it looks like RESTMapper from controller-runtime client does what we are looking for, take a look

pkg/operator/operator.go
tallaxes: Take a look at the version with RESTMapper

matthchr: Updated RESTMapper is nice but doesn't actually solve what I was getting at here which was: we're hardcoding the CRD list here that we want to watch which means we need to update it when a new CRD (or a new version) is added, and that would be easy to miss because it would _mostly_ work (since CRDs are generally applied together) but would be incorrect.

Sourcing from scheme with a filter, or having a test, protects against future evolution where we add CRDs or versions and forget to add them here.

matthchr (approved):

pkg/operator/operator.go
tallaxes: Agree on the value, ran into some minor issues when filtering from Scheme, for expediency will merge this for now and circle back to this later.

PR #874, opened 2025-05-09
feat: Replacing fmt errors for cloudProvider.Create with corresponding cloudprovider CreateError types.

### LINKED ISSUE #858: Have `cloudprovider.Create` return `clouprovider.CreateError` on failures
`cloudprovider.CreateError` has better fidelity `ConditionReason` and `ConditionMessage`, which are also not subject to message length constraints ([evidence](https://github.com/kubernetes-sigs/karpenter/blob/9764a3a96a953bd58f60ab4416d43ac88d8eda8f/pkg/controllers/nodeclaim/lifecycle/launch.go#L105-L108)). One could worry about potential etcd storage impact at scale, but the transparency is likely worth it. And we can constrain it better later.

aagusuab:
Fixes #858

**Description**
Replacing the error messages for Create command in CloudProvder, so that the error messages aren't truncated to 300 characters.

The new cloudprovider.Error types will keep the error messages in conditions, allowing up to around 23000 characters to be kept. We will truncate this to 1200 for now.

* Manual Testing / Unit Testing
* E2E: https://github.com/Azure/karpenter-provider-azure/actions/runs/15057585120 ✅ (one integration failure does not appear to be related)

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note

```


pkg/cloudprovider/cloudprovider.go
```diff
@@ -109,22 +109,22 @@ func (c *CloudProvider) Create(ctx context.Context, nodeClaim *karpv1.NodeClaim)
...
-		return nil, fmt.Errorf("resolving instance types, %w", err)
+		return nil, cloudprovider.NewInsufficientCapacityError(fmt.Errorf("resolving instance types, %w", err))
 	}
 	if len(instanceTypes) == 0 {
 		return nil, cloudprovider.NewInsufficientCapacityError(fmt.Errorf("all requested instance types were unavailable during launch"))
 	}
 	instancePromise, err := c.instanceProvider.BeginCreate(ctx, nodeClass, nodeClaim, instanceTypes)
 	if err != nil {
-		return nil, fmt.Errorf("creating instance, %w", err)
+		return nil, cloudprovider.NewInsufficientCapacityError(fmt.Errorf("creating instance, %w", err))
```
tallaxes, line 127: This won't necessarily be insufficient capacity, should be a more generic error

pkg/cloudprovider/cloudprovider.go
```diff
@@ -109,22 +109,22 @@ func (c *CloudProvider) Create(ctx context.Context, nodeClaim *karpv1.NodeClaim)
 		return nil, cloudprovider.NewNodeClassNotReadyError(stderrors.New(nodeClassReady.Message))
 	}
 	if nodeClassReady.IsUnknown() {
-		return nil, fmt.Errorf("resolving NodeClass readiness, NodeClass is in Ready=Unknown, %s", nodeClassReady.Message)
+		return nil, cloudprovider.NewNodeClassNotReadyError(fmt.Errorf("resolving NodeClass readiness, NodeClass is in Ready=Unknown, %s", nodeClassReady.Message))
```
tallaxes, line 112: This should be just `cloudprovider.NewCreateError` (does not neccesarily mean it is not ready)

pkg/cloudprovider/cloudprovider.go
```diff
@@ -109,22 +109,22 @@ func (c *CloudProvider) Create(ctx context.Context, nodeClaim *karpv1.NodeClaim)
...
 	}
 	if _, err = nodeClass.GetKubernetesVersion(); err != nil {
 		return nil, err
 	}
 
 	instanceTypes, err := c.resolveInstanceTypes(ctx, nodeClaim, nodeClass)
 	if err != nil {
-		return nil, fmt.Errorf("resolving instance types, %w", err)
+		return nil, cloudprovider.NewInsufficientCapacityError(fmt.Errorf("resolving instance types, %w", err))
```
tallaxes, line 120: Error here does not neccesarily mean insufficient cacpacity, should use `cloudprovider.NewCreateError`

tallaxes (changes requested):
Left some comments, please take a look. Consider using AWS provider [implementation](https://github.com/aws/karpenter-provider-aws/blob/12d7d5a3aa7b0b1d45877ec8b7fa9a7037d24fa4/pkg/cloudprovider/cloudprovider.go#L90-L137) for reference. It looks like we need to update NodeClass resolution code as well (only return `NewInsufficientCapacityError` if NodeClass is not found, and return a regular error otherwise.

pkg/cloudprovider/cloudprovider.go
aagusuab: Changed this to be the NewCreateError.
Added Reason Constant to be more specific.

pkg/cloudprovider/cloudprovider.go
aagusuab: Changed this to be the NewCreateError.
Added Reason Constant to be more specific.

pkg/cloudprovider/cloudprovider.go
aagusuab: Changed this to be the NewCreateError.
Added Reason Constant to be more specific.

pkg/test/environment.go
```diff
@@ -191,6 +191,7 @@ func (env *Environment) Reset() {
 	env.PricingAPI.Reset()
 	env.PricingProvider.Reset()
 	env.ImageProvider.Reset()
+	env.MockSkuClientSignalton.SKUClient.Reset()
```
tallaxes, line 194:
```suggestion
	env.MockSkuClientSignleton.SKUClient.Reset()
```
Looks like an old type (would need fixing in the struct as well)

pkg/providers/instancetype/suite_test.go
```diff
@@ -1459,6 +1460,139 @@ var _ = Describe("InstanceType Provider", func() {
...
+			fmt.Printf("  Message: %s\n", nodeClassReady.Message)
+			fmt.Printf("  LastTransitionTime: %v\n", nodeClassReady.LastTransitionTime)
+			fmt.Printf("  ObservedGeneration: %d\n", nodeClassReady.ObservedGeneration)
+
+			fmt.Printf("NodeClassReady is %+v", nodeClassReady.IsUnknown())
+
+			claim, err := cloudProvider.Create(ctx, nodeClaim)
+			Expect(err).To(HaveOccurred())
+			Expect(claim).To(BeNil())
+			Expect(err.Error()).To(ContainSubstring("resolving NodeClass readiness, NodeClass is in Ready=Unknown"))
```
tallaxes, line 1524: Consider checking for err type (here and elsewhere), since that's what this PR implements

pkg/cloudprovider/cloudprovider.go
```diff
@@ -109,7 +110,7 @@ func (c *CloudProvider) Create(ctx context.Context, nodeClaim *karpv1.NodeClaim)
 		return nil, cloudprovider.NewNodeClassNotReadyError(stderrors.New(nodeClassReady.Message))
 	}
 	if nodeClassReady.IsUnknown() {
-		return nil, fmt.Errorf("resolving NodeClass readiness, NodeClass is in Ready=Unknown, %s", nodeClassReady.Message)
+		return nil, cloudprovider.NewCreateError(fmt.Errorf("resolving NodeClass readiness, NodeClass is in Ready=Unknown, %s", nodeClassReady.Message), consts.NodeUnreadyReason, nodeClassReady.Message)
```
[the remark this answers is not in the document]
aagusuab: Correcting the error messages / error constant.

pkg/cloudprovider/cloudprovider.go
```diff
@@ -124,14 +125,14 @@ func (c *CloudProvider) Create(ctx context.Context, nodeClaim *karpv1.NodeClaim)
 
 	instanceTypes, err := c.resolveInstanceTypes(ctx, nodeClaim, nodeClass)
 	if err != nil {
-		return nil, fmt.Errorf("resolving instance types, %w", err)
+		return nil, cloudprovider.NewCreateError(fmt.Errorf("resolving instance types, %w", err), consts.InstanceTypeNotResolvedReason, err.Error())
```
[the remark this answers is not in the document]
aagusuab: Fixed it to be more precise and matching the example.

pkg/cloudprovider/cloudprovider.go
```diff
@@ -124,14 +125,14 @@ func (c *CloudProvider) Create(ctx context.Context, nodeClaim *karpv1.NodeClaim)
...
 	}
 	if len(instanceTypes) == 0 {
 		return nil, cloudprovider.NewInsufficientCapacityError(fmt.Errorf("all requested instance types were unavailable during launch"))
 	}
 	instancePromise, err := c.instanceProvider.BeginCreate(ctx, nodeClass, nodeClaim, instanceTypes)
 	if err != nil {
-		return nil, fmt.Errorf("creating instance, %w", err)
+		return nil, cloudprovider.NewCreateError(fmt.Errorf("creating instance, %w", err), consts.CreateInstanceFailedReason, err.Error())
 	}
```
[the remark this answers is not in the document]
aagusuab: Will update this to at least say "instance failed" for now. Definitely good idea to make more specific errors in teh future.

pkg/test/environment.go
aagusuab: Fixed the struct as well.

pkg/providers/instancetype/suite_test.go
aagusuab: Added checking for error types (of cloudprovider.CreateError)

pkg/cloudprovider/cloudprovider.go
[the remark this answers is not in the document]
tallaxes: AWS provider just uses "NodeClass is in Ready=Unknown" for the message. Do you have a good argument for propagating nodeClassReady.Message instead?

pkg/cloudprovider/cloudprovider.go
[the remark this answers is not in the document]
tallaxes: Not convinced that  err.Error() has much value here over just "Error resolving instance types", but willing to wait and see.

pkg/cloudprovider/cloudprovider.go
```diff
@@ -109,7 +110,7 @@ func (c *CloudProvider) Create(ctx context.Context, nodeClaim *karpv1.NodeClaim)
 		return nil, cloudprovider.NewNodeClassNotReadyError(stderrors.New(nodeClassReady.Message))
 	}
 	if nodeClassReady.IsUnknown() {
-		return nil, fmt.Errorf("resolving NodeClass readiness, NodeClass is in Ready=Unknown, %s", nodeClassReady.Message)
+		return nil, cloudprovider.NewCreateError(fmt.Errorf("resolving NodeClass readiness, NodeClass is in Ready=Unknown, %s", nodeClassReady.Message), consts.NodeClassReadinessUnknownReason, nodeClassReady.Message)
```
tallaxes, line 113: `consts.NodeClassReadinessUnknownReason` (and other new consts) are only used here. I don't see a good reason to tuck them away in const package. Would suggest moving them in here (or even getting rid of the const altogether).

pkg/cloudprovider/cloudprovider.go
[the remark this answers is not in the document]
tallaxes: I am a little worried about not constraining the string size here _at all_. Some of the CRP etc. errors could be quite long. At the same time, we do want to capture the main cause, and fishing it out of the error may be tricky or impossible. @matthchr, @Bryce-Soghigian - thoughts?

pkg/cloudprovider/cloudprovider.go
aagusuab: Moved all of the consts to the cloudprovider.go. (Didn't delete it for now, since there were other unrelated const variables) 

Later when we get too much error reasons, we could move them to a separate error consts file just to keep track of them, but that's in the far future from what I can see.

pkg/cloudprovider/cloudprovider.go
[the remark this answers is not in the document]
aagusuab: Kept it as NodeClass is in Ready=Unknown.
I was thinking since this was already being repeated in the error field, nodeClassReady.Message may include more information if anything. (this really just returns some {condition: boolean})

pkg/cloudprovider/cloudprovider.go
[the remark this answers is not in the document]
aagusuab: Naive solution is setting the limit to be let's say... 2000? (Previous kubernetes event default being 500). 
It really depends on how wrapped the errors would be.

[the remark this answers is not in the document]
matthchr: IIRC I added cloudprovider/events/events.go `truncateMessage` to truncate some messages from CRP at 500 characters (which is more than the 300 upstream does which annoyed both Bryce and I because it's not enough to get useful information).

Really what we ought to do I _think_ here define an error processor that forces the relevant bits to the top so that if we truncate it's truncating something less useful.

See https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azcore/internal/exported/response_error.go#L127 which puts the HTTP details first (which is annoying because you probably want the code/message first and the HTTP details are really long). We can just use `errors.As` in a helper and reorder things to be more user-friendly by putting code/message first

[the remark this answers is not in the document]
matthchr: And we probably don't want the `-------` separator they have because it just eats space and doesn't lay out all that well in events anyway?

I guess they don't actually even try to parse the error into code/message/details either which is interesting... but probably because services don't all follow the recommended structure for errors (I learned that AKS does _not_ earlier this year)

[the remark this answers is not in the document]
tallaxes: Ok, how about this: for now we truncate to 500, and get this PR in. Then follow-up with checking for errors.As(ResponseError) and teasing out the compact version with important bits first in that case. WDYT?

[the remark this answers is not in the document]
Bryce-Soghigian: We can also query the existing errors and check for their length we have seen returned from azure. 

Then use that as a baseline truncation number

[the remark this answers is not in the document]
aagusuab: as per discussion from the teams, Will truncate to 1200 characters for now. 
(Bryce found the maximum error to be around 7140 characters long)

pkg/cloudprovider/cloudprovider.go
```diff
@@ -124,14 +130,14 @@ func (c *CloudProvider) Create(ctx context.Context, nodeClaim *karpv1.NodeClaim)
 
 	instanceTypes, err := c.resolveInstanceTypes(ctx, nodeClaim, nodeClass)
 	if err != nil {
-		return nil, fmt.Errorf("resolving instance types, %w", err)
+		return nil, cloudprovider.NewCreateError(fmt.Errorf("resolving instance types, %s", truncateMessage(err.Error())), InstanceTypeResolutionFailedReason, err.Error())
```
tallaxes, line 133: Here and below: Please truncate the message (last argument to NewCreateError) and not the error (first argument). The actual error passed is not even used when CreatedError is returned ([evidence](https://github.com/kubernetes-sigs/karpenter/blob/7bcbabf822b2ed2a1eab8750763f7e0c405874f4/pkg/controllers/nodeclaim/lifecycle/launch.go#L105-L106))

```suggestion
		return nil, cloudprovider.NewCreateError(fmt.Errorf("resolving instance types, %s", err.Error()), InstanceTypeResolutionFailedReason, truncateMessage(err.Error()))
```

aagusuab: Fixed all 3 to be truncating the last argument. (Except the first one, the first one doesn't really have a full error)

PR #883, opened 2025-05-13
feat(MW): add support for Maintenance Window NodeOS channel

charliedmcb:
Fixes # 

**Description**
Adding support for the `aksManagedNodeOSUpgradeSchedule` Maintenance Window channel for Managed Karpenter.

- E2E runs:
    - https://github.com/Azure/karpenter-provider-azure/actions/runs/15008912813
    - https://github.com/Azure/karpenter-provider-azure/actions/runs/15028990798
    - https://github.com/Azure/karpenter-provider-azure/actions/runs/15032503198 (latest)

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
Node Image Upgrades will respect the "aksManagedNodeOSUpgradeSchedule" Maintenance Window channel for Managed Karpenter.
```


pkg/operator/options/options.go
```diff
@@ -106,6 +107,7 @@ func (o *Options) AddFlags(fs *coreoptions.FlagSet) {
 	fs.StringVar(&o.SIGAccessTokenServerURL, "sig-access-token-server-url", env.WithDefaultString("SIG_ACCESS_TOKEN_SERVER_URL", ""), "The URL for the SIG access token server. Only used for AKS managed karpenter. UseSIG must be set tot true for this to take effect.")
 	fs.StringVar(&o.SIGAccessTokenScope, "sig-access-token-scope", env.WithDefaultString("SIG_ACCESS_TOKEN_SCOPE", ""), "The scope for the SIG access token. Only used for AKS managed karpenter. UseSIG must be set to true for this to take effect.")
 	fs.StringVar(&o.SIGSubscriptionID, "sig-subscription-id", env.WithDefaultString("SIG_SUBSCRIPTION_ID", ""), "The subscription ID of the shared image gallery.")
+	fs.BoolVar(&o.AKSControlPlane, "aks-control-plane", env.WithDefaultBool("AKS_CONTROL_PLANE", false), "Used to signal that Karpenter is running in Managed AKS, as NAP, with certain features only currently available in that environment.") // Note (charliedmcb): I think this should be more explicate about NAP only, as certain features enabled this way will not work in self-hosted at the moment.
```
charliedmcb, line 110: I'm still questioning the naming of this Flag. I feel like it should be more explicate about being for Managed only features.

pkg/controllers/nodeclass/status/images.go
```diff
@@ -147,9 +171,50 @@ func imageVersionsUnready(nodeClass *v1alpha2.AKSNodeClass) bool {
...
 // Handles case 4: check if the maintenance window is open
-func isMaintenanceWindowOpen() bool {
-	// TODO: need to add the actual logic for handling maintenance windows once it is in the ConfigMap.
-	return true
+func (r *NodeImageReconciler) isMaintenanceWindowOpen(ctx context.Context) (bool, error) {
+	if !r.aksControlPlane {
+		// Note: when aksControlPlane is false, continuous upgrade is enabled without checking the maintenane windows.
+		return true, nil
+	}
+	systemNamespace := strings.TrimSpace(os.Getenv("SYSTEM_NAMESPACE"))
```
charliedmcb, line 179: Is pulling this directly from the Env here best, or should we do it at an earlier point/in a different way?

charliedmcb: I do somewhat prefer doing it here, as if it is missing from the env we still get come update paths, and will only fail when checking the maintenance windows.

cmd/controller/main.go
```diff
@@ -78,6 +78,7 @@ func main() {
 			// TODO: refactor so we are passing in kubernetesVersionProvider and nodeImageProvider. Currently ImageProvider just implements both.
 			op.ImageProvider,
 			op.ImageProvider,
+			op.InClusterKubernetesInterface,
```
tallaxes, line 81: Don't forget to update main_ccp (we will get rid of it soon)

pkg/controllers/nodeclass/status/images.go
tallaxes: Reading from environment on every call seems unnecesary. Whatever logic you need when it is not set can still be used if it is initialized once.

pkg/controllers/nodeclass/status/images.go
```diff
@@ -147,9 +171,54 @@ func imageVersionsUnready(nodeClass *v1alpha2.AKSNodeClass) bool {
...
+	if systemNamespace == "" {
+		return false, fmt.Errorf("SYSTEM_NAMESPACE not set")
+	}
+
+	mwConfigMap, err := r.inClusterKubernetesInterface.CoreV1().ConfigMaps(systemNamespace).Get(ctx, "upcoming-maintenance-window", metav1.GetOptions{})
```
tallaxes, line 188: A constant for ConfigMap name?

pkg/controllers/nodeclass/status/images.go
```diff
@@ -147,9 +171,54 @@ func imageVersionsUnready(nodeClass *v1alpha2.AKSNodeClass) bool {
...
 
 // Handles case 4: check if the maintenance window is open
-func isMaintenanceWindowOpen() bool {
-	// TODO: need to add the actual logic for handling maintenance windows once it is in the ConfigMap.
-	return true
+// TODO (charliedmcb): remove nolint on gocyclo. Added for now in order to pass "make verify"
+// I think the best way to get rid of gocyclo is to break the section retrieving the maintenance window
+// range from the ConfigMap into its own helper function using channel as a parameter.
+// nolint: gocyclo
+func (r *NodeImageReconciler) isMaintenanceWindowOpen(ctx context.Context) (bool, error) {
```
tallaxes, line 178: It feels like some form of caching would be appropriate here, but it is probably fine to revisit this later.
(I am assuming getting ConfigMap on every call is Ok for now - as long as we use caching client.)

pkg/controllers/nodeclass/status/images_test.go
```diff
@@ -98,13 +136,11 @@ var _ = Describe("NodeClass NodeImage Status Controller", func() {
...
-		Expect(len(nodeClass.Status.Images)).To(Equal(3))
-		Expect(nodeClass.Status.Images).To(HaveExactElements(getExpectedTestCommunityImages(newCIGImageVersion)))
-		Expect(nodeClass.StatusConditions().IsTrue(v1alpha2.ConditionTypeImagesReady)).To(BeTrue())
+		ExpectReadyWithCIGImages(nodeClass, newCIGImageVersion)
 	})
 
-	It("should update Images and its readiness on AKSNodeClass when in an open maintenance window", func() {
-		// TODO: once maintenance window support is added we need to actually add test code here causing it to be open.
+	It("should update Images and its readiness on AKSNodeClass when aksControlPlane is false", func() {
+		// Note: when aksControlPlane is false, continuous upgrade is enabled without checking the maintenane windows.
```
tallaxes, line 143: "maintenance" misspelled (here and elsewhere, at least in this file)

tallaxes (dismissed):

pkg/controllers/nodeclass/status/images.go
charliedmcb: Good callout. Moved to the constructor, and simply preforms the same logic with the stored value, instead of getting it from the env each time.

pkg/operator/options/options.go
charliedmcb: No longer important, since flag is removed.

pkg/controllers/nodeclass/status/images.go
charliedmcb: So, this is actually an instance where I think we need to be rechecking this about every 5 minutes, to catch updated to it. Another pattern we were considering was to look at setting up a watch, but we want this catch updates pretty quickly.

tallaxes (approved):

pkg/controllers/nodeclass/status/images.go
```diff
@@ -42,17 +46,32 @@ import (
...
+	nodeImageProvider            imagefamily.NodeImageProvider
+	inClusterKubernetesInterface kubernetes.Interface
+	systemNamespace              string
+	cm                           *pretty.ChangeMonitor
 }
 
-func NewNodeImageReconciler(provider imagefamily.NodeImageProvider) *NodeImageReconciler {
+func NewNodeImageReconciler(
+	provider imagefamily.NodeImageProvider,
+	inClusterKubernetesInterface kubernetes.Interface,
```
matthchr, line 66: I almost wonder if we want this to be a MaintenanceWindow provider? It could abstract the configmap-ness away, which feels good given that nodeclass/status doesn't feel like the right place for the MW logic to live - i.e. fail open or fail closed, etc.

matthchr: Plus it better matches the pattern of passing providers around not passing raw clients?

pkg/controllers/nodeclass/status/images.go
matthchr: We can (and possibly _should_?) set up a watch to just tell us when this configmap is updated and trigger a reconcile?

You can see how ASO does this here: https://github.com/Azure/azure-service-operator/blob/main/v2/internal/controllers/controller_resources.go#L343

matthchr: Ok I see we have requeue every 5m so it should be fine, though it might one day be better to just have this be a watch on the configmap I'm not super hung up on it (it gets you caching too which I guess is nice)

matthchr: also this is tricky w/ the fact we have 2 API servers in some cases so the controller-runtime clients are all pointed at the overlay one, so a watch via controller-runtime is actually harder than it sounds here.

pkg/controllers/nodeclass/status/images.go
```diff
@@ -147,9 +173,50 @@ func imageVersionsUnready(nodeClass *v1alpha2.AKSNodeClass) bool {
...
+	}
+	if len(mwConfigMap.Data) == 0 {
+		// An empty configmap means there's no maintenance windows defined, and its up to us when to preform maintenance
+		return true, nil
+	}
+
+	nextNodeOSMWStartStr, okStart := mwConfigMap.Data[fmt.Sprintf(configMapStartTimeFormat, nodeOSMaintenanceWindowChannel)]
+	nextNodeOSMWEndStr, okEnd := mwConfigMap.Data[fmt.Sprintf(configMapEndTimeFormat, nodeOSMaintenanceWindowChannel)]
+	if !okStart && !okEnd {
+		// No maintenance window defined for aksManagedNodeOSUpgradeSchedule, so its up to us when to preform maintenance
```
matthchr, line 202:
```suggestion
		// No maintenance window defined for aksManagedNodeOSUpgradeSchedule, so its up to us when to perform maintenance
```

matthchr: and elsewhere too =)

pkg/controllers/nodeclass/status/images.go
```diff
@@ -147,9 +173,50 @@ func imageVersionsUnready(nodeClass *v1alpha2.AKSNodeClass) bool {
...
+	nextNodeOSMWStart, err := time.Parse(time.RFC3339, nextNodeOSMWStartStr)
+	if err != nil {
+		return false, fmt.Errorf("error parsing maintenance window start time for channel %s, %w", nodeOSMaintenanceWindowChannel, err)
+	}
+	nextNodeOSMWEnd, err := time.Parse(time.RFC3339, nextNodeOSMWEndStr)
+	if err != nil {
+		return false, fmt.Errorf("error parsing maintenance window end time for channel %s, %w", nodeOSMaintenanceWindowChannel, err)
+	}
+
+	now := time.Now().UTC()
```
matthchr, line 217: minor: A trick I've often seen used, and definitely recommend, is having some sort of `TimeProvider` rather than just reading system time directly here.

Probably not worth fixing here because we call time.Now() in other places too, but generally being able to swap your time provider out for unit testing can be very helpful. Curious what @tallaxes thinks about this... maybe a low pri future improvement...

matthchr (approved):

PR #886, opened 2025-05-13
chore: move KubeletClientID and NodeResourceGroup from Config to Options

matthchr:
We should probably move other stuff and/or reconcile what the difference between options and config is at some point, but for now just reducing the duplicate stuff and we can worry about the rest later.

Fixes # 

**Description**

* Unit tests - can run E2Es if we think it's a good idea.

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
NONE
```


tallaxes (approved):
LGTM

PR #888, opened 2025-05-14
feat: drift for Kubelet Client ID

matthchr:
* Add kubernetes.azure.com/kubelet-identity-client-id label to all nodes, to match what AKS does.
* Drift nodes if the expected kubelet-identity-client-id has changed.
* Does not drift if the node doesn't have the AKS Kubelet ClientID label set.

* Unit tests
* Manual test with NPS bootstrap, shows nodes now come up with this label: `kubernetes.azure.com/kubelet-identity-client-id=fcd9a498-b00b-4757-8c2a-90f6ede47610`. Changing it at the configmap scope results in drift and the new node coming up with the new label.

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
Added the kubernetes.azure.com/kubelet-identity-client-id to all nodes and drifts nodes if the expected kubelet-identity-client-id has changed.
```

### DIFF
--- a/pkg/cloudprovider/drift.go
+++ b/pkg/cloudprovider/drift.go
@@ -23,12 +23,14 @@ import (
 
 	sdkerrors "github.com/Azure/azure-sdk-for-go-extensions/pkg/errors"
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/network/armnetwork"
+	"github.com/samber/lo"
+	v1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+
 	"github.com/Azure/karpenter-provider-azure/pkg/apis/v1beta1"
 	"github.com/Azure/karpenter-provider-azure/pkg/operator/options"
 	"github.com/Azure/karpenter-provider-azure/pkg/providers/instance"
 	"github.com/Azure/karpenter-provider-azure/pkg/utils"
-	"github.com/samber/lo"
-	"sigs.k8s.io/controller-runtime/pkg/log"
 
 	karpv1 "sigs.k8s.io/karpenter/pkg/apis/v1"
 	"sigs.k8s.io/karpenter/pkg/cloudprovider"
@@ -36,58 +38,61 @@ import (
 )
 
 const (
-	NodeClassDrift  cloudprovider.DriftReason = "NodeClassDrift"
-	K8sVersionDrift cloudprovider.DriftReason = "K8sVersionDrift"
-	ImageDrift      cloudprovider.DriftReason = "ImageDrift"
-	SubnetDrift     cloudprovider.DriftReason = "SubnetDrift"
+	NodeClassDrift       cloudprovider.DriftReason = "NodeClassDrift"
+	K8sVersionDrift      cloudprovider.DriftReason = "K8sVersionDrift"
+	ImageDrift           cloudprovider.DriftReason = "ImageDrift"
+	SubnetDrift          cloudprovider.DriftReason = "SubnetDrift"
+	KubeletIdentityDrift cloudprovider.DriftReason = "KubeletIdentityDrift"
 
 	// TODO (charliedmcb): Use this const across code and test locations which are signaling/checking for "no drift"
 	NoDrift cloudprovider.DriftReason = ""
 )
 
 func (c *CloudProvider) isNodeClassDrifted(ctx context.Context, nodeClaim *karpv1.NodeClaim, nodeClass *v1beta1.AKSNodeClass) (cloudprovider.DriftReason, error) {
-	// First check if the node class is statically staticFieldsDrifted to save on API calls.
-	if staticFieldsDrifted := c.areStaticFieldsDrifted(nodeClaim, nodeClass); staticFieldsDrifted != "" {
-		return staticFieldsDrifted, nil
+	// TODO: if we find more expensive checks, such as reading VMs or NICs from Azure, are being duplicated between checks, we should
+	//       produce a lazy at-most-once that allows a check to cache a value for later checks to read.
+	checks := []func(ctx context.Context, nodeClaim *karpv1.NodeClaim, nodeClass *v1beta1.AKSNodeClass) (cloudprovider.DriftReason, error){
+		c.areStaticFieldsDrifted,
+		c.isK8sVersionDrifted,
+		c.isKubeletIdentityDrifted,
+		c.isImageVersionDrifted,
+		c.isSubnetDrifted,
 	}
-	k8sVersionDrifted, err := c.isK8sVersionDrifted(ctx, nodeClaim, nodeClass)
-	if err != nil {
-		return "", err
+	for _, check := range checks {
+		driftReason, err := check(ctx, nodeClaim, nodeClass)
+		if err != nil {
+			return "", err
+		}
+		if driftReason != "" {
+			return driftReason, nil
+		}
 	}
-	if k8sVersionDrifted != "" {
-		return k8sVersionDrifted, nil
-	}
-	imageVersionDrifted, err := c.isImageVersionDrifted(ctx, nodeClaim, nodeClass)
-	if err != nil {
-		return "", err
-	}
-	if imageVersionDrifted != "" {
-		return imageVersionDrifted, nil
-	}
-	subnetDrifted, err := c.isSubnetDrifted(ctx, nodeClaim, nodeClass)
-	if err != nil {
-		return "", err
-	}
-	if subnetDrifted != "" {
-		return subnetDrifted, nil
-	}
+
 	return "", nil
 }
 
-func (c *CloudProvider) areStaticFieldsDrifted(nodeClaim *karpv1.NodeClaim, nodeClass *v1beta1.AKSNodeClass) cloudprovider.DriftReason {
+func (c *CloudProvider) areStaticFieldsDrifted(ctx context.Context, nodeClaim *karpv1.NodeClaim, nodeClass *v1beta1.AKSNodeClass) (cloudprovider.DriftReason, error) {
+	logger := log.FromContext(ctx)
+
 	nodeClassHash, foundNodeClassHash := nodeClass.Annotations[v1beta1.AnnotationAKSNodeClassHash]
 	nodeClassHashVersion, foundNodeClassHashVersion := nodeClass.Annotations[v1beta1.AnnotationAKSNodeClassHashVersion]
 	nodeClaimHash, foundNodeClaimHash := nodeClaim.Annotations[v1beta1.AnnotationAKSNodeClassHash]
 	nodeClaimHashVersion, foundNodeClaimHashVersion := nodeClaim.Annotations[v1beta1.AnnotationAKSNodeClassHashVersion]
 
 	if !foundNodeClassHash || !foundNodeClaimHash || !foundNodeClassHashVersion || !foundNodeClaimHashVersion {
-		return ""
+		return "", nil
 	}
 	// validate that the hash version for the AKSNodeClass is the same as the NodeClaim before evaluating for static drift
 	if nodeClassHashVersion != nodeClaimHashVersion {
-		return ""
+		return "", nil
 	}
-	return lo.Ternary(nodeClassHash != nodeClaimHash, NodeClassDrift, "")
+
+	if nodeClassHash != nodeClaimHash {
+		logger.V(1).Info(fmt.Sprintf("drift triggered for %s, as nodeClassHash (%s) != nodeClaimHash (%s)", NodeClassDrift, nodeClassHash, nodeClaimHash))
+		return NodeClassDrift, nil
+	}
+
+	return "", nil
 }
 
 func (c *CloudProvider) isK8sVersionDrifted(ctx context.Context, nodeClaim *karpv1.NodeClaim, nodeClass *v1beta1.AKSNodeClass) (cloudprovider.DriftReason, error) {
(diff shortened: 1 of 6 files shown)


pkg/cloudprovider/suite_test.go
```diff
@@ -280,6 +282,23 @@ var _ = Describe("CloudProvider", func() {
 				Expect(err).ToNot(HaveOccurred())
 				Expect(drifted).To(Equal(ImageDrift))
 			})
+
+			It("should NOT trigger drift if node doesn't have kubelet client ID label", func() {
```
matthchr, line 286: Have UTs but don't have an E2E. It is probably possible to write an E2E but it would require:
1. Provisioning a new Kubelet identity (calling ARM APIs, which I don't think we currently import the SDKs to do this, though it's not that hard to do).
2. Using the `ExpectSettingsReplaced` helper (thanks @Bryce-Soghigian for pointing me to it), which can't be tested in NAP.

It's not clear to me if we should add an E2E for this or if we'd prefer just UTs.

pkg/cloudprovider/drift.go
```diff
@@ -221,6 +204,65 @@ func (c *CloudProvider) isSubnetDrifted(ctx context.Context, nodeClaim *karpv1.N
...
+	if node == nil {
+		return "", nil
+	}
+
+	kubeletIdentityClientID := node.Labels[v1alpha2.AKSLabelKubeletIdentityClientID]
+	// The kubelet identity label is supposed to be set on every node, but prior to
+	// 1.4.0 it was not set by Karpenter. In order to avoid rolling all existing nodes,
+	// we don't count a missing kubelet identity as drift. This situation should resolve itself as
+	// image version and Kubernetes version drift is performed.
+	// TODO: This short-circuit should be removed post 1.4.0 (~2025-07-01)
```
charliedmcb, line 225: Discussed offline. Agreed, this approach works. Just need to come back to this in a later release.

pkg/providers/imagefamily/bootstrap/aksbootstrap.go
```diff
@@ -307,7 +308,8 @@ func (a AKS) applyOptions(nbv *NodeBootstrapVariables) {
 
 	// merge and stringify labels
 	kubeletLabels := lo.Assign(getBaseKubeletNodeLabels(), a.Labels)
-	getAgentbakerGeneratedLabels(a.ResourceGroup, kubeletLabels)
+	getAgentBakerGeneratedLabels(a.ResourceGroup, kubeletLabels)
+	kubeletLabels[v1alpha2.AKSLabelKubeletIdentityClientID] = a.KubeletIdentityClientID
```
charliedmcb, line 312: This handles the `provisionMode`, `aksscriptless` case, which is the default. 

We also need to handle:
https://github.com/Azure/karpenter-provider-azure/blob/11be7f237150c9d0299b4d9081bbac740ff707ae/pkg/providers/launchtemplate/launchtemplate.go#L205

Need to ensure the `bootstrappingclient` case is handled as well.

charliedmcb (commented):
Overall, looks good, but need to check the `bootstrappingclient` case

pkg/cloudprovider/suite_test.go
charliedmcb, line 286: nit: I was liking breaking the drift tests into `Context` based on their drift type.

charliedmcb: Yea, realizing this is in the `Node Image Drift` Context. Would be great to split it into its own `Context("Kubelet Identity Drift"`

matthchr: Oh I msised that in the rebase, can fix.

pkg/providers/imagefamily/bootstrap/aksbootstrap.go
matthchr: Good callout:

Looks like it's not there with bootstrappingClient mode today:
```
                    beta.kubernetes.io/arch=amd64
                    beta.kubernetes.io/instance-type=Standard_D16_v3
                    beta.kubernetes.io/os=linux
                    failure-domain.beta.kubernetes.io/region=westcentralus
                    failure-domain.beta.kubernetes.io/zone=
                    karpenter.azure.com/sku-cpu=16
                    karpenter.azure.com/sku-family=D
                    karpenter.azure.com/sku-gpu-count=0
                    karpenter.azure.com/sku-hyperv-generation=1
                    karpenter.azure.com/sku-memory=65536
                    karpenter.azure.com/sku-name=Standard_D16_v3
                    karpenter.azure.com/sku-networking-accelerated=true
                    karpenter.azure.com/sku-version=3
                    karpenter.azure.com/zone=
                    karpenter.sh/capacity-type=on-demand
                    karpenter.sh/initialized=true
                    karpenter.sh/nodepool=default
                    karpenter.sh/registered=true
                    kubernetes.azure.com/agentpool=
                    kubernetes.azure.com/azure-cni-overlay=true
                    kubernetes.azure.com/cluster=MC_matthchr-temp_matthchr-test-1_westcentralus
                    kubernetes.azure.com/ebpf-dataplane=cilium
                    kubernetes.azure.com/mode=user
                    kubernetes.azure.com/network-subnet=aks-subnet
                    kubernetes.azure.com/nodenetwork-vnetguid=3615f57c-f813-4cb7-b5a8-a27a312c7adb
                    kubernetes.azure.com/podnetwork-type=overlay
                    kubernetes.azure.com/role=agent
                    kubernetes.io/arch=amd64
                    kubernetes.io/hostname=aks-default-r6fkt
                    kubernetes.io/os=linux
                    node.kubernetes.io/instance-type=Standard_D16_v3
                    topology.disk.csi.azure.com/zone=
                    topology.kubernetes.io/region=westcentralus
                    topology.kubernetes.io/zone=

```

matthchr: Ok I think I've pushed a fix to this... @comtalyst how do I test it? I am not seeing any UTs that run with this mode, nor a mock for it if we set it?

Bryce-Soghigian: we don't have a fake for NPS so we can't really right now. I have encountered this a few times :(

pkg/cloudprovider/drift.go
```diff
@@ -39,58 +40,51 @@ import (
...
 
 	// TODO (charliedmcb): Use this const across code and test locations which are signaling/checking for "no drift"
 	NoDrift cloudprovider.DriftReason = ""
 )
 
 func (c *CloudProvider) isNodeClassDrifted(ctx context.Context, nodeClaim *karpv1.NodeClaim, nodeClass *v1alpha2.AKSNodeClass) (cloudprovider.DriftReason, error) {
-	// First check if the node class is statically staticFieldsDrifted to save on API calls.
-	if staticFieldsDrifted := c.areStaticFieldsDrifted(nodeClaim, nodeClass); staticFieldsDrifted != "" {
-		return staticFieldsDrifted, nil
+	checks := []func(ctx context.Context, nodeClaim *karpv1.NodeClaim, nodeClass *v1alpha2.AKSNodeClass) (cloudprovider.DriftReason, error){
```
jackfrancis, line 54: Is the specific serialization of these checks meaningful in terms of either optimizing for "least expensive mean computation time", or "functional correctness"? If not, should we add a TODO or issue to parallelize them?

pkg/cloudprovider/drift.go
```diff
@@ -39,58 +40,51 @@ import (
...
+		if driftReason != "" {
+			return driftReason, nil
+		}
 	}
+
 	return "", nil
 }
 
-func (c *CloudProvider) areStaticFieldsDrifted(nodeClaim *karpv1.NodeClaim, nodeClass *v1alpha2.AKSNodeClass) cloudprovider.DriftReason {
+func (c *CloudProvider) areStaticFieldsDrifted(_ context.Context, nodeClaim *karpv1.NodeClaim, nodeClass *v1alpha2.AKSNodeClass) (cloudprovider.DriftReason, error) {
```
jackfrancis, line 74: should we take advantage of the added context and log a warning if we the terminal outcome is `nodeClassHash != nodeClaimHash` ?

pkg/cloudprovider/drift.go
matthchr: My understanding is that order matters here, and cheaper checks are checked first. @charliedmcb could answer this more completely though.

I _think_ that these checks are all served from controller-runtime cached clients, which means that the general level of expense here is not actually that high, because the data we're examining is already in memory, so this is just a bunch of in-memory comparisons. The only thing that isn't true for is the maintenance windows stuff that @charliedmcb just added I believe.

pkg/cloudprovider/drift.go
```diff
@@ -39,58 +39,51 @@ import (
...
+	K8sVersionDrift      cloudprovider.DriftReason = "K8sVersionDrift"
+	ImageDrift           cloudprovider.DriftReason = "ImageDrift"
+	SubnetDrift          cloudprovider.DriftReason = "SubnetDrift"
+	KubeletIdentityDrift cloudprovider.DriftReason = "KubeletIdentityDrift"
 
 	// TODO (charliedmcb): Use this const across code and test locations which are signaling/checking for "no drift"
 	NoDrift cloudprovider.DriftReason = ""
 )
 
 func (c *CloudProvider) isNodeClassDrifted(ctx context.Context, nodeClaim *karpv1.NodeClaim, nodeClass *v1beta1.AKSNodeClass) (cloudprovider.DriftReason, error) {
```
jackfrancis, line 52: Is there a case to be made to determine this before calling the funcs below (looks like not all of them need it but more than one does):

```
	node, err := c.getNodeForDrift(ctx, nodeClaim)
	if err != nil {
		return "", err
	}
	if node == nil {
		return "", nil
	}
```

The current flow is a bit weird in the sense that for the funcs that require the above to make a determination, we are kind of overloading an implicit ternary return semantic ("not enough input to actually determine drift") in the function itself. Maybe we hoist that out of the function and instead make it a precondition of calling the func at all, like a pre-step where we gather "which types are drift are we actually able to evaluate?"

pkg/providers/imagefamily/bootstrap/aksbootstrap.go
matthchr: I tested manually for now.

pkg/cloudprovider/drift.go
matthchr: Added one but @charliedmcb was telling me he thinks we maybe should remove some of these. For now having it makes more sense than not though, so added.

pkg/cloudprovider/drift.go
charliedmcb: The intention of ordering is to start with the easier to complete (more reliable, less costly etc), with short circuiting on any drift condition being true. 

I think this space is cheap enough, there isn't a need to parallelize them.

Also, I do lean towards returning a consistent ordering for the drift conditions, even if it's not necessary the exact same one (maybe an earlier condition is later returned, once 2 cases become true).

matthchr: Correcting this some.

The MW stuff is in disruption not drift so doesn't apply here. The risky (high cost?) calls here are GET VM and GET NIC sent to Azure. The rest are served from controller-runtime clients and so cached. GetVM and GetNIC are the last pieces of drift checked.

I think above means we don't want to parallelize. AWS also doesn't parallelize these checks AFAIK.

charliedmcb: The two checks, which I think would be more expensive, are the checks for `isImageVersionDrifted` and `isSubnetDrifted` which have calls to `instanceProvider.Get` for the VM and `instanceProvider.GetNic` for the Nic respectively. However, they are located at the end of each set of drift checks.

pkg/cloudprovider/drift.go
charliedmcb: I wouldn't log a warning, since Drift is a feature that's expected for resolving things based on customer configuration changes, but logging an `info` for now I think makes sense.

Chatted with Matthew about this offline, and might remove the logging on drift cases here, as core already logs when a nodeclaim becomes drifted, or stops being drifted:
- [log for becoming drifted](https://github.com/kubernetes-sigs/karpenter/blob/79c4a7ea60ca95b36f37226dcb868f50ba0fd25c/pkg/controllers/nodeclaim/disruption/drift.go#L73)
- [log for losing the drift condition](https://github.com/kubernetes-sigs/karpenter/blob/79c4a7ea60ca95b36f37226dcb868f50ba0fd25c/pkg/controllers/nodeclaim/disruption/drift.go#L66)

pkg/cloudprovider/drift.go
matthchr: If we do end up wanting to do this "lazy at-most-once cacher" I don't think it makes sense to do it here because nodes are already cached by the controller-runtime client, the things we'd want to cache are like VMs or NICs or whatever, which today are only checked once so don't need to be cached.

charliedmcb: I wouldn't put this as a pre-check for all `is___Drifted` checks, but could see it being added before `isK8sVersionDrifted`, and `isKubeletIdentityDrifted`.

The reasoning I wouldn't put it before `areStaticFieldsDrifted` is that part of the current drift pattern is to short circuit on success, meaning that if we don't have to check for the `getNodeForDrift`, we might as well not.

My opinion is to leave it the way Matthew has it (I think its fine as is), but leave a TODO for possibly hoisting it out. Would only want it preformed after `areStaticFieldsDrifted` though.

charliedmcb: Actually, thinking on this more, I'm more mixed on it being better or not. I kind of like the individual cases doing the checks themselves, and leaving the easy loop over all cases Matthew has. 

The fact it is being used in two separate places does make it feel worth "reusing" and doing in one location, but as Matthew said with point 2 `OTOH, it's not expensive to do this check because it's reading the node from a cached client anyway.`, I think its fine? And extracting it would make the pass over things a bit more clunky keeping the priority ordering, and such.

charliedmcb: > lazy at-most-once cacher

- I think this is the solution I would lean towards, if we start finding this pattern creep up more, but agree its not necessary for nodes here.

matthchr: I've left a TODO comment here.

jackfrancis: all sgtm!

pkg/cloudprovider/drift.go
```diff
@@ -39,58 +40,61 @@ import (
...
-	// First check if the node class is statically staticFieldsDrifted to save on API calls.
-	if staticFieldsDrifted := c.areStaticFieldsDrifted(nodeClaim, nodeClass); staticFieldsDrifted != "" {
-		return staticFieldsDrifted, nil
+	// TODO: if we find more expensive checks, such as reading VMs or NICs from Azure, are being duplicated between checks, we should
+	// TODO: produce a lazy at-most-once that allows a check to cache a value for later checks to read.
```
charliedmcb, line 55: Nit: I like leaving it just so one of the lines has a TODO in it, with indent, so searching only leaves with one occurrence showing up, but just my feeling of it.
```suggestion
	//     produce a lazy at-most-once that allows a check to cache a value for later checks to read.
```

matthchr: Fixed

pkg/providers/imagefamily/customscriptsbootstrap/provisionclientbootstrap.go
```diff
@@ -90,7 +91,7 @@ func (p ProvisionClientBootstrap) GetCustomDataAndCSE(ctx context.Context) (stri
 		OsType:                   lo.ToPtr(lo.Ternary(p.IsWindows, models.OSTypeWindows, models.OSTypeLinux)),
 		VMSize:                   lo.ToPtr(p.InstanceType.Name),
 		Distro:                   lo.ToPtr(p.ImageDistro),
-		CustomNodeLabels:         labels,
+		CustomNodeLabels:         nodeLabels,
```
charliedmcb, line 94: Dropping a comment just for tracking on what we covered offline, ensuring the ordering of kubelet identity used with `bootstrappingclient`, and `label` added will guaranteed always have the label at a <= version used.

charliedmcb: Maybe leave a comment somewhere about this?

charliedmcb (dismissed):
lgtm, assuming `bootstrappingclient` ordering is entirely covered.

pkg/cloudprovider/drift.go
```diff
@@ -221,6 +215,62 @@ func (c *CloudProvider) isSubnetDrifted(ctx context.Context, nodeClaim *karpv1.N
...
+	if err != nil || node == nil {
+		return "", err
+	}
+
+	kubeletIdentityClientID := node.Labels[v1beta1.AKSLabelKubeletIdentityClientID]
+	// The kubelet identity label is supposed to be set on every node, but prior to
+	// 1.4.0 it was not set by Karpenter. In order to avoid rolling all existing nodes,
+	// we don't count a missing kubelet identity as drift. This situation should resolve itself as
+	// image version and Kubernetes version drift is performed.
+	// TODO: This short-circuit should be removed post 1.4.0 (~2025-07-01)
```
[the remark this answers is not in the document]
matthchr: I think post 1.4.0 is still technically correct?

pkg/providers/imagefamily/customscriptsbootstrap/provisionclientbootstrap.go
matthchr: Did this

tallaxes (approved):

PR #893, opened 2025-05-16
chore: refresh (and backfill missing) pricing

tallaxes:
**Description**

* Update the list of regions for static pricing
* Update pricing API version
* Regenerate pricing
* Add test for missing pricing info (any supported SKU should have pricing in at least one region)
* Backfill missing pricing (with a high price for now) - using the list produced by the above test

* `make presubmit`

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


pkg/providers/instancetype/skus.go
```diff
@@ -22,10 +22,10 @@ import (
...
 )
 
-// getKarpenterWorkingSKUs returns a the list of SKUs that are
+// GetKarpenterWorkingSKUs returns a the list of SKUs that are
 // allowed to be used by Karpenter. This is a subset of the
 // SKUs that are available in Azure.
-func getKarpenterWorkingSKUs() []skewer.SKU {
+func GetKarpenterWorkingSKUs() []skewer.SKU {
 	workingSKUs := []skewer.SKU{}
 	for _, sku := range allAzureVMSkus {
```
charliedmcb, line 30: Does the `allAzureVMSkus` need to be updated? Is it automatically tested for updates in our CI somehow?

pkg/providers/pricing/skus_missing_price.go
```diff
@@ -0,0 +1,100 @@
...
+	"Standard_A2":               missingPrice,
+	"Standard_A3":               missingPrice,
+	"Standard_A4":               missingPrice,
+	"Standard_A5":               missingPrice,
+	"Standard_A6":               missingPrice,
+	"Standard_A7":               missingPrice,
+	"Standard_D1":               missingPrice,
+	"Standard_D11":              missingPrice,
+	"Standard_D11_v2":           missingPrice,
+	"Standard_D11_v2_Promo":     missingPrice,
```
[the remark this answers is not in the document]
charliedmcb: Ok, looked over how the set of skus are determined, and calls to the pricing provider is performed. I think this is ok, but do still think its worth an assessment of the skus we allow, but can be beyond this PR.

pkg/providers/pricing/skus_missing_price.go
```diff
@@ -0,0 +1,100 @@
...
+// limitations under the License.
+
+package pricing
+
+// setting to a high value means it won't be chosen automatically,
+// but will still be available if chosen explicitly
+const missingPrice float64 = 999.0
+
+// skusWithMissingPrice is a map of SKUs that are known to have missing prices in the Azure pricing API.
+var skusWithMissingPrice = map[string]float64{
```
charliedmcb, line 22: This still doesn't necessarily catch all the case, since some skus might only be supported in one region, for instance?

I don't really see a harm in adding this in though, since with such high pricing it should only select these if the selectors are only for skus with missing prices.

Does feel a bit like a more complete assessment of skus might be worth it though.

pkg/providers/pricing/pricing.go
```diff
@@ -155,6 +159,10 @@ func (p *Provider) SpotPrice(instanceType string) (float64, bool) {
 	defer p.mu.RUnlock()
 	price, ok := p.spotPrices[instanceType]
 	if !ok {
+		// if we don't have a price, check if it's a known SKU with missing price
+		if price, ok = skusWithMissingPrice[instanceType]; ok {
```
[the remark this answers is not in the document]
tallaxes: Yes. The names come from the test that only prints SKUs that don't have _either_.

pkg/providers/instancetype/skus.go
tallaxes: Not updated automatically yet, but will be in the future.

pkg/providers/pricing/skus_missing_price.go
tallaxes: SKUs that are not supported in a particular region should already be filtered out on the skewer / resource API side, andn won't have available offerings, so should not be a problem.

tallaxes:
> feels like its putting a band aid on a bigger issue of why these skus are missing from the pricing API

Ack, being taken care of separately

pkg/providers/pricing/skus_missing_price.go
charliedmcb: Ah, make sense. Good point.

pkg/providers/pricing/client/pricingapi.go
```diff
@@ -27,7 +27,7 @@ import (
 )
 
 const (
-	apiVersion = "2021-10-01-preview"
+	apiVersion = "2023-01-01-preview"
```
Bryce-Soghigian, line 30: why 2023? Whats the latest version? Also why preview? Just following convention?

charliedmcb (approved):
Approving as a fix for now, but need follow up on fixing root of why this is occurring.

Bryce-Soghigian (approved):

pkg/providers/pricing/client/pricingapi.go
tallaxes: That's the latest version

PR #894, opened 2025-05-17
test: spot E2E

comtalyst:
Fixes # 

**Description**
Add a new E2E for spot.

* `make az-e2etests`
* https://github.com/Azure/karpenter-provider-azure/actions/runs/15455686427/job/43507459585

- [ ] Yes, issue opened: # 
- [X] No

**Release Note**

```release-note

```

### DIFF
--- a/test/pkg/environment/azure/expectations.go
+++ b/test/pkg/environment/azure/expectations.go
@@ -17,9 +17,7 @@ limitations under the License.
 package azure
 
 import (
-	"context"
 	"fmt"
-	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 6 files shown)


rakechill:
remind me -- can we run the test on this PR before merge? and if not, can you post the output of it working?

test/pkg/environment/azure/expectations.go
```diff
@@ -30,33 +28,10 @@ import (
...
-	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/compute/armcompute"
 	containerservice "github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/containerservice/armcontainerservice/v4"
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/network/armnetwork"
 )
 
-func (env *Environment) GetVM(nodeName string) armcompute.VirtualMachine {
-	GinkgoHelper()
-	node := env.Environment.GetNode(nodeName)
-	return env.GetVMByName(env.ExpectParsedProviderID(node.Spec.ProviderID))
-}
```
Bryce-Soghigian, line 42: this is used by other tests we cant remove this

test/pkg/environment/azure/expectations.go
```diff
@@ -87,28 +62,6 @@ func (env *Environment) ExpectCreatedInterface(networkInterface armnetwork.Inter
...
-		if err != nil {
-			return nil, fmt.Errorf("failed to list virtual networks: %w", err)
-		}
-		if len(resp.VirtualNetworkListResult.Value) > 0 {
-			return resp.VirtualNetworkListResult.Value[0], nil
-		}
-	}
-	return nil, fmt.Errorf("no virtual networks found in resource group: %s", vnetRG)
-}
-
```
Bryce-Soghigian, line 111: Why?

test/pkg/environment/azure/expectations.go
Bryce-Soghigian: oh i see you moved it to utils.go, maybe change it to 


azclient_utils.go?

test/suites/interruption/suite_test.go
```diff
@@ -0,0 +1,118 @@
...
+
+		// Simulate spot eviction
+		env.SimulateVMEviction(nodes[0].Name)
+
+		// Verify that a node is deleted
+		env.EventuallyExpectNotFound(nodes[0])
+		env.EventuallyExpectNotFound(pods[0])
+
+		// Verify pods are scheduled and running after replacement
+		env.EventuallyExpectHealthyPodCount(labels.SelectorFromSet(dep.Spec.Selector.MatchLabels), 1)
```
Bryce-Soghigian, line 113: interruption is a confusing test name, since karpenter for aws has an interruption controller, and we don't implement it ourselves.

test/pkg/environment/azure/expectations.go
comtalyst: Don't think it needs to be specific to Azure client utils as of now. The idea was more of a separation between validators (which fails the test on failing expectations) and any non-validating util function.

test/pkg/environment/azure/expectations.go
comtalyst: What do you mean?

rakechill: I think it's a follow up to "why did you remove these functions?" but you didn't...just moved them to another file

comtalyst:
@Bryce-Soghigian parity with AWS on interruption suite was my original intention. But, given that the name comes from interruption controller as you mentioned, then we should not have interruption suite in the first place.

I don't think we are keeping that consistency with AWS in integration suite(?), at least not to the extent that prevents the test of integration with Azure-specific functionalities. Also don't think spot should be a standalone suite, at least at this point where it is small and doesn't have explicit handler/controller in the code.

Bryce-Soghigian:
@comtalyst  good to go if we move it into the chaos suite, integration takes too long as it is, we need to reduce the runtime.

Bryce-Soghigian (dismissed):

Bryce-Soghigian (approved):

PR #908, opened 2025-05-23
fix: adjust TTL for unavailable offerings when allocation error happens

UtheMan:
**Description**
Adjusts TTL for unavailable offerings when allocation related failures are reported back by Azure, adds unit tests

* [x] Unit Tests
* [x] E2E: https://github.com/Azure/karpenter-provider-azure/actions/runs/15224626927

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note
Adjusts TTL for unavailable offerings when Azure reports Allocation related errors from 3 minutes to 15 minutes.
```

### DIFF
--- a/pkg/providers/instance/instance.go
+++ b/pkg/providers/instance/instance.go
@@ -67,7 +67,7 @@ var (
 	SKUNotAvailableReason                       = "SKUNotAvailable"
 
 	SubscriptionQuotaReachedTTL = 1 * time.Hour
-	AllocationFailureTTL        = 15 * time.Minute
+	AllocationFailureTTL        = 1 * time.Hour
 	SKUNotAvailableSpotTTL      = 1 * time.Hour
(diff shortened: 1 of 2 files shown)


pkg/providers/instance/instance.go
```diff
@@ -67,6 +67,7 @@ var (
 	SKUNotAvailableReason                       = "SKUNotAvailable"
 
 	SubscriptionQuotaReachedTTL = 1 * time.Hour
+	AllocationFailureTTL        = 15 * time.Minute
```
UtheMan, line 70: 15 minutes based on the behavior I saw when investigating the ZonalAllocationFailures - after around 15~ minutes, node claims were getting deleted by the controller, the goal is to block SKUs that failed with ZonalAllocation for long enough time that more different SKUs are tried before node claim is deleted.

UtheMan: Adjusted to 1 hr based on offline discussion

PR #910, opened 2025-05-23
fix: VM client options should use default ops + aux token ops

rakechill:
**Description**
Updated VM Client to use both default arm opts + auxiliary token policy. Defaults were accidentally removed in #540.

* make test

- [ ] Yes, issue opened: # 
- [X] No

**Release Note**

```release-note
NONE
```

### DIFF
--- a/pkg/auth/auxiliarytoken.go
+++ b/pkg/auth/auxiliarytoken.go
@@ -23,15 +23,15 @@ import (
 	"net/http"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
-	"github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
-	"github.com/Azure/karpenter-provider-azure/pkg/operator/options"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 )
 
 type AuxiliaryTokenServer interface {
 	Do(req *http.Request) (*http.Response, error)
 }
+
+var _ policy.Policy = &AuxiliaryTokenPolicy{}
 
 // AuxiliaryTokenPolicy provides a custom policy used to authenticate
 // with shared node image galleries.
@@ -44,16 +44,7 @@ func (p *AuxiliaryTokenPolicy) Do(req *policy.Request) (*http.Response, error) {
 	return req.Next()
 }
 
-func AddAuxiliaryTokenPolicyClientOptions(ctx context.Context, client AuxiliaryTokenServer, opts *options.Options, clientOptions *arm.ClientOptions) error {
-	auxPolicy, err := getAuxiliaryTokenPolicy(ctx, client, opts.SIGAccessTokenServerURL, opts.SIGAccessTokenScope)
-	if err != nil {
-		return fmt.Errorf("failed to add auxiliary token policy to virtual machine client options: %w", err)
-	}
-	clientOptions.ClientOptions.PerRetryPolicies = append(clientOptions.ClientOptions.PerRetryPolicies, auxPolicy)
-	return nil
-}
-
-func getAuxiliaryTokenPolicy(ctx context.Context, client AuxiliaryTokenServer, url string, scope string) (*AuxiliaryTokenPolicy, error) {
+func NewAuxiliaryTokenPolicy(ctx context.Context, client AuxiliaryTokenServer, url string, scope string) (*AuxiliaryTokenPolicy, error) {
 	token, err := getAuxiliaryToken(client, url, scope)
 	if err != nil {
 		return nil, fmt.Errorf("failed to get auxiliary token: %w", err)
--- a/pkg/providers/instance/azure_client.go
+++ b/pkg/providers/instance/azure_client.go
@@ -140,10 +140,11 @@ func NewAZClient(ctx context.Context, cfg *auth.Config, env *azclient.Environmen
 	if o.UseSIG {
 		klog.V(1).Info("Using SIG for image versions")
 		client := &http.Client{Timeout: 10 * time.Second}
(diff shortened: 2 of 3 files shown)


pkg/providers/instance/azure_client.go
```diff
@@ -134,16 +133,18 @@ func NewAZClient(ctx context.Context, cfg *auth.Config, env *azure.Environment)
...
+	var vmClientOptions = *opts
 	o := options.FromContext(ctx)
 	if o.UseSIG {
 		klog.V(1).Info("Using SIG for image versions")
-		vmClientOptions, err = getVirtualMachinesClientOptions(ctx, o.SIGAccessTokenServerURL, o.SIGAccessTokenScope)
+		auxOptions, err := getAuxiliaryTokenPolicyOptions(ctx, o.SIGAccessTokenServerURL, o.SIGAccessTokenScope)
 		if err != nil {
 			return nil, err
 		}
+		vmClientOptions.ClientOptions = auxOptions
```
tallaxes, line 145: This would still replace everything in ClientOptions, and wipe out PerCallPolicies? The additional PerRetryPolicy needs to be appended to whatever default PerRetryPolicies are there; you can see an example of appending to PerCallPolicies in DefaultArmOpts().

rakechill: Yeah reached out to alex offline. Accidentally left out some local changes that switched this to appending to that struct. Was adding some tests and a fake today. 

It's a little weirder since this client is used to initialize another client + then not really used again. It doesn't make as much sense to add it into our larger acceptance testing framework either, since VirtualMachinesAPI will be initialized without it as part of NewAZClientFromAPI(). However, I think it's still worthwhile to sort of acceptance test the auth.GetAuxiliaryTokenPolicy flow.

pkg/auth/auxiliarytoken.go
```diff
@@ -0,0 +1,88 @@
...
+type AuxiliaryTokenPolicy struct {
+	Token string
+}
+
+func (p *AuxiliaryTokenPolicy) Do(req *policy.Request) (*http.Response, error) {
+	req.Raw().Header.Add("x-ms-authorization-auxiliary", "Bearer "+p.Token)
+	return req.Next()
+}
+
+func GetAuxiliaryTokenPolicy(ctx context.Context, url string, scope string) (policy.Policy, error) {
```
rakechill, line 42: most of the logic was just moved over from cred.go

pkg/auth/cred.go
```diff
@@ -36,43 +33,6 @@ type expireEarlyTokenCredential struct {
 	cred azcore.TokenCredential
 }
 
-func GetAuxiliaryToken(ctx context.Context, url string, scope string) (azcore.AccessToken, error) {
```
rakechill, line 39: moved to auxiliarytoken.go

matthchr (approved):
Approved after discussion about separate "dealing with token expiry" PR.

pkg/auth/auxiliarytoken.go
```diff
@@ -0,0 +1,101 @@
...
+	auxPolicy, err := getAuxiliaryTokenPolicy(ctx, client, opts.SIGAccessTokenServerURL, opts.SIGAccessTokenScope)
+	if err != nil {
+		return fmt.Errorf("failed to add auxiliary token policy to virtual machine client options: %w", err)
+	}
+	clientOptions.ClientOptions.PerRetryPolicies = append(clientOptions.ClientOptions.PerRetryPolicies, auxPolicy)
+	return nil
+}
+
+func getAuxiliaryTokenPolicy(ctx context.Context, client AuxiliaryTokenServer, url string, scope string) (*AuxiliaryTokenPolicy, error) {
+	token, err := getAuxiliaryToken(client, url, scope)
```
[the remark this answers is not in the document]
rakechill: Will update to handle expiration properly in a follow-up PR

pkg/auth/auxiliarytoken.go
```diff
@@ -0,0 +1,101 @@
...
+	req.Raw().Header.Add("x-ms-authorization-auxiliary", "Bearer "+p.Token)
+	return req.Next()
+}
+
+func AddAuxiliaryTokenPolicyClientOptions(ctx context.Context, client AuxiliaryTokenServer, opts *options.Options, clientOptions *arm.ClientOptions) error {
+	auxPolicy, err := getAuxiliaryTokenPolicy(ctx, client, opts.SIGAccessTokenServerURL, opts.SIGAccessTokenScope)
+	if err != nil {
+		return fmt.Errorf("failed to add auxiliary token policy to virtual machine client options: %w", err)
+	}
+	clientOptions.ClientOptions.PerRetryPolicies = append(clientOptions.ClientOptions.PerRetryPolicies, auxPolicy)
```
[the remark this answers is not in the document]
rakechill: > There are actually multiple places you could put this policy (per retry or per request)   

I looked into this originally and think `PerRetry` is a bit better since it ensures this policy is called on each _call_ regardless of if it's first try or retry:

// PerRetryPolicies contains custom policies to inject into the pipeline.
// Each policy is executed once per request, and for each retry of that request.
PerRetryPolicies [][Policy](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azcore/policy#Policy)


> Which makes me think actually just NewAuxiliaryTokenPolicy(...) (*AuxiliaryTokenPolicy, error) + at the call site doing


Yeah I went back and forth on the structure here a few times... my final reasoning was around trying to keep as much of the aux token logic to this file rather than polluting `azure_client.go`, so appending to options there would just leave a larger footprint. I do think reducing the # of nested helper functions here would be nice though...I'll look into this

matthchr (dismissed):

matthchr (approved):

PR #912, opened 2025-05-24
fix: capture both values for boolean requirements

tallaxes:
**Description**

Boolean-valued requirements (like `sku-storage-premium-capable`) should allow both "true" and "false" value in selectors. (Only "true" was supported.)

* `make presubmit`

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/providers/instancetype/instancetype.go
+++ b/pkg/providers/instancetype/instancetype.go
@@ -166,8 +166,8 @@ func computeRequirements(sku *skewer.SKU, vmsize *skewer.VMSizeType, architectur
 
 		// SKU capabilities
 		scheduling.NewRequirement(v1beta1.LabelSKUStorageEphemeralOSMaxSize, corev1.NodeSelectorOpDoesNotExist),
-		scheduling.NewRequirement(v1beta1.LabelSKUStoragePremiumCapable, corev1.NodeSelectorOpDoesNotExist),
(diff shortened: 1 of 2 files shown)


pkg/providers/instancetype/instancetype.go
```diff
@@ -166,8 +166,8 @@ func computeRequirements(sku *skewer.SKU, vmsize *skewer.VMSizeType, architectur
 
 		// SKU capabilities
 		scheduling.NewRequirement(v1beta1.LabelSKUStorageEphemeralOSMaxSize, corev1.NodeSelectorOpDoesNotExist),
-		scheduling.NewRequirement(v1beta1.LabelSKUStoragePremiumCapable, corev1.NodeSelectorOpDoesNotExist),
```
matthchr, line 169: Just confirming, setting this here is enough to result in nodes whose SKUs have this capability getting the label correctly set, even if the user doesn't actually _specify_ this requirement, is that right?

matthchr (approved):

pkg/providers/instancetype/instancetype.go
tallaxes: Yes. All single-valued requirements of instance type used are converted to labels: https://github.com/Azure/karpenter-provider-azure/blob/67902180fa6a5d01fce0f03e1d19f78d8fa2a83a/pkg/cloudprovider/cloudprovider.go#L438

PR #914, opened 2025-05-26
feat: support Azure Linux v3 (from AKS 1.32)

tallaxes:
**Description**

Add support for Azure Linux v3, which should be used from AKS/Kubernetes v1.32 onwards. This does not expose direct control over v2 vs v3, just uses the appropriate version internally when `AzureLinux` image family is used - matching AKS behavior.

> [!WARNING]  
> This will result in drift and node replacement on existing deployments of AzureLinux v2 on Kubernetes 1.32 and later.

Follow-up https://github.com/Azure/karpenter-provider-azure/pull/952 regenerates nodeimageversions fake data.

* `make presubmit`
* manual test
* E2E: https://github.com/Azure/karpenter-provider-azure/actions/runs/15336652219 ✅

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


pkg/providers/imagefamily/resolver.go
```diff
@@ -99,8 +101,12 @@ func (r *defaultResolver) Resolve(
 	if err != nil {
 		return nil, err
 	}
+	kubernetesVersion, err := nodeClass.GetKubernetesVersion()
+	if err != nil {
+		return nil, err
+	}
 
-	imageFamily := getImageFamily(nodeClass.Spec.ImageFamily, staticParameters)
+	imageFamily := getImageFamily(nodeClass.Spec.ImageFamily, kubernetesVersion, staticParameters)
```
charliedmcb, line 109: I think this would work, but I don't loving having the added dependency on the k8s version here. 

imageFamily is just being used to map to the distro on the provisioning side of things. Not necessarily critical to get in before this, but the area could use some cleanup, and would prefer to get that done before expanding on some of the mucky code already.

tallaxes: Agree that this area could use some cleanup, but the added dependency on Kubernetes version just reflects the reality that, for Azure Linux at least, the distro explicitly depends on Kubernetes version, using a different one from 1.32 on.

pkg/fake/nodeimageversionsapi.go
```diff
@@ -43,10 +43,10 @@ var (
 			Version:  "17763.2019.221114",
 		},
 		{
-			FullName: "AKSAzureLinux-V3-202409.23.0",
+			FullName: "AKSAzureLinux-V3-202410.09.0",
```
charliedmcb, line 46: I'm assuming this was manually modified. The data in this file, from my understanding was produced from a real API call, so I'd hesitate from manually modifying it. Feel this should be commented somewhere though.

pkg/providers/imagefamily/azlinux3.go
```diff
@@ -0,0 +1,135 @@
...
+	karpv1 "sigs.k8s.io/karpenter/pkg/apis/v1"
+	"sigs.k8s.io/karpenter/pkg/cloudprovider"
+	"sigs.k8s.io/karpenter/pkg/scheduling"
+)
+
+const (
+	AzureLinux3Gen2ImageDefinition    = "V3gen2"
+	AzureLinux3Gen1ImageDefinition    = "V3"
+	AzureLinux3Gen2ArmImageDefinition = "V3gen2arm64"
+)
```
charliedmcb, line 36: Looks correct to me based on our current support, and the previous real world data from the NodeImageVersions API. However, any need to double check with Node SIG?

pkg/providers/imagefamily/azlinux3.go
```diff
@@ -0,0 +1,135 @@
...
+
+type AzureLinux3 struct {
+	Options *parameters.StaticParameters
+}
+
+func (u AzureLinux3) Name() string {
+	return v1beta1.AzureLinuxImageFamily
+}
+
+func (u AzureLinux3) DefaultImages() []DefaultImageOutput {
```
charliedmcb, line 46: Is anything different here, between v2, and v3 beyond `DefaultImages()`?

If we don't expect the bootstrapping, or naming to change, it might possibly make more sense for the DefaultImages to have `k8s` as a parameter?

pkg/providers/imagefamily/azlinux3.go
```diff
@@ -0,0 +1,135 @@
...
+	// image provider will select these images in order, first match wins
+	return []DefaultImageOutput{
```
charliedmcb, line 48: Assuming simply adopted the same ordering as AzLinuxv2?

@Bryce-Soghigian, how did you determine the priority ordering to begin with?

pkg/providers/imagefamily/azlinux3.go
```diff
@@ -0,0 +1,135 @@
...
+		{
+			PublicGalleryURL:     AKSAzureLinuxPublicGalleryURL,
```
charliedmcb, line 50: @Bryce-Soghigian are these supported in CIG?

pkg/providers/imagefamily/azlinux3.go
```diff
@@ -0,0 +1,135 @@
...
+			GalleryResourceGroup: AKSAzureLinuxResourceGroup,
+			GalleryName:          AKSAzureLinuxGalleryName,
+			ImageDefinition:      AzureLinux3Gen2ImageDefinition,
+			Requirements: scheduling.NewRequirements(
+				scheduling.NewRequirement(v1.LabelArchStable, v1.NodeSelectorOpIn, karpv1.ArchitectureAmd64),
+				scheduling.NewRequirement(v1beta1.LabelSKUHyperVGeneration, v1.NodeSelectorOpIn, v1beta1.HyperVGenerationV2),
+			),
+			Distro: "aks-azurelinux-v3-gen2",
```
charliedmcb, line 58: Double checked distros looks good:
[AgentBaker/pkg/agent/datamodel/types.go at 7d14215035b51860de586510afd396f19d7005be · Azure/AgentBaker](https://github.com/Azure/AgentBaker/blob/7d14215035b51860de586510afd396f19d7005be/pkg/agent/datamodel/types.go#L145C1-L199C81)

pkg/providers/imagefamily/resolver.go
```diff
@@ -188,23 +194,38 @@ func prepareKubeletConfiguration(ctx context.Context, instanceType *cloudprovide
...
 	}
 }
 
+// UseAzureLinux3 checks if the Kubernetes version is 1.32.0 or higher,
+// which is when Azure Linux 3 support starts
+func UseAzureLinux3(kubernetesVersion string) bool {
+	// Parse version, stripping any 'v' prefix if present
+	version, err := semver.Parse(strings.TrimPrefix(kubernetesVersion, "v"))
+	if err != nil {
+		// If we can't parse the version, default to AzureLinux (false)
```
charliedmcb: Slightly wonder if this should be an error, but fine with it being a soft failure too.

pkg/providers/imagefamily/azlinux3.go
tallaxes: They match what AgentBaker has, e.g. https://github.com/Azure/AgentBaker/blob/c03652307949f9259dfc84aca858afdaf788e692/pkg/agent/datamodel/sig_config.go#L768

pkg/providers/imagefamily/resolver.go
```diff
@@ -188,23 +194,38 @@ func prepareKubeletConfiguration(ctx context.Context, instanceType *cloudprovide
...
+		}
 		return &AzureLinux{Options: parameters}
 	default:
 		return &Ubuntu2204{Options: parameters}
 	}
 }
 
+// UseAzureLinux3 checks if the Kubernetes version is 1.32.0 or higher,
+// which is when Azure Linux 3 support starts
+func UseAzureLinux3(kubernetesVersion string) bool {
```
charliedmcb: nit: would prefer this in a `utils.go` file in the package, rather than adding it to the `resolver.go`

pkg/providers/imagefamily/azlinux3.go
tallaxes: We prefer Gen2 to Gen1. Would also probably prefer AMD over ARM for compatibility if both were supported - but this is constrained by VM SKU architecture.

pkg/providers/imagefamily/nodeimage_test.go
```diff
@@ -62,12 +67,17 @@ func getExpectedTestCIGImages(imageFamily string, version string) []imagefamily.
 	return nodeImages
 }
 
-func getExpectedTestSIGImages(imageFamily string, version string) []imagefamily.NodeImage {
+//nolint:unparam // might always be using the same version in test, but could change in the future
```
charliedmcb, line 70: `kubernetesVersion` will actually change in the different ci tests.

See line 111:
https://github.com/Azure/karpenter-provider-azure/blob/5679a16276876a86af042290a4f857e7527c92d5/pkg/providers/imagefamily/nodeimage_test.go#L111

pkg/providers/imagefamily/nodeimage_test.go
```diff
@@ -97,32 +108,52 @@ var _ = Describe("NodeImageProvider tests", func() {
...
+			Expect(foundImages).To(ContainElements(getExpectedTestCIGImages(*nodeClass.Spec.ImageFamily, cigImageVersion, "1.31.0")))
+		})
+
+		It("should match expected images for AzureLinux with version >= 1.32", func() {
+			nodeClass.Spec.ImageFamily = lo.ToPtr(v1beta1.AzureLinuxImageFamily)
+			nodeClass.Status.KubernetesVersion = "1.32.0"
+
+			foundImages, err := nodeImageProvider.List(ctx, nodeClass)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(foundImages).To(ContainElements(getExpectedTestCIGImages(*nodeClass.Spec.ImageFamily, cigImageVersion, "1.32.0")))
```
charliedmcb: With these being explicate version tests, do you want to directly check for at least one exact ID in the foundImages, rather than just calling `getExpectedTestCIGImages`?

Just since as you mentioned above the tests themselves check over the different versions, and this is more for clarity/local testing one one version.

pkg/providers/imagefamily/nodeimage_test.go
```diff
@@ -139,49 +170,67 @@ var _ = Describe("NodeImageProvider tests", func() {
...
-			imageFamily := v1beta1.AzureLinuxImageFamily
-			nodeClass.Spec.ImageFamily = &imageFamily
+		It("should match expected images for AzureLinux with version >= 1.32", func() {
+			nodeClass.Spec.ImageFamily = lo.ToPtr(v1beta1.AzureLinuxImageFamily)
+			nodeClass.Status.KubernetesVersion = "1.32.0"
 
 			foundImages, err := nodeImageProvider.List(ctx, nodeClass)
 			Expect(err).ToNot(HaveOccurred())
-			Expect(foundImages).To(ContainElements(getExpectedTestSIGImages(*nodeClass.Spec.ImageFamily, sigImageVersion)))
+			Expect(foundImages).To(ContainElements(getExpectedTestSIGImages(*nodeClass.Spec.ImageFamily, sigImageVersion, "1.32.0")))
```
charliedmcb, line 204: Same question on if you want to have an explicate ID match here?

test/suites/kubernetesupgrade/suite_test.go
```diff
@@ -114,5 +120,17 @@ var _ = Describe("KubernetesUpgrade", func() {
 		By(fmt.Sprintf("new nodes having upgraded kubernetes version: %s", kubernetesUpgradeVersion))
 		node = env.ExpectCreatedNodeCount("==", 1)[0]
 		Expect(strings.TrimPrefix(node.Status.NodeInfo.KubeletVersion, "v")).To(Equal(kubernetesUpgradeVersion))
+
+		if testAzureLinux {
```
charliedmcb, line 124: Assuming you tested this locally?

pkg/providers/imagefamily/azlinux3.go
tallaxes: Based on what I have seen - Yes; and I think we publish essentially everything to CIG (with very specific exceptions).

test/suites/kubernetesupgrade/suite_test.go
tallaxes: Yes, and also in CI E2E: https://github.com/Azure/karpenter-provider-azure/actions/runs/15336652219/job/43159611973#step:18:121

pkg/providers/imagefamily/resolver.go
tallaxes: Ack, moved.

pkg/providers/imagefamily/nodeimage_test.go
tallaxes: Frankly does not seem necessary to me; version-specificity here refers to Kubernetes version (not image version) - and specific Kubernetes version is indeed tested. So `ContainElements` ends up checking _all_ of the exact image IDs for that Kubernetes version / image family combination.

pkg/providers/imagefamily/nodeimage_test.go
tallaxes: I don't think so; see if the answer to the first question makes sense.

pkg/providers/imagefamily/azlinux3.go
tallaxes: No, there are no other differences - _for now_. Agree it can be handled differently, and I think I tried that - and having a separate family was cleaner / more straightforward. Plus, right now Azure Linux is the only family where images depend on Kubernetes version, so would be good to have more examples before changing the pattern.

pkg/fake/nodeimageversionsapi.go
tallaxes: Ack. Regenerated (and added poor-man's codegen) in https://github.com/Azure/karpenter-provider-azure/pull/952

pkg/providers/imagefamily/azlinux3.go
charliedmcb: I'm a little torn here, as I'm still somewhat leaning towards this being a combined family, as we are currently handling AzureLinux generically, and if we had a generic Ubuntu family, than I think we'd want to handle it the same way, and need to differentiate based on k8s. 

I think this split handling since one of our image families is generic and the other is versioned makes things a bit strange.

Its behind the scenes, so can be refactored later. Non-blocking, and we can always decide to adapt as we support other image families as well. I just expect if we add the generic option of Ubuntu as well, we'll want a combined overarching structure of it, to allow for reuse of bootstrapping, etc.

pkg/providers/imagefamily/nodeimage_test.go
charliedmcb: Aren't these tests for the expected image family as well, no? I mean for < 1.32 we are expecting v2, and => 1.32 we are expecting v3 to be hard coupled here.

I was more so thinking of these tests for readability, just so reading the tests you see that x k8s version should produce y AzureLinux version. Could also be called out in the test name itself (instead of `expected` be the expected AzureLinux version). Very minor, just was thinking of how to expand them as reader clarity tests.
 
However, fine with it as is too.

pkg/fake/nodeimageversionsapi.go
charliedmcb: Drop the changes here to avoid conflict, or do you need the version bumps here for some reason?

pkg/providers/imagefamily/nodeimage.go
```diff
@@ -44,7 +44,7 @@ func (p *Provider) List(ctx context.Context, nodeClass *v1beta1.AKSNodeClass) ([
 		return []NodeImage{}, err
 	}
 
-	supportedImages := getSupportedImages(nodeClass.Spec.ImageFamily)
+	supportedImages := getSupportedImages(nodeClass.Spec.ImageFamily, kubernetesVersion)
```
charliedmcb, line 47: Note: this will trigger any customer currently using AzureLinux and on k8s version 1.32.x or later to Drift and roll all of their nodes when we release this.

charliedmcb: https://learn.microsoft.com/en-us/azure/aks/supported-kubernetes-versions?tabs=azure-cli#aks-kubernetes-release-calendar

1.32 is already GA:
image

charliedmcb (approved):

charliedmcb (dismissed):
Assuming, release, and rolling nodes are ok in this instance, lgtm

pkg/providers/imagefamily/nodeimage.go
tallaxes: Ack. Still the right thing to do IMO, to align with AKS.

pkg/providers/imagefamily/azlinux3.go
tallaxes: Good points, let's revisit when we add more image families / options.

pkg/fake/nodeimageversionsapi.go
tallaxes: Should be no conflict, that PR targets this branch, will be retargeted after merge.

pkg/providers/imagefamily/nodeimage_test.go
tallaxes: Updated for readability, PTAL

charliedmcb: lgtm

charliedmcb (approved):
lgtm

PR #918, opened 2025-05-27
chore: MISE setup for 2a GetNodeBootstrapping

comtalyst:
Fixes # 

**Description**
Included cluster token in GET NodeBootstrapping requests for MISE enablement.

* Inspecting provisioner logs from standard NAP E2Es, not seeing dry-run warnings compared to without the change
* (TODO) Token not expiring after 24h

- [ ] Yes, issue opened: # 
- [X] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/providers/imagefamily/customscriptsbootstrap/provisionclientbootstrap.go
+++ b/pkg/providers/imagefamily/customscriptsbootstrap/provisionclientbootstrap.go
@@ -27,7 +27,9 @@ import (
 	"time"
 
 	"github.com/Azure/aks-middleware/http/client/direct/restlogger"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
+	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
 
 	"github.com/Azure/karpenter-provider-azure/pkg/apis/v1beta1"
 	"github.com/Azure/karpenter-provider-azure/pkg/operator/options"
@@ -178,6 +180,19 @@ func (p ProvisionClientBootstrap) GetCustomDataAndCSE(ctx context.Context) (stri
 
 func (p *ProvisionClientBootstrap) getNodeBootstrappingFromClient(ctx context.Context, provisionProfile *models.ProvisionProfile, provisionHelperValues *models.ProvisionHelperValues, bootstrapToken string) (string, string, error) {
 	transport := httptransport.New(options.FromContext(ctx).NodeBootstrappingServerURL, "/", []string{"http"})
+
+	// Add Authorization Bearer token header
+	cred, err := azidentity.NewDefaultAzureCredential(nil)
+	if err != nil {
+		return "", "", err
+	}
+	token, err := cred.GetToken(ctx, policy.TokenRequestOptions{
+		Scopes: []string{"https://management.azure.com/.default"},
+	})
+	if err != nil {
+		return "", "", fmt.Errorf("failed to get token: %w", err)
+	}
+	transport.DefaultAuthentication = httptransport.BearerToken(token.Token)
 
 	logger := slog.New(slog.NewJSONHandler(os.Stdout, nil))
 	loggingClient := restlogger.NewLoggingClient(logger)


pkg/providers/imagefamily/nodebootstrappingclient.go
```diff
@@ -32,21 +35,75 @@ import (
...
+// However, this is never made clear in the interface of this layer nor its documentation, thus relying on that assumption may not be perfect.
+// Alternatively, we could try to pair our cache implementation with what's in Azure clients as below. However, the cost makes it arguably not worth it, given the current circumstances.
+// https://github.com/Azure/azure-sdk-for-go/blob/f72e2ad4f23b02eba6387dc31580c0e66333f2ae/sdk/internal/temporal/resource.go#L78-L140
+type tokenCache struct {
+	mu          sync.Mutex         // Mutex to ensure thread-safety
+	token       azcore.AccessToken // The cached token
+	refreshTime time.Time          // Time after which we should refresh the token
+}
+
+func (t *tokenCache) getTokenAtTime(ctx context.Context, credential azcore.TokenCredential, curTime time.Time) (azcore.AccessToken, error) {
```
matthchr: minor: Another way to test this where you control the time is to have the construction of tokenCache take an `timeProvider TimeProvider` interface that offers a single method which returns the time, which you can then pass a mock impl or a "real" impl to.

pkg/providers/imagefamily/nodebootstrappingclient.go
```diff
@@ -32,21 +35,75 @@ import (
 	"github.com/go-openapi/strfmt"
 )
 
+// This simple, short-lived caching is mainly to prevent excessive token fetches on a large scale up, rather than guaranteeing the optimal number of calls and 100% valid token.
+//
+// As of this time, for managed identity, token caching exists in the implementation beneath MSAL GetToken(...) as below. This means our caching is not needed if MSAL GetToken(...) caching is valid.
+// https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/b4b8bfc9569042572ccb82b648ea509075fadb74/apps/managedidentity/managedidentity.go#L318
+// However, this is never made clear in the interface of this layer nor its documentation, thus relying on that assumption may not be perfect.
```
matthchr: At this point I'd imagine it's guaranteed because it would probably be a breaking change for them to change it.

Also, won't we have a ton of other issues (in other clients) if MSAL token caching isn't there? For example in all of the other calls we make to Azure? Seems like maybe we can just rely on their caching?

If we're really worried, I have contacts on the MSAL/Go SDK team we can ask if it's guaranteed to be cached.

comtalyst: > it's guaranteed because it would probably be a breaking change for them to change it

That's a good point. Although I'm hesitant to join into the mess(?)...

> Also, won't we have a ton of other issues (in other clients) if MSAL token caching isn't there? For example in all of the other calls we make to Azure? 

Other Azure clients we are using also implemented their own caching, which means even if MSAL caching isn't there, there is another spare layer.

comtalyst: (discussed offline) the decision is to not implement caching, mainly from the expectation that it is unlikely to change while this code exists, but will keep the lock as there is no MSAL-layer lock to prevent concurrent requests to token provider.

pkg/providers/imagefamily/nodebootstrappingclient.go
```diff
@@ -32,21 +35,43 @@ import (
 	"github.com/go-openapi/strfmt"
 )
 
+// As of this time, for managed identity, token caching exists in the implementation beneath MSAL GetToken(...) as below. This means our caching is not needed if MSAL GetToken(...) caching is valid.
+// https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/b4b8bfc9569042572ccb82b648ea509075fadb74/apps/managedidentity/managedidentity.go#L318
+// However, this is never made clear in the interface of this layer nor its documentation, thus relying on that assumption may not be perfect. Still, the likelihood of change is low given history.
+// Alternatively, we could try to pair our cache implementation with what's in Azure clients as below. However, the cost makes it arguably not worth it, given the current circumstances.
+// https://github.com/Azure/azure-sdk-for-go/blob/f72e2ad4f23b02eba6387dc31580c0e66333f2ae/sdk/internal/temporal/resource.go#L78-L140
+type tokenProvider struct {
+	mu sync.Mutex // credential.GetToken(...) doesn't have locks, thus could have resulted in multiple token requests at the same time.
```
matthchr: minor: For the comment here, consider explaining why we're worried about this. Something like this:

> Multiple token requests at the same time is not a functional problem, but could produce more load on the upstream AAD/Entra service, which we want to avoid. See <issue here> for issue tracking this in MSAL.

Basically both:
1. Giving the reason we have concern
2. A pointer to where we might be looking at MSAL fixing this, eventually.

pkg/providers/imagefamily/nodebootstrappingclient.go
```diff
@@ -32,21 +35,43 @@ import (
...
+}
+
 // NodeBootstrappingClient implements the NodeBootstrappingAPI interface using the swagger-generated client.
 type NodeBootstrappingClient struct {
 	serverURL         string
 	subscriptionID    string
 	resourceGroupName string
 	resourceName      string
+	credential        azcore.TokenCredential
+	tokenProvider     *tokenProvider // Cache for Azure AD tokens to improve performance
```
matthchr: comment is not really accurate anymore (I mean there is a cache but its "inside" - unsure if worth calling out here)

pkg/providers/imagefamily/nodebootstrappingclient.go
```diff
@@ -58,6 +83,13 @@ func (c *NodeBootstrappingClient) Get(
 ) (types.NodeBootstrapping, error) {
 	transport := httptransport.New(c.serverURL, "/", []string{"http"})
```
matthchr, line 87: minor: I think that you could probably write this in such a way that you don't need to construct a new httptransport and client for every request.

Everything that gets fed into the client (the transport) is actually scoped at the `NodeBootstrappingClient` struct level, except the token, but you could write a token provider like:
```
// This interface we need to impl:
type ClientAuthInfoWriter interface {
	AuthenticateRequest(ClientRequest, strfmt.Registry) error
}


type BearerTokenCallbackAuthenticator struct {
	credential azcore.TokenCredential
}

func (b *BearerTokenCallbackAuthenticator) AuthenticateRequest(r ClientRequest, reg strfmt.Registry) error {
	token, err := c.tokenProvider.getToken(ctx, c.credential)
	if err != nil {
		return fmt.Errorf("failed to get token: %w", err)
	}
	return r.SetHeaderParam(runtime.HeaderAuthorization, "Bearer "+token)
}
```

The advantage of doing this is that it avoids recreating the transport + client + logger/etc for every request - which at the very least saves some GC cycles, and also just avoids the somewhat antipattern of "new client for each request".

Not something you need to fix in this PR necessarily, but it would be good to get it fixed in a future PR.

matthchr (dismissed):
Approved with some comments.

pkg/providers/imagefamily/nodebootstrappingclient.go
comtalyst: mb

matthchr (approved):

PR #919, opened 2025-05-27
fix: Fix getAKSClusterID function

matthchr:
DNSPrefix is not always 33 characters

This fixes a bug which could cause the ClusterID to be created/parsed incorrectly when the cluster has a non-default DNS prefix, resulting in incorrect azure.json entries in standalone mode (no impact on managed Karpenter via NAP)

* Unit tests

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
Fix bug which could cause the ClusterID to be created/parsed incorrectly when the cluster has a non-default DNS prefix, resulting in incorrect azure.json entries in standalone mode (no impact on managed Karpenter via NAP)
```

### DIFF
--- a/pkg/operator/options/options.go
+++ b/pkg/operator/options/options.go
@@ -26,6 +26,7 @@ import (
 	"math/rand"
 	"net/url"
 	"os"
+	"regexp"
 	"strings"
 
 	"k8s.io/apimachinery/pkg/util/sets"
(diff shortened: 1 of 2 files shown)


pkg/operator/options/options.go
```diff
@@ -175,11 +176,25 @@ func FromContext(ctx context.Context) *Options {
...
-// with the additional assumption of DNS prefix being the first 33 chars of FQDN
 func getAKSClusterID(apiServerFQDN string) string {
-	dnsPrefix := apiServerFQDN[:33]
+	// TODO: This doesn't work for clusters with BYO private dns zone with fqdnSubdomain. For those clusters
+	//       we need to hash the emptyString to get the cluster ID (that's what AKS RP does), but
+	//       you can't actually tell by examining a fqdn record if it's using dnsPrefix or fqdnSubdomain.
+	//       You ALMOST could, by looking for <dnsprefix>-<8 character random string>.privatelink.<region>.azmk8s.io, but
+	//       There's nothing stopping users from making their dns subdomain with an 8 character random string in the same
+	//       location that AKS put it.
+	//       I think it would be cleaner to just read this value directly from the nsg in the MC RG, rather than
```
matthchr, line 189: If there's agreement on this I can just delete this function in a followup PR and we can source it from the NSG name directly instead. The two main points of awkwardness there are:
1. It means that we have to add it to options later as opposed to computing it inline as part of options parse, which is a bit uglier.
2. If there are any cases where a cluster doesn't have an NSG, it wouldn't work. I don't think that ever happens though.

tallaxes (approved):

pkg/operator/options/options.go
matthchr: We need the NSG provider in the case that the cluster is BYO VNET, because we have to attach the NSG to the NIC ourselves in that case.

Of course, if we had a durable way to generate the NSG name (using a reliable cluster ID) we could have done that instead, but given that we already have access to the node resource group, and we don't (as far as I can tell) have a reliable way to get cluster ID due to the points I called out in this code comment, we do need the NSG provider so my thinking is more that we might as well use it all the time as opposed to only in BYO VNET.

I'll follow up with a PR to just remove this method entirely... may hold off on merging this one in the meantime.

matthchr:
Closing this in favor of https://github.com/Azure/karpenter-provider-azure/pull/1684, which does this more durably.

PR #610, opened 2024-12-04
Attach Agent Pool network security group to karpenter node nics..

johnthompson-ybor:
Fixes # [](https://github.com/Azure/karpenter-provider-azure/issues/562)

**Description**

Your logic for determining the cluster id seems to be wrong, and through experimentation it seems like it was just based on the name of the cluster.  In addition, while you add some code to generate the name of the agent pool security group, you didn't actually attach it anywhere.  I added the code to attach the security group to the NIC.

**How was this change tested?**

I tested it on one of my clusters, you should probably do some more thorough testing.

**Does this change impact docs?**
- [ ] Yes, PR includes docs updates
- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

PR #924, opened 2025-05-29
fix: Attach Managed NSG to NIC when in BYO VNET mode

### LINKED ISSUE #562: Load balancing doesn't work correctly in BYO VNET scenarios
### Version

__Karpenter Version:__ v0.7.0

__Kubernetes Version:__ v1.3.0

If you use BYO VNET for your AKS cluster with Standard Public LB, the ingress traffic only hits the nodes of the system nodepool and not the karpenter nodes. In this scenario AKS automatically creates a NSG and attaches it to the network interfaces of the system nodes and not to the karpenter nodes. This behavior ultimately blocks the ingress traffic to reach karpenter nodes. If the externalTrafficPolicy is set to Cluster, the traffic can reach the nodes through an extra hop from the system nodes but never directly. However if you set the externalTrafficPolicy to Local, there is no way to forward ingress traffic to the karpenter nodes. 

### Expected Behavior

An NSG should be associated with the karpenter nodes also, not just with the system nodes to enable ingress traffic from the public LB. 

### Actual Behavior

See above, due to the missing NSG, traffic can't reach the karpenter nodes directly. 

Create a cluster with karpenter.
Install NGINX ingress controller with externalTrafficPolicy set to local. 
Force the NGINX pods to move to the karpenter nodes. 
Create an ingress and a backend app. 
Try to reach the public LB externally. 

### Resource Specs and Logs

-

### Community Note

matthchr:
This mirrors what AKS does. This is not required in managed VNET mode because then the NSG is attached to the Subnet (created by AKS), so it doesn't need to be attached to each NIC individually.

Fixes #562.
This supersedes #610, which should be closed.

* Unit tests
* E2E tests (see newly added test). Full E2E run results: https://github.com/Azure/karpenter-provider-azure/actions/runs/15312805268
* Manually tested the scenario reported in #562 using `make az-all-custom-vnet`

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
Fix bug which caused load balancing to not work correctly in BYO VNET scenarios.
```

### DIFF
--- a/pkg/providers/instance/instance.go
+++ b/pkg/providers/instance/instance.go
@@ -306,6 +306,14 @@ func (p *DefaultProvider) newNetworkInterfaceForVM(opts *createNICOptions) armne
 	if err := opts.InstanceType.Requirements.Compatible(skuAcceleratedNetworkingRequirements); err == nil {
 		enableAcceleratedNetworking = true
 	}
+
+	var nsgRef *armnetwork.SecurityGroup
+	if opts.NetworkSecurityGroupID != "" {
+		nsgRef = &armnetwork.SecurityGroup{
+			ID: &opts.NetworkSecurityGroupID,
+		}
+	}
+
 	nic := armnetwork.Interface{
 		Location: lo.ToPtr(p.location),
 		Properties: &armnetwork.InterfacePropertiesFormat{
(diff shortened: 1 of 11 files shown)


pkg/providers/instance/instance.go
```diff
@@ -306,6 +306,14 @@ func (p *DefaultProvider) newNetworkInterfaceForVM(opts *createNICOptions) armne
 	if err := opts.InstanceType.Requirements.Compatible(skuAcceleratedNetworkingRequirements); err == nil {
 		enableAcceleratedNetworking = true
 	}
+
+	var nsgRef *armnetwork.SecurityGroup
```
matthchr, line 310: Note to self that I need to run manual test one more time to ensure the full LB scenario works E2E now.

PR #927, opened 2025-05-30
refactor: mimic bootstrapping client flow to other Azure clients

comtalyst:
Fixes # 

**Description**
In preparation for https://github.com/Azure/karpenter-provider-azure/pull/918. This should make the implementation more consistent with other Azure clients. 
In addition:
- Add a fake nodebootstrapping API
- Separate imagefamily/types.go into `types` sub-package, mainly to allow wider uses without import cycle

* added unit + acceptance tests
* NAP E2E
* Scale test

- [ ] Yes, issue opened: # 
- [X] No

**Release Note**

```release-note

```


pkg/fake/nodebootstrappingapi_test.go
```diff
@@ -0,0 +1,118 @@
...
+	distro := "AzureLinux"
+	orchestratorVersion := "1.26.0"
+	vnetCidrs := []string{"10.0.0.0/8"}
+	vnetSubnetID := "/subscriptions/sub/resourceGroups/rg/providers/Microsoft.Network/virtualNetworks/vnet/subnets/subnet"
+	mode := models.AgentPoolModeSystem
+	architecture := "amd64"
+	maxPods := int32(110)
+
+	return &models.ProvisionProfile{
+		Name:                &name,
```
tallaxes: `lo.ToPtr("test-node")` etc. would make this easier to read

pkg/providers/imagefamily/types/types.go
```diff
@@ -82,3 +72,13 @@ type NodeImageVersionsResponse struct {
...
+
+// NodeBootstrappingAPI defines the interface for retrieving node bootstrapping data
+type NodeBootstrappingAPI interface {
+	Get(ctx context.Context, parameters *models.ProvisionValues) (NodeBootstrapping, error)
+}
+
+type NodeBootstrapping struct {
+	CustomDataEncodedDehydratable string
+	CSEDehydratable               string
+}
```
tallaxes: Comments on the structure and the fields would be good

pkg/providers/imagefamily/types/types.go
tallaxes, line 1: I find this "types" subpackage a little awkward. Wondering what exactly created the import loop? And maybe "imagefamilytypes" would work better as a name?

pkg/providers/imagefamily/image.go
```diff
@@ -38,15 +39,16 @@ import (
...
+	kubernetesVersionCache    *cache.Cache
+	cm                        *pretty.ChangeMonitor
+	location                  string
+	kubernetesInterface       kubernetes.Interface
+	imageCache                *cache.Cache
+	nodeImagesCache           *cache.Cache
+	imageVersionsClient       types.CommunityGalleryImageVersionsAPI
+	subscription              string
+	NodeImageVersions         types.NodeImageVersionsAPI
+	NodeBootstrappingProvider types.NodeBootstrappingAPI
```
tallaxes: Does it need to be exported?

pkg/providers/imagefamily/customscriptsbootstrap/provisionclientbootstrap.go
```diff
@@ -68,15 +55,40 @@ type ProvisionClientBootstrap struct {
 	InstanceType                   *cloudprovider.InstanceType
 	StorageProfile                 string
 	ImageFamily                    string
+	NodeBootstrappingProvider      types.NodeBootstrappingAPI
```
tallaxes, line 58: I find it a bit awkward to have to flow provider API through multiple layers with what's otherwise just a set of configuration ...

pkg/providers/imagefamily/nodebootstrappingclient.go
```diff
@@ -0,0 +1,98 @@
...
+) (types.NodeBootstrapping, error) {
+	transport := httptransport.New(c.serverURL, "/", []string{"http"})
+
+	// Middleware logging
+	logger := slog.New(slog.NewJSONHandler(os.Stdout, nil))
+	loggingClient := restlogger.NewLoggingClient(logger)
+	transport.Transport = loggingClient.Transport
+
+	// Create the client
+	client := client.New(transport, strfmt.Default)
```
tallaxes, line 67: Is there a reason we need to construct client for every call?

tallaxes (commented):
Overall LGTM, minor comments. 

One thing I do find a bit awkward is flowing NodeBootstrappingProvider API through multiple layers, sometimes with what's otherwise just a set of configuration values. Wondering if there is way instead to pass it directly into where it is actually needed in the end? (Also wondering if simplifying the flow might eliminate the need for types subpackage, as a side effect ...)

Also, should there be at least one unit/acceptance test exercising end-to-end flow with the fake NodeBootstrappingProvider?

pkg/providers/imagefamily/types/types.go
comtalyst: `imagefamily` is importing `customscriptsbootstrap` to use `GetCSE...(...)`. `customscriptsbootstrap` wants to use `NodeBootstrappingAPI`, which cannot be in `imagefamily`, but can be in `imagefamily/types`

An alternative is to abolish `customscriptsbootstrap` package and merge it into `imagefamily`, which makes sense given that the current pattern is built on top of Scriptless bootstrapping. But I don't think it is worth doing until we switch to use Azure client, also given we still need to maintain Scriptless.

For the name, not sure if that's better. The folder would have to be `imagefamily/imagefamilytypes` if we are to remain this hierarchy.

pkg/providers/imagefamily/image.go
comtalyst: No, so does `NodeImageVersions`. Changed both.

pkg/providers/imagefamily/customscriptsbootstrap/provisionclientbootstrap.go
comtalyst: (see another comment)

pkg/providers/imagefamily/nodebootstrappingclient.go
comtalyst: In the next PR, #918, the safer way to inject the token here is to reconstruct the transport, which is used to construct the client. Note that client reconstruction per request is also the current behavior.

pkg/providers/imagefamily/customscriptsbootstrap/provisionclientbootstrap.go
tallaxes: Which one? :-) One other comment addresses the import loop, but not the overall flow ...

pkg/providers/imagefamily/types/types.go
tallaxes: > customscriptsbootstrap wants to use NodeBootstrappingAPI

Not flowing NodeBootstrappingAPI through might remove the need for its use by customscriptbootstrap?

pkg/providers/imagefamily/nodebootstrappingclient.go
tallaxes: So, is the answer Yes we have to (or it is best practice / safer to) reconstruct the client per call? I find this a bit surprising.

pkg/providers/imagefamily/types/types.go
comtalyst: `customscriptsbootstrap` is modeled after `bootstrap` (i.e., Scriptless). Their responsibility is to retrieve bootstrapping data, which is custom data, and CSE for `customscriptsbootstrap` case. NodeBootstrappingAPI provides them, so it needs to be flow through as long as we are relying on this model. `bootstrap` doesn't need it given we have the "template".
The alternative I mentioned is to change this model. Currently, I don't think it is good to spend time on that. An example concern is how self-hosted and NAP may further diverges due to the further loss of `customscriptsbootstrap`-`bootstrap` interface parallelism. Well, unless we think flowing the API through is a worse pattern.

pkg/providers/imagefamily/nodebootstrappingclient.go
comtalyst: It is "an apparently better choice", rather than "yes, we have to". That is due to 1) token needs to be "injected" to transport object, which is used to construct the client; injection without reconstruction is not safe as the old token might already be used/copied at that point; and even if it is a no, there is no guarantee; 2) this is the current behavior in which there is no motivation to make change (or research on it) at this point

pkg/providers/imagefamily/customscriptsbootstrap/provisionclientbootstrap.go
comtalyst: Yes, that one. One of the contributors to the import loop is this "awkward" flow.

pkg/providers/imagefamily/types/types.go
tallaxes: Hmm, the actual retrieval (call to GetCustomDataAndCSE) is done from launch template, which is already aware of the differences between the two variants, and could pass the NodeBootstrappingAPI in, which would simplify the flow. But your encapsulation/responsibility argument has merit too. Approving, happy to discuss further later.

pkg/providers/imagefamily/nodebootstrappingclient.go
tallaxes: Would be interested in more reasoning, let's discuss offline, but not a blocker.

tallaxes (approved):

pkg/providers/imagefamily/resolver.go
```diff
@@ -156,6 +158,7 @@ func (r *defaultResolver) Resolve(
 			instanceType,
 			imageDistro,
 			storageProfile,
+			r.imageProvider.nodeBootstrappingProvider,
```
charliedmcb, line 161: nodeBootstrappingProvider should be a field on the resolver, if used there, not the `imageProvider`.


The imageProvider has no reason to have it as the field, and accessing a private field on the other doesn't seem great for isolation.

I think the resolver and provider should possibly be split into separate packages, but that's beyond this. Just would make it more clear

PR #928, opened 2025-06-01
fix: Propagate missing kubelet flags to node bootstrap variables

nandorKollar:
Fixes # 

**Description**

This change propagates the following missing flags to node bootsrap vars: 
```
--cpu-manager-policy
--topology-manager-policy
--container-log-max-size
--container-log-max-files
--pod-max-pids
--allowed-unsafe-sysctls
```

* Ran `make test`, every test passed.
* Verified that a selected kubelet flag (allowed-unsafe-sysctls) on a live cluster, the flag worked.
* Added a new unit test `KubeletConfigToMap` function.

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
NONE
```

### DIFF
--- a/pkg/providers/imagefamily/bootstrap/aksbootstrap.go
+++ b/pkg/providers/imagefamily/bootstrap/aksbootstrap.go
@@ -350,7 +350,7 @@ func (a AKS) applyOptions(nbv *NodeBootstrapVariables) {
 	nodeclaimKubeletConfig := KubeletConfigToMap(a.KubeletConfig)
 	kubeletFlags = lo.Assign(kubeletFlags, nodeclaimKubeletConfig)
 
-	// striginify kubelet flags (including taints)
+	// stringify kubelet flags (including taints)
 	nbv.KubeletFlags = strings.Join(lo.MapToSlice(kubeletFlags, func(k, v string) string {
 		return fmt.Sprintf("%s=%s", k, v)
 	}), " ")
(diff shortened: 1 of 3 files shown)


tallaxes (commented):
Looks good, thank you! A couple of asks:
* Please also update corresponding bootstrap acceptance/unit tests in instancetype/suite_test.go (not a great location, I know); look for ExpectKubeletFlags
* Please compare value propagation to the logic we use in provisioningclientbootstrap, which does a couple of extra checks/conversions. I don't think these would actually be needed in this context, but good to double check

nandorKollar:
> * // nolint: gocyclo

Thanks for the review, I can easily address the last one, will look at the rest later. I think splitting the method doesn't make much sense, as it has a rather well defined, single purpose, so adding linter exception sounds a better option. For some reason `make presubmit` always triggers a code generation, which results in compile error (I don't use Codespaces):
```
pkg/providers/imagefamily/customscriptsbootstrap/provisionclientbootstrap.go:37:2: could not import github.com/Azure/karpenter-provider-azure/pkg/provisionclients/client/operations (-: # github.com/Azure/karpenter-provider-azure/pkg/provisionclients/client/operations
pkg/provisionclients/client/operations/node_bootstrapping_get_responses_override.go:26:39: method NodeBootstrappingGetDefault.readResponse already declared at pkg/provisionclients/client/operations/node_bootstrapping_get_responses.go:192:39) (typecheck)
	"github.com/Azure/karpenter-provider-azure/pkg/provisionclients/client/operations"
	^
pkg/providers/instance/suite_test.go:196:18: undefined: ExpectDecodedCustomData (typecheck)

```

Do you happen to know what could be the problem?

nandorKollar:
@tallaxes as the added flags are now covered pretty well in `instancetype/suite_test.go`, do you think my extra test case for `KubeletConfigToMap` provides any extra value? If not, I can remove it. Also, do you mind if I change the visibility of `KubeletConfigToMap` to private (kubeletConfigToMap)? I feel that what it does is rather an implementation detail, better to keep it private, so that nobody uses it by accident.

tallaxes:
> always triggers a code generation, which results in compile error 

Not sure, but would suspect it is the `make az-swagger-generate-clients-raw` step? Maybe check that it does not pick up some other version of swagger in your environment?

The latest CI failure is on order of imports in aksboostrap_test.go ("testing" and "time" should be right after "fmt") - but hopefully you will see it locally once you fix (or bypass) the above problem.

> do you think my extra test case for KubeletConfigToMap provides any extra value? If not, I can remove it.

Your call :-) Yes, we generally think the lower-level tests have less value once acceptance unit tests that exercise the logic E2E are in place, but no harm leaving it. Could also take a look at test coverage to see if all code paths are covered well by acceptance tests.

> Also, do you mind if I change the visibility of KubeletConfigToMap to private (kubeletConfigToMap)?

+1, makes sense

nandorKollar:
> > always triggers a code generation, which results in compile error
> 
> Not sure, but would suspect it is the `make az-swagger-generate-clients-raw` step? Maybe check that it does not pick up some other version of swagger in your environment?
> 
> The latest CI failure is on order of imports in aksboostrap_test.go ("testing" and "time" should be right after "fmt") - but hopefully you will see it locally once you fix (or bypass) the above problem.
> 
> > do you think my extra test case for KubeletConfigToMap provides any extra value? If not, I can remove it.
> 
> Your call :-) Yes, we generally think the lower-level tests have less value once acceptance unit tests that exercise the logic E2E are in place, but no harm leaving it. Could also take a look at test coverage to see if all code paths are covered well by acceptance tests.
> 
> > Also, do you mind if I change the visibility of KubeletConfigToMap to private (kubeletConfigToMap)?
> 
> +1, makes sense

Thanks, just realised that there are still style problems, fixed it. Also change KubeletConfigToMap to a private method.

PR #934, opened 2025-06-02
feat(perf): pprof & monitoring config refresh

tallaxes:
**Description**

Refresh pprof & monitoring dev config. Manual/scripted/programmatic dump + `go tool pprof` is nice for control & automation; Grafana w/Pyroscope best for continuous / interactive profiling (sans graph & source). Sample use:

```
make az-all-perftest
make az-mon-access  # and open Grafana for continuous profiling
# run your test (from a different terminal)
make az-pprof # and open pprof UI for snapshot profiling
```

In Grafana, point dashboards at Karpenter; for profiles go to Drilldown | Profiles.

#### Perf test cluster / config

`make az-all-perftest` creates test cluster, enables profiling, deploys monitoring stack, runs karpenter, taints system nodes, deploys test nodepool and workload. Test cluster: `make az-mkaks-perftest` - Azure Overlay (w/o Cilium), 2 larger VMs for system pool, larger pods CIDR.

#### Test & workloads

NodePool example/v1/perftest.yaml. Note that it has both on-demand and spot nodepools; you may want to enable just one of them. Manual scaling of inflate workload should be enough for pod-per-node (workload requests 1 CPU, NodePool is constrained to 2-CPU VMs).

Alternatively, one could run any of the existing E2E tests. Utilization (or subset) is good candidate. Scale suite would be even better. We could consider instrumenting those to capture profiles (either scripted externally, or dumped programmatically).

#### pprof instrumentation

Already in place, nothing extra is needed, just needs to be enabled with `ENABLE_PROFILING`. (`make az-pprof-enable` takes care of this.) We could consider dumping profiles automatically from perf-oriented E2E tests. Pyroscope supports push instrumentation, but we likely don't need it.

#### Grafana - dashboards

Several dashboards are pre-deployed (you can add more as needed):
* Karpenter Capacity & Performance dashboards from AWS provider
* Karpenter Overview / Activity / Performance dashboards from community. These are actually done right, using jsonnet / mixins, would be great candidates for adoption & iteration (including for upstream)
* Kubernetes (just Pods)
* Go runtime

#### Grafana - Pyroscope

Pyroscope (for profiling) is deployed and configured with the rest. Alloy (which is required for pull consumption of profiles) is deployed automatically as a subcomponent. Automatic scraping for Karpenter is configured via pod annotations. Source (GitHub) integration does not work (either too new, or only available in Grafana Cloud, or I configured it wrong) - use `go tool pprof` instead if desired. The relevant views in Grafana are under Drilldown | Profiles.

* `make az-all-perftest`

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/Makefile-az.mk
+++ b/Makefile-az.mk
@@ -31,6 +31,8 @@ az-all-cni-overlay:  az-login az-create-workload-msi az-mkaks-overlay     az-cre
 az-all-perftest:     az-login az-create-workload-msi az-mkaks-perftest    az-create-federated-cred az-perm               az-perm-acr az-configure-values
 	$(MAKE) az-mon-deploy
 	$(MAKE) az-pprof-enable
+	yq -i '.manifests.helm.releases[0].overrides.controller.resources.requests = {"cpu":4,"memory":"3Gi"}' skaffold.yaml
+	yq -i '.manifests.helm.releases[0].overrides.controller.resources.limits   = {"cpu":4,"memory":"3Gi"}' skaffold.yaml
 	$(MAKE) az-run
 	$(MAKE) az-taintsystemnodes
 	kubectl apply -f examples/v1/perftest.yaml


rakechill (approved):

Makefile-az.mk
```diff
@@ -361,11 +385,15 @@ az-argvmlist: ## List current VMs owned by Karpenter
...
-	yq -i '.manifests.helm.releases[0].overrides.controller.env += [{"name":"ENABLE_PROFILING","value":"true"}]' skaffold.yaml
+	yq -i '.controller.env += [{"name":"ENABLE_PROFILING","value":"true"}]' karpenter-values.yaml
 
+# remove -source_path=/go/pkg/mod to focus source on provider
+SOURCE=-source_path=/go/pkg/mod -trim_path=github.com/Azure/karpenter-provider-azure
 az-pprof: ## Profile
-	kubectl port-forward service/karpenter -n "${KARPENTER_NAMESPACE}" 8000 &
-	sleep 2 && go tool pprof -http 0.0.0.0:9000 localhost:8000/debug/pprof/heap
+	kubectl port-forward service/karpenter -n "${KARPENTER_NAMESPACE}" 8080 &
+	sleep 2
```
aagusuab, line 394: Not a blocker, but out of curiosity, did it make a difference to remove the && from the script?

tallaxes: No functional difference, just clarity. `cmd1 && cmd2` is an idiomatic way to run cmd2 only if cmd1 succeeds, which is not the intent here, so I was abusing it a bit; cleaned up.

PR #935, opened 2025-06-02
WIP: Adding sku scaling test

aagusuab:
Fixes # 

**Description**

*

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/providers/instancetype/suite_test.go
+++ b/pkg/providers/instancetype/suite_test.go
@@ -1683,9 +1683,10 @@ var _ = Describe("InstanceType Provider", func() {
 			Expect(workingSKUs).ToNot(BeEmpty())
 
 			// Test with a few specific SKUs
-			testSKUs := []string{"Standard_D2_v2", "Standard_D4_v2", "Standard_D8_v2"}
+			testSKUs := []string{"Standard_A2", "Standard_B1s"}
 
 			for _, skuName := range testSKUs {
+				fmt.Println("Current SKU:", skuName)
 				skuNodePool := coretest.NodePool(karpv1.NodePool{
 					Spec: karpv1.NodePoolSpec{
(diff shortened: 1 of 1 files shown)


pkg/providers/instancetype/sku_scaling_test.go
```diff
@@ -0,0 +1,163 @@
...
+				})
+
+				// Create the node pool and pod
+				ExpectApplied(ctx, env.Client, skuNodePool, pod)
+
+				// Expect the pod to be scheduled
+				ExpectProvisioned(ctx, env.Client, cluster, cloudProvider, provisioner, pod)
+				node := ExpectScheduled(ctx, env.Client, pod)
+				Expect(node.Labels[v1beta1.LabelSKUName]).To(Equal(*sku.Name))
+
```
Bryce-Soghigian: Shouldn't we do this in parallel?

aagusuab: Yes, and we will try to do just the smallest skus in the VM family to reduce the time.

pkg/providers/instancetype/suite_test.go
```diff
@@ -1679,6 +1675,57 @@ var _ = Describe("InstanceType Provider", func() {
...
+											Key:      v1beta1.LabelSKUName,
+											Operator: v1.NodeSelectorOpIn,
+											Values:   []string{skuName},
+										},
+									},
+								},
+							},
+						},
+					},
+				})
```
Bryce-Soghigian, line 1711: why create nodepools instead of creating a nodeclaim directly? 

Also should we be testing that these skus work with specific image families for both ubuntu + azure linux? Some GPUs are only supported for ubuntu for example?

PR #747, opened 2025-03-22
fix: Fix Gt and Lt for EphemerealOSMaxSize (Fixes #683)

Sturgelose:
Fixes #683 

**Description**

Context: https://github.com/Azure/karpenter-provider-azure/issues/683#issuecomment-2729583707

GT and LT operators do not work for `karpenter.azure.com/sku-storage-ephemeralos-maxsize`.
The reason is (or at least seems to be for me) that K8s parses the annotation's value as integer and that is a float. So, the GT and LT won't work...

> Gt and Lt operators will not work with non-integer values. If the given value doesn't parse as an integer, the pod will fail to get scheduled. Also, Gt and Lt are not available for podAffinity.

Issue request for float support in K8s core here: https://github.com/kubernetes/kubernetes/issues/44932

I guess that the annotation should be (at least for now) an integer, and not a float... Or this filtering won't work at all in current K8s versions!

This PR includes changes to write the annotation as integer. When reading the annotation, it is already parsed as int32!

This is my first PR in this repo. If needed I can provide an update to the docs!

**How was this change tested?**

* Updated the tests in order to make sure it reflects integers instead of floats.

**Does this change impact docs?**
- [ ] Yes, PR includes docs updates
- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note
GT and LT operators are honored and do work for `karpenter.azure.com/sku-storage-ephemeralos-maxsize` annotation. This changes the type from float64 to int32 in the annotation.
```

PR #937, opened 2025-06-03
fix: should select placement ourselves rather than delegating placement to CRP

Bryce-Soghigian:
**Description**
AKS + Karpenter have deviated in how they select the ephemeral disk, and determine it can be used. 

AKS prioritizes specific placement types, and determines placement itself, rather than delegating it to CRP. The order of placement priority is as follows. 

If supported for sku and large enough are applied to all os disk placements. 
1. NVME disk is used first, and is the only option considered by v6 skus
2. Cache disk is used and prefered over temp disk 
3. temp disk 
4. Fall back on a Managed Disk

Karpenter will copy the logic of using the max size to determine placement that AKS traditionally does, but it will not copy the preference ordering of Cache disk over temp disk, and not yet solve the NVME + V6 sku stuff. 

This PR will just follow in AKS's footsteps, and select placement ourselves rather than delegating to CRP, as some of the data for B20ms for example is incorrect for the placement type and requires we do that selection.  

This PR also fixes the Gt + Lt node selector operations for requirements: see [this](https://github.com/kubernetes/kubernetes/issues/44932) as requested in https://github.com/Azure/karpenter-provider-azure/pull/747

- make battletest

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


pkg/providers/instancetype/suite_test.go
```diff
@@ -1047,7 +1108,7 @@ var _ = Describe("InstanceType Provider", func() {
 				v1beta1.LabelSKUName:                      "Standard_NC24ads_A100_v4",
 				v1beta1.LabelSKUFamily:                    "N",
 				v1beta1.LabelSKUVersion:                   "4",
-				v1beta1.LabelSKUStorageEphemeralOSMaxSize: "53.6870912",
+				v1beta1.LabelSKUStorageEphemeralOSMaxSize: "53",
```
Bryce-Soghigian, line 1111: floats aren't supported for Lt + Gt node selector operations, so we should just use int for now.

Bryce-Soghigian:
Closing in favor of PR that has v6 support too

PR #939, opened 2025-06-03
fix: Fix double role assignment in custom-vnet case

matthchr:
Removed the (I believe unneeded) NetworkContributor at cluster scope in in the non-byo-vnet case.

This was found due to us double-assigning the same role assignment in the BYO case, which doesn't work on unmanaged devices due to a bug in az cli, see https://github.com/Azure/azure-cli/issues/31579

* E2E

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note
NONE
```


matthchr:
See E2E here: https://github.com/Azure/karpenter-provider-azure/actions/runs/15428927032.

edit: This E2E had some failures but I don't think any that indicate an issue with this change (just flakes?)

Bryce-Soghigian (approved):

PR #940, opened 2025-06-04
fix: supporting Ephemeral OS Disk placement

Bryce-Soghigian:
**Description**
AKS + Karpenter have deviated in how they select the ephemeral disk, and determine it can be used. 

AKS prioritizes specific placement types, and determines placement itself, rather than delegating it to CRP. The order of placement priority is as follows. 

If supported for sku and large enough are applied to all os disk placements. 
1. NVME disk is used first, and is the only option considered by v6 skus
2. Cache disk is used and prefered over temp disk 
3. temp disk 
4. Fall back on a Managed Disk

Karpenter will copy the logic of using the max size to determine placement that AKS traditionally does, but it will not copy the preference ordering of Cache disk over temp disk. 

This PR will just follow in AKS's footsteps, and select placement ourselves rather than delegating to CRP, as some of the data for B20ms for example is incorrect for the placement type and requires we do that selection.  

This PR also fixes the Gt + Lt node selector operations for requirements: see [this](https://github.com/kubernetes/kubernetes/issues/44932) as requested in https://github.com/Azure/karpenter-provider-azure/pull/747

- make battletest
- E2E: https://github.com/Azure/karpenter-provider-azure/actions/runs/15842389269

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


test/suites/nodeclaim/eph_osdisk_test.go
```diff
@@ -0,0 +1,78 @@
...
+		env.ExpectCreated(nodeClass, nodePool, pod)
+		env.EventuallyExpectHealthy(pod)
+		env.ExpectCreatedNodeCount("==", 1)
+
+		vm := env.GetVM(pod.Spec.NodeName)
+		Expect(vm.Properties.StorageProfile.OSDisk).ToNot(BeNil())
+		Expect(vm.Properties.StorageProfile.OSDisk.DiffDiskSettings).ToNot(BeNil())
+		// We should be specifying os disk placement now
+		Expect(vm.Properties.StorageProfile.OSDisk.DiffDiskSettings.Placement).ToNot(BeNil())
+		Expect(string(lo.FromPtr(vm.Properties.StorageProfile.OSDisk.DiffDiskSettings.Option))).To(Equal("Local"))
```
tallaxes, line 52:
```suggestion
		Expect(string(lo.FromPtr(vm.Properties.StorageProfile.OSDisk.DiffDiskSettings.Option))).To(Equal(armcompute.DiffDiskOptionsLocal))
```

test/suites/nodeclaim/eph_osdisk_test.go
tallaxes, line 1: Why is this under nodeclaim suite?

pkg/providers/instancetype/instancetypes.go
```diff
@@ -349,20 +369,24 @@ func (p *DefaultProvider) isConfidential(sku *skewer.SKU) bool {
...
 	if sku == nil {
-		return 0
+		return 0, armcompute.DiffDiskPlacementResourceDisk
 	}
 	maxCachedDiskBytes, _ := sku.MaxCachedDiskBytes()
 	maxResourceVolumeMB, _ := sku.MaxResourceVolumeMB() // NOTE: this is a misnomer, MB is actually MiB, hence the conversion below
+	maxNvmeDiskSize, _ := sku.NvmeDiskSizeInMiB()
+
+	if maxNvmeDiskSize > 0 {
+		return int32(maxNvmeDiskSize / 1024), armcompute.DiffDiskPlacementNvmeDisk
```
tallaxes: This has to also check `SupportedEphemeralOSDiskPlacements` to be able to pick NVME

pkg/providers/instancetype/instancetypes.go
```diff
@@ -349,20 +369,24 @@ func (p *DefaultProvider) isConfidential(sku *skewer.SKU) bool {
 // For Ephemeral disk creation, CRP will use the larger of the two values to ensure we have enough space for the ephemeral disk.
 // Note that generally only older SKUs use the Temp Disk space for ephemeral disks, and newer SKUs use the Cached Disk in most cases.
 // The ephemeral OS disk is created with the free space of the larger of the two values in that place.
-func MaxEphemeralOSDiskSizeGB(sku *skewer.SKU) float64 {
+func MaxEphemeralOSDiskSizeGB(sku *skewer.SKU) (sizeGB int32, placement armcompute.DiffDiskPlacement) {
```
tallaxes: The comment is outdated, and the name of the function should reflect the new functionality

pkg/providers/instancetype/instancetypes.go
```diff
@@ -349,20 +369,24 @@ func (p *DefaultProvider) isConfidential(sku *skewer.SKU) bool {
...
 	if sku == nil {
-		return 0
+		return 0, armcompute.DiffDiskPlacementResourceDisk
```
tallaxes: Returning placement on error does not make sense and is confusing, maybe nil instead?

pkg/providers/instancetype/instancetypes.go
tallaxes: Also, if NVME is supported Cache/Temp won't even be considered - because they are not supported for SKUs that support NVMe. This warrants a comment.

pkg/providers/instancetype/instancetypes.go
```diff
@@ -349,20 +369,24 @@ func (p *DefaultProvider) isConfidential(sku *skewer.SKU) bool {
...
 
 	maxResourceVolumeBytes := maxResourceVolumeMB * int64(units.Mebibyte)
 	maxDiskBytes := math.Max(float64(maxCachedDiskBytes), float64(maxResourceVolumeBytes))
 	if maxDiskBytes == 0 {
-		return 0
+		return 0, armcompute.DiffDiskPlacementResourceDisk
 	}
-	// convert bytes to GB
-	return maxDiskBytes / float64(units.Gigabyte)
+	return int32(maxDiskBytes / float64(units.Gigabyte)), lo.Ternary(maxDiskBytes == float64(maxResourceVolumeBytes), armcompute.DiffDiskPlacementResourceDisk, armcompute.DiffDiskPlacementCacheDisk)
```
tallaxes: Please extract placement determination for readability - at least into its own var, but probaby best to determine size and placement together. (Also, AKS appears to pick Cache over Resource, can we align with that?)

pkg/providers/instancetype/instancetype.go
```diff
@@ -187,8 +187,10 @@ func computeRequirements(sku *skewer.SKU, vmsize *skewer.VMSizeType, architectur
 }
 
 func setRequirementsEphemeralOSDiskSupported(requirements scheduling.Requirements, sku *skewer.SKU, vmsize *skewer.VMSizeType) {
+	// Dlds_v5 && E4d_v5, say they support ephemeral disk, but they do not
 	if sku.IsEphemeralOSDiskSupported() && vmsize.Series != "Dlds_v5" { // Dlds_v5 does not support ephemeral OS disk, contrary to what it claims
```
tallaxes: The comment does not match the logic. Also, can we double check the facts? I am not seeing these exceptions in AKS ...

pkg/providers/instancetype/instancetypes.go
```diff
@@ -349,20 +369,24 @@ func (p *DefaultProvider) isConfidential(sku *skewer.SKU) bool {
...
 	maxCachedDiskBytes, _ := sku.MaxCachedDiskBytes()
 	maxResourceVolumeMB, _ := sku.MaxResourceVolumeMB() // NOTE: this is a misnomer, MB is actually MiB, hence the conversion below
+	maxNvmeDiskSize, _ := sku.NvmeDiskSizeInMiB()
+
+	if maxNvmeDiskSize > 0 {
+		return int32(maxNvmeDiskSize / 1024), armcompute.DiffDiskPlacementNvmeDisk
+	}
 
 	maxResourceVolumeBytes := maxResourceVolumeMB * int64(units.Mebibyte)
 	maxDiskBytes := math.Max(float64(maxCachedDiskBytes), float64(maxResourceVolumeBytes))
```
tallaxes: It would be really nice to get rid of any float-based logic around OS disk size ...

pkg/providers/instancetype/instancetypes.go
```diff
@@ -349,20 +369,24 @@ func (p *DefaultProvider) isConfidential(sku *skewer.SKU) bool {
...
 	if maxDiskBytes == 0 {
-		return 0
+		return 0, armcompute.DiffDiskPlacementResourceDisk
```
tallaxes: Same comment, returning placement (even if fake/unused) is confusing here

pkg/providers/launchtemplate/parameters/types.go
```diff
@@ -55,6 +56,8 @@ type Parameters struct {
 	ScriptlessCustomData           bootstrap.Bootstrapper
 	CustomScriptsNodeBootstrapping customscriptsbootstrap.Bootstrapper
 	ImageID                        string
-	StorageProfile                 string
+	StorageProfileDiskType         string
+	StorageProfilePlacement        armcompute.DiffDiskPlacement
+	StorageProfileSizeGB           float64
```
tallaxes: Why float?

pkg/providers/imagefamily/resolver.go
```diff
@@ -151,16 +153,35 @@ func (r *defaultResolver) Resolve(
...
 		),
-		ImageID:        imageID,
-		StorageProfile: storageProfile,
-		IsWindows:      false, // TODO(Windows)
+		StorageProfileDiskType:  diskType,
+		StorageProfilePlacement: placement,
+
+		// TODO: We could potentially use the instance type to do defaulting like
+		// traditional AKS, so putting this here along with the other settings
+		StorageProfileSizeGB: float64(lo.FromPtr(nodeClass.Spec.OSDiskSizeGB)),
```
tallaxes: Is there any way we can not add any more to the template? For one thing, we should be going away from using it altogether. Perhaps more importantly, in this case, it distributes the (already less than obvious) size & placement logic and settings between many different places ...

pkg/providers/instance/instance.go
```diff
@@ -468,11 +468,15 @@ func newVMObject(opts *createVMOptions) *armcompute.VirtualMachine {
 }
 
 // setVMPropertiesOSDiskType enables ephemeral os disk for instance types that support it
-func setVMPropertiesOSDiskType(vmProperties *armcompute.VirtualMachineProperties, storageProfile string) {
-	if storageProfile == "Ephemeral" {
+func setVMPropertiesOSDiskType(vmProperties *armcompute.VirtualMachineProperties, launchTemplate *launchtemplate.Template) {
+	diskType, placement, sizeGB := launchTemplate.StorageProfileDiskType, launchTemplate.StorageProfilePlacement, launchTemplate.StorageProfileSizeGB
+
+	if diskType == "Ephemeral" {
+		vmProperties.StorageProfile.OSDisk.DiskSizeGB = lo.ToPtr(int32(sizeGB))
```
tallaxes: DiskSizeGB is already assigned in newVMObject? If this is specific override for ephemeral, it deserves a comment - but I think I would just remove the size aspect here altogether for now

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -53,7 +54,9 @@ type Template struct {
 	CustomScriptsCustomData string
 	CustomScriptsCSE        string
 	IsWindows               bool
-	StorageProfile          string
+	StorageProfileDiskType  string
```
tallaxes: Can this be just bool, like StorageProfileIsEphemeral?

pkg/providers/instance/instance.go
```diff
@@ -468,11 +468,15 @@ func newVMObject(opts *createVMOptions) *armcompute.VirtualMachine {
 }
 
 // setVMPropertiesOSDiskType enables ephemeral os disk for instance types that support it
-func setVMPropertiesOSDiskType(vmProperties *armcompute.VirtualMachineProperties, storageProfile string) {
-	if storageProfile == "Ephemeral" {
+func setVMPropertiesOSDiskType(vmProperties *armcompute.VirtualMachineProperties, launchTemplate *launchtemplate.Template) {
+	diskType, placement, sizeGB := launchTemplate.StorageProfileDiskType, launchTemplate.StorageProfilePlacement, launchTemplate.StorageProfileSizeGB
+
+	if diskType == "Ephemeral" {
```
tallaxes: "Ephemeral" needs to be at least a shared constant here, but probably best to use boolean instead

pkg/providers/imagefamily/resolver.go
```diff
@@ -151,16 +153,35 @@ func (r *defaultResolver) Resolve(
...
+		return "", "", err
+	}
+
+	_, placement = instancetype.MaxEphemeralOSDiskSizeGB(sku)
+
+	if instancetype.UseEphemeralDisk(sku, nodeClass) {
+		return "Ephemeral", placement, nil
+	}
+	return "ManagedDisks", placement, nil
+}
```
tallaxes: "Ephemeral"/"ManagedDisks" should be at least shared constants, though I think in this context I would recommend just using a bool. If CSE bootstrap nees to translate it to corresponding strings - that logic would be contained to that package.

hack/find_sub.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+
+SUBSCRIPTIONS=()
+
+REGIONS=($(az account list-locations --query "[?metadata.regionType=='Physical'].name" -o tsv))
```
tallaxes: https://www.shellcheck.net/wiki/SC2207

pkg/providers/imagefamily/resolver.go
tallaxes: For size defaulting, good callout - this should be part of short description of how this logic difference from AKS - but the design & implementation would require more thought and I would suggest not trying to anticipate it here.

pkg/providers/imagefamily/resolver.go
```diff
@@ -151,16 +153,35 @@ func (r *defaultResolver) Resolve(
...
+		StorageProfileSizeGB: float64(lo.FromPtr(nodeClass.Spec.OSDiskSizeGB)),
+		ImageID:              imageID,
+		IsWindows:            false, // TODO(Windows)
 	}
 
 	return template, nil
 }
 
+func (r *defaultResolver) getStorageProfile(ctx context.Context, instanceType *cloudprovider.InstanceType, nodeClass *v1beta1.AKSNodeClass) (diskType string, placement armcompute.DiffDiskPlacement, err error) {
+	sku, err := r.instanceTypeProvider.Get(ctx, nodeClass, instanceType.Name)
```
tallaxes, line 184: It seems unnecessary to have to flow intanceTypeProvider in here. Is there a problem with existing implementation, which gets the data it needs from just instance type and nodeclass?

pkg/providers/instance/instance.go
tallaxes: Also, it looks like there is an overall limit of 2040GiB on ephemeral disk size that we should be checking as well? (And using managed disk otherwise?)

pkg/providers/instancetype/suite_test.go
```diff
@@ -551,6 +568,44 @@ var _ = Describe("InstanceType Provider", func() {
...
 		})
+		It("should fail to provision if ephemeral disk ask for is too large", func() {
+			nodePool.Spec.Template.Spec.Requirements = append(nodePool.Spec.Template.Spec.Requirements, karpv1.NodeSelectorRequirementWithMinValues{
+				NodeSelectorRequirement: v1.NodeSelectorRequirement{
+					Key:      v1beta1.LabelSKUStorageEphemeralOSMaxSize,
+					Operator: v1.NodeSelectorOpGt,
+					Values:   []string{"100000"},
+				},
+			}) // No InstanceType will match this requirement
+			nodeClass.Spec.OSDiskSizeGB = lo.ToPtr[int32](100001)
```
tallaxes: Does this setting matter in this context?

pkg/providers/instancetype/suite_test.go
```diff
@@ -551,6 +568,44 @@ var _ = Describe("InstanceType Provider", func() {
...
+
+			ExpectApplied(ctx, env.Client, nodePool, nodeClass)
+			pod := coretest.UnschedulablePod()
+			ExpectProvisioned(ctx, env.Client, cluster, cloudProvider, coreProvisioner, pod)
+			ExpectNotScheduled(ctx, env.Client, pod)
+
+		})
+
+		It("should select an ephemeral disk if LabelSKUStorageEphemeralOSMaxSize is set", func() {
```
tallaxes: ... and OS disk size fits?

pkg/providers/instancetype/suite_test.go
```diff
@@ -551,6 +568,44 @@ var _ = Describe("InstanceType Provider", func() {
...
+			ExpectApplied(ctx, env.Client, nodePool, nodeClass)
+			pod := coretest.UnschedulablePod()
+			ExpectProvisioned(ctx, env.Client, cluster, cloudProvider, coreProvisioner, pod)
+			ExpectScheduled(ctx, env.Client, pod)
+
+			vm := azureEnv.VirtualMachinesAPI.VirtualMachineCreateOrUpdateBehavior.CalledWithInput.Pop().VM
+			Expect(vm).NotTo(BeNil())
+			Expect(vm.Properties.StorageProfile.OSDisk.DiskSizeGB).NotTo(BeNil())
+			Expect(*vm.Properties.StorageProfile.OSDisk.DiskSizeGB).To(Equal(int32(30)))
+			Expect(lo.FromPtr(vm.Properties.StorageProfile.OSDisk.DiffDiskSettings.Option)).To(Equal(armcompute.DiffDiskOptionsLocal))
```
tallaxes, line 787: Shouldn't all ephemeral tests now validate placement as well? Or maybe at least those that use specific SKU?

pkg/providers/instancetype/suite_test.go
```diff
@@ -596,6 +651,55 @@ var _ = Describe("InstanceType Provider", func() {
 			Expect(*vm.Properties.StorageProfile.OSDisk.DiskSizeGB).To(Equal(int32(128)))
 			Expect(vm.Properties.StorageProfile.OSDisk.DiffDiskSettings).To(BeNil())
 		})
+		It("should select NvmeDisk for v6 skus with maxNvmeDiskSize > 0", func() {
+			nodePool.Spec.Template.Spec.Requirements = append(nodePool.Spec.Template.Spec.Requirements, karpv1.NodeSelectorRequirementWithMinValues{
+				NodeSelectorRequirement: v1.NodeSelectorRequirement{
+					Key:      "node.kubernetes.io/instance-type",
```
tallaxes, line 836: There is a constant for this label key (here and elsewhere)

pkg/providers/instancetype/suite_test.go
```diff
@@ -99,15 +99,19 @@ func TestAzure(t *testing.T) {
 	ctx, stop = context.WithCancel(ctx)
 	azureEnv = test.NewEnvironment(ctx, env)
 	azureEnvNonZonal = test.NewEnvironmentNonZonal(ctx, env)
+	azureEnvSouthCentralUS = test.NewRegionalEnvironment(ctx, env, "southcentralus", true)
```
tallaxes: It is unfortunate to have to use several region-specific environments in this context. We may not have had a choice with zonal-vs-nonzonal, but for regular ones ... Can we use only one zonal? Does southcentral have everything we need to test for? (And this is fake data anyway, maybe we can just cook it up ...)

test/suites/nodeclaim/eph_osdisk_test.go
```diff
@@ -0,0 +1,78 @@
...
+	//		})
+
+	//	pod := test.Pod()
+	//	env.ExpectCreated(nodeClass, nodePool, pod)
+	//	env.EventuallyExpectHealthy(pod)
+	//	env.ExpectCreatedNodeCount("==", 1)
+
+	//	vm := env.GetVM(nodePool.Spec.Template.Spec.NodeName)
+	//	Expect(vm.Properties.StorageProfile.OSDisk.DiffDiskSettings.Option).To(Equal("Local"))
+	//	Expect(vm.Properties.StorageProfile.OSDisk.DiffDiskSettings.Placement).To(Equal("NVMe"))
```
tallaxes:
```suggestion
	//	Expect(vm.Properties.StorageProfile.OSDisk.DiffDiskSettings.Placement).To(Equal(armcompute.DiffDiskPlacementNvmeDisk))
```
(and the constant is actually "NvmeDisk")

go.mod
```diff
@@ -167,3 +168,5 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.4.2 // indirect
 	sigs.k8s.io/yaml v1.4.0 // indirect
 )
+
+replace github.com/Azure/skewer => github.com/Bryce-Soghigian/skewer v0.0.0-20250603213716-64caf6daff73
```
tallaxes: Please don't forget to cleanup. Might be better to just implement the logic client side, until skewer changes merge?

tallaxes (commented):
(Apologies for comments on draft, started to review the previous PR, so moving comments here). The relevant logic right now is dispersed across many places, making it hard to review and maintain. Would be great to consolidate it if possible - and maybe extract it from template altogether?

pkg/providers/instancetype/instancetypes.go
Bryce-Soghigian: ack

pkg/providers/instancetype/suite_test.go
Bryce-Soghigian: yeah southcentralus does have everything we need

test/suites/nodeclaim/eph_osdisk_test.go
Bryce-Soghigian: Just seemed like thats where some other azure level validation was for gc, ideally was thinking of moving some storage e2es to their own suite

Bryce-Soghigian: I want to eventually write some other storage e2es

pkg/providers/instance/instance.go
Bryce-Soghigian: We should have a short discussion on some of the logic around ephemeral, should we add a selector through requiremetns, or AKSNodeClass, that allows someone to force ephemeral and fail provisioning otherwise? 

The defaulting or fallback may be confusing sometimes.

pkg/providers/instance/instance.go
Bryce-Soghigian: good catch

pkg/providers/instancetype/instancetype.go
Bryce-Soghigian: remant of previous work

pkg/providers/instancetype/instancetype.go
```diff
@@ -187,8 +187,10 @@ func computeRequirements(sku *skewer.SKU, vmsize *skewer.VMSizeType, architectur
...
-		requirements[v1beta1.LabelSKUStorageEphemeralOSMaxSize].Insert(fmt.Sprint(MaxEphemeralOSDiskSizeGB(sku)))
+		sizeGB, _ := MaxEphemeralOSDiskSizeGB(sku)
+		requirements[v1beta1.LabelSKUStorageEphemeralOSMaxSize].Insert(fmt.Sprint(sizeGB))
```
Bryce-Soghigian, line 192: is there a problem using fmt.Sprint() rather than %d?

pkg/providers/instancetype/instancetypes.go
Bryce-Soghigian: > This has to also check SupportedEphemeralOSDiskPlacements to be able to pick NVME

Seems I forgot to push that part up

pkg/providers/instancetype/instancetypes.go
Bryce-Soghigian: We can align there

pkg/providers/instancetype/suite_test.go
Bryce-Soghigian: some of them should yes

pkg/providers/launchtemplate/parameters/types.go
Bryce-Soghigian: should change this, this was added before

pkg/providers/instancetype/instancetypes.go
Bryce-Soghigian: might need ot keep some of it for gigabyte conversion

pkg/providers/imagefamily/resolver.go
Bryce-Soghigian: Just following the existing NPS logic which stores the size gb here, maybe there is a better way to do this in general?

pkg/providers/imagefamily/resolver.go
Bryce-Soghigian: yes, I would like to have access ot the sku, so that we can get rid of the logic where we are parsing values from requirements. 

Thoughts?

pkg/providers/instancetype/suite_test.go
Bryce-Soghigian: removed everything for eastus

pkg/providers/instancetype/instancetype.go
tallaxes: I don't think so

pkg/providers/imagefamily/resolver.go
tallaxes: > so that we can get rid of the logic where we are parsing values from requirements.

Not convinced this is a good reason. Have you seen how it gets the value from SKU capability? It has to iterate over all capabilities, and then ends up - wait for it - parsing it from string.

I do agree there is place for refactoring things here, we likely need to add all the info we need to the instanceType - and even just keeping sku reference there would simplify the flow, though I'd rather we did not have to go via sku at all past a certain point. Will take more work, not for now.

(Not blocking, want to land this.)

test/suites/nodeclaim/eph_osdisk_test.go
Bryce-Soghigian: Ill have a separate PR out to fix this test along with 3 other tests

hack/find_sub.sh
Bryce-Soghigian: i will just remove this file all together

pkg/providers/instancetype/suite_test.go
```diff
@@ -612,6 +625,44 @@ var _ = Describe("InstanceType Provider", func() {
...
+			nodeClass.Spec.OSDiskSizeGB = lo.ToPtr[int32](100001)
+
+			ExpectApplied(ctx, env.Client, nodePool, nodeClass)
+			pod := coretest.UnschedulablePod()
+			ExpectProvisioned(ctx, env.Client, cluster, cloudProvider, coreProvisioner, pod)
+			ExpectNotScheduled(ctx, env.Client, pod)
+
+		})
+
+		It("should select an ephemeral disk if LabelSKUStorageEphemeralOSMaxSize is set", func() {
```
Bryce-Soghigian:
```suggestion
		It("should select an ephemeral disk if LabelSKUStorageEphemeralOSMaxSize is set and os disk size fits", func() {
```

go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/Azure/karpenter-provider-azure
 
 go 1.24.4
 
+require github.com/Azure/skewer v0.0.20
```
tallaxes: Should be under require block, just undo changes?

Bryce-Soghigian: updated

pkg/providers/launchtemplate/launchtemplate.go
Bryce-Soghigian: addressed this feedback on storageprofileisephemeral

pkg/providers/instancetype/suite_test.go
tallaxes: Not sure why this is resolved. Shouldn't _all_ ephemeral tests now validate placement? I see a single test for DiffDiskPlacementNvmeDisk, what about others? This is what the PR is about, right? Are these tests elsewhere?

Bryce-Soghigian: I added placement tests in a placement block that tests some of the placement permutations, do we want to ensure these tests are tied to a specific sku? Otherwise this is brittle to code in what if we add in a new cheaper sku that gets selected?

Bryce-Soghigian: oh... seems i somehow deleted my placement tests in my rebase......

Bryce-Soghigian:
```
Context("Placement", func() {
			It("should prefer NVMe disk if supported for ephemeral", func() {
				nodePool.Spec.Template.Spec.Requirements = append(nodePool.Spec.Template.Spec.Requirements, karpv1.NodeSelectorRequirementWithMinValues{
					NodeSelectorRequirement: v1.NodeSelectorRequirement{
						Key:      "node.kubernetes.io/instance-type",
						Operator: v1.NodeSelectorOpIn,
						Values:   []string{"Standard_D128ds_v6"},
					},
				})

				ExpectApplied(ctx, env.Client, nodePool, nodeClass)
				pod := coretest.UnschedulablePod()

				vm := azureEnv.VirtualMachinesAPI.VirtualMachineCreateOrUpdateBehavior.CalledWithInput.Pop().VM
				Expect(vm).NotTo(BeNil())
				Expect(vm.Properties.StorageProfile.OSDisk.DiffDiskSettings).NotTo(BeNil())
				Expect(lo.FromPtr(vm.Properties.StorageProfile.OSDisk.DiffDiskSettings.Placement)).To(Equal(armcompute.DiffDiskPlacementNvmeDisk))
			})

			It("should not select NVMe ephemeral disk placement if the sku has an nvme disk, supports ephemeral os disk, but doesnt support NVMe placement", func() {
				nodePool.Spec.Template.Spec.Requirements = append(nodePool.Spec.Template.Spec.Requirements, karpv1.NodeSelectorRequirementWithMinValues{
					NodeSelectorRequirement: v1.NodeSelectorRequirement{
						Key:      "node.kubernetes.io/instance-type",
						Operator: v1.NodeSelectorOpIn,
						Values:   []string{"Standard_NC24ads_A100_v4"},
					},
				})

				ExpectApplied(ctx, env.Client, nodePool, nodeClass)
				pod := coretest.UnschedulablePod()

				vm := azureEnv.VirtualMachinesAPI.VirtualMachineCreateOrUpdateBehavior.CalledWithInput.Pop().VM
				Expect(vm).NotTo(BeNil())
				Expect(vm.Properties.StorageProfile.OSDisk.DiffDiskSettings).NotTo(BeNil())
				Expect(lo.FromPtr(vm.Properties.StorageProfile.OSDisk.DiffDiskSettings.Placement)).ToNot(Equal(armcompute.DiffDiskPlacementNvmeDisk))
			})


```

test/suites/nodeclaim/eph_osdisk_test.go
Bryce-Soghigian: just deleted this test

pkg/providers/instancetype/instancetypes.go
```diff
@@ -341,27 +356,59 @@ func (p *DefaultProvider) isConfidential(sku *skewer.SKU) bool {
...
+// safeConvert checks for overflow and then converts.
+// #nosec G115  // we’ve manually guarded against overflow
+func safeConvert(val int64) (int32, bool) {
+	if val > maxInt32 || val < minInt32 {
+		return 0, false
+	}
+	return int32(val), true
+}
+
+func bytesToGBWithPlacement(sizeInBytes int64, placement *armcompute.DiffDiskPlacement) (int32, *armcompute.DiffDiskPlacement, bool) {
```
tallaxes: Why would this ever have placement as parameter and return??

Bryce-Soghigian: maybe time to put down the LLM, first time using agent mode in vscode rather than cursor and although the test passed when I prompted about cyclomatic complexity, I should have taken a closer look.

pkg/providers/instancetype/instancetypes.go
```diff
@@ -341,27 +356,33 @@ func (p *DefaultProvider) isConfidential(sku *skewer.SKU) bool {
...
+
+func FindMaxEphemeralSizeGBAndPlacement(sku *skewer.SKU) (sizeGB int64, placement *armcompute.DiffDiskPlacement) {
 	if sku == nil {
-		return 0
+		return 0, nil
 	}
-	maxCachedDiskBytes, _ := sku.MaxCachedDiskBytes()
-	maxResourceVolumeMB, _ := sku.MaxResourceVolumeMB() // NOTE: this is a misnomer, MB is actually MiB, hence the conversion below
+	maxCacheDiskBytes, _ := sku.MaxCachedDiskBytes()
+	maxResourceDiskMiB, _ := sku.MaxResourceVolumeMB() // NOTE: MaxResourceVolumeMB is actually in MiBs
```
tallaxes: Let's also move these right before they are needed, don't need to lookup (relatively expensive, iterates over capabilities) unless will actually be used.

Bryce-Soghigian: ACK

pkg/providers/instancetype/instancetypes.go
```diff
@@ -288,6 +302,7 @@ func (p *DefaultProvider) getInstanceTypes(ctx context.Context) (map[string]*ske
 		atomic.AddUint64(&p.instanceTypesSeqNum, 1)
 		log.FromContext(ctx).V(1).Info("discovered instance types", "instanceTypeCount", len(instanceTypes))
 	}
+	log.FromContext(ctx).Info(fmt.Sprintf("InstanceTypes: %v", instanceTypes))
```
tallaxes: Leftover?

Bryce-Soghigian: Yes let me remove it

Bryce-Soghigian:
```suggestion
```

pkg/providers/instancetype/instancetypes.go
```diff
@@ -430,3 +446,19 @@ func isCompatibleImageAvailable(sku *skewer.SKU, useSIG bool) bool {
 
 	return useSIG || hasSCSISupport(sku) // CIG images are not currently tagged for NVMe
 }
+
+func SupportsNVMeEphemeralOSDisk(sku *skewer.SKU) bool {
```
tallaxes: Does not need to be exported

pkg/providers/instancetype/instancetypes.go
Bryce-Soghigian: removed this

pkg/providers/instancetype/suite_test.go
```diff
@@ -589,6 +590,145 @@ var _ = Describe("InstanceType Provider", func() {
...
+		AfterEach(func() {
+			ctx = options.ToContext(ctx, originalOptions)
+		})
+
+		Context("FindMaxEphemeralSizeGBAndPlacement(sku *skewer.SKU) -> diskSizeGB, *placement", func() {
+			// B20ms:
+			// NvmeDiskSizeInMiB == 0
+			// CacheDiskBytes == 32212254720 -> 32.21225472 GB .. we should select this as the ephemeral disk size
+			// placement == CacheDisk
+			// MaxResourceVolumeMB == 163840 MiB -> 160 GB,
```
tallaxes:
```suggestion
			// MaxResourceVolumeMB == 163840 MiB -> 171.80 GB,
```

pkg/providers/instancetype/suite_test.go
```diff
@@ -589,6 +590,145 @@ var _ = Describe("InstanceType Provider", func() {
...
+			// NvmeDiskSizeInMiB == 7208960 -> 7559.142441 GB // SupportedEphemeralOSDiskPlacments == NvmeDisk
+			// and this is greater than 0, so we select 7559, placement == NvmeDisk
+			// Standard_NC24ads_A100_v4:
+			// {Name: lo.ToPtr("SupportedEphemeralOSDiskPlacements"), Value: lo.ToPtr("ResourceDisk,CacheDisk")},
+			// NvmeDiskSizeInMiB == 915527 -> 959.99964 GB  but no SupportedEphemeralOSDiskPlacments == NvmeDisk so we move to cache disk
+			// CacheDiskBytes == 274877906944 -> 274.877906944 GB so we select cache disk + 274
+			// MaxResourceVolumeMB == 65536 MiB
+			// Standard_D64s_v3:
+			// NvmeDiskSizeInMiB == 0
+			// CacheDiskBytes == 171798691840 -> 1717.9869184 GB, this is greater than zero, so we select this as the ephemeral disk size
```
tallaxes:
```suggestion
			// CacheDiskBytes == 1717986918400 -> 1717.9869184 GB, this is greater than zero, so we select this as the ephemeral disk size
```

pkg/providers/instancetype/instancetypes.go
```diff
@@ -430,3 +446,19 @@ func isCompatibleImageAvailable(sku *skewer.SKU, useSIG bool) bool {
...
+	const nvme = "NvmeDisk"
+	return sku.HasCapabilityWithSeparator(ephemeralOSDiskPlacementCapability, nvme)
+}
+
+func UseEphemeralDisk(sku *skewer.SKU, nodeClass *v1beta1.AKSNodeClass) bool {
+	sizeGB, _ := FindMaxEphemeralSizeGBAndPlacement(sku)
+	return int64(*nodeClass.Spec.OSDiskSizeGB) <= sizeGB // use ephemeral disk if it is large enough
+}
+
+func NvmeDiskSizeInMiB(s *skewer.SKU) (int64, error) {
```
tallaxes: Does not need to be exported

Bryce-Soghigian:
Im gonna run the e2e tests and address comments before merging

pkg/providers/instancetype/suite_test.go
tallaxes: Marking unresolved

tallaxes (approved):

pkg/providers/instancetype/suite_test.go
```diff
@@ -589,6 +590,145 @@ var _ = Describe("InstanceType Provider", func() {
...
+				func(sku *skewer.SKU, expectedSize int64, expectedPlacement *armcompute.DiffDiskPlacement) {
+					sizeGB, placement := instancetype.FindMaxEphemeralSizeGBAndPlacement(sku)
+					Expect(sizeGB).To(Equal(expectedSize))
+					Expect(placement).To(Equal(expectedPlacement))
+
+				}, Entry("Standard_B20ms", SkewerSKU("Standard_B20ms"), int64(32), lo.ToPtr(armcompute.DiffDiskPlacementCacheDisk)),
+				Entry("Standard_D128ds_v6", SkewerSKU("Standard_D128ds_v6"), int64(7559), lo.ToPtr(armcompute.DiffDiskPlacementNvmeDisk)),
+				Entry("Standard_D16plds_v5", SkewerSKU("Standard_D16plds_v5"), int64(429), lo.ToPtr(armcompute.DiffDiskPlacementCacheDisk)),
+				Entry("Standard_D2as_v6", SkewerSKU("Standard_D2as_v6"), int64(0), nil),
+				Entry("Standard_D64s_v3", SkewerSKU("Standard_D128ds_v6"), int64(7559), lo.ToPtr(armcompute.DiffDiskPlacementNvmeDisk)),
```
tallaxes, line 650:
```suggestion
				Entry("Standard_D128ds_v6", SkewerSKU("Standard_D128ds_v6"), int64(7559), lo.ToPtr(armcompute.DiffDiskPlacementNvmeDisk)),
```

tallaxes: Name and SKU differ. Did you mean to test Standard_D64s_v3? Need to fix size / placement, otherwise just duplicate test with the wrong name. Edit: Ok, just a duplicate to be removed

Bryce-Soghigian:
EDIT: thats a duplicate

We actually test D64s_v3 below, we do mean to test that sku there, ill update the comment rather than the SkewerSKU()

Bryce-Soghigian:
image
https://github.com/Azure/karpenter-provider-azure/actions/runs/15848421023

Everything passing here except for scheduling, will trigger a run to see scheduling pass here: https://github.com/Azure/karpenter-provider-azure/actions/runs/15865761400/job/44732221239

Bryce-Soghigian:
Scheduling e2e passed

PR #948, opened 2025-06-09
removing version from project

MaximilianoUribe:
Fixes # 

**Description**
project version was not populated correctly, taking operator.Version from Makefile to populate the value

*Checked user agent with deployment in kusto:

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


pkg/utils/project/project.go
```diff
@@ -18,4 +18,3 @@ package project
 
 // Version is the karpenter app version injected during compilation
 // when using the Makefile
```
tallaxes: Let's remove these comments as well?

pkg/utils/project/project.go
```diff
@@ -18,4 +18,3 @@ package project
 
 // Version is the karpenter app version injected during compilation
```
matthchr: Shouldn't this whole file be removed, given that there's nothing else in it?

pkg/utils/project/project.go
MaximilianoUribe: deleted the file and path

tallaxes (approved):

PR #951, opened 2025-06-10
fix: reducing flake due to LRO tail end errors and fixing termination grace period e2e

Bryce-Soghigian:
Fixes # 

**Description**

The drift-blocking E2E test + other e2e tests with similar codepaths sometimes fail with NodeClaim.karpenter.sh "<id>" not found.
1.	The test captures the first NodeClaim that appears:

nodeClaim := env.EventuallyExpectCreatedNodeClaimCount("==", 1)[0]
	
2.	That NodeClaim (f5zvx) can fail to launch (e.g., ARM allocation error) and is deleted by Karpenter.
3.	Karpenter immediately creates a replacement NodeClaim (6ds7l) that succeeds:

[DELETED] NODECLAIM …-f5zvx
[CREATED] NODECLAIM …-6ds7l

4. env.EventuallyExpectCreatedNodeCount("==", 1)  this condition is met by nodeclaim 6ds7l and will pass 
5. env.EventuallyExpectDrifted(nodeClaim) // points at f5zvx and will fail because that nodeclaim never actually reached a registered state

- make az-e2etest 

Gonna do some pipeline runs too 

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


test/suites/consolidation/suite_test.go
```diff
@@ -160,10 +160,10 @@ var _ = Describe("Consolidation", Ordered, func() {
 			}
 			env.ExpectCreated(nodeClass, nodePool, job)
 
-			nodeClaims := env.EventuallyExpectCreatedNodeClaimCount("==", 1)
 			env.EventuallyExpectCreatedNodeCount("==", 1)
 			pods := env.EventuallyExpectHealthyPodCount(selector, int(1))
 
+			nodeClaims := env.EventuallyExpectRegisteredNodeClaimCount("==", 1)
```
tallaxes: Why move this after `EventuallyExpectCreatedNodeCount`? Seems to break the logical sequence (even if does not matter for the test itself)

test/suites/drift/suite_test.go
```diff
@@ -305,8 +305,8 @@ var _ = Describe("Drift", func() {
 
 			dep.Spec.Template.Annotations = nil
 			env.ExpectCreated(nodeClass, nodePool, dep)
-
-			nodeClaim := env.EventuallyExpectCreatedNodeClaimCount("==", 1)[0]
+			// we want to make sure the reference we get for the nodeclaim is a registered nodeclaim
```
tallaxes: This is a good comment for PR, but I don't think we need to keep this in code? (Also, should be consistent, and you don't have this comment everywhere - I would suggest just removing everywhere)

test/suites/drift/suite_test.go
```diff
@@ -592,10 +594,9 @@ var _ = Describe("Drift", func() {
 			env.ExpectCreated(dep, nodeClass, nodePool)
 
 			By("deploying multiple replicas, pod per node")
-			startingNodeClaimState := env.EventuallyExpectCreatedNodeClaimCount("==", int(numPods))
 			env.EventuallyExpectCreatedNodeCount("==", int(numPods))
 			env.EventuallyExpectHealthyPodCount(selector, int(numPods))
-
+			startingNodeClaimState := env.EventuallyExpectRegisteredNodeClaimCount("==", int(numPods))
```
tallaxes: Same, why move to later?

test/suites/drift/suite_test.go
```diff
@@ -726,10 +727,11 @@ var _ = Describe("Drift", func() {
 
 	It("should disrupt nodes that have drifted due to VNETSubnetID", func() {
 		env.ExpectCreated(dep, nodeClass, nodePool)
-		nodeClaim := env.EventuallyExpectCreatedNodeClaimCount("==", 1)[0]
 		env.EventuallyExpectHealthyPodCount(selector, numPods)
 		By("expect created node count to be 1")
 		env.ExpectCreatedNodeCount("==", 1)
+		nodeClaim := env.EventuallyExpectCreatedNodeClaimCount("==", 1)[0]
```
tallaxes, line 733: Why not "Registered" in this particular case? (And also why move, as elsewhere?)

test/suites/integration/emptiness_test.go
```diff
@@ -62,7 +62,9 @@ var _ = Describe("Emptiness", func() {
 
 			env.ExpectCreated(nodeClass, nodePool, dep)
 
-			nodeClaim := env.EventuallyExpectCreatedNodeClaimCount("==", 1)[0]
+			// we want to make sure the reference we get for the nodeclaim is a registered nodeclaim
```
tallaxes: Same comment on comment as elsewhere

test/suites/integration/subnet_test.go
```diff
@@ -81,9 +81,12 @@ var _ = Describe("Subnets", func() {
 		Expect(nic.Properties.IPConfigurations[0].Properties.Subnet).ToNot(BeNil())
 		Expect(nic.Properties.IPConfigurations[0].Properties.Subnet.ID).To(Equal(subnet.ID))
 
-		// The NIC should have the right NSG
-		Expect(nic.Properties.NetworkSecurityGroup).ToNot(BeNil())
-		Expect(nic.Properties.NetworkSecurityGroup.ID).ToNot(BeNil())
-		Expect(*nic.Properties.NetworkSecurityGroup.ID).To(MatchRegexp(`aks-agentpool-\d{8}-nsg`))
+		// TODO: Unblock when 1.4.1 is released
```
tallaxes: This is about having NSG client? Let's reference what is actually needed, rather than a tentative version

tallaxes (commented):
Overall looks good, nice catch! Left some comments, most nits, but one place still uses "Created" instead of "Registered"? Also please update the description (it logs the problematic test sequence, but does not describe the fix), the title (what's "tgp"?), and leave some evidence of (consistently) passing E2Es.

Also please

test/suites/drift/suite_test.go
Bryce-Soghigian: Its fine to have the reference after nodes have been created, as it ensures the nodes have registered meaning nodeclaim will have registered and it wont hit the bug from earlier.

tallaxes: Ack, so there are two solution patterns, use EventuallyExpect**Registered**NodeClaimCount, or move after ExpectCreatedNodeCount. (Three patterns, if you count "use Registered _and_ move after ExpectCreatedNodeCount.) Let's pick one and stay consistent?

Bryce-Soghigian: sure we can do that

test/suites/nodeclaim/nodeclaim_test.go
```diff
@@ -54,8 +54,8 @@ var _ = Describe("StandaloneNodeClaim", func() {
 			},
 		})
 		env.ExpectCreated(nodeClass, nodeClaim)
+		nodeClaim = env.EventuallyExpectRegisteredNodeClaimCount("==", 1)[0]
```
Bryce-Soghigian, line 57: this scenario turned flaky after the v1.4.0 release, i suspect its because of LRO things, and this doesnt really fix it, unlike other tests, this assumes we will create the nodeclaim with that name, but in the case of Zonal Allocation errors we sometimes(1/15 runs) see this one fail. 


I suspect that we need to rewrite these tests to support the new paradigm

Bryce-Soghigian: cc: @matthchr

test/suites/drift/suite_test.go
tallaxes: So which did you choose? All other places use Registered ...

tallaxes (approved):
LGTM, one unaddressed comment. Also please update the description (it logs the problematic test sequence, but does not describe the fix), the title (what's "tgp"?), and leave some evidence of (consistently) passing E2Es.

test/suites/nodeclaim/nodeclaim_test.go
tallaxes: Where exactly does it fail? Would be surprised it fails after registered NodeClaim is observed ...

tallaxes: Not sure if related, but standalone NodeClaim might be the special case where deleting NodeClaim on error - which is what we now do in async Create implementation - might not be the right approach. With NodePools, there will be something that would create another NodeClaim on failure. With standalone NodeClaim there won't be anything - or, rather, there might not be something to recreate it. The desired behavior might depend on the use case. This may be a good topic for follow-up with other Karpenter maintainers.

test/suites/drift/suite_test.go
Bryce-Soghigian: Registered is what i chose, if i missed one good catch

tallaxes: Not addressed? Marking unresolved

tallaxes: @Bryce-Soghigian please clean this up in a follow-up, don't want to hold this on that.

PR #955, opened 2025-06-11
chore(logging): update logging for nodeclass status controllers

charliedmcb:
Fixes # 

**Description**
Doing some testing surrounding the maintenance window integration, and the nodeclass status controllers in general, had me uncovering multiple ways for improvement, which this PR aims to do.

- `make presubmit`
- manual testing of `make az-all` + inspecting of logs for different cases.
- E2E Runs:
    - https://github.com/Azure/karpenter-provider-azure/actions/runs/15592313549
    - https://github.com/Azure/karpenter-provider-azure/actions/runs/15594986110 (latest)

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/controllers/nodeclass/status/images.go
+++ b/pkg/controllers/nodeclass/status/images.go
@@ -109,7 +109,6 @@ func (r *NodeImageReconciler) Register(_ context.Context, m manager.Manager) err
 func (r *NodeImageReconciler) Reconcile(ctx context.Context, nodeClass *v1beta1.AKSNodeClass) (reconcile.Result, error) {
 	ctx = log.IntoContext(ctx, log.FromContext(ctx).WithName(nodeImageReconcilerName))
 	logger := log.FromContext(ctx)
-	logger.V(1).Info("starting reconcile")
 
 	nodeImages, err := r.nodeImageProvider.List(ctx, nodeClass)
 	if err != nil {
@@ -141,7 +140,7 @@ func (r *NodeImageReconciler) Reconcile(ctx context.Context, nodeClass *v1beta1.
 	shouldUpdate := imageVersionsUnready(nodeClass)
 	if !shouldUpdate {
 		// Case 4: Check if the maintenance window is open
-		shouldUpdate, err = r.isMaintenanceWindowOpen(ctx)
+		shouldUpdate, err = r.isMaintenanceWindowOpen(ctx, nodeClass.Name)
 		if err != nil {
 			return reconcile.Result{}, fmt.Errorf("checking maintenance window, %w", err)
 		}
@@ -154,14 +153,17 @@ func (r *NodeImageReconciler) Reconcile(ctx context.Context, nodeClass *v1beta1.
 	if len(goalImages) == 0 {
 		nodeClass.Status.Images = nil
 		nodeClass.StatusConditions().SetFalse(v1beta1.ConditionTypeImagesReady, "ImagesNotFound", "ImageSelectors did not match any Images")
-		logger.Info("no node images")
+		logger.Info("no available node images")
 		return reconcile.Result{RequeueAfter: 5 * time.Minute}, nil
 	}
 
 	nodeClass.Status.Images = goalImages
 	nodeClass.StatusConditions().SetTrue(v1beta1.ConditionTypeImagesReady)
-
-	logger.V(1).Info("success")
+	// Note: the ChangeMonitor uses SlicesAsSets=true. However, we do actually care about the ordering of the Images, meaning we'll
+	//     miss logging potential updates on the ordering. Although, this is unexpected to occur.
+	if r.cm.HasChanged(fmt.Sprintf("nodeclass-%s-images", nodeClass.Name), nodeClass.Status.Images) {
+		logger.WithValues("images", nodeClass.Status.Images).Info("new available images updated for nodeclass")
+	}
 	return reconcile.Result{RequeueAfter: 5 * time.Minute}, nil
 }
 
@@ -177,7 +179,8 @@ func imageVersionsUnready(nodeClass *v1beta1.AKSNodeClass) bool {
 // I think the best way to get rid of gocyclo is to break the section retrieving the maintenance window
 // range from the ConfigMap into its own helper function using channel as a parameter.
 // nolint: gocyclo
-func (r *NodeImageReconciler) isMaintenanceWindowOpen(ctx context.Context) (bool, error) {
+func (r *NodeImageReconciler) isMaintenanceWindowOpen(ctx context.Context, nodeClassName string) (bool, error) {
+	logger := log.FromContext(ctx)
 	if r.systemNamespace == "" {
 		// We fail open here, since the default case should be to upgrade
 		return true, nil
@@ -190,6 +193,11 @@ func (r *NodeImageReconciler) isMaintenanceWindowOpen(ctx context.Context) (bool
 			return true, nil
 		}
 		return false, fmt.Errorf("error getting maintenance window configmap, %w", err)
+	}
+	// Monitoring the entire ConfigMap's data might catch more data changes than we care about. However, I think it makes sense to monitor
+	//     here as catches the entire spread of cases we care about, and will give us direct insight on the raw data.
+	if r.cm.HasChanged(fmt.Sprintf("nodeclass-%s-mwdata", nodeClassName), mwConfigMap.Data) {
+		logger.WithValues("maintenancewindowdata", mwConfigMap.Data).Info("new maintenance window data discovered")
 	}
 	if len(mwConfigMap.Data) == 0 {
 		// An empty configmap means there's no maintenance windows defined, and its up to us when to preform maintenance
--- a/pkg/controllers/nodeclass/status/images.go
+++ b/pkg/controllers/nodeclass/status/images.go
@@ -161,6 +161,7 @@ func (r *NodeImageReconciler) Reconcile(ctx context.Context, nodeClass *v1beta1.
 	nodeClass.StatusConditions().SetTrue(v1beta1.ConditionTypeImagesReady)
 	// Note: the ChangeMonitor uses SlicesAsSets=true. However, we do actually care about the ordering of the Images, meaning we'll
 	//     miss logging potential updates on the ordering. Although, this is unexpected to occur.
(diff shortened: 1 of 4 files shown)


pkg/controllers/nodeclass/status/images.go
```diff
@@ -154,14 +153,18 @@ func (r *NodeImageReconciler) Reconcile(ctx context.Context, nodeClass *v1beta1.
...
-		logger.Info("no node images")
+		logger.Info("no available node images")
 		return reconcile.Result{RequeueAfter: 5 * time.Minute}, nil
 	}
 
+	// Note: the ChangeMonitor uses SlicesAsSets=true. However, we do actually care about the ordering of the Images, meaning we'll
+	//     miss logging potential updates on the ordering. Although, this is unexpected to occur.
+	//     https://github.com/kubernetes-sigs/karpenter/blob/349487633193bced541ad05bb76d02e633e73473/pkg/utils/pretty/changemonitor.go#L42
+	if r.cm.HasChanged(fmt.Sprintf("nodeclass-%s-images", nodeClass.Name), goalImages) {
+		logger.WithValues("existingimages", nodeClass.Status.Images).WithValues("newimages", goalImages).Info("new available images updated for nodeclass")
```
matthchr: minor:
```suggestion
		logger.WithValues("existingImages", nodeClass.Status.Images).WithValues("newImages", goalImages).Info("new available images updated for nodeclass")
```

maybe?

pkg/controllers/nodeclass/status/kubernetesversion.go
```diff
@@ -71,8 +71,7 @@ func (r *KubernetesVersionReconciler) Register(_ context.Context, m manager.Mana
 //     - Note: We will indirectly trigger an upgrade to latest image version as well, by resetting the Images readiness.
 func (r *KubernetesVersionReconciler) Reconcile(ctx context.Context, nodeClass *v1beta1.AKSNodeClass) (reconcile.Result, error) {
 	ctx = log.IntoContext(ctx, log.FromContext(ctx).WithName(kubernetesVersionReconcilerName))
-	logger := log.FromContext(ctx)
-	logger.V(1).Info("starting reconcile")
+	logger := log.FromContext(ctx).WithValues("existingkubernetesversion", nodeClass.Status.KubernetesVersion)
```
matthchr:
```suggestion
	logger := log.FromContext(ctx).WithValues("existingKubernetesVersion", nodeClass.Status.KubernetesVersion)
```

maybe?

pkg/controllers/nodeclass/status/kubernetesversion.go
```diff
@@ -94,16 +93,18 @@ func (r *KubernetesVersionReconciler) Reconcile(ctx context.Context, nodeClass *
...
+			logger.Info(fmt.Sprintf("WARN: detected potential kubernetes downgrade: from %s (current), to %s (discovered)", currentK8sVersion.String(), newK8sVersion.String()))
 			// We do not currently support downgrading, so keep the kubernetes version the same
 			goalK8sVersion = nodeClass.Status.KubernetesVersion
 		}
 	}
 	nodeClass.Status.KubernetesVersion = goalK8sVersion
 	nodeClass.StatusConditions().SetTrue(v1beta1.ConditionTypeKubernetesVersionReady)
-	logger.V(1).Info("successful reconcile")
+	if r.cm.HasChanged(fmt.Sprintf("nodeclass-%s-kubernetesversion", nodeClass.Name), nodeClass.Status.KubernetesVersion) {
+		logger.WithValues("newkubernetesversion", nodeClass.Status.KubernetesVersion).Info("new kubernetes version updated for nodeclass")
```
matthchr:
```suggestion
		logger.WithValues("newKubernetesVersion", nodeClass.Status.KubernetesVersion).Info("new kubernetes version updated for nodeclass")
```

pkg/controllers/nodeclass/status/images.go
```diff
@@ -191,6 +195,11 @@ func (r *NodeImageReconciler) isMaintenanceWindowOpen(ctx context.Context) (bool
 		}
 		return false, fmt.Errorf("error getting maintenance window configmap, %w", err)
 	}
+	// Monitoring the entire ConfigMap's data might catch more data changes than we care about. However, I think it makes sense to monitor
+	//     here as it does catch the entire spread of cases we care about, and will give us direct insight on the raw data.
+	if r.cm.HasChanged(fmt.Sprintf("nodeclass-%s-mwdata", nodeClassName), mwConfigMap.Data) {
+		logger.WithValues("maintenancewindowdata", mwConfigMap.Data).Info("new maintenance window data discovered")
```
matthchr:
```suggestion
		logger.WithValues("maintenanceWindowData", mwConfigMap.Data).Info("new maintenance window data discovered")
```

pkg/controllers/nodeclass/status/images.go
```diff
@@ -191,6 +195,11 @@ func (r *NodeImageReconciler) isMaintenanceWindowOpen(ctx context.Context) (bool
 		}
 		return false, fmt.Errorf("error getting maintenance window configmap, %w", err)
 	}
+	// Monitoring the entire ConfigMap's data might catch more data changes than we care about. However, I think it makes sense to monitor
+	//     here as it does catch the entire spread of cases we care about, and will give us direct insight on the raw data.
+	if r.cm.HasChanged("nodeclass-maintenancewindowdata", mwConfigMap.Data) {
```
matthchr, line 205: Needs nodeclass name, or keys will collide.

pkg/controllers/nodeclass/status/images.go
```diff
@@ -154,14 +153,18 @@ func (r *NodeImageReconciler) Reconcile(ctx context.Context, nodeClass *v1beta1.
...
 		nodeClass.StatusConditions().SetFalse(v1beta1.ConditionTypeImagesReady, "ImagesNotFound", "ImageSelectors did not match any Images")
-		logger.Info("no node images")
+		logger.Info("no available node images")
 		return reconcile.Result{RequeueAfter: 5 * time.Minute}, nil
 	}
 
+	// Note: the ChangeMonitor uses SlicesAsSets=true. However, we do actually care about the ordering of the Images, meaning we'll
+	//     miss logging potential updates on the ordering. Although, this is unexpected to occur.
+	//     https://github.com/kubernetes-sigs/karpenter/blob/349487633193bced541ad05bb76d02e633e73473/pkg/utils/pretty/changemonitor.go#L42
+	if r.cm.HasChanged(fmt.Sprintf("nodeclass-%s-images", nodeClass.Name), goalImages) {
```
matthchr: In this instance maybe it doesn't make sense to use cm at all?

A lot of the value of cm is that it caches the old value, so if you're in a situation where you don't know what the old value was, that's OK it knows the old value. In this instance though, I think we know both the old and the new values (we're logging them).

So we could just use `hv, _ := hashstructure.Hash(value, hashstructure.FormatV2, &hashstructure.HashOptions{SlicesAsSets: false})` and compare them ourselves, rather than needing to ask the `cm` about it?

pkg/controllers/nodeclass/status/images.go
charliedmcb: So, I actually just dropped the nodeclass name from here, as I want them to collide.

MW data is at the cluster level, so should be fine to be shared across nodeclasses, and only really want to know when it changes for the cluster as a whole.

pkg/controllers/nodeclass/status/images.go
matthchr: (possibly in a little helper method)

pkg/controllers/nodeclass/status/images.go
charliedmcb: Discussed a bit further offline. Expectedly maintenance window information will eventually exist as its own sharable provider. However, for now its grouped inside of the nodeclass level controller which makes the logging pattern a bit strange. Will leave a comment explaining this a bit more.

pkg/controllers/nodeclass/status/images.go
charliedmcb: Super reasonable. Will look at updating.

charliedmcb: Added a new func to the utils pkg. lmk what you think :)

pkg/utils/utils.go
```diff
@@ -178,3 +179,14 @@ func IsAKSManagedVNET(nodeResourceGroup string, subnetID string) (bool, error) {
 		strings.EqualFold(nodeResourceGroup, id.ResourceGroupName) &&
 		strings.EqualFold(id.SubnetName, managedSubnetName), nil
 }
+
+// HasChanged returns if the given value has changed, given the existing and new instance
+//
+// This option is accessible in place of using a ChangeMonitor, when there's access to both
+// the existing and new data.
+func HasChanged(existing, new any, options *hashstructure.HashOptions) bool {
```
matthchr, line 187: minor: a single UT might be good too? Copilot can probably generate one for you

pkg/controllers/nodeclass/status/images.go
```diff
@@ -154,14 +155,15 @@ func (r *NodeImageReconciler) Reconcile(ctx context.Context, nodeClass *v1beta1.
 	if len(goalImages) == 0 {
 		nodeClass.Status.Images = nil
 		nodeClass.StatusConditions().SetFalse(v1beta1.ConditionTypeImagesReady, "ImagesNotFound", "ImageSelectors did not match any Images")
-		logger.Info("no node images")
+		logger.Info("no available node images")
 		return reconcile.Result{RequeueAfter: 5 * time.Minute}, nil
 	}
 
+	if utils.HasChanged(nodeClass.Status.Images, goalImages, &hashstructure.HashOptions{SlicesAsSets: false}) {
```
matthchr, line 163: minor: Possibly worth a comment explaining why you set SlicesAsSets here to false?

```
// Not treating these slices as sets because we care about order - it impacts which one will eventually be chosen
```

or something like that?

pkg/utils/utils.go
matthchr: Tests find bugs and all that

charliedmcb: Yea, good callout. Added some copilot tests with minor adjustments.

PR #956, opened 2025-06-11
refactor: handling of different response errors to separate functions

UtheMan:
**Description**
Moves Azure's response error handling to separate functions depending on error type, adds tests and extra logging in case of errors that didn't matc to any of the handlers.

* Unit Tests

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
NONE
```


matthchr (commented):
This LGTM, but full disclosure I didnt read the refactored code really closely because my assumption is that it hasn't materially changed, just how it's called has changed, is that right?

UtheMan:
Other than the "error matching" pattern and extra logging, the behavior stays the same. I also moved the tests around, since response error handling no longer happens in `instance.go`

matthchr (approved):

PR #958, opened 2025-06-12
fix: update devcontainer features

tallaxes:
**Description**

Fixes several issues with outdated or broken devcontainer features

* Rebuilding devcontainer

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -70,10 +70,10 @@
 		// custom
 		"ghcr.io/devcontainers/features/sshd:1": {},
 		"ghcr.io/devcontainers/features/docker-outside-of-docker:1": {},
-		"ghcr.io/devcontainers/features/kubectl-helm-minikube:1": {},
+		"ghcr.io/devcontainers/features/kubectl-helm-minikube:1.2.1": {}, // 1.2.2 seems to be broken
 		"ghcr.io/devcontainers/features/azure-cli:1": {},
-		"ghcr.io/devcontainers-contrib/features/pre-commit:2": {},
-		"ghcr.io/marcozac/devcontainer-features/shellcheck:1": {},
+		"ghcr.io/devcontainers-extra/features/pre-commit:2": {},
+		"ghcr.io/froazin/devcontainer-features/shellcheck:0": {},
 		"ghcr.io/rio/features/skaffold:2": { "version": "v2.14.2" }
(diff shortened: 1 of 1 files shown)


.devcontainer/devcontainer.json
```diff
@@ -70,10 +70,10 @@
 		// custom
 		"ghcr.io/devcontainers/features/sshd:1": {},
 		"ghcr.io/devcontainers/features/docker-outside-of-docker:1": {},
-		"ghcr.io/devcontainers/features/kubectl-helm-minikube:1": {},
+		"ghcr.io/devcontainers/features/kubectl-helm-minikube:1.2.1": {}, // 1.2.2 seems to be broken
```
charliedmcb, line 73: nit: add TODO to the comment, to circle back on?

.devcontainer/devcontainer.json
```diff
@@ -70,10 +70,10 @@
...
 		"ghcr.io/devcontainers/features/azure-cli:1": {},
-		"ghcr.io/devcontainers-contrib/features/pre-commit:2": {},
-		"ghcr.io/marcozac/devcontainer-features/shellcheck:1": {},
+		"ghcr.io/devcontainers-extra/features/pre-commit:2": {},
+		"ghcr.io/froazin/devcontainer-features/shellcheck:0": {},
```
charliedmcb, line 76: Seems quite minimal on amount of usage. Sure this is good as a backing?

charliedmcb (commented):
not entirely following these changes, so have a few questions.

.devcontainer/devcontainer.json
tallaxes: We have dependabot on this, which should be bugging us
https://github.com/Azure/karpenter-provider-azure/blob/42a991e1b4cec3b51b34c54e60603ebebb1ab161/.github/dependabot.yaml#L32-L35

.devcontainer/devcontainer.json
tallaxes: I think it will work for now (and I did take a look at the install logic). Can go back to the other one when fixed, or just create our own feature if none of them prove stable.

charliedmcb: So long as you've covered the install, and think it's good, fine with this as the backing for now.

PR #963, opened 2025-06-13
feat: add VM family level cache to the unvailable offerings cache

UtheMan:
**Description**
Adds VM family level cache to give an option of blocking a part of, or entire, vm SKU family when certain allocation errors are encountered.

* Unit tests
* E2E tests(??) TODO

- [x] Yes, PR includes docs updates  - TODO
- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note
TODO
```

### DIFF
--- a/pkg/cache/unavailableofferings.go
+++ b/pkg/cache/unavailableofferings.go
@@ -84,7 +84,7 @@ func (u *UnavailableOfferings) IsUnavailable(sku *skewer.SKU, zone, capacityType
 	if err != nil {
 		skuVCPUCount = 0 // default to 0 if we can't determine VCPU count
 	}
-	if u.isFamilyUnavailable(sku.GetFamilyName(), capacityType, zone, skuVCPUCount) {
+	if u.isFamilyUnavailable(sku.GetFamilyName(), zone, capacityType, skuVCPUCount) {
 		return true
 	}
 
@@ -93,9 +93,9 @@ func (u *UnavailableOfferings) IsUnavailable(sku *skewer.SKU, zone, capacityType
 	return found
 }
 
-func (u *UnavailableOfferings) isFamilyUnavailable(skuFamilyName, capacityType, zone string, cpuCount int64) bool {
+func (u *UnavailableOfferings) isFamilyUnavailable(skuFamilyName, zone, capacityType string, cpuCount int64) bool {
 	// Check if VM family is blocked in the specific zone
-	if val, found := u.vmFamilyCache.Get(vmFamilyKey(skuFamilyName, capacityType, zone)); found {
+	if val, found := u.vmFamilyCache.Get(vmFamilyKey(skuFamilyName, zone, capacityType)); found {
 		if blockedCPUCount, ok := val.(int64); ok {
 			if blockedCPUCount == wholeVMFamilyBlockedSentinel {
 				// Entire VM family is blocked in this zone
@@ -120,7 +120,7 @@ func (u *UnavailableOfferings) MarkFamilyUnavailableWithTTL(ctx context.Context,
 				return
 			}
 			// If new limit would be less restrictive, keep the existing one
-			// TODO - should we adjust TTL here? 
+			// TODO - should we adjust TTL here for the existing entry in cache?
 			if cpuCount != wholeVMFamilyBlockedSentinel && currentBlockedCPUCount <= cpuCount {
 				return
 			}
--- a/pkg/cache/unavailableofferings.go
+++ b/pkg/cache/unavailableofferings.go
@@ -22,13 +22,19 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/Azure/skewer"
 	"github.com/patrickmn/go-cache"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 	karpv1 "sigs.k8s.io/karpenter/pkg/apis/v1"
 )
 
+const (
+	// wholeVMFamilyBlockedSentinel means that entire SKU family is blocked, not just certain instance types with a CPU count above a threshold
+	wholeVMFamilyBlockedSentinel = -1
+)
+
 var (
-	spotKey = key("", "", karpv1.CapacityTypeSpot)
+	spotKey = singleInstanceKey("", "", karpv1.CapacityTypeSpot)
 )
 
 // UnavailableOfferings stores any offerings that return ICE (insufficient capacity errors) when
@@ -36,16 +42,22 @@ var (
 // GetInstanceTypes responses
 type UnavailableOfferings struct {
 	// key: <capacityType>:<instanceType>:<zone>, value: struct{}{}
-	cache  *cache.Cache
-	SeqNum uint64
+	singleOfferingCache *cache.Cache
+	// key: <skuFamilyName>:<zone>:<capacityType>, value: int64 (CPU count at or above which we block, or wholeVMFamilyBlockedSentinel if entire family is blocked)
+	vmFamilyCache *cache.Cache
+	SeqNum        uint64
 }
 
-func NewUnavailableOfferingsWithCache(c *cache.Cache) *UnavailableOfferings {
+func NewUnavailableOfferingsWithCache(singleOfferingCache, vmFamilyCache *cache.Cache) *UnavailableOfferings {
 	uo := &UnavailableOfferings{
-		cache:  c,
-		SeqNum: 0,
+		singleOfferingCache: singleOfferingCache,
+		vmFamilyCache:       vmFamilyCache,
+		SeqNum:              0,
 	}
-	uo.cache.OnEvicted(func(_ string, _ interface{}) {
+	uo.singleOfferingCache.OnEvicted(func(_ string, _ interface{}) {
+		atomic.AddUint64(&uo.SeqNum, 1)
+	})
+	uo.vmFamilyCache.OnEvicted(func(_ string, _ interface{}) {
 		atomic.AddUint64(&uo.SeqNum, 1)
 	})
 	return uo
@@ -53,18 +65,87 @@ func NewUnavailableOfferingsWithCache(c *cache.Cache) *UnavailableOfferings {
 
 func NewUnavailableOfferings() *UnavailableOfferings {
 	return NewUnavailableOfferingsWithCache(
-		cache.New(UnavailableOfferingsTTL, UnavailableOfferingsCleanupInterval))
+		cache.New(UnavailableOfferingsTTL, UnavailableOfferingsCleanupInterval),
+		// TODO - should we adjust cleanup interval for vmFamilyCache?
+		cache.New(UnavailableOfferingsTTL, UnavailableOfferingsCleanupInterval),
+	)
 }
 
 // IsUnavailable returns true if the offering appears in the cache
-func (u *UnavailableOfferings) IsUnavailable(instanceType, zone, capacityType string) bool {
+func (u *UnavailableOfferings) IsUnavailable(sku *skewer.SKU, zone, capacityType string) bool {
 	if capacityType == karpv1.CapacityTypeSpot {
-		if _, found := u.cache.Get(spotKey); found {
+		if _, found := u.singleOfferingCache.Get(spotKey); found {
 			return true
 		}
 	}
-	_, found := u.cache.Get(key(instanceType, zone, capacityType))
+
+	// check if there offering is marked as unavailable on vm family level
+	skuVCPUCount, err := sku.VCPU()
+	if err != nil {
+		skuVCPUCount = 0 // default to 0 if we can't determine VCPU count
+	}
+	if u.isFamilyUnavailable(sku.GetFamilyName(), capacityType, zone, skuVCPUCount) {
+		return true
+	}
+
+	// lastly check if the offering is marked as unavailable for the specific instance type, zone and capacity type
+	_, found := u.singleOfferingCache.Get(singleInstanceKey(sku.GetName(), zone, capacityType))
 	return found
+}
+
+func (u *UnavailableOfferings) isFamilyUnavailable(skuFamilyName, capacityType, zone string, cpuCount int64) bool {
+	// Check if VM family is blocked in the specific zone
+	if val, found := u.vmFamilyCache.Get(vmFamilyKey(skuFamilyName, capacityType, zone)); found {
+		if blockedCPUCount, ok := val.(int64); ok {
+			if blockedCPUCount == wholeVMFamilyBlockedSentinel {
+				// Entire VM family is blocked in this zone
+				return true
+			}
+			// VM sizes from this family are blocked for CPU counts >= blockedCPUCount in this zone
+			return cpuCount >= blockedCPUCount
+		}
+	}
+	return false
+}
+
+// MarkFamilyUnavailableWithTTL marks a VM family with custom TTL in a specific zone
+func (u *UnavailableOfferings) MarkFamilyUnavailableWithTTL(ctx context.Context, versionedSKUFamily, zone, capacityType string, cpuCount int64, ttl time.Duration) {
+	key := vmFamilyKey(versionedSKUFamily, zone, capacityType)
+
+	// Check if we already have a more restrictive limit
+	if existing, found := u.vmFamilyCache.Get(key); found {
+		if currentBlockedCPUCount, ok := existing.(int64); ok {
+			// If entire family is already blocked, don't override
+			if currentBlockedCPUCount == wholeVMFamilyBlockedSentinel {
+				return
+			}
+			// If new limit would be less restrictive, keep the existing one
+			// TODO - should we adjust TTL here? 
+			if cpuCount != wholeVMFamilyBlockedSentinel && currentBlockedCPUCount <= cpuCount {
+				return
+			}
+		}
+	}
+
+	log.FromContext(ctx).WithValues(
+		"family", versionedSKUFamily,
+		"capacity-type", capacityType,
+		"zone", zone,
+		"max-cpu", cpuCount,
+		"ttl", ttl).V(1).Info("marking VM family unavailable in zone")
+
+	u.vmFamilyCache.Set(key, cpuCount, ttl)
+	atomic.AddUint64(&u.SeqNum, 1)
+}
+
+// MarkWholeVMFamilyUnavailable marks an entire VM family as unavailable in a specific zone
+func (u *UnavailableOfferings) MarkWholeVMFamilyUnavailable(ctx context.Context, family, capacityType, zone string) {
+	u.MarkFamilyUnavailableWithTTL(ctx, family, zone, capacityType, wholeVMFamilyBlockedSentinel, UnavailableOfferingsTTL)
+}
+
+// MarkVMFamilyUnavailableAtCPUCount marks a VM family as unavailable for CPU counts above the threshold in a specific zone
+func (u *UnavailableOfferings) MarkVMFamilyUnavailableAtCPUCount(ctx context.Context, family, capacityType, zone string, maxCPUCount int64) {
+	u.MarkFamilyUnavailableWithTTL(ctx, family, zone, capacityType, maxCPUCount, UnavailableOfferingsTTL)
 }
 
 // MarkSpotUnavailable communicates recently observed temporary capacity shortages for spot
@@ -81,7 +162,7 @@ func (u *UnavailableOfferings) MarkUnavailableWithTTL(ctx context.Context, unava
 		"zone", zone,
 		"capacity-type", capacityType,
 		"ttl", ttl).V(1).Info("removing offering from offerings")
-	u.cache.Set(key(instanceType, zone, capacityType), struct{}{}, ttl)
+	u.singleOfferingCache.Set(singleInstanceKey(instanceType, zone, capacityType), struct{}{}, ttl)
 	atomic.AddUint64(&u.SeqNum, 1)
 }
 
@@ -91,11 +172,17 @@ func (u *UnavailableOfferings) MarkUnavailable(ctx context.Context, unavailableR
 }
 
 func (u *UnavailableOfferings) Flush() {
-	u.cache.Flush()
+	u.singleOfferingCache.Flush()
+	u.vmFamilyCache.Flush()
 	atomic.AddUint64(&u.SeqNum, 1)
 }
 
-// key returns the cache key for all offerings in the cache
-func key(instanceType string, zone string, capacityType string) string {
+// singleInstanceKey returns the cache singleInstanceKey for all offerings in the cache
+func singleInstanceKey(instanceType string, zone string, capacityType string) string {
 	return fmt.Sprintf("%s:%s:%s", capacityType, instanceType, zone)
+}
+
+// vmFamilyKey returns the cache key for VM family blocks in a specific zone
+func vmFamilyKey(skuFamilyName, zone, capacityType string) string {
+	return fmt.Sprintf("skuFamily:%s:%s:%s", skuFamilyName, zone, capacityType)
 }
--- a/pkg/cache/unavailableofferings_test.go
+++ b/pkg/cache/unavailableofferings_test.go
@@ -21,16 +21,26 @@ import (
 	"testing"
 	"time"
 
+	"github.com/Azure/skewer"
 	"github.com/patrickmn/go-cache"
 )
 
+func createTestSKU(name, family string) *skewer.SKU {
+	return &skewer.SKU{
+		Name:   &name,
+		Family: &family,
+	}
+}
+
 func TestUnavailableOfferings(t *testing.T) {
 	// create a new cache with a short TTL
-	c := cache.New(time.Second, time.Second)
-	u := NewUnavailableOfferingsWithCache(c)
+	singleInstanceCache := cache.New(time.Second, time.Second)
+	vmFamilyCache := cache.New(time.Second, time.Second)
+	u := NewUnavailableOfferingsWithCache(singleInstanceCache, vmFamilyCache)
+	testSKU := createTestSKU("NV16as_v4", "NVas_v4")
 
 	// test that an offering is not marked as unavailable initially
-	if u.IsUnavailable("NV16as_v4", "westus", "spot") {
+	if u.IsUnavailable(testSKU, "westus", "spot") {
 		t.Error("Offering should not be marked as unavailable initially")
 	}
 
@@ -38,7 +48,7 @@ func TestUnavailableOfferings(t *testing.T) {
 	u.MarkUnavailableWithTTL(context.TODO(), "test reason", "NV16as_v4", "westus", "spot", time.Second)
 
 	// test that the offering is now marked as unavailable
-	if !u.IsUnavailable("NV16as_v4", "westus", "spot") {
+	if !u.IsUnavailable(testSKU, "westus", "spot") {
 		t.Error("Offering should be marked as unavailable after being marked as such")
 	}
 
@@ -46,14 +56,14 @@ func TestUnavailableOfferings(t *testing.T) {
 	time.Sleep(time.Second)
 
 	// test that the offering is no longer marked as unavailable
-	if u.IsUnavailable("NV16as_v4", "westus", "spot") {
+	if u.IsUnavailable(testSKU, "westus", "spot") {
 		t.Error("Offering should not be marked as unavailable after cache entry has expired")
 	}
 }
 
 func TestUnavailableOfferings_KeyGeneration(t *testing.T) {
 	expectedKey := "spot:NV16as_v4:westus"
-	key := key("NV16as_v4", "westus", "spot")
+	key := singleInstanceKey("NV16as_v4", "westus", "spot")
 	if key != expectedKey {
 		t.Errorf("Expected key to be %s, but got %s", expectedKey, key)
 	}
--- a/pkg/cache/unavailableofferings_test.go
+++ b/pkg/cache/unavailableofferings_test.go
@@ -18,17 +18,36 @@ package cache
 
 import (
 	"context"
+	"strconv"
 	"testing"
 	"time"
 
+	"github.com/Azure/azure-sdk-for-go/profiles/latest/compute/mgmt/compute"
 	"github.com/Azure/skewer"
 	"github.com/patrickmn/go-cache"
+	"github.com/samber/lo"
+	karpv1 "sigs.k8s.io/karpenter/pkg/apis/v1"
 )
 
-func createTestSKU(name, family string) *skewer.SKU {
+func createTestSKU(name, family string, cpuCount int) *skewer.SKU {
 	return &skewer.SKU{
-		Name:   &name,
-		Family: &family,
+		Name:         &name,
+		Family:       &family,
+		Capabilities: &[]compute.ResourceSkuCapabilities{{Name: lo.ToPtr(skewer.VCPUs), Value: lo.ToPtr(strconv.Itoa(cpuCount))}},
+	}
+}
+
+func assertOfferingUnavailable(t *testing.T, u *UnavailableOfferings, sku *skewer.SKU, zone, capacityType, message string) {
+	t.Helper()
+	if !u.IsUnavailable(sku, zone, capacityType) {
+		t.Errorf("%s: %s", sku.GetName(), message)
+	}
+}
+
+func assertOfferingAvailable(t *testing.T, u *UnavailableOfferings, sku *skewer.SKU, zone, capacityType, message string) {
+	t.Helper()
+	if u.IsUnavailable(sku, zone, capacityType) {
+		t.Errorf("%s: %s", sku.GetName(), message)
 	}
 }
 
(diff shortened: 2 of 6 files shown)


pkg/cache/unavailableofferings.go
```diff
@@ -22,51 +22,126 @@ import (
...
+	if existing, found := u.vmFamilyCache.Get(key); found {
+		if currentBlockedCPUCount, ok := existing.(int64); ok {
+			// Keep the more restrictive limit for CPU count(lower value, with -1 being most restrictive - wholeVMFamilyBlockedSentinel)
+			if currentBlockedCPUCount <= cpuCount {
+				cpuCount = currentBlockedCPUCount
+			}
+		}
+	}
+
+	log.FromContext(ctx).WithValues(
```
matthchr: Dont use `WithValues` if you're not gonna use the resulting logger it creatures, instead just pass the values to `Info`

pkg/cache/unavailableofferings_test.go
```diff
@@ -18,43 +18,182 @@ package cache
...
 
 	// mark the offering as unavailable
-	u.MarkUnavailableWithTTL(context.TODO(), "test reason", "NV16as_v4", "westus", "spot", time.Second)
+	u.MarkUnavailableWithTTL(context.TODO(), "test reason", "NV16as_v4", "westus", karpv1.CapacityTypeSpot, time.Second)
 
 	// test that the offering is now marked as unavailable
-	if !u.IsUnavailable("NV16as_v4", "westus", "spot") {
-		t.Error("Offering should be marked as unavailable after being marked as such")
-	}
+	assertOfferingUnavailable(t, u, testSKU, "westus", karpv1.CapacityTypeSpot, "Offering should be marked as unavailable after being marked as such")
```
matthchr:
```suggestion
	assertOfferingUnavailable(t, u, testSKU, "westus", karpv1.CapacityTypeSpot, "Offering should be marked as unavailable")
```

pkg/cache/unavailableofferings_test.go
```diff
@@ -18,43 +18,182 @@ package cache
...
+	// Test that offerings are not marked as unavailable initially
+	for _, sku := range skus {
+		assertOfferingAvailable(t, u, sku, "westus-1", karpv1.CapacityTypeOnDemand, "Offering should not be marked as unavailable initially")
+	}
+
+	// Mark part of the VM family as unavailable (>= 16 CPUs)
+	u.MarkFamilyUnavailableWithTTL(context.TODO(), skus[0].GetFamilyName(), "westus-1", karpv1.CapacityTypeOnDemand, 16, time.Second)
+
+	// Test that 16+ CPU offerings are marked as unavailable, but 8 CPU is not
+	assertOfferingAvailable(t, u, skus[0], "westus-1", karpv1.CapacityTypeOnDemand, "8 CPU offering should remain available after partial family marking")
```
matthchr: Name the `skus` rather than `[0]` 1, 2... clearer for readers? You can put the named SKUs in a collection still.

pkg/cache/unavailableofferings_test.go
```diff
@@ -18,43 +18,182 @@ package cache
...
+	// Wait for cache expiration
+	time.Sleep(time.Second)
+
+	// Test that offerings are no longer marked as unavailable after expiration
+	for _, sku := range skus {
+		assertOfferingAvailable(t, u, sku, "westus-1", karpv1.CapacityTypeOnDemand, "Offering should not be marked as unavailable after cache expiration")
+	}
+
+	// Mark the entire VM family as unavailable
+	u.MarkFamilyUnavailableWithTTL(context.TODO(), skus[0].GetFamilyName(), "westus-1", karpv1.CapacityTypeOnDemand, -1, time.Second)
```
matthchr: I think this is a different test.

pkg/cache/unavailableofferings_test.go
```diff
@@ -18,43 +18,182 @@ package cache
...
+	}
+
+	// Test that offerings from same VM family but different version are available
+	differentVersionSKU := createTestSKU("NV16as_v5", "NVasv5Family", 16)
+	assertOfferingAvailable(t, u, differentVersionSKU, "westus-1", karpv1.CapacityTypeOnDemand, "Offering from a different version of the same VM family should be available")
+
+	// Wait for cache expiration
+	time.Sleep(time.Second)
+
+	// // Test that offerings with both more and fewer than 16 CPUs are now marked as available after expiration
```
matthchr:
```suggestion
	// Test that all offerings are now marked as available after expiration
```

pkg/cache/unavailableofferings.go
```diff
@@ -22,51 +22,126 @@ import (
...
 // UnavailableOfferings stores any offerings that return ICE (insufficient capacity errors) when
 // attempting to launch the capacity. These offerings are ignored as long as they are in the cache on
 // GetInstanceTypes responses
 type UnavailableOfferings struct {
 	// key: <capacityType>:<instanceType>:<zone>, value: struct{}{}
-	cache  *cache.Cache
-	SeqNum uint64
+	singleOfferingCache *cache.Cache
+	// key: <skuFamilyName>:<zone>:<capacityType>, value: int64 (CPU count at or above which we block, or wholeVMFamilyBlockedSentinel if entire family is blocked)
+	vmFamilyCache *cache.Cache
```
matthchr, line 52: Looking at this more: wondering if we actually need 2 caches. Is singleOfferingCache not the same as `MarkFamilyUnavailableWithTTL` w/ the core-count specified? i.e. if I get explicit denial for Standard_D4_v5, is that the same as denial for Stanadard_D_v5` family, >=4 cores?

matthchr: Maybe this has to do with the questions you raised in teams about having incomplete access to the details you need though?

pkg/cache/unavailableofferings_test.go
```diff
@@ -18,43 +18,193 @@ package cache
...
+
+	// Wait for cache expiration
+	time.Sleep(2 * time.Second)
+
+	// Test that offerings are no longer marked as unavailable after expiration
+	for _, sku := range skus {
+		assertOfferingAvailable(t, u, sku, "westus-1", karpv1.CapacityTypeOnDemand, "Offering should not be marked as unavailable after cache expiration")
+	}
+
+	// Mark the entire VM family as unavailable
```
matthchr, line 138: I think this is a different test?

pkg/cache/unavailableofferings_test.go
```diff
@@ -18,43 +18,193 @@ package cache
...
+func getVersionedSKUFamily(sku *skewer.SKU) string {
+	skuVMSize, _ := sku.GetVMSize()
+	skuVersion := "1"
+	if skuVMSize.Version != "" {
+		skuVersion = skuVMSize.Version[1:]
+	}
+	return skuVMSize.Family + skuVersion
+}
+
+func TestUnavailableOfferingsVMFamilyLevel(t *testing.T) {
```
matthchr: minor: Consider naming this test a bit better?

IMO tests should say what they are doing and what they expect to happen (we don't follow that consistently in the Karpenter codebase).

I think this is actually two tests:
`TestUnavailableOfferings_VMFamilyCoreLimit_AllowsFewerCores`
and
`TestUnavailableOfferings_VMFamily_BlocksAll`

(or something -- free free to fiddle with the names but I think you get the idea).

pkg/cache/unavailableofferings_test.go
```diff
@@ -18,43 +18,193 @@ package cache
...
+	skus := []*skewer.SKU{
+		createTestSKU("NV8as_v4", "NVasv4Family", 8),
+		createTestSKU("NV16as_v4", "NVasv4Family", 16),
+		createTestSKU("NV24as_v4", "NVasv4Family", 24),
+		createTestSKU("NV32as_v4", "NVasv4Family", 32),
+	}
+
+	// Test 1: Set a limit at 16 CPUs, then try to set a less restrictive limit at 24 CPUs
+	// The 16 CPU limit should be preserved
+	u.MarkFamilyUnavailableWithTTL(context.TODO(), getVersionedSKUFamily(skus[0]), "westus-1", karpv1.CapacityTypeOnDemand, 16, 2*time.Second)
```
matthchr: You set 10s up above but then override at 2s here... consider just hardcoding these 10s/2s times in a const for all the tests and reusing? Also possibly this should be 10s here, or there should be a variant for family that doesn't require you to specify the TTL when specifying the cores?

pkg/providers/instance/resperrorhandlers.go
```diff
@@ -143,10 +145,33 @@ func handleSKUNotAvailableError(ctx context.Context, provider *DefaultProvider,
...
+	if len(instanceTypeSKUCPURequirements) > 0 {
+		parsedCPUCount, parseErr := strconv.ParseInt(instanceType.Requirements.Get(v1beta1.LabelSKUCPU).Values()[0], 10, 64)
+		if parseErr != nil {
+			skuCPUCount = 0 // fallback to 0 if parsing fails, but this should not happen
+		} else {
+			skuCPUCount = parsedCPUCount
+		}
+	}
+
+	provider.unavailableOfferings.MarkFamilyUnavailableWithTTL(ctx, skuFamily+skuFamilyVersion, zone, karpv1.CapacityTypeOnDemand, skuCPUCount, AllocationFailureTTL)
```
matthchr: Does `skuFamily+skuFamilyVersion` actually build the correct "full family"?

Feel like this method deserves some comments explaining why we're doing this building here (unless this is going to change to a better way?)

pkg/cache/unavailableofferings_test.go
UtheMan: Yeah good idea to split these two, initially I just included all of the "VM family cache" functionality in that one test. I separated them into two distinct ones, kept the names you suggested since they seemed to explain everything properly.

pkg/cache/unavailableofferings_test.go
UtheMan: Changed to use `testUnavailableOfferingsTTL = 2 * time.Second` across the board

pkg/cache/unavailableofferings_test.go
UtheMan: Could you share an example of what you mean? :) I'm a bit confused about what you have in mind here to be honest (mostly about the `You can put the named SKUs in a collection still.` part)

pkg/providers/instance/resperrorhandlers.go
UtheMan: So it depends on what we consider "full family". In `skewer.SKU` type, there is a `Family` field which holds what I would consider the "full family". Problem is, as we discussed offline, is that this value can be formatted in a way that makes it trickier to use as a label.

For example, the data we have from skewer (from `SKU` and `VMSizeType` types), some of which we propagate to `InstanceType` in `computeRequirements(...)` function:

Standard_D2_v3
- SKU: `"Standard_D2_v3"` (available in `skewer.SKU`)
- Family: `"standardDv3Family"` (available in `skewer.SKU`)
- SubFamily: `""` (available in `skewer.VMSizeType`, currently not passed to `InstanceType`)
- Version: `"3"` (available in `skewer.VMSizeType`)


Standard_DS2_v2
- SKU `"Standard_DS2_v2"` (available in `skewer.SKU`)
- Family: `"standardDSv2Family"` (available in `skewer.SKU`)
- SubFamily: `"S"` (available in `skewer.VMSizeType`, currently not passed to `InstanceType`)
- Version: `"2"`(available in `skewer.VMSizeType`)


So the combination of `skuFamily + skuFamilyVersion` would give us:
- Standard_D2_v3 -> `D3`
- Standard_DS2_v2 -> `D2` (we lose subfamily information)

I'm thinking we should probably add `SubFamily` to requirements here: https://github.com/Azure/karpenter-provider-azure/blob/ea180666df80fd50a72f7bbc5a1cdb1831194828/pkg/providers/instancetype/instancetype.go#L179

to differentiate between cases like `Standard_D2_v3` and `Standard_D2S_v3`

UtheMan: Actually, after digging a bit more, there is also another field which i believe we should take into account, namely `additiveFeatures` inside of `skewer.VMSizeType`, which, for some SKUs, holds relevant information. But it appears to not be entirely consistent across different SKUs and versions... 

So I think we should have something like:

`requirements[v1beta1.LabelSKUFamily].Insert(vmsize.Family + vmsize.SubFamily + vmsize.additiveFeatures)`

pkg/cache/unavailableofferings.go
UtheMan: `if I get explicit denial for Standard_D4_v5, is that the same as denial for Stanadard_D_v5` family, >=4 cores?` 
Yes, technically it could be the same. If we failed to allocate 4 cores of D series v5, then it means there was not enough capacity for that -> we should not be able to allocate more. .

I went ahead with two separate caches, because in `singleOfferingCache` we use empty struct keys and for the VM family one I use cpu count. We could potentially change the original cache to also pass number of CPUs as value, to which we have access in `instanceType`. Another reason why I wanted to separate them was because of different key values (e.g. Standard_D4_v5 fails, then we'd need to adjust the key to know that Standard_D32_v5 will fail too). I guess it's premature optimization to worry about that now, but adding cpu count to every cache entry might be a bit overkill since zonal allocation failures (currently the only error I add family-level block for) seem fairly infrequent... 

Would definitely be simpler to have one cache, I agree. We don't expose the details of what's inside to "consumer" outside of the `IsUnavailable` call, so maybe different kinds of values/keys won't be that big of an issue? Let me know your thoughts.

pkg/providers/instance/resperrorhandlers.go
UtheMan: Changed the approach as per discussion offline with @matthchr , we have access to complete `skewer.SKU` data through `instanceTypeProvider.Get(...)` so I'm using that instead to retrieve vCPU count and full SKU family name.

pkg/cache/unavailableofferings.go
UtheMan: As discussed offline, we decided to stick to separate caches for now. I explained the reasoning behind this in a comment here: https://github.com/Azure/karpenter-provider-azure/blob/0ee568e9305ba30baa00dc680912c61dc8826ee5/pkg/cache/unavailableofferings.go#L44 

```
We maintain two caches (singleOfferingCache and vmFamilyCache) to better handle error cases in which we know that allocation from a specific VM family + capacity type + zone combination will not work.
Information available from skewer.SKU is used to determine details about the VM SKU for which we encountered allocation errors.
We don't bundle the two caches together into one to avoid accidentally bundling together different SKUs while handling errors which don't necessarily warrant blocking more than just the single instance type.
This could be adjusted in the future, as we gather more data and get more confidence in information available in skewer.SKU.
```

pkg/cache/unavailableofferings.go
```diff
@@ -19,54 +19,136 @@ package cache
...
+func (u *UnavailableOfferings) isFamilyUnavailable(sku *skewer.SKU, zone, capacityType string) bool {
+	skuVCPUCount, err := sku.VCPU()
+	if err != nil {
+		// default to 0 if we can't determine VCPU count, this shouldn't happen as long as data in skewer.SKU is correct
+		skuVCPUCount = 0
+	}
+	// Check if VM family is blocked in the specific zone
+	if val, found := u.vmFamilyCache.Get(vmFamilyKey(sku.GetFamilyName(), zone, capacityType)); found {
+		if blockedCPUCount, ok := val.(int64); ok {
+			if blockedCPUCount == wholeVMFamilyBlockedSentinel {
```
matthchr, line 105: minor: This code is technically not required as long as skewer doesn't ever report negative cores, right? (Since `<sku count >= 0> >= -1` will always return true?)

If your intent is to leave this for clarity that's fine (the code is correct either way).

pkg/utils/utils.go
```diff
@@ -53,6 +54,19 @@ func GetVMName(providerID string) (string, error) {
 	return "", fmt.Errorf("parsing vm name %s", providerID)
 }
 
+// extractVersionFromVMSize extracts and normalizes the version from VMSizeType, dropping "v" prefix and backfilling "1"
+func ExtractVersionFromVMSize(vmsize *skewer.VMSizeType) string {
```
matthchr, line 58: Missing unit test?

pkg/cache/unavailableofferings.go
```diff
@@ -19,54 +19,136 @@ package cache
...
+		"max-cpu", cpuCount,
+		"ttl", ttl)
+
+	// call Set to update the cache entry, even if it already exists, to extend its TTL
+	u.vmFamilyCache.Set(key, cpuCount, ttl)
+	atomic.AddUint64(&u.SeqNum, 1)
+}
+
+// MarkWholeVMFamilyUnavailable marks an entire VM family as unavailable in a specific zone
+func (u *UnavailableOfferings) MarkWholeVMFamilyUnavailable(ctx context.Context, skuFamilyName, capacityType, zone string) {
```
matthchr: Am I missing where these two functions are actually called? I don't see any usage.

pkg/cache/unavailableofferings.go
```diff
@@ -19,54 +19,136 @@ package cache
...
+			// VM sizes from this family are blocked for CPU counts >= blockedCPUCount in this zone
+			return skuVCPUCount >= blockedCPUCount
+		}
+	}
+	return false
+}
+
+// MarkFamilyUnavailableWithTTL marks a VM family with custom TTL in a specific zone
+// skuFamilyName e.g. "StandardDv2Family" for "Standard_D2_v2" VM SKU
+func (u *UnavailableOfferings) MarkFamilyUnavailableWithTTL(ctx context.Context, skuFamilyName, zone, capacityType string, cpuCount int64, ttl time.Duration) {
```
matthchr: nit, naming consistency:

`VMFamily` vs `Family` -- you don't say `VMFamily` here but you do elsewhere. Feels like the public interface should be consistent in what it calls it. Given other methods that already existed here didn't bother with VM, I'd go for:

`MarkFamilyUnavailableWithTTL`
`MarkFamilyUnavailable` // was MarkWhole...
`MarkFamilyUnavailableAtCPUCount`

The other thing you _could_ do here (this is not a requirement, just making sure you're aware of the pattern) is some Options variant, where you offer a single `MarkFamilyUnavailable` that takes an either a variadic last arg for opts, or a Struct with the `CPUCount` and `TTL` parameters, which callers can optionally pass -- if the struct is empty they get the defaults for those values (== marking the whole family unavailable).
The advantage of doing it that way from a callers perspective is you get all the overloads without needing to specify them (here you have 3/4, there's no "Whole" variant with a TTL)

pkg/cache/unavailableofferings_test.go
```diff
@@ -18,43 +18,213 @@ package cache
...
+const (
+	testUnavailableOfferingsTTL = 2 * time.Second
+)
+
+func createTestSKU(name, familyName, size string, cpuCount int) *skewer.SKU {
+	return &skewer.SKU{
+		Name:         &name,
+		Family:       &familyName,
+		Size:         &size,
+		Capabilities: &[]compute.ResourceSkuCapabilities{{Name: lo.ToPtr(skewer.VCPUs), Value: lo.ToPtr(strconv.Itoa(cpuCount))}},
```
matthchr: minor, style: hard to read - consider splitting onto multiple lines for ease of editing/reading.

pkg/cache/unavailableofferings_test.go
matthchr: Sorry missed this comment, would've replied quicker!

```
nv8 := createTestSKU("Standard_NV8as_v4", "standardNVasv4Family", "NV8as_v4", 8)
nv16 := createTestSKU("Standard_NV16as_v4", "standardNVasv4Family", "NV16as_v4", 16)
nv24 := createTestSKU("Standard_NV24as_v4", "standardNVasv4Family", "NV24as_v4", 24)
skus := []*skewer.SKU{nv8, nv16, nv24}
```

It's a really minor nit though so if you don't like it ignore me.

pkg/cache/unavailableofferings_test.go
```diff
@@ -18,43 +18,213 @@ package cache
...
+		createTestSKU("Standard_NV24as_v4", "standardNVasv4Family", "NV24as_v4", 24),
+	}
+
+	// Test that offerings are not marked as unavailable initially
+	for _, sku := range skus {
+		assertOfferingAvailable(t, u, sku, "westus-1", karpv1.CapacityTypeOnDemand, "Offering should not be marked as unavailable initially")
+	}
+
+	// Mark the entire VM family as unavailable
+	u.MarkFamilyUnavailableWithTTL(context.TODO(), skus[0].GetFamilyName(), "westus-1", karpv1.CapacityTypeOnDemand, -1, testUnavailableOfferingsTTL)
```
matthchr: Not using the `WholeFamily` variant, should it be?

pkg/providers/instancetype/instancetype.go
```diff
@@ -211,15 +211,10 @@ func setRequirementsGPU(requirements scheduling.Requirements, sku *skewer.SKU, v
 	}
 }
 
-// setRequirementsVersion sets the SKU version label, dropping "v" prefix and backfilling "1"
```
matthchr, line 214: Why remove the comment?

pkg/providers/instance/resperrorhandlers.go
```diff
@@ -143,24 +144,30 @@ func handleSKUNotAvailableError(ctx context.Context, provider *DefaultProvider,
...
+func handleZonalAllocationFailureError(ctx context.Context, provider *DefaultProvider, sku *skewer.SKU, instanceType *corecloudprovider.InstanceType, zone, capacityType string, err error) error {
 	log.FromContext(ctx).Error(err, "zonal allocation failure", "instance-type", instanceType.Name, "zone", zone)
-	provider.unavailableOfferings.MarkUnavailableWithTTL(ctx, ZonalAllocationFailureReason, instanceType.Name, zone, karpv1.CapacityTypeOnDemand, AllocationFailureTTL)
-	provider.unavailableOfferings.MarkUnavailableWithTTL(ctx, ZonalAllocationFailureReason, instanceType.Name, zone, karpv1.CapacityTypeSpot, AllocationFailureTTL)
+	vCPU, err := sku.VCPU() // versionedSKUFamily e.g. "N4" for "NV8as_v4"
+	if err != nil {
+		// default to 0 if we can't determine VCPU count, this shouldn't happen as long as data in skewer.SKU is correct
+		vCPU = 0
+	}
+	provider.unavailableOfferings.MarkFamilyUnavailableWithTTL(ctx, sku.GetFamilyName(), zone, karpv1.CapacityTypeOnDemand, vCPU, AllocationFailureTTL)
```
matthchr: One thing that I notice on our calls here.

There's only a _single_ call to `unavailableOfferings` with the default TTL. There's 5 with `AllocationFailureTTL`, and one with `skuNotAvailableTTL` (which is a bit dynamic).

This makes me wonder if we just want the unavailableOfferings interface to always require a TTL, since there seem to be a couple different options and most aren't using the default.

We _could_ make the default `AllocationFailureTTL` since that's used most often, which would be another angle that would be IMO better than what we have now.

pkg/providers/instance/resperrorhandlers_test.go
```diff
@@ -35,15 +39,112 @@ import (
...
+			expectedUnavailableOfferingsInformation: []offeringToCheck{},
+			expectedAvailableOfferingsInformation:   []offeringToCheck{},
+		},
+	}
+}
+
+func (b *testCaseBuilder) withInstanceType(offerings ...offering) *testCaseBuilder {
+	offeringTypes := make([]struct{ zone, capacityType string }, len(offerings))
+	for i, o := range offerings {
+		offeringTypes[i] = struct{ zone, capacityType string }{zone: o.zone, capacityType: o.capacityType}
```
matthchr: minor: newlines?

pkg/providers/instance/resperrorhandlers_test.go
```diff
@@ -35,15 +39,112 @@ import (
...
+	}
+	b.tc.instanceType = defaultCreateInstanceType(offeringTypes...)
```
matthchr: Should this helper `defaultCreateInstanceType` just take `offering`? Bascially go through your whole test here and drop `struct{ zone, capacityType string }` in favor of the named type?

pkg/providers/instance/resperrorhandlers_test.go
```diff
@@ -35,15 +39,112 @@ import (
...
+	b.tc.responseErr = nil
+	return b
+}
+
+func (b *testCaseBuilder) expectError(err error) *testCaseBuilder {
+	b.tc.expectedErr = err
+	return b
+}
+
+func (b *testCaseBuilder) expectNilError() *testCaseBuilder {
```
matthchr: This is just `expectError(nil)` right? Maybe can just remove in favor of that?

pkg/providers/instance/resperrorhandlers_test.go
```diff
@@ -35,15 +39,112 @@ import (
...
+	b.tc.capacityType = capacityType
+	return b
+}
+
+func (b *testCaseBuilder) withResponseError(errorCode, errorMessage string) *testCaseBuilder {
+	b.tc.responseErr = createResponseError(errorCode, errorMessage)
+	return b
+}
+
+func (b *testCaseBuilder) withNilResponseError() *testCaseBuilder {
```
matthchr: Seems implied/default?

Any actual need for it?

pkg/cache/unavailableofferings.go
matthchr: See my other comment about "maybe we only want WithTTL variants" too.

matthchr: I think given the actual usage maybe just: `MarkFamilyUnavailable(ctx context.Context, skuFamilyName, zone, capacityType string, ttl time.Duration)` and `MarkFamilyUnavailableAtCPUCount(ctx context.Context, skuFamilyName, zone, capacityType string, cpuCount int64, ttl time.Duration)` is enough, if you don't wanna go the options route.

pkg/cache/unavailableofferings.go
UtheMan: Yes that's right, we could remove it, but I wanted to be explicit about the option of blocking the entire family as well. I think it's a bit more readable this way

pkg/cache/unavailableofferings.go
UtheMan: Good catch, they were leftovers from previous versions of my code, removing them for now.

pkg/cache/unavailableofferings_test.go
UtheMan: We could use that or just pass `-1`. But since there is no use for the whole family variant right now outside of tests, I will just keep it as is for the time being.

pkg/providers/instance/resperrorhandlers_test.go
UtheMan: Yeah can be removed, good point!

pkg/utils/utils.go
UtheMan: Added one in `utils_test.go`

pkg/providers/instance/resperrorhandlers_test.go
UtheMan: Good point, this test file went through like 10 iterations. Changing to use `offering`

pkg/cache/unavailableofferings_test.go
UtheMan: Gotcha, that is definitely a lot nicer, changing !

pkg/providers/instancetype/instancetype.go
UtheMan: Removed by accident, adding back

pkg/cache/unavailableofferings.go
UtheMan: Will hold off with adding options for now, but it could be a good follow-up for this - I'm guessing people might prefer to have more control over this behavior, but probably out of scope for this PR, since it's already pretty sizeable. 
I changed to `MarkFamilyUnavailable(ctx context.Context, skuFamilyName, zone, capacityType string, ttl time.Duration) and MarkFamilyUnavailableAtCPUCount(ctx context.Context, skuFamilyName, zone, capacityType string, cpuCount int64, ttl time.Duration)` like you suggested.

pkg/cache/unavailableofferings_test.go
UtheMan: Actually, I decided to bring it back and to use the naming you suggested in your another comment, it's `MarkFamilyUnavailableWithTTL` now.

pkg/providers/instance/resperrorhandlers.go
UtheMan: True, it is kind of all over the place now. I think we could maybe address it by adding options parameters like you mentioned, but I'd prefer to do it in a separate PR, this one is pretty bloated already.

pkg/cache/unavailableofferings.go
matthchr: happy leaving it this way.

pkg/cache/unavailableofferings.go
```diff
@@ -139,14 +139,8 @@ func (u *UnavailableOfferings) MarkFamilyUnavailableWithTTL(ctx context.Context,
...
 
-// MarkWholeVMFamilyUnavailable marks an entire VM family as unavailable in a specific zone
-func (u *UnavailableOfferings) MarkWholeVMFamilyUnavailable(ctx context.Context, skuFamilyName, capacityType, zone string) {
-	u.MarkFamilyUnavailableWithTTL(ctx, skuFamilyName, zone, capacityType, wholeVMFamilyBlockedSentinel, UnavailableOfferingsTTL)
-}
-
-// MarkVMFamilyUnavailableAtCPUCount marks a VM family as unavailable for CPU counts above the threshold in a specific zone
-func (u *UnavailableOfferings) MarkVMFamilyUnavailableAtCPUCount(ctx context.Context, skuFamilyName, capacityType, zone string, maxCPUCount int64) {
-	u.MarkFamilyUnavailableWithTTL(ctx, skuFamilyName, zone, capacityType, maxCPUCount, UnavailableOfferingsTTL)
+func (u *UnavailableOfferings) MarkFamilyUnavailable(ctx context.Context, skuFamilyName, zone, capacityType string, ttl time.Duration) {
```
matthchr, line 144: minor: Missing comment.

pkg/cache/unavailableofferings.go
```diff
@@ -113,9 +113,9 @@ func (u *UnavailableOfferings) isFamilyUnavailable(sku *skewer.SKU, zone, capaci
 	return false
 }
 
-// MarkFamilyUnavailableWithTTL marks a VM family with custom TTL in a specific zone
+// MarkFamilyUnavailableAtCPUCount marks a VM family with custom TTL in a specific zone
```
matthchr: Minor: correct comment -- it should probably reference a cpuCount somehow.

Also maybe should reference that -1 is "all"

pkg/cache/unavailableofferings_test.go
matthchr: I think you mean `MarkFamilyUnavailableAtCPUCount`? But happy with this.

pkg/providers/instance/resperrorhandlers.go
matthchr: Agree doing it in a separate PR would be better.

pkg/providers/instance/resperrorhandlers_test.go
```diff
@@ -87,12 +87,8 @@ func newTestCase(name string) *testCaseBuilder {
 	}
 }
 
-func (b *testCaseBuilder) withInstanceType(offerings ...offering) *testCaseBuilder {
-	offeringTypes := make([]struct{ zone, capacityType string }, len(offerings))
-	for i, o := range offerings {
-		offeringTypes[i] = struct{ zone, capacityType string }{zone: o.zone, capacityType: o.capacityType}
-	}
-	b.tc.instanceType = defaultCreateInstanceType(offeringTypes...)
+func (b *testCaseBuilder) withDefaultInstanceType(offerings ...offering) *testCaseBuilder {
```
matthchr: minor: Why `defaultInstanceType` vs `instanceType`? What's the difference between a "default type" and "the instance type"?

matthchr (dismissed):
LGTM, had a few minor doc improvement suggestions

pkg/providers/instance/resperrorhandlers_test.go
UtheMan: Initially there were tests that used different instance type names, right now they all use an empty instance type  or  
```
responseErrorTestInstanceName  = "Standard_D2s_v3"
```

will change to 

`func (b *testCaseBuilder) withInstanceType(instanceTypeName string, offerings ...offering) *testCaseBuilder `

To make it more extendible in the future.

UtheMan: Scratch that, will just rename the function in the end. Since all the tests would pass the same argument then and linter is unhappy.

pkg/cache/unavailableofferings.go
UtheMan: Added

Issue #833, opened 2025-04-24
General logging improvements


tallaxes:
Logging can use a good quality pass. Some suggestions:

* Remove unnecessary logging in places where error is returned and would be logged later (e.g. by controller) - and develop guidance for when to log when not to log
* Review logging levels (e.g. maybe drop per-request levels to debug) - and develop guidance for use of log levels
* Review for good use of structural logging (WithValues etc.)
* Consider structured errors (https://github.com/awslabs/operatorpkg/tree/main/serrors)
* Find a clean way to make middleware-driven logging optional
* New: reduce/eliminate use of klog in favor of structured logging

alimaazamat:
For this issue:
- #968
- #1010
- #1028
- #1031 
- Docs: #1001 came originally from #968

PR #968, opened 2025-06-13
chore: Standardized logging with consistent keys and WithValues usage

alimaazamat:
Contributes to #833

**Description**
Consistent keys in structured logging across the code space, keys listed below.

* `make az-klogs-watch` show JSON structured logging successfully
* deployed resource intensive workload forcing new Azure VMs, kubectl logs show JSON structured logging successfully

- [ ] Yes, issue opened: # 
- [X] No

**Release Note**

```release-note
Standardized logging with consistent key naming and optimized WithValues usage.
```

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,4 +1,4 @@
-FROM mcr.microsoft.com/oss/go/microsoft/golang:1.24.4-bookworm@sha256:c7171283a29bbeafeaf7abba08c7cd9868a67033537c1cffbdf8ac1d65080b5b
+FROM mcr.microsoft.com/oss/go/microsoft/golang:1.24.4-bookworm@sha256:d9559f122feadff27b4adb8e491bc073ac9b41a80d6e38c0b0fae7224dfcf512
 
 # [Optional] Uncomment this section to install additional OS packages.
 # graphviz for pprof
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/Azure/karpenter-provider-azure
 go 1.24.4
 
 require (
-	github.com/Azure/aks-middleware v0.0.37
+	github.com/Azure/aks-middleware v0.0.40
 	github.com/Azure/azure-kusto-go v0.16.1
 	github.com/Azure/azure-sdk-for-go v68.0.0+incompatible
 	github.com/Azure/azure-sdk-for-go-extensions v0.1.9
--- a/go.sum
+++ b/go.sum
@@ -1,5 +1,5 @@
-github.com/Azure/aks-middleware v0.0.37 h1:WH46dEtAU3bJsGHHLvgefsqa0SBFznsWwnP53K43wNQ=
-github.com/Azure/aks-middleware v0.0.37/go.mod h1:s03PzzzlJm1+9TerNBdCDb6p3GTYgWU9056IlpgTThc=
+github.com/Azure/aks-middleware v0.0.40 h1:eFRuAxCcIAZoy/6+FvumDl2KOWnSPxXcAeCSOA4+aTo=
+github.com/Azure/aks-middleware v0.0.40/go.mod h1:7Y+wxZmS7p1K0FPreiO3+6Wr8YhYjWz9c50YohDQIQ4=
 github.com/Azure/azure-kusto-go v0.16.1 h1:vCBWcQghmC1qIErUUgVNWHxGhZVStu1U/hki6iBA14k=
 github.com/Azure/azure-kusto-go v0.16.1/go.mod h1:9F2zvXH8B6eWzgI1S4k1ZXAIufnBZ1bv1cW1kB1n3D0=
 github.com/Azure/azure-sdk-for-go v68.0.0+incompatible h1:fcYLmCpyNYRnvJbPerq7U0hS+6+I79yEDJBqVNcqUzU=


alimaazamat:
**List of all the keys in the structured logging:**
1. actualHash
2. actualK8sVersion
3. actualKubeletIdentityClientId
4. capacityType
5. currentK8sVersion
6. discoveredK8sVersion
7. driftType
8. error
9. expectedK8sVersion
10. expectedKubeletIdentityClientId
11. extensionId
12. goalHash
13. goalK8sVersion
14. imageId
15. instanceType
16. instanceTypeCount
17. ipv4PoolCount
18. lastOnDemandUpdateTime
19. lastSpotUpdateTime
20. loadBalancerCount
21. nicId
22. nicName
23. nodeClassHash
24. nodeClaimName
25. nsgCount
26. objectType
27. providerID
28. region
29. resourceGroup
30. resourceName
31. skuCount
32. vmName
33. vmPatch
34. zone

**GROUPED TOGETHER BY FUNCTION:**

AZURE:
- region
- resourceGroup
- resourceName
- zone

VM DATA:
- extensionId
- instanceType
- instanceTypeCount
- vmName
- vmPatch
- skuCount
- capacityType
- providerID
- imageId

NETWORKING:
- nicName
- nicId
- ipv4PoolCount
- loadBalancerCount
- nsgCount

K8S:
- nodeClaimName
- nodeClassHash
- currentK8sVersion
- discoveredK8sVersion
- expectedK8sVersion
- goalK8sVersion
- actualK8sVersion
- expectedKubeletIdentityClientId
- actualKubeletIdentityClientId
- objectType

OPERATIONS:
- error

VERSIONING/DRIFT:
- driftType
- goalHash
- actualHash
- lastOnDemandUpdateTime
- lastSpotUpdateTime

pkg/providers/instance/instance.go
```diff
@@ -270,26 +270,41 @@ func (p *DefaultProvider) DeleteNic(ctx context.Context, nicName string) error {
...
+	log.FromContext(ctx).WithValues(
+		"vmName", vmName,
+	).V(1).Info("creating virtual machine AKS identifying extension")
 	v, err := createVirtualMachineExtension(ctx, p.azClient.virtualMachinesExtensionClient, p.resourceGroup, vmName, vmExtName, *vmExt)
 	if err != nil {
-		log.FromContext(ctx).Error(err, fmt.Sprintf("Creating VM AKS identifying extension for VM %q failed", vmName))
+		log.FromContext(ctx).WithValues(
+			"vmName", vmName,
+			"extensionName", vmExtName,
+		).Error(err, "failed to create VM AKS identifying extension")
```
alimaazamat: remove this only if this gets logged later, will error returned be logged later

alimaazamat: return error as long as it gets logged later

pkg/cloudprovider/cloudprovider.go
```diff
@@ -190,13 +190,13 @@ func (c *CloudProvider) waitOnPromise(ctx context.Context, promise *instance.Vir
 		vmName := lo.FromPtr(promise.VM.Name)
 		err = c.instanceProvider.Delete(ctx, vmName)
 		if cloudprovider.IgnoreNodeClaimNotFoundError(err) != nil {
-			log.FromContext(ctx).Error(err, fmt.Sprintf("failed to delete VM %s", vmName))
+			log.FromContext(ctx).WithValues("vmName", vmName).Error(err, "failed to delete VM")
```
tallaxes: Wondering if for names (which is probably the majority) we should use "vm" rather than "vmName". Not a strong opinon, wondering if there is a best practice here, and what others think ...

pkg/cloudprovider/cloudprovider.go
```diff
@@ -446,7 +446,7 @@ func (c *CloudProvider) instanceToNodeClaim(ctx context.Context, vm *armcompute.
 	}
 
 	if zone, err := utils.GetZone(vm); err != nil {
-		log.FromContext(ctx).Info(fmt.Sprintf("WARN: Failed to get zone for VM %s, %v", *vm.Name, err))
+		log.FromContext(ctx).WithValues("vmName", *vm.Name).V(1).Info("failed to get zone for VM", "error", err)
```
tallaxes: This also drop the level, is that intentional? Also why different pattern for err (why not in WithValues)?

pkg/cloudprovider/drift.go
```diff
@@ -104,7 +108,7 @@ func (c *CloudProvider) isK8sVersionDrifted(ctx context.Context, nodeClaim *karp
 		// Note: we don't consider this a hard failure for drift if the KubernetesVersion is invalid/not ready to use, so we ignore returning the error here.
 		// We simply ensure the stored version is valid and ready to use, if we are to calculate potential Drift based on it.
 		// TODO (charliedmcb): I'm wondering if we actually want to have these soft-error cases switch to return an error if no-drift condition was found across all of IsDrifted.
-		logger.Info(fmt.Sprintf("WARN: Kubernetes version readiness invalid when checking drift: %s", err))
+		logger.Info("kubernetes version readiness invalid when checking drift", "error", err)
```
tallaxes: Same here, wondering on difference/pros/cons of this vs WithValues

pkg/cloudprovider/drift.go
```diff
@@ -170,7 +178,7 @@ func (c *CloudProvider) isImageVersionDrifted(
 		// Note: we don't consider this a hard failure for drift if the Images are not ready to use, so we ignore returning the error here.
 		// The stored Images must be ready to use if we are to calculate potential Drift based on them.
 		// TODO (charliedmcb): I'm wondering if we actually want to have these soft-error cases switch to return an error if no-drift condition was found across all of IsDrifted.
-		logger.Info(fmt.Sprintf("WARN: NodeImage readiness invalid when checking drift: %s", err))
+		logger.Info("WARN: NodeImage readiness invalid when checking drift", "error", err)
```
tallaxes: Elsewhere you removed "WARN:", let's pick up a pattern and stay consistent.

pkg/controllers/nodeclaim/inplaceupdate/controller.go
```diff
@@ -184,6 +184,6 @@ func logVMPatch(ctx context.Context, update *armcompute.VirtualMachineUpdate) {
 			raw, _ := json.Marshal(update)
 			rawStr = string(raw)
 		}
-		log.FromContext(ctx).V(1).Info(fmt.Sprintf("applying patch to Azure VM: %s", rawStr))
+		log.FromContext(ctx).WithValues("vmPatch", rawStr).V(1).Info("applying patch to Azure VM")
```
tallaxes: This really feels like it should include vm name in the log message - unless patch includes it already? (Maybe not for this PR ...)

pkg/providers/instance/resperrorhandlers.go
```diff
@@ -97,14 +97,20 @@ func defaultResponseErrorHandlers() []responseErrorHandler {
 func handleLowPriorityQuotaError(ctx context.Context, provider *DefaultProvider, instanceType *corecloudprovider.InstanceType, zone, capacityType string, err error) error {
 	// Mark in cache that spot quota has been reached for this subscription
 	provider.unavailableOfferings.MarkSpotUnavailableWithTTL(ctx, SubscriptionQuotaReachedTTL)
-	log.FromContext(ctx).Error(err, "")
+	log.FromContext(ctx).WithValues(
+		"instanceType", instanceType.Name,
+		"capacityType", capacityType,
+	).Error(err, "low priority quota reached")
```
tallaxes: Great to add values for context 👍 Not totally sure about adding a message (here and elsewhere), I suspect the reason it was omitted is that the error itself basically contains it, but maybe does not hurt, given that we don't quite have control over err, but have control over our message. Would be interested in opinion of others.

pkg/providers/instancetype/instancetypes.go
```diff
@@ -131,12 +131,12 @@ func (p *DefaultProvider) List(
 	for _, sku := range skus {
 		vmsize, err := sku.GetVMSize()
 		if err != nil {
-			log.FromContext(ctx).Error(err, fmt.Sprintf("parsing VM size %s", *sku.Size))
+			log.FromContext(ctx).WithValues("skuSize", *sku.Size).Error(err, "parsing VM size")
```
tallaxes: vmSize _might_ be better here (and elsewhere) ...

pkg/providers/instancetype/instancetypes.go
```diff
@@ -269,11 +269,11 @@ func (p *DefaultProvider) getInstanceTypes(ctx context.Context) (map[string]*ske
 	}
 
 	skus := cache.List(ctx, skewer.IncludesFilter(GetKarpenterWorkingSKUs()))
-	log.FromContext(ctx).V(1).Info(fmt.Sprintf("Discovered %d SKUs", len(skus)))
+	log.FromContext(ctx).WithValues("skuCount", len(skus)).V(1).Info("discovered SKUs")
```
tallaxes: Hmm, Yes, I think I prefer "skuCount" over more generic "count"; again wondering if there are best practices for this ...

pkg/providers/pricing/pricing.go
```diff
@@ -215,7 +224,10 @@ func (p *Provider) UpdateOnDemandPricing(ctx context.Context, onDemandPrices map
 	p.onDemandPrices = lo.Assign(onDemandPrices)
 	p.onDemandUpdateTime = time.Now()
 	if p.cm.HasChanged("on-demand-prices", p.onDemandPrices) {
-		log.FromContext(ctx).WithValues("instance-type-count", len(p.onDemandPrices)).Info(fmt.Sprintf("updated on-demand pricing for region %s", p.region))
+		log.FromContext(ctx).WithValues(
+			"region", p.region,
+			"instanceTypeCount", len(p.onDemandPrices),
```
tallaxes, line 225: I guess we are going with camelCasing for keys everywhere? Don't have a strong opinon, but let's make sure this is a concious choice.

test/pkg/environment/common/expectations.go
```diff
@@ -94,7 +94,7 @@ func (env *Environment) ExpectUpdated(objects ...client.Object) {
 			current := o.DeepCopyObject().(client.Object)
 			g.Expect(env.Client.Get(env.Context, client.ObjectKeyFromObject(current), current)).To(Succeed())
 			if current.GetResourceVersion() != o.GetResourceVersion() {
-				log.FromContext(env).Info(fmt.Sprintf("detected an update to an object (%s) with an outdated resource version, did you get the latest version of the object before patching?", lo.Must(apiutil.GVKForObject(o, env.Client.Scheme()))))
+				log.FromContext(env).WithValues("objectType", lo.Must(apiutil.GVKForObject(o, env.Client.Scheme())).String()).Info("detected an update to an object with an outdated resource version, did you get the latest version of the object before patching?")
```
tallaxes: Maybe break down the line - on values? - has complex expression ...

test/pkg/environment/common/expectations.go
```diff
@@ -115,7 +115,7 @@ func (env *Environment) ExpectStatusUpdated(objects ...client.Object) {
...
```
tallaxes: Same, maybe break down the line?

tallaxes (approved):
Looks good (and approved), left some minor questions / suggestions.

pkg/cloudprovider/cloudprovider.go
alimaazamat: I did this because some resources have a name and an id. For example, `nicId` and `nicName` are two keys logged.

pkg/cloudprovider/cloudprovider.go
alimaazamat: This was a mistake! Fixed to `Error(err, "message")`

alimaazamat: Better practice to have error in `.Error(err, "message" )` so error is automatically included in structured log

pkg/cloudprovider/drift.go
alimaazamat: Heard! Keeping "WARN"

pkg/cloudprovider/drift.go
alimaazamat: For consistency with rest of codebase I changed this to WithValues

alimaazamat: Made this V(1) as well

pkg/controllers/nodeclaim/inplaceupdate/controller.go
alimaazamat: Added a TODO message for this... assuming its changing the function call to take in something like `vm *armcompute.VirtualMachine, ` to get the vm name

pkg/providers/pricing/pricing.go
alimaazamat: I'm just used to Go using camel case and I think the first file I started with used camel case. I do see that AWS uses-this-kind-of-syntax but then also has a few camel case keys as well.

pkg/cloudprovider/cloudprovider.go
tallaxes: @matthchr, thoughts?

tallaxes: @matthchr, thoughts?

pkg/cloudprovider/cloudprovider.go
tallaxes: Using .Error also "elevates" this from warning to error - which is actually appropriate in this case, I think, but in some other places we might keep using WithValues("error", err) maybe?

pkg/cloudprovider/drift.go
tallaxes: Probably fine, @charliedmcb for opinion

alimaazamat:
Tried squashing my commits but then it doesn't have the clean [Merge branch 'main' into refactor-logging](https://github.com/Azure/karpenter-provider-azure/pull/968/commits/da89ddb63dc667fcc03dd6a9e45aa5f53234dc99) in one commit

pkg/cloudprovider/drift.go
tallaxes: Don't love "WARN", maybe "warning" (or nothing) could be better. Does not seem to be standard pattern for warnings in this logging API, maybe Info + dropping one level (and no "warning" in the message)? Not critical, consistency is already good, and can revisit later, but also @matthchr for opinion

tallaxes:
> Tried squashing my commits but then it doesn't have the clean [Merge branch 'main' into refactor-logging](https://github.com/Azure/karpenter-provider-azure/pull/968/commits/da89ddb63dc667fcc03dd6a9e45aa5f53234dc99) in one commit

Will be squashed on merge to main

pkg/cloudprovider/cloudprovider.go
matthchr: IMO if we log an ID we shouldn't even bother logging a name, because the ID is a superset of the name? At least, assuming we're talking about Azure ARM IDs here.

tallaxes: In this case we don't log the ID, just the name, and the question is whether the key can be "vm" or "vmName". Minor, but trying to establish a pattern and consistency in key names.

pkg/cloudprovider/drift.go
charliedmcb: I think this is fine switching to `V(1)`. I'm assuming that's in case we're stuck in a case of the kubernetes version being not ready we'd be concerned about this being spammy at a higher level?

I do think this log is relatively important for visibility, since its handling the error, as we omit returning the error after logging here.

matthchr: edit: moved this comment elsewhere.

pkg/cloudprovider/cloudprovider.go
matthchr: Ahh and the point is we want consistency between when we log `vmId` and when we log `vmName`.

I don't have too much of an opinion as long as we're consistent - i.e. if we call it `vm` we should also call it `nic` (not `nicName`. Maybe a _slight_ preference towards `xName` because in _theory_ we could log the actual "whole JSON payload" which might be `vm` or `nic`? as opposed to just the name? In practice not sure that's ever going to happen though.

pkg/cloudprovider/drift.go
matthchr: I put the `WARN` because at the time I swapped to consistent logging stack we had like 2 `logger.Warning` which didn't exist in the unified interface and this was my quick and dirty way of maintaining that.

A better solution would be to examine the logs we have with a warn prefix and just rephrase them such that they work as a normal V(0) or V(1) (or whatever) log. Presumably V0 because they were "important".

pkg/cloudprovider/drift.go
tallaxes: That's a good point! @alimaazamat, if you agree, let's replace all WithValues with just passing the keys/values to Info call at the end.

pkg/cloudprovider/cloudprovider.go
```diff
@@ -190,13 +190,13 @@ func (c *CloudProvider) waitOnPromise(ctx context.Context, promise *instance.Vir
...
 		if cloudprovider.IgnoreNodeClaimNotFoundError(err) != nil {
-			log.FromContext(ctx).Error(err, fmt.Sprintf("failed to delete VM %s", vmName))
+			log.FromContext(ctx).WithValues("vmName", vmName).Error(err, "failed to delete VM")
 		}
 
 		if err = c.kubeClient.Delete(ctx, nodeClaim); err != nil {
 			err = client.IgnoreNotFound(err)
 			if err != nil {
-				log.FromContext(ctx).Error(err, "failed to delete nodeclaim %s, will wait for liveness TTL", nodeClaim.Name)
+				log.FromContext(ctx).WithValues("nodeClaimName", nodeClaim.Name).Error(err, "failed to delete nodeclaim, will wait for liveness TTL")
```
matthchr: I thought the point of `WithValues` was to return a logger that consistently logged those values.

It feels weird to me to use `WithValues` and then throw the result of the `WithValues` away after a single call, because `WithValues` allocates and just passing the value to `Info` in the `keysAndVals` section doesn't (AFAIK) allocate as much.

pkg/cloudprovider/cloudprovider.go
tallaxes: Ok, let's keep "*Name" suffix

pkg/cloudprovider/cloudprovider.go
tallaxes: That's a good point! @alimaazamat, if you agree, let's replace all WithValues with just passing the keys/values to Info call at the end. (Except where we actually capture WithValues logger - which I don't think we do anywhere in our provider yet ...)

matthchr: I think there were a few places where we use `WithValues` and `WithName` to stash a logger back in the ctx, like here in `CloudProvider.Delete`:
`ctx = log.IntoContext(ctx, log.FromContext(ctx).WithValues("nodeclaim", nodeClaim.Name))`

pkg/cloudprovider/drift.go
tallaxes: FWIW, it looks like with go-logr/zapr, the levels are essentially reversed, V(1) is debug (-1 in zapr), and the right way to emit warning (more important than Info) is to use V(-1) - which would be 1 in zapr.

matthchr: I didn't realize that the logr interface supported negative V levels, although in practice maybe it doesn't matter because as long as our default V level is >= 0 (which it is), V0 and V-1 would both be the same in that they're always logged? I guess it matters in log-gathering because they'll show up as different levels there?

pkg/cloudprovider/cloudprovider.go
alimaazamat: I agree thanks for pointing this out, this went over my head. Going to alter refactoring to use direct key value pairs for single log calls!

pkg/providers/imagefamily/resolver.go
```diff
@@ -115,7 +115,7 @@ func (r *defaultResolver) Resolve(
 		return nil, err
 	}
 
-	log.FromContext(ctx).Info(fmt.Sprintf("Resolved image %s for instance type %s", imageID, instanceType.Name))
+	log.FromContext(ctx).WithValues("imageId", imageID, "instanceType", instanceType.Name).Info("resolved image")
```
alimaazamat: Not sure if it would be overkill to do `instanceTypeName` here

pkg/cloudprovider/drift.go
alimaazamat: Okay I changed it to V(0)... just so we don't impact log-gathering. If we change our mind lmk!

alimaazamat: Update: ^We want the WARN label in Zapr so -1 is better

tallaxes: > I guess it matters in log-gathering because they'll show up as different levels there?

Yes, was just discussing this with @alimaazamat

pkg/cloudprovider/cloudprovider.go
```diff
@@ -259,7 +259,6 @@ func (c *CloudProvider) Get(ctx context.Context, providerID string) (*karpv1.Nod
 	if err != nil {
 		return nil, fmt.Errorf("getting vm name, %w", err)
 	}
-	ctx = log.IntoContext(ctx, log.FromContext(ctx).WithValues("id", vmName))
```
tallaxes, line 262: This one needs to stay, stashes new logger into context. Also need to update "id" with "vmName" for consistency

pkg/cloudprovider/cloudprovider.go
```diff
@@ -295,8 +294,6 @@ func (c *CloudProvider) GetInstanceTypes(ctx context.Context, nodePool *karpv1.N
 }
 
 func (c *CloudProvider) Delete(ctx context.Context, nodeClaim *karpv1.NodeClaim) error {
-	ctx = log.IntoContext(ctx, log.FromContext(ctx).WithValues("nodeclaim", nodeClaim.Name))
```
tallaxes, line 298: Needs to stay (and let's check "nodeclaim" is consistent)

pkg/controllers/nodeclaim/inplaceupdate/controller.go
```diff
@@ -184,6 +184,7 @@ func logVMPatch(ctx context.Context, update *armcompute.VirtualMachineUpdate) {
 			raw, _ := json.Marshal(update)
 			rawStr = string(raw)
 		}
-		log.FromContext(ctx).V(1).Info(fmt.Sprintf("applying patch to Azure VM: %s", rawStr))
+		// TODO: Include vm name in log message
```
tallaxes: Hmm, I know I called this out, but maybe it is already included in the logger earlier via WithValues?

pkg/cloudprovider/cloudprovider.go
alimaazamat: Sounds good, mistakingly thought that vmName wasn't useful here

pkg/providers/kubernetesversion/kubernetesversion.go
```diff
@@ -59,7 +59,7 @@ func (p *kubernetesVersionProvider) KubeServerVersion(ctx context.Context) (stri
 	version := strings.TrimPrefix(serverVersion.GitVersion, "v") // v1.24.9 -> 1.24.9
 	p.kubernetesVersionCache.SetDefault(kubernetesVersionCacheKey, version)
 	if p.cm.HasChanged("kubernetes-version", version) {
-		log.FromContext(ctx).WithValues("kubernetes-version", version).V(1).Info("discovered kubernetes version")
+		log.FromContext(ctx).V(1).Info("discovered kubernetes version", "kubernetes-version", version)
```
tallaxes: Let's keep logging camelCased?

test/pkg/environment/common/expectations.go
```diff
@@ -94,7 +94,9 @@ func (env *Environment) ExpectUpdated(objects ...client.Object) {
 			current := o.DeepCopyObject().(client.Object)
 			g.Expect(env.Client.Get(env.Context, client.ObjectKeyFromObject(current), current)).To(Succeed())
 			if current.GetResourceVersion() != o.GetResourceVersion() {
-				log.FromContext(env).Info(fmt.Sprintf("detected an update to an object (%s) with an outdated resource version, did you get the latest version of the object before patching?", lo.Must(apiutil.GVKForObject(o, env.Client.Scheme()))))
+				log.FromContext(env).WithValues(
```
tallaxes: WithValues -> keysAndVals arg

test/pkg/environment/common/expectations.go
```diff
@@ -115,7 +117,9 @@ func (env *Environment) ExpectStatusUpdated(objects ...client.Object) {
...
+					"objectType", lo.Must(apiutil.GVKForObject(o, env.Client.Scheme())).String(),
+				).Info("detected an update to an object with an outdated resource version, did you get the latest version of the object before patching?")
```
tallaxes: WithValues -> kkeysAndVals arg

pkg/controllers/nodeclaim/inplaceupdate/controller.go
alimaazamat: It's not included in logger earlier, but I can add it with context enrichment after we get vmName here: https://github.com/alimaazamat/karpenter-provider-azure/blob/refactor-logging/pkg/controllers/nodeclaim/inplaceupdate/controller.go#L94

pkg/providers/instance/instance.go
```diff
@@ -161,22 +161,22 @@ func (p *DefaultProvider) BeginCreate(
 		// There may be orphan NICs (created before promise started)
 		// This err block is hit only for sync failures. Async (VM provisioning) failures will be returned by the vmPromise.Wait() function
 		if cleanupErr := p.cleanupAzureResources(ctx, GenerateResourceName(nodeClaim.Name)); cleanupErr != nil {
-			log.FromContext(ctx).Error(cleanupErr, fmt.Sprintf("failed to cleanup resources for node claim %s", nodeClaim.Name))
+			log.FromContext(ctx).Error(cleanupErr, "failed to cleanup resources for node claim", "nodeClaimName", nodeClaim.Name)
```
tallaxes: Ok, hopefully the last one: For node claims (here and elsewhere), let's use "nodeClaim" key rather than "nodeClaimName". Rationale: *Name is justified for Azure resources (like VMs), which often have name and ID, but not for (cluster-scoped) Kubernetes resources.

pkg/controllers/nodeclaim/garbagecollection/instance_garbagecollection.go
```diff
@@ -107,7 +107,7 @@ func (c *VirtualMachine) garbageCollect(ctx context.Context, nodeClaim *karpv1.N
 		if err := c.kubeClient.Delete(ctx, &node); err != nil {
 			return client.IgnoreNotFound(err)
 		}
-		log.FromContext(ctx).WithValues("node", node.Name).V(1).Info("garbage collected node")
+		log.FromContext(ctx).V(1).Info("garbage collected node", "nodeName", node.Name)
```
tallaxes: Node is also a kubernetes resource, so let's just use "node". Edit: Actually, checking (and aligning with) logging in Karpenter core (and AWS provider), let's use capitalized keys for Kubernetes resources, so "Node", "NodeClaim", "NodePool". Apologies for not catching this earlier / multiple iterations.

pkg/controllers/nodeclass/status/kubernetesversion.go
```diff
@@ -93,10 +93,10 @@ func (r *KubernetesVersionReconciler) Reconcile(ctx context.Context, nodeClass *
 		}
 		// Handles case 2: Upgrade kubernetes version [Note: we set node image to not ready, since we upgrade node image when there is a kubernetes upgrade]
 		if newK8sVersion.GT(currentK8sVersion) {
-			logger.V(1).Info(fmt.Sprintf("kubernetes upgrade detected: from %s (current), to %s (discovered)", currentK8sVersion.String(), newK8sVersion.String()))
+			logger.V(1).Info("kubernetes upgrade detected", "currentK8sVersion", currentK8sVersion.String(), "discoveredK8sVersion", newK8sVersion.String())
```
tallaxes: Let's expand K8s in keys (already expanded in several places)

pkg/providers/instance/instance.go
```diff
@@ -758,15 +767,15 @@ func (p *DefaultProvider) pickSkuSizePriorityAndZone(
 func (p *DefaultProvider) cleanupAzureResources(ctx context.Context, resourceName string) error {
 	vmErr := deleteVirtualMachineIfExists(ctx, p.azClient.virtualMachinesClient, p.resourceGroup, resourceName)
 	if vmErr != nil {
-		log.FromContext(ctx).Error(vmErr, fmt.Sprintf("virtualMachine.Delete for %s failed", resourceName))
+		log.FromContext(ctx).Error(vmErr, "virtualMachine.Delete failed", "resourceName", resourceName)
```
tallaxes: This probably should be vmName?

pkg/providers/instance/instance.go
```diff
@@ -758,15 +767,15 @@ func (p *DefaultProvider) pickSkuSizePriorityAndZone(
...
 	}
 	// The order here is intentional, if the VM was created successfully, then we attempt to delete the vm, the
 	// nic, disk and all associated resources will be removed. If the VM was not created successfully and a nic was found,
 	// then we attempt to delete the nic.
 
 	nicErr := deleteNicIfExists(ctx, p.azClient.networkInterfacesClient, p.resourceGroup, resourceName)
 	if nicErr != nil {
-		log.FromContext(ctx).Error(nicErr, fmt.Sprintf("networkinterface.Delete for %s failed", resourceName))
+		log.FromContext(ctx).Error(nicErr, "networkinterface.Delete failed", "resourceName", resourceName)
```
tallaxes: This probably should be nicName?

pkg/providers/instance/instance.go
alimaazamat: Yes agreed and I also changed resourceName to vmName here https://github.com/alimaazamat/karpenter-provider-azure/blob/refactor-logging/pkg/providers/instance/instance.go#L234 because the Get method signature says vmName too.

alimaazamat:
**Updated List of Keys**
====================================

WithValues (Context Enrichment) Keys:
1. region
2. vmName
3. NodeClaim
4. existingKubernetesVersion
5. newKubernetesVersion
6. providerID

Direct Key-Value Keys:
1. actualHash
2. actualImageID
3. actualKubeletIdentityClientID
4. actualKubernetesVersion
5. capacity-type
6. currentKubernetesVersion
7. discoveredKubernetesVersion
8. driftType
9. error
10. existingImages
11. expectedKubeletIdentityClientID
12. expectedKubernetesVersion
13. extensionID
14. extensionName
15. goalHash
16. goalKubernetesVersion
17. hostname
18. imageID
19. instance-type
20. instanceTypeCount
21. ipv4PoolCount
22. ipv4PoolIDs
23. kubernetesVersion
24. lastOnDemandUpdateTime
25. lastSpotUpdateTime
26. launchedInstance
27. length
28. loadBalancerCount
29. maintenanceWindowData
30. newImages
31. nicID
32. nicName
33. Node
34. NodeClaim
35. nodeClassHash
36. nodeClaimHash
37. nsgCount
38. objectType
39. options
40. providerID
41. resourceGroup
42. skuCount
43. ttl
44. type
45. unavailable
46. vmName
47. vmPatch
48. vmSize
49. zone

**Keys Subset that I believe are Stable/Common/Basic**
====================================
Karpenter/Kubernetes:
1. Node
2. NodeClaim
3. providerID
4. driftType
5. instance-type
6. capacity-type

Azure:
1. vmName
2. resourceGroup
3. region
4. nicName
6. zone

docs/logging-guidelines.md
```diff
@@ -0,0 +1,209 @@
...
+{"level":"ERROR","time":"2024-01-01T12:00:10Z","message":"failed to cleanup VM, may cause resource leak","error":"timeout waiting for deletion"}
+```
+
+### Warning Level (`V(-1)`)
+For non-fatal issues that operators should be aware of
+
+**When to use V(-1):**
+- Recoverable errors that don't prevent operation but indicate potential issues
+- Invariant violations that suggest system inconsistencies
+- Fallback scenarios where default behavior is used
```
matthchr: Is this "fallback scenarios..." really worth a warning? Seems like it's pretty expected?

matthchr: Same question I think applies to a bunch of the below warning options, and actually even to the first one:
"Recoverable errors that don't prevent operation but indicate potential issues" -- the .Error() (at V0 -- though maybe it doesn't matter?) seems sufficient for error cases because even in error cases the error is probably retryable?

tl;dr: Some more clarity between when .Error and when .V(-1) with maybe guidance to _generally_ avoid `V(-1)` unless it's really warranted?

docs/logging-guidelines.md
```diff
@@ -0,0 +1,209 @@
...
+- **Drift Detection**: `driftType`, `nodeClassHash`, `nodeClaimHash`, `actualImageID`
+- **Operations**: `capacityType`, `instanceType`, `goalHash`, `actualHash`
+- **Counts**: `instanceTypeCount`, `skuCount`, `loadBalancerCount`
+- **Identity**: `expectedKubeletIdentityClientID`, `actualKubeletIdentityClientID`
+
+
+
+## Summary
+
+| Level    | Purpose                              | Examples                                 | Zapr Output Level |
```
matthchr: minor/nit: Prettify the markdown table by lining things up? I think vscode has a format-helper for this.

docs/logging-guidelines.md
alimaazamat: I said fallback scenarios cause of this line here: https://github.com/alimaazamat/karpenter-provider-azure/blob/refactor-logging/pkg/utils/utils.go#L62 
Should I leave this or maybe change this to V(1)?

alimaazamat: In my opinion the distinction is that for an .Error() the operation/function failed, while for .V(-1) the operation/function will continue/suceed with unexpected behavior. 
The retryability doesn't matter: for example an Error could be retryable later, but for that specific attempt in the scope of the operation/function it fails.
I updated docs to reflect this but let me know if more needs to be said!

tallaxes: Agree that it is hard to pin down guidance for -1 / Warning, and that it probably needs to be generally avoided; this is just trying to faithfully translate what was originally logged with ".Warn". (Also agree that "fallback scenarios" is probably too generic, feels like it might apply in some cases and not apply in others.) One difference from Error() may be that Error() does not depend on level, is always issued. I'd also be Ok with just dropping -1 for now.

Have not yet looked at the rest of guidance; suspect it might take some more iterations to pin down, maybe we can handle in a separate PR?

(Was wondering if this guidance
https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/logging.md#what-method-to-use could be useful - though levels are different ...)

alimaazamat: Okay I dropped V(-1) to .Info for all instances of V(-1).
I will continue this conversation on seperate PR for guidelines documentation here: https://github.com/Azure/karpenter-provider-azure/pull/1001

tallaxes (approved):

pkg/cloudprovider/cloudprovider.go
```diff
@@ -190,13 +190,13 @@ func (c *CloudProvider) waitOnPromise(ctx context.Context, promise *instance.Vir
...
 		if cloudprovider.IgnoreNodeClaimNotFoundError(err) != nil {
-			log.FromContext(ctx).Error(err, fmt.Sprintf("failed to delete VM %s", vmName))
+			log.FromContext(ctx).Error(err, "failed to delete VM", "vmName", vmName)
 		}
 
 		if err = c.kubeClient.Delete(ctx, nodeClaim); err != nil {
 			err = client.IgnoreNotFound(err)
 			if err != nil {
-				log.FromContext(ctx).Error(err, "failed to delete nodeclaim %s, will wait for liveness TTL", nodeClaim.Name)
+				log.FromContext(ctx).Error(err, "failed to delete nodeclaim, will wait for liveness TTL", "NodeClaim", nodeClaim.Name)
```
matthchr, line 199: minor: should have case consistency. `nodeClaim`?

matthchr: Or rather, why are some things `vmName` and some `NodeClaim` -- different casing seems weird?

pkg/controllers/nodeclaim/garbagecollection/instance_garbagecollection.go
```diff
@@ -107,7 +107,7 @@ func (c *VirtualMachine) garbageCollect(ctx context.Context, nodeClaim *karpv1.N
 		if err := c.kubeClient.Delete(ctx, &node); err != nil {
 			return client.IgnoreNotFound(err)
 		}
-		log.FromContext(ctx).WithValues("node", node.Name).V(1).Info("garbage collected node")
+		log.FromContext(ctx).V(1).Info("garbage collected node", "Node", node.Name)
```
matthchr, line 110: another casing weirdness - `node`?

pkg/providers/imagefamily/resolver.go
```diff
@@ -115,7 +115,7 @@ func (r *defaultResolver) Resolve(
 		return nil, err
 	}
 
-	log.FromContext(ctx).Info(fmt.Sprintf("Resolved image %s for instance type %s", imageID, instanceType.Name))
+	log.FromContext(ctx).Info("resolved image", "imageID", imageID, "instance-type", instanceType.Name)
```
matthchr, line 118: `instanceType`, rather than `instance-type`?

pkg/providers/instance/instance.go
```diff
@@ -161,22 +161,22 @@ func (p *DefaultProvider) BeginCreate(
...
 
-	log.FromContext(ctx).WithValues(
-		"launched-instance", *vm.ID,
+	log.FromContext(ctx).Info("launched new instance",
+		"launchedInstance", *vm.ID,
 		"hostname", *vm.Name,
 		"type", string(*vm.Properties.HardwareProfile.VMSize),
 		"zone", zone,
-		"capacity-type", GetCapacityType(vm)).Info("launched new instance")
+		"capacity-type", GetCapacityType(vm))
```
matthchr, line 179: `capacityType`?

matthchr (approved):
Approved but I think we should be consistent for lowerCamel versus dash-case versus UpperCamel, which we don't seem to be (we've got some using all 3).

pkg/cloudprovider/cloudprovider.go
alimaazamat: After discussing with Alex, Kubernetes resources will have uppercase CamelCase. Things like Node, NodeClaim, NodePool, Pod..

pkg/providers/imagefamily/resolver.go
alimaazamat: After discussing with Alex we decided to keep kebab-case for Kubernetes labels like instance-type and capacity-type to match Karpenter Core

alimaazamat:
View matching documentation here: https://github.com/Azure/karpenter-provider-azure/pull/1001/files?short_path=ab8e388#diff-ab8e388409d486d398b131afcdea48cd2892bf8787631e14848e44871283d07c

pkg/cloudprovider/cloudprovider.go
tallaxes: Trying to align with core. But the rationale is that Kubernetes resources already have a naming convention to align to

pkg/controllers/nodeclaim/garbagecollection/instance_garbagecollection.go
tallaxes: Same, Kubernetes resource

pkg/providers/imagefamily/resolver.go
tallaxes: Same, aligning with core. The rationale is likely that it is coming from well-known labels, which already have a naming convention.

pkg/cloudprovider/cloudprovider.go
tallaxes: Discussed offline, we are on the same page, not great to have a mix but is worth matching core; resolving

PR #969, opened 2025-06-13
chore(imagefamily): refactor switch nodeimage provider to use interface

charliedmcb:
Fixes # 

**Description**
Part of a series of refactors of the imagefamily package cleaning up the area after the logic changes to provider and resolver. 

This change aims to unify everything in the same file, and switch to using the interface outside the package.

This also introduces a new handling for interfaces in prod vs testing, when we need access to the Reset method in testing on the Provider, which we don't want to expose in the prod interface, and allows us to keep things encapulated.

- `make presubmit`
- E2E Runs:
    - https://github.com/Azure/karpenter-provider-azure/actions/runs/15644292328
    - https://github.com/Azure/karpenter-provider-azure/actions/runs/15644389335
    - https://github.com/Azure/karpenter-provider-azure/actions/runs/15644616767
    - https://github.com/Azure/karpenter-provider-azure/actions/runs/15645155687 (latest) 

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
NONE
```


pkg/cloudprovider/cloudprovider.go
```diff
@@ -73,12 +73,12 @@ type CloudProvider struct {
 	instanceTypeProvider instancetype.Provider
 	instanceProvider     instance.Provider
 	kubeClient           client.Client
-	imageProvider        *imagefamily.Provider
+	imageProvider        imagefamily.NodeImageProvider
 	recorder             events.Recorder
 }
 
 func New(instanceTypeProvider instancetype.Provider, instanceProvider instance.Provider, recorder events.Recorder,
```
matthchr: minor: 1 arg per line

pkg/providers/imagefamily/nodeimage.go
```diff
@@ -40,8 +50,30 @@ type NodeImageProvider interface {
...
+	cm              *pretty.ChangeMonitor
+}
+
+func NewProvider(versionsClient types.CommunityGalleryImageVersionsAPI, location, subscription string, nodeImageVersionsClient types.NodeImageVersionsAPI) *provider {
+	return &provider{
+		subscription:        subscription,
+		location:            location,
+		imageVersionsClient: versionsClient,
+		nodeImageVersions:   nodeImageVersionsClient,
+		nodeImagesCache:     cache.New(imageExpirationInterval, imageCacheCleaningInterval),
```
matthchr: Consider passing the cache as an argument (could be an optional argument where you new it up here if it's not provided)

pkg/test/environment.go
```diff
@@ -76,7 +76,7 @@ type Environment struct {
 	InstanceProvider             instance.Provider
 	PricingProvider              *pricing.Provider
 	KubernetesVersionProvider    kubernetesversion.KubernetesVersionProvider
```
matthchr, line 79: Note that other caches are above (LoadBalancerCache, etc)

pkg/test/testinterfaces.go
```diff
@@ -0,0 +1,32 @@
...
+
+import (
+	"context"
+
+	"github.com/Azure/karpenter-provider-azure/pkg/apis/v1beta1"
+	"github.com/Azure/karpenter-provider-azure/pkg/providers/imagefamily"
+)
+
+type TestNodeImageProvider interface {
+	List(ctx context.Context, nodeClass *v1beta1.AKSNodeClass) ([]imagefamily.NodeImage, error)
```
matthchr: You can just embed the public `NodeImageProvider` here.

pkg/providers/imagefamily/nodeimage.go
charliedmcb: Updated. Not optional for now.

pkg/test/environment.go
charliedmcb: Yea, I leaned away from that patterning, but we discussed offline, and decided that IoC seems solid for better testing options.

pkg/test/testinterfaces.go
charliedmcb: Dropped this file, so no longer relevant

pkg/test/environment.go
```diff
@@ -114,14 +115,15 @@ func NewRegionalEnvironment(ctx context.Context, env *coretest.Environment, regi
 	azureResourceGraphAPI := fake.NewAzureResourceGraphAPI(resourceGroup, virtualMachinesAPI, networkInterfacesAPI)
 	// Cache
 	kubernetesVersionCache := cache.New(azurecache.KubernetesVersionTTL, azurecache.DefaultCleanupInterval)
+	nodeImagesCache := cache.New(imagefamily.ImageExpirationInterval, imagefamily.ImageCacheCleaningInterval)
```
matthchr, line 118: Totally happy with this, but if you don't like having to use the cache consts here, you could have the `imagefamily` pkg offer a `NewDefaultCache` function. Just throwing ideas around really.

matthchr (approved):

pkg/test/environment.go
charliedmcb: I was considering that, but didn't want to bundle too much into this PR, and matches the other styling for now. If we decide to adjust cache handling with more defensive patterning, and reuse/helpers that I'd look at this potential.

PR #970, opened 2025-06-13
fix: issue where container can come up but be unable to access token

### LINKED ISSUE #964: NAP CreateInstanceFailed on nodeclaims
### Version

```
aks-managed-karpenter-overlay     	kube-system	1421    	2025-06-13 05:36:18.076307149 +0000 UTC	deployed	karpenter-overlay-addon-0.1.0-67ffb6c0ffda7bf054bdff6fbbd34e5c83f1dc81     	           
aks-managed-karpenter-overlay-base	kube-system	1422    	2025-06-13 05:42:09.081121036 +0000 UTC	deployed	karpenter-overlay-base-addon-0.1.0-9aa25dba81eee53fafe1b096c2d24821eda83616	           
```
__Kubernetes Version:__ v1.31.8

### Expected Behavior

Nodes would provision.

### Actual Behavior

Seeing the following in my nodeclaims:
```  - lastTransitionTime: "2025-06-13T05:34:52Z"
    message: getting backend pools: failed to get next loadbalancer page: DefaultAzureCredential:
      failed to acquire a token.\nAttempted credentials:\n\tEnvironmentCredential:
      missing environment variable AZURE_TENANT_ID\n\tWorkloadIdentityCredential:
      no client ID specified. Check pod configuration or set ClientID in the options\n\tManagedIdentityCredential:
      failed to authenticate a system assigned identity. The endpoint responded with
      {\"error\":\"invalid_request\",\"error_description\":\"Multiple user assigned
      identities exist, please specify the clientId / resourceId of the identity in
      the token request\"}\n\tAzureCLICredential: fork/exec /bin/sh: no such file
      or directory\n\tAzureDeveloperCLICredential: fork/exec /bin/sh: no such file
      or directory
    observedGeneration: 1
    reason: CreateInstanceFailed
    status: Unknown
    type: Launched
```

Enabled NAP for an existing AKS cluster, it was working but started seeing this issue.

Seeing this on multiple node pools.

### Resource Specs and Logs

node pool definition:
```
apiVersion: karpenter.sh/v1
kind: NodePool
metadata:
  name: gpu-batch
spec:
  weight: 10
  disruption:
    consolidationPolicy: WhenEmpty
    consolidateAfter: 1m
  template:
    metadata:
      labels:
        worlds.io/batch: "true"
    spec:
      nodeClassRef:
        group: karpenter.azure.com
        kind: AKSNodeClass
        name: default
      taints:
        - key: nvidia.com/gpu
          value: "true"
          effect: NoSchedule
      requirements:
        - key: karpenter.azure.com/sku-name
          operator: In
          values:
            - Standard_NC4as_T4_v3
            - Standard_NC8as_T4_v3
        # - key: karpenter.azure.com/sku-family
        #   operator: In
        #   values:
        #     - N
        # - key: karpenter.azure.com/sku-gpu-name
        #   operator: In
        #   values:
        #     - t4
        # - key: karpenter.azure.com/sku-version
        #   operator: In
        #   values:
        #     - "3"
        # - key: karpenter.azure.com/sku-storage-premium-capable
        #   operator: In
        #   values:
        #     - "true"
        - key: kubernetes.io/arch
          operator: In
          values:
            - amd64
        - key: kubernetes.io/os
          operator: In
          values:
            - linux
          operator: In
          values:
            - on-demand
      startupTaints:
        - effect: NoExecute
          key: node.cilium.io/agent-not-ready
          value: "true"
```

### Community Note

matthchr:
When the process launches, if it doesn't check that it can retrieve a token, then if it's unable to retrieve a token later it will continue to retry getting a token, but won't ever exit the process.

This is a problem especially in the managed NAP case, because there's an IMDS proxy that needs to be set up before the NAP pod can get a token, but if the Karpenter pod starts too quickly it will open a connection in the transport connection pool to the wrong address (not the proxy), which will never serve it the correct token. This will mostly self-recover if request load is low, because the transport idle timeout will be hit and a new connection will go to the proxy, but if request load is high enough it can keep the connection open forever.

Better to lose the race gracefully and just exit the pod if we can't get a token.

Fixes #964

Tested: https://github.com/Azure/karpenter-provider-azure/actions/runs/15645616204 -- few flakes but the pods launched which is the main thing

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
Fix bug impacting managed Karpenter (aka NAP) which could cause the pod to very occasionally fail to authenticate with Azure for long periods of time
```

### DIFF
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -22,6 +22,8 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
 	"github.com/go-logr/logr"
 	"github.com/patrickmn/go-cache"
 	"github.com/samber/lo"
@@ -35,6 +37,7 @@ import (
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/transport"
 	"k8s.io/client-go/util/flowcontrol"
+	"sigs.k8s.io/cloud-provider-azure/pkg/azclient"
 	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 	karpv1 "sigs.k8s.io/karpenter/pkg/apis/v1"
 	"sigs.k8s.io/karpenter/pkg/operator"
@@ -90,10 +93,16 @@ func NewOperator(ctx context.Context, operator *operator.Operator) (context.Cont
 	azConfig, err := GetAZConfig()
 	lo.Must0(err, "creating Azure config") // NOTE: we prefer this over the cleaner azConfig := lo.Must(GetAzConfig()), as when initializing the client there are helpful error messages in initializing clients and the azure config
 
-	azClient, err := instance.CreateAZClient(ctx, azConfig)
+	cred, err := getCredential()
+	lo.Must0(err, "getting Azure credential")
+
+	// Get a token to ensure we can
+	lo.Must0(ensureToken(cred, azConfig), "ensuring Azure token can be retrieved")
+
+	azClient, err := instance.CreateAZClient(ctx, azConfig, cred)
 	lo.Must0(err, "creating Azure client")
 	if options.FromContext(ctx).VnetGUID == "" && options.FromContext(ctx).NetworkPluginMode == consts.NetworkPluginModeOverlay {
-		vnetGUID, err := getVnetGUID(azConfig, options.FromContext(ctx).SubnetID)
+		vnetGUID, err := getVnetGUID(cred, azConfig, options.FromContext(ctx).SubnetID)
 		lo.Must0(err, "getting VNET GUID")
 		options.FromContext(ctx).VnetGUID = vnetGUID
 	}
@@ -210,11 +219,7 @@ func getCABundle(restConfig *rest.Config) (*string, error) {
 	return lo.ToPtr(base64.StdEncoding.EncodeToString(transportConfig.TLS.CAData)), nil
 }
 
-func getVnetGUID(cfg *auth.Config, subnetID string) (string, error) {
-	creds, err := azidentity.NewDefaultAzureCredential(nil)
-	if err != nil {
-		return "", err
-	}
+func getVnetGUID(creds azcore.TokenCredential, cfg *auth.Config, subnetID string) (string, error) {
 	opts := armopts.DefaultArmOpts()
 	vnetClient, err := armnetwork.NewVirtualNetworksClient(cfg.SubscriptionID, creds, opts)
 	if err != nil {
@@ -281,4 +286,33 @@ func WaitForCRDs(ctx context.Context, timeout time.Duration, config *rest.Config
 
 	log.Info("all required CRDs are available")
 	return nil
+}
+
+// ensureToken ensures we can get a token for the Azure environment. Note that this doesn't actually
+// use the token for anything, it just checks that we can get one.
+func ensureToken(cred azcore.TokenCredential, cfg *auth.Config) error {
+	cloudEnv := azclient.EnvironmentFromName(cfg.Cloud)
+
+	// Short timeout to avoid hanging forever if something bad happens
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+
+	_, err := cred.GetToken(ctx, policy.TokenRequestOptions{
+		Scopes: []string{cloudEnv.ServiceManagementEndpoint + "/.default"},
+	})
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func getCredential() (azcore.TokenCredential, error) {
+	// TODO: Don't use NewDefaultAzureCredential
+	cred, err := azidentity.NewDefaultAzureCredential(nil)
+	if err != nil {
+		return nil, err
+	}
+
+	return auth.NewTokenWrapper(cred), nil
 }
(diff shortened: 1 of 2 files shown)


pkg/operator/operator.go
```diff
@@ -90,6 +92,9 @@ func NewOperator(ctx context.Context, operator *operator.Operator) (context.Cont
 	azConfig, err := GetAZConfig()
 	lo.Must0(err, "creating Azure config") // NOTE: we prefer this over the cleaner azConfig := lo.Must(GetAzConfig()), as when initializing the client there are helpful error messages in initializing clients and the azure config
 
+	// Get a token to ensure we can
+	lo.Must0(ensureToken(azConfig), "ensuring Azure token can be retrieved")
```
matthchr: I was torn if we should actually use the token here, or just get one. I went with just get one as it seemed slightly more targeted and also easier to truly assert than "the token is usable" -- because the token being usable gets at permissions and so would realistically require more than a single check to assert all of the permissions that we need.

With that said, the downside here is that we can retrieve the token but can't use it at all... which isn't _quite_ what we were doing before with the vnetGUID stuff, where we actually used the token to do something.

pkg/operator/operator.go
```diff
@@ -282,3 +287,21 @@ func WaitForCRDs(ctx context.Context, timeout time.Duration, config *rest.Config
...
+
+// ensureToken ensures we can get a token for the Azure environment. Note that this doesn't actually
+// use the token for anything, it just checks that we can get one.
+func ensureToken(cfg *auth.Config) error {
+	creds, err := azidentity.NewDefaultAzureCredential(nil)
+	if err != nil {
+		return err
+	}
+	cloudEnv := azclient.EnvironmentFromName(cfg.Cloud)
+	_, err = creds.GetToken(context.Background(), policy.TokenRequestOptions{
```
bcho: recommend to set a context with short timeout as the GetToken call could block for unknown reason

matthchr: Good call

pkg/operator/operator.go
```diff
@@ -282,3 +287,21 @@ func WaitForCRDs(ctx context.Context, timeout time.Duration, config *rest.Config
 	log.Info("all required CRDs are available")
 	return nil
 }
+
+// ensureToken ensures we can get a token for the Azure environment. Note that this doesn't actually
+// use the token for anything, it just checks that we can get one.
+func ensureToken(cfg *auth.Config) error {
+	creds, err := azidentity.NewDefaultAzureCredential(nil)
```
bcho: @comtalyst maybe we should have a unified usage of constructing credentials to use here. I believe in managed mode we should only use ManagedIdentityCredential . The worst part of `NewDefaultAzureCredential` is it could block for CLI input when it fallbacks to AzCLI login (if the environment has shell and azcli installed)

pkg/operator/operator.go
matthchr: So I could use the token to list LBs at the NRG level for example and just throw the response away

pkg/operator/operator.go
matthchr: Yeah we have a backlog item to fix this and so I didn't take a crack at doing anything better here. Agree with what you're saying though that we should have our own `GetCredential` that makes a ChainedCredential w/ the options we support.

pkg/operator/operator.go
tallaxes: Just getting one seems good enough / a good compromise to me, given the kind of failures we are fixing here

bcho: getting the token once is fine as there are multiple cache layers in IMDS / managed environment. It would be great if we could reuse the azidentity instance created in this ensure call as azidentity has its own cache layer too.

matthchr: There is no `azidentity` instance? `azidentity` is the package - do you mean reuse the actual credentials?

That's... fair.

bcho: yep I mean the azidentity credentials instance

matthchr: k I fixed this.

pkg/operator/operator.go
matthchr: Ok I ended up sorta needing to almost fix this to fix @bcho's other comment. There's now a single method in operator that gets the token. It still uses default credential but we could swap it to something else very easily in a follow up PR.

bcho (approved):
LGTM

pkg/operator/operator.go
```diff
@@ -90,10 +93,16 @@ func NewOperator(ctx context.Context, operator *operator.Operator) (context.Cont
 	azConfig, err := GetAZConfig()
 	lo.Must0(err, "creating Azure config") // NOTE: we prefer this over the cleaner azConfig := lo.Must(GetAzConfig()), as when initializing the client there are helpful error messages in initializing clients and the azure config
 
-	azClient, err := instance.CreateAZClient(ctx, azConfig)
+	cred, err := getCredential()
+	lo.Must0(err, "getting Azure credential")
```
bcho, line 97: nit(non-blocking): lo has `Must1` for this too:

```
cred := lo.Must1(getCredential(), "getting Azure credential)
```

matthchr: We purposefully don't use it above, so I followed that pattern here -- I'm actually not 100% sure I understand the comment on the one above so happy to swap to `Must1`/`Must` if @tallaxes prefers here.

tallaxes: The comment in its current form is confusing, but the original intent was just to have the extra message on failure

tallaxes (approved):

PR #991, opened 2025-06-18
fix: auxiliary token refresh and use sig in acceptance tests

rakechill:
**Description**
* Follow-up from #540 and #910 
* AuxialiaryTokenPolicy will now call a function before every call made by the VM Client
* This function call aux token server if the token is expired or will expire soon. Otherwise, it will use a cached token

* acceptance tests (WIP)
* further e2e testing will be done w/ a test image used by managed Karpenter

- [ ] Yes, issue opened: # 
- [X] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/auth/auxiliarytoken.go
+++ b/pkg/auth/auxiliarytoken.go
@@ -17,7 +17,6 @@ limitations under the License.
 package auth
 
 import (
-	"context"
 	"encoding/json"
 	"fmt"
 	"net/http"
@@ -70,7 +69,7 @@ func (p *AuxiliaryTokenPolicy) Do(req *policy.Request) (*http.Response, error) {
 	return req.Next()
 }
 
-func NewAuxiliaryTokenPolicy(ctx context.Context, client AuxiliaryTokenServer, url string, scope string) (*AuxiliaryTokenPolicy, error) {
+func NewAuxiliaryTokenPolicy(client AuxiliaryTokenServer, url string, scope string) *AuxiliaryTokenPolicy {
 	auxPolicy := AuxiliaryTokenPolicy{
 		Token:  azcore.AccessToken{},
 		url:    url,
@@ -78,8 +77,7 @@ func NewAuxiliaryTokenPolicy(ctx context.Context, client AuxiliaryTokenServer, u
 		client: client,
 		lock:   sync.Mutex{},
 	}
-	log.FromContext(ctx).V(1).Info("Will use auxiliary token policy for creating virtual machines")
-	return &auxPolicy, nil
+	return &auxPolicy
 }
 
 func getAuxiliaryToken(client AuxiliaryTokenServer, url string, scope string) (azcore.AccessToken, error) {
--- a/pkg/auth/auxiliarytoken.go
+++ b/pkg/auth/auxiliarytoken.go
@@ -38,47 +38,45 @@ var _ policy.Policy = &AuxiliaryTokenPolicy{}
 // AuxiliaryTokenPolicy provides a custom policy used to authenticate
 // with shared node image galleries.
 type AuxiliaryTokenPolicy struct {
-	auxToken AuxiliaryToken
-	url      string
-	scope    string
-	client   AuxiliaryTokenServer
+	Token  azcore.AccessToken
+	url    string
+	scope  string
+	client AuxiliaryTokenServer
+	lock   sync.Mutex
 }
 
-type AuxiliaryToken struct {
-	token azcore.AccessToken
-	lock  sync.Mutex
-}
-
-func (p *AuxiliaryTokenPolicy) GetAuxiliaryToken() {
-	p.auxToken.lock.Lock()
-	defer p.auxToken.lock.Unlock()
-	// If the token is expired or close to expiration, fetch a new one
-	if p.auxToken.token.ExpiresOn.IsZero() || p.auxToken.token.RefreshOn.Before(time.Now()) || p.auxToken.token.ExpiresOn.Before(time.Now().Add(5*time.Minute)) {
+func (p *AuxiliaryTokenPolicy) GetAuxiliaryToken() error {
+	p.lock.Lock()
+	defer p.lock.Unlock()
+	// If the token is uninitialized or close to expiration, fetch a new one
+	currentTime := time.Now()
+	if p.Token.ExpiresOn.IsZero() || p.Token.RefreshOn.Before(currentTime) || p.Token.ExpiresOn.Before(currentTime.Add(5*time.Minute)) {
 		newToken, err := getAuxiliaryToken(p.client, p.url, p.scope)
 		if err != nil {
-			log.FromContext(context.Background()).Error(err, "Failed to get auxiliary token")
-			return
+			return err
 		}
-		p.auxToken.token = newToken
+		p.Token = newToken
 	}
+	return nil
 }
 
 func (p *AuxiliaryTokenPolicy) Do(req *policy.Request) (*http.Response, error) {
-	p.GetAuxiliaryToken()
-	req.Raw().Header.Add("x-ms-authorization-auxiliary", "Bearer "+p.auxToken.token.Token)
+	err := p.GetAuxiliaryToken()
+	if err != nil {
+		log.FromContext(req.Raw().Context()).Error(err, "Failed to get auxiliary token")
+		return nil, err
+	}
+	req.Raw().Header.Add("x-ms-authorization-auxiliary", "Bearer "+p.Token.Token)
 	return req.Next()
 }
 
 func NewAuxiliaryTokenPolicy(ctx context.Context, client AuxiliaryTokenServer, url string, scope string) (*AuxiliaryTokenPolicy, error) {
-	token, err := getAuxiliaryToken(client, url, scope)
-	if err != nil {
-		return nil, fmt.Errorf("failed to get auxiliary token: %w", err)
-	}
 	auxPolicy := AuxiliaryTokenPolicy{
-		auxToken: AuxiliaryToken{token: token, lock: sync.Mutex{}},
-		url:      url,
-		scope:    scope,
-		client:   client,
+		Token:  azcore.AccessToken{},
+		url:    url,
+		scope:  scope,
+		client: client,
+		lock:   sync.Mutex{},
 	}
 	log.FromContext(ctx).V(1).Info("Will use auxiliary token policy for creating virtual machines")
 	return &auxPolicy, nil
--- a/pkg/auth/auxiliarytoken.go
+++ b/pkg/auth/auxiliarytoken.go
@@ -21,6 +21,8 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"sync"
+	"time"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
@@ -36,11 +38,34 @@ var _ policy.Policy = &AuxiliaryTokenPolicy{}
 // AuxiliaryTokenPolicy provides a custom policy used to authenticate
 // with shared node image galleries.
 type AuxiliaryTokenPolicy struct {
-	Token string
+	auxToken AuxiliaryToken
+	url      string
+	scope    string
+	client   AuxiliaryTokenServer
+}
+
+type AuxiliaryToken struct {
+	token azcore.AccessToken
+	lock  sync.Mutex
+}
+
+func (p *AuxiliaryTokenPolicy) GetAuxiliaryToken() {
+	p.auxToken.lock.Lock()
+	defer p.auxToken.lock.Unlock()
+	// If the token is expired or close to expiration, fetch a new one
+	if p.auxToken.token.ExpiresOn.IsZero() || p.auxToken.token.RefreshOn.Before(time.Now()) || p.auxToken.token.ExpiresOn.Before(time.Now().Add(5*time.Minute)) {
+		newToken, err := getAuxiliaryToken(p.client, p.url, p.scope)
+		if err != nil {
+			log.FromContext(context.Background()).Error(err, "Failed to get auxiliary token")
+			return
+		}
+		p.auxToken.token = newToken
+	}
 }
 
 func (p *AuxiliaryTokenPolicy) Do(req *policy.Request) (*http.Response, error) {
-	req.Raw().Header.Add("x-ms-authorization-auxiliary", "Bearer "+p.Token)
+	p.GetAuxiliaryToken()
+	req.Raw().Header.Add("x-ms-authorization-auxiliary", "Bearer "+p.auxToken.token.Token)
 	return req.Next()
 }
 
@@ -49,7 +74,12 @@ func NewAuxiliaryTokenPolicy(ctx context.Context, client AuxiliaryTokenServer, u
 	if err != nil {
 		return nil, fmt.Errorf("failed to get auxiliary token: %w", err)
 	}
-	auxPolicy := AuxiliaryTokenPolicy{Token: token.Token}
+	auxPolicy := AuxiliaryTokenPolicy{
+		auxToken: AuxiliaryToken{token: token, lock: sync.Mutex{}},
+		url:      url,
+		scope:    scope,
+		client:   client,
+	}
 	log.FromContext(ctx).V(1).Info("Will use auxiliary token policy for creating virtual machines")
 	return &auxPolicy, nil
 }
--- a/pkg/fake/auxiliarytokenserver_test.go
+++ b/pkg/fake/auxiliarytokenserver_test.go
@@ -17,13 +17,13 @@ limitations under the License.
 package fake
 
 import (
-	"context"
+	"net/http"
+	"net/url"
 	"testing"
 	"time"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/Azure/karpenter-provider-azure/pkg/auth"
-	armopts "github.com/Azure/karpenter-provider-azure/pkg/utils/opts"
 	"github.com/stretchr/testify/assert"
 )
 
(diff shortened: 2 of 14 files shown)


pkg/auth/auxiliarytoken.go
```diff
@@ -36,11 +38,34 @@ var _ policy.Policy = &AuxiliaryTokenPolicy{}
...
 // with shared node image galleries.
 type AuxiliaryTokenPolicy struct {
-	Token string
+	auxToken AuxiliaryToken
+	url      string
+	scope    string
+	client   AuxiliaryTokenServer
+}
+
+type AuxiliaryToken struct {
```
matthchr: minor: I personally probably wouldn't bother with this struct and would put the lock at the policy scope rather than the token scope.

i.e.

```
	auxToken azcore.AccessToken
	url      string
	scope    string
	client   AuxiliaryTokenServer

	 lock sync.Mutex
```

pkg/auth/auxiliarytoken.go
```diff
@@ -49,7 +74,12 @@ func NewAuxiliaryTokenPolicy(ctx context.Context, client AuxiliaryTokenServer, u
 	if err != nil {
```
matthchr: minor: Usually when you make a token policy like this we don't actually bother to get the token right away, we get it on first use instead. Your refresh code seems to handle that well w/ checking the token `ExpiresOn.IsZero`, so you could consider dropping this initial token request.

pkg/auth/auxiliarytoken.go
```diff
@@ -36,11 +38,34 @@ var _ policy.Policy = &AuxiliaryTokenPolicy{}
...
+
+func (p *AuxiliaryTokenPolicy) GetAuxiliaryToken() {
+	p.auxToken.lock.Lock()
+	defer p.auxToken.lock.Unlock()
+	// If the token is expired or close to expiration, fetch a new one
+	if p.auxToken.token.ExpiresOn.IsZero() || p.auxToken.token.RefreshOn.Before(time.Now()) || p.auxToken.token.ExpiresOn.Before(time.Now().Add(5*time.Minute)) {
+		newToken, err := getAuxiliaryToken(p.client, p.url, p.scope)
+		if err != nil {
+			log.FromContext(context.Background()).Error(err, "Failed to get auxiliary token")
+			return
```
matthchr: You should return this error up via `Do` I _think_

pkg/auth/auxiliarytoken.go
```diff
@@ -36,11 +38,34 @@ var _ policy.Policy = &AuxiliaryTokenPolicy{}
...
+type AuxiliaryToken struct {
+	token azcore.AccessToken
+	lock  sync.Mutex
+}
+
+func (p *AuxiliaryTokenPolicy) GetAuxiliaryToken() {
+	p.auxToken.lock.Lock()
+	defer p.auxToken.lock.Unlock()
+	// If the token is expired or close to expiration, fetch a new one
+	if p.auxToken.token.ExpiresOn.IsZero() || p.auxToken.token.RefreshOn.Before(time.Now()) || p.auxToken.token.ExpiresOn.Before(time.Now().Add(5*time.Minute)) {
```
matthchr: minor: Consider calling `time.Now()` a single time.

`now := time.Now()`.

pkg/auth/auxiliarytoken.go
```diff
@@ -36,11 +38,34 @@ var _ policy.Policy = &AuxiliaryTokenPolicy{}
...
+			log.FromContext(context.Background()).Error(err, "Failed to get auxiliary token")
+			return
+		}
+		p.auxToken.token = newToken
+	}
 }
 
 func (p *AuxiliaryTokenPolicy) Do(req *policy.Request) (*http.Response, error) {
-	req.Raw().Header.Add("x-ms-authorization-auxiliary", "Bearer "+p.Token)
+	p.GetAuxiliaryToken()
```
matthchr: Should probably consider passing `req.Context()` to this as ctx -- though worth checking if that's what other impls o token auth are doing I assume that it is.

matthchr: Yeah, they use the request ctx: https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azcore/runtime/policy_bearer_token.go#L47.

Something worth noting: In general in (modern) Go anytime you find yourself using the background context when you're doing something that's a network (or probably filesystem) call, you should be looking for "which context should I use". Sometimes the answer is the background context, but usually there's a context stashed someplace you can/should use. net/http is just a bit interesting in that it puts it on the actual request object.

pkg/auth/auxiliarytoken.go
matthchr: Oh also the log.FromContext above isn't really gonna work/do anything because the background context won't have a logger attached so you'll get the default logger which I don't think we want. Doing what I mention elsewhere of using the http req context may solve this but I am not sure if we want a separate log here or if it's better to just return the error and let the caller log.

pkg/auth/auxiliarytoken.go
```diff
@@ -36,20 +38,46 @@ var _ policy.Policy = &AuxiliaryTokenPolicy{}
 // AuxiliaryTokenPolicy provides a custom policy used to authenticate
 // with shared node image galleries.
 type AuxiliaryTokenPolicy struct {
-	Token string
+	Token  azcore.AccessToken
```
matthchr, line 40: minor: does this need to be exported?

pkg/auth/auxiliarytoken.go
```diff
@@ -36,20 +38,46 @@ var _ policy.Policy = &AuxiliaryTokenPolicy{}
...
+	err := p.GetAuxiliaryToken()
+	if err != nil {
+		log.FromContext(req.Raw().Context()).Error(err, "Failed to get auxiliary token")
+		return nil, err
+	}
+	req.Raw().Header.Add("x-ms-authorization-auxiliary", "Bearer "+p.Token.Token)
 	return req.Next()
 }
 
 func NewAuxiliaryTokenPolicy(ctx context.Context, client AuxiliaryTokenServer, url string, scope string) (*AuxiliaryTokenPolicy, error) {
```
matthchr: minor: you could drop the `ctx` parameter here if you didn't log below - FWIW this log feels more appropriate for the caller to be writing than the token itself, so maybe that makes sense?

pkg/fake/virtualmachinesapi.go
```diff
@@ -88,7 +90,17 @@ func (c *VirtualMachinesAPI) Reset() {
 	})
 }
 
-func (c *VirtualMachinesAPI) BeginCreateOrUpdate(_ context.Context, resourceGroupName string, vmName string, parameters armcompute.VirtualMachine, options *armcompute.VirtualMachinesClientBeginCreateOrUpdateOptions) (*runtime.Poller[armcompute.VirtualMachinesClientCreateOrUpdateResponse], error) {
+func (c *VirtualMachinesAPI) UseAuxiliaryPolicy() error {
```
matthchr: Confused as to what this is doing.

pkg/providers/instance/azure_client.go
```diff
@@ -134,10 +134,11 @@ func NewAZClient(ctx context.Context, cfg *auth.Config, env *azclient.Environmen
 
 	// copy the options to avoid modifying the original
 	var vmClientOptions = *opts
+	var auxiliaryTokenClient auth.AuxiliaryTokenServer
 	if o.UseSIG {
 		klog.V(1).Info("Using SIG for image versions")
-		client := &http.Client{Timeout: 10 * time.Second}
-		auxPolicy, err := auth.NewAuxiliaryTokenPolicy(ctx, client, o.SIGAccessTokenServerURL, o.SIGAccessTokenScope)
+		auxiliaryTokenClient = &http.Client{Timeout: 10 * time.Second}
```
matthchr: I missed that we were doing this before, but I dont think that a bare http client here is correct, because it will use the Go-defined default transport and I think we want to share transports with the other clients we have constructed.

We should be setting our http client here to have `Transport: opts.Transport`

pkg/providers/imagefamily/suite_test.go
```diff
@@ -27,12 +27,14 @@ import (
...
+	"github.com/Azure/karpenter-provider-azure/pkg/operator/options"
 	"github.com/Azure/karpenter-provider-azure/pkg/test"
 )
 
 var (
-	ctx context.Context
-	env *coretest.Environment
+	ctx         context.Context
+	env         *coretest.Environment
+	testOptions *options.Options
```
matthchr: Not used anywhere in this file?

pkg/test/nodeclass.go
```diff
@@ -62,8 +62,12 @@ func AKSNodeClass(overrides ...v1beta1.AKSNodeClass) *v1beta1.AKSNodeClass {
 	}
 }
 
-func ApplyDefaultStatus(nodeClass *v1beta1.AKSNodeClass, env *coretest.Environment) {
-	ApplyCIGImages(nodeClass)
+func ApplyDefaultStatus(nodeClass *v1beta1.AKSNodeClass, env *coretest.Environment, useSIG bool) {
```
matthchr, line 66: Does it make sense to just pass the whole test.Options here rather than just UseSIG? The thinking being that there may be other options that configure behavior that we want to have access to in here at some point?

pkg/auth/auxiliarytoken.go
```diff
@@ -21,6 +21,8 @@ import (
 	"encoding/json"
 	"fmt"
```
matthchr, line 21: supermeganit: Should this file be `auxiliary_token.go`?

pkg/fake/auxiliarytokenserver_test.go
```diff
@@ -27,6 +28,11 @@ import (
 )
 
 func Test_AddAuxiliaryTokenPolicyClientOptions(t *testing.T) {
```
matthchr, line 30: Is this test in the right place?

Its location implies its testing the `auxiliarytokenserver`, but it actually really seems to be more about testing the token itself.

Can you define the auxiliarytokenserver here and then use it in `auth/auxiliary_token_test.go`?

pkg/providers/instance/suite_test.go
```diff
@@ -164,6 +169,92 @@ var _ = Describe("InstanceProvider", func() {
...
 
+	When("getting the auxiliary token", func() {
+		var originalOptions *options.Options
+		var originalEnv *test.Environment
+		var originalCloudProvider *cloudprovider.CloudProvider
+		newOptions := test.Options(test.OptionsFields{
+			UseSIG: lo.ToPtr(true),
+		})
+		BeforeEach(func() {
+			originalOptions = options.FromContext(ctx)
```
matthchr, line 180: I dont think you should need the originalOptions stuff here (and then "unrolling" your options customization in aftereach). Or rather I see why you need them but I dont think it's great/right to do it.

It would be good if there was a more obvious/official way of per-test configuration, but it looks like the pattern right now is to just construct the environment up above, like with azureEnvNonZonal:
```
	cloudProvider = cloudprovider.New(azureEnv.InstanceTypesProvider, azureEnv.InstanceProvider, events.NewRecorder(&record.FakeRecorder{}), env.Client, azureEnv.ImageProvider)
	cloudProviderNonZonal = cloudprovider.New(azureEnvNonZonal.InstanceTypesProvider, azureEnvNonZonal.InstanceProvider, events.NewRecorder(&record.FakeRecorder{}), env.Client, azureEnvNonZonal.ImageProvider)
```

Maybe it makes more sense to do that then this for now?

cc @tallaxes if there's a better way to do per-test configuration...

tallaxes: I don't think we have a good pattern this yet. Neither is great, but yes, separate environment is cleaner (though heavier). With options set/undo pattern I am also concerned about its effect on sometime running tests in parallel. But if this test currently works, for expediency I think I would vote for keeping it this way, merging and revisiting (perhaps right away).

pkg/auth/auxiliarytoken.go
rakechill: I think we mostly seem to be only adding `_` when in the context of test files
image

pkg/auth/auxiliarytoken.go
rakechill: It makes testing a bit easier which is why I did it. I played around with using a setter method instead, but it made the tests more complicated + I felt like if I was already exposing a setter method, might as well let it be directly edited

pkg/auth/auxiliarytoken.go
rakechill: yeah I think it's maybe a bit better here, especially since it has access to the req's context too + makes more sense in the context of the vm create call

pkg/fake/auxiliarytokenserver_test.go
rakechill: Yeah I think this is leftover from my original tests when I didn't have vm api aware of auth token policy. I'll update these to be a proper unit test of the fake itself

pkg/fake/virtualmachinesapi.go
rakechill: Faking this behavior is a little awkward, because the policy is added as part of the vm client itself + then not accessible and used transparently on each request made by the vm client. 

So in order to the mock this behavior, I needed a way to have this policy be called at the beginning of each request to the fake api. 

And I've only added it on vm create, but should probably add it to the other requests as well to accurately reflect the real behavior. I removed it for a short time, because it was making it harder to reason about # of calls to the AuxiliaryTokenServer, but I think now the logic is correct, so it would only be called once if it needs to be refreshed (on create)

The nil check is there so that we don't do anything if the policy/server isn't set (this will be the case for USE_SIG= false).

rakechill: TODO for me: (1) add a comment above this function to explain intended behavior and (2) ensure it's called before each vm api call.

Lmk if you need more explanation, though.

pkg/providers/imagefamily/suite_test.go
rakechill: yeah I was debating adding this as a var in the nodeimage_test.go file (no vars yet) or including here. I'll move there.

pkg/providers/instance/suite_test.go
rakechill: Yeah I heavily debated between separate environment, opts, ctx, and cloudprovider vs. making temp copies here. In this file alone, we have one instance of each of those patterns so I ended up landing on this one. But maybe since it is affecting all of these components, having a separate one makes more sense...

pkg/test/nodeclass.go
rakechill: I think I'm good either way in the case, could leave as a TODO if we want to incorporate more options later?

pkg/auth/auxiliarytoken.go
matthchr: This call can't fail anymore so I think no `err` needed

pkg/providers/instance/suite_test.go
```diff
@@ -164,6 +169,92 @@ var _ = Describe("InstanceProvider", func() {
...
+				ExpectScheduled(ctx, env.Client, pod)
+
+				Expect(azureEnv.VirtualMachinesAPI.VirtualMachineCreateOrUpdateBehavior.CalledWithInput.Len()).To(Equal(1))
+				Expect(azureEnv.AuxiliaryTokenServer.AuxiliaryTokenDoBehavior.CalledWithInput.Len()).To(Equal(1))
+			})
+		})
+
+		Context("token is cached by previous vmClient call", func() {
+			BeforeEach(func() {
+				azureEnv.VirtualMachinesAPI.AuxiliaryTokenPolicy.Token = azureEnv.AuxiliaryTokenServer.Token
```
matthchr: Isn't the better way to do this to configure the AuxTokenServer to serve a token with these configurations, and then you can actually test that the policy is "getting the token from the server and doing the right thing with it"?

For example here you're not testing that you ever assign the token in the policy from the GetToken call - if you just never assigned the token AFAIK all your tests I see here would pass?

matthchr: Actually I see how what you're doing here is basically doing that.

pkg/fake/virtualmachinesapi.go
matthchr: That makes sense. I think a comment here to that effect would help.

pkg/providers/instance/azure_client.go
rakechill: We actually have a DefaultHTTPClient that i'll use

pkg/providers/instance/suite_test.go
matthchr: Fine if we file an item for this. My concern was the same as yours Alex with parallel tests and pollution of further test state (part of the reason why I am not a huge fan of Ginkgo btw).

pkg/providers/instance/suite_test.go
rakechill: The goal here was to simulate that the get token call already happened. These tests are more around "how can we ensure it gets called again due to refresh/expires being invalid." The tests ensure that the token server was called (if a refresh was needed), a vm was created, a node was provisioning, and a pod got scheduled onto the node.

But if the BeforeEach also calls GetAuxiliaryTokenPolicy, the # of calls to aux token server gets a little harder to reason about. Should be 2 calls (one for init + one for refresh)

pkg/test/nodeclass.go
matthchr: Happy with a TODO for now

pkg/auth/auxiliarytoken.go
matthchr: My thinking on this log being awkward here is that it makes assumptions about where this policy is being used (virtual machines), which is obviously what it's for but there's nothing in the shape of the policy itself that _requires_ that it be used only on a VM client.

It's a minor nit though.

pkg/providers/instance/suite_test.go
rakechill: A setter func on aux token server might make more sense than exporting aux token policy's Token field tho...

pkg/fake/auxiliarytokenserver_test.go
rakechill: fixed

matthchr (approved):
I think worth a followup possibly on how we're constructing the test env (see previous comment), but approving to unblock 1.5 release. The rest LGTM

PR #1004, opened 2025-06-25
feat(metrics): add metric for cloudprovider async operations, starting with Create()

comtalyst:
Fixes # 

**Description**
[karpenter_cloudprovider_duration_seconds](https://github.com/kubernetes-sigs/karpenter/blob/0611f9b77d794f26491e0f037df09b9061231ff3/pkg/cloudprovider/metrics/cloudprovider.go#L48-L61) does not cover Azure's case where `Create()` has asynchronous executions introduced by https://github.com/Azure/karpenter-provider-azure/pull/836. In addition, it does not categorize by errors, in which the duration is affected by. 

This new provider-level metric `karpenter_azure_cloudprovider_duration_seconds_with_async` behaves similarly to `karpenter_cloudprovider_duration_seconds`, but ends the measurement on async operation completion instead of method call return. It also has an extra error field to categorize errors.

The only use case introduced by this PR is `Create()`. It has only one extra routine. The measurement only occurs at the end of that extra routine if it reaches that state. An alternative could be to wait for the end of every routine, including the primary one, which can include the case where async phase is not reached. But, that complexity is unnecessary as of now.

* Unit tests added
* [E2E](https://github.com/Azure/karpenter-provider-azure/actions/runs/15884922267/job/44794435358)

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note
* Add a new metric `karpenter_azure_cloudprovider_duration_seconds_with_async` for async Create()
* Rename metric `karpenter_image_selection_error_count` to `karpenter_azure_image_selection_error_count`

```


pkg/cloudprovider/cloudprovider.go
```diff
@@ -165,12 +168,25 @@ func (c *CloudProvider) Create(ctx context.Context, nodeClaim *karpv1.NodeClaim)
...
 			log.FromContext(ctx).Error(err, "panic during waitOnPromise")
+			mainErrForMetrics = err
 		}
+
+		defer func() {
+			if r := recover(); r != nil {
+				err := fmt.Errorf("%v", r)
+				log.FromContext(ctx).Error(err, "panic during waitOnPromise defer, likely during metrics recording")
+			}
+		}()
```
comtalyst, line 186: I'm debating whether this is worth it. Technically the metrics method below can panic if the labels are not matched what's defined (which is annoying, as the definition is not at compile time, but with strings).

pkg/metrics/constants.go
```diff
@@ -18,8 +18,20 @@ package metrics
 
 const (
 	// Namespace(s).
-	Namespace = "karpenter"
+	Namespace = "karpenter_azure"
```
comtalyst, line 21: This will diverge from aws provider, but I think it is better to make it clear if a metric is provider or core, per abstractions.

comtalyst: Although, it will rename ImageSelectionErrorCount metric we have.

matthchr, line 21: This is technically a breaking change as metrics published are a contract and users may be consuming them, this will change the existing metric we have published.

With that said, I am not against the idea of differentiating from upstream. I'll defer to @tallaxes on if we should actually take this breaking change or not though.

pkg/cloudprovider/cloudprovider.go
```diff
@@ -165,12 +168,25 @@ func (c *CloudProvider) Create(ctx context.Context, nodeClaim *karpv1.NodeClaim)
...
+
+		duration := time.Since(startTime)
+		azuremetrics.MethodDurationWithAsync.With(azuremetrics.GetLabelsMapForCloudProviderDurationWithAsync(ctx, "Create", mainErrForMetrics)).Observe(duration.Seconds())
```
matthchr, line 189: I think we want these labels to be `snake_case`, so `Create` should be `create`?

pkg/metrics/metrics.go
```diff
@@ -31,10 +36,38 @@ var (
...
+
+	MethodDurationWithAsync = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Namespace: Namespace,
+			Subsystem: cloudProviderSubsystem,
+			Name:      "duration_seconds_with_async",
+			Help:      "Duration of cloud provider method calls. Includes async/LRO operations in routines that run beyond the initial call.",
+			Buckets:   prometheus.DefBuckets,
+		},
+		[]string{metricLabelController, metricLabelMethod, metricLabelError},
```
matthchr, line 48: Upstream duration histogram doesn't have an error label... should we?

pkg/metrics/metrics.go
```diff
@@ -31,10 +36,38 @@ var (
 		},
 		[]string{"family"},
 	)
+
+	MethodDurationWithAsync = prometheus.NewHistogramVec(
```
matthchr, line 40: What's the actual objective with this metric - what are we trying to track?

The upstream metric seems most useful to me because CP.Create is effectively "blocking". The additional time from the async side of things is nonblocking, which isn't to say that we don't want to track it, but I'm wondering if we want to track it at a level that exposes more interesting details, such as the HTTP level.

For example, instead of `duration_seconds_with_async` we could have `duration_seconds` and `duration_lro_seconds` in the `http` (or `client`?) subsystem, with labels: `resource_type`, `status_code` (and possibly `error_code`?)

Also in that same subsystem we could have `requests_total` and possibly other azure/http focused metrics.

It feels like there a quite a few advantages to that angle: 
`status_code` or `error_code` are _relatively_ finite sets, such that it's probably OK to put them in prom metrics as the cardinality isn't going to be super high, whereas in this approach you've got an error label but that isn't all that useful because it's hard at this level to classify errors (other than there was one).

It _seems_ like this angle would still probably give us what we wanted here, but with the advantage of also having other uses? I might be wrong though.

pkg/cloudprovider/cloudprovider.go
comtalyst: That is not what's happening with core metrics (e.g., `karpenter_cloudprovider_duration_seconds`). I would lean towards parity.

pkg/metrics/metrics.go
comtalyst: > What's the actual objective with this metric - what are we trying to track?

It will track end-to-end provisioning latency for each node by error categories, which will include error count in the process. Yes, this means we want to track the async phase for each node, even though they are not blocking each other.

> I'm wondering if we want to track it at a level that exposes more interesting details, such as the HTTP level

Sounds like that relies on the lower abstraction unnecessarily? Especially considering the fact that we have multiple CRP/NRP requests, and that would change.

> `status_code`

At least currently, I wouldn't categorize/determine correctness by that directly. For example, some bad requests could be we misusing the services (e.g., ephemeral disk issue), and some 200 returns with errors. 

> it's hard at this level to classify errors

I think we are okay putting effort into error categorization, in which is already done for the most part (not in this code). Unknown errors would serve as a fallback for new ones, which will be dev's responsibility to categorize them.

pkg/metrics/metrics.go
comtalyst: It is intentional---this will allow us to exclude failed scale ups from being measured in scale up latency (or allow them to be processed separately however we want to).

PR #1010, opened 2025-06-26
chore: use context logging instead of klogs

alimaazamat:
Contributes to #833

**Description**

*

- [ ] Yes, issue opened: # 
- [X] No

**Release Note**

```release-note
Removed klogs to structured contextual logging for SKU and azure client.
```

### DIFF
--- a/pkg/providers/instance/azure_client.go
+++ b/pkg/providers/instance/azure_client.go
@@ -20,6 +20,7 @@ import (
 	"context"
 
 	"sigs.k8s.io/cloud-provider-azure/pkg/azclient"
+	"sigs.k8s.io/controller-runtime/pkg/log"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
@@ -130,6 +131,7 @@ func NewAZClient(ctx context.Context, cfg *auth.Config, env *azclient.Environmen
 	var vmClientOptions = *opts
 	var auxiliaryTokenClient auth.AuxiliaryTokenServer
 	if o.UseSIG {
+		log.FromContext(ctx).Info("using SIG for image versions with auxiliary token policy for creating virtual machines")
 		auxiliaryTokenClient = armopts.DefaultHTTPClient()
 		auxPolicy := auth.NewAuxiliaryTokenPolicy(auxiliaryTokenClient, o.SIGAccessTokenServerURL, o.SIGAccessTokenScope)
 		vmClientOptions.ClientOptions.PerRetryPolicies = append(vmClientOptions.ClientOptions.PerRetryPolicies, auxPolicy)
--- a/pkg/providers/imagefamily/nodeimage.go
+++ b/pkg/providers/imagefamily/nodeimage.go
@@ -97,7 +97,7 @@ func (p *provider) List(ctx context.Context, nodeClass *v1beta1.AKSNodeClass) ([
 
 	var nodeImages []NodeImage
 	if useSIG {
-		log.FromContext(ctx).V(1).Info("Using SIG to list node images")
+		log.FromContext(ctx).V(1).Info("using SIG to list node images")
 		nodeImages, err = p.listSIG(ctx, supportedImages)
 		if err != nil {
 			return []NodeImage{}, err


pkg/providers/instance/azure_client.go
```diff
@@ -121,55 +121,55 @@ func NewAZClient(ctx context.Context, cfg *auth.Config, env *azclient.Environmen
...
 	}
-	klog.V(5).Infof("Created network interface client %v using token credential", interfacesClient)
+	log.FromContext(ctx).V(1).Info("created network interface client using token credential", "client", interfacesClient)
 
 	// copy the options to avoid modifying the original
 	var vmClientOptions = *opts
 	var auxiliaryTokenClient auth.AuxiliaryTokenServer
 	if o.UseSIG {
-		klog.V(1).Info("Using SIG for image versions")
+		log.FromContext(ctx).V(1).Info("using SIG for image versions")
```
tallaxes: We likely want this by default / at V(0) - and I would expect that applies to any klog.V(1)

pkg/providers/instance/azure_client.go
```diff
@@ -121,55 +121,55 @@ func NewAZClient(ctx context.Context, cfg *auth.Config, env *azclient.Environmen
 	if err != nil {
 		return nil, err
 	}
-	klog.V(5).Infof("Created virtual machine extensions client %v using token credential", extensionsClient)
+	log.FromContext(ctx).V(1).Info("created virtual machine extensions client using token credential", "client", extensionsClient)
```
tallaxes: What is actually logged here? We probably have not observed this much, since it is / was at V(5). Wondering if these - logging all *Clients - are actually useful ...

pkg/providers/instance/azure_client.go
rakechill: we maybe don't have any V(0) logs?
image
>https://github.com/search?q=repo%3AAzure%2Fkarpenter-provider-azure+V%280%29&type=code

rakechill: I think we have some info logs that have no verbosity specified, like [this](https://github.com/Azure/karpenter-provider-azure/blob/112462d1d9e31fec982fc90f2460b5d24c79db51/pkg/cloudprovider/cloudprovider.go#L448):
`log.FromContext(ctx).Info("failed to get zone for VM, zone label will be empty", "vmName", *vm.Name, "error", err)`

pkg/providers/instance/azure_client.go
rakechill: I see what you mean, like the absence of an error log for most of these is probably enough.

pkg/providers/instance/azure_client.go
tallaxes: It is V(0) if not specified

pkg/providers/instance/azure_client.go
alimaazamat: Yeah I agree with the principle that "no news is good news."

pkg/providers/instance/skuclient/skuclient.go
```diff
@@ -64,25 +64,29 @@ func (sc *skuClient) updateInstance() {
...
-	klog.V(5).Infof("Created sku client with authorizer: %v", skuClient)
 
 	sc.instance = skuClient
 }
 
 func NewSkuClient(ctx context.Context, cfg *auth.Config, env *azclient.Environment) SkuClient {
+	ctx = log.IntoContext(ctx, log.FromContext(ctx).WithName("skuclient").WithValues(
+		"subscriptionID", cfg.SubscriptionID,
+		"environment", env.Name,
+	))
```
alimaazamat, line 75: To match [pricing.go](https://github.com/alimaazamat/karpenter-provider-azure/blob/cleanup-klogs/pkg/providers/pricing/pricing.go#L86) I created this logging context and just used values that felt the most important base don the inputs... Let me know if these values feel useful or

pkg/providers/instance/azure_client.go
tallaxes: The latest iteration [dropped](https://github.com/Azure/karpenter-provider-azure/pull/1010/files#diff-8712d91345ec5a646969fcb26dfb605268403e8887c9f48e9e9c3a10238cb366L136-L138) these two altogether, I think we want to keep them (fine to combine), with log.FromContext(ctx).Info

alimaazamat: Ah okay, I thought those two were also considered success logging.

rakechill (approved):

PR #1011, opened 2025-06-28
fix: don't use ephemeral disk if not supported

tallaxes:
Follow-up to #940 

**Description**

Check whether SKU supports ephemeral disk. Fixes regression in provisioning of selected SKUs that don't support ephemeral OS disk, but might have cache or temp disk that is larger than OS disk (such as Standard_D8pls_v5) - a combination that would cause attempted use of ephemeral OS disk and fail the provisioning.

* Add the check for ephemeral disk support
* Use the same exact logic for ephemeral OS disk size requirement population
* Drop outdated restriction on `Dlds_v5` (which was likely in place due to lack of placement support); tested
* Update unit tests & add E2E tests

There are some additional refactoring, SKU codegen, and cleanup of comments possible and desirable; keeping changes in this PR to the minimum to focus on fixing the regression and facilitate easy review

* `make presubmit`
* manual testing of Standard_D8pls_v5 (original failure) & Dlds_v5 (removed restriction)
* E2E: https://github.com/Azure/karpenter-provider-azure/actions/runs/15945735374

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/providers/instancetype/instancetype.go
+++ b/pkg/providers/instancetype/instancetype.go
@@ -178,7 +178,7 @@ func computeRequirements(sku *skewer.SKU, vmsize *skewer.VMSizeType, architectur
 	// size parts
 	requirements[v1beta1.LabelSKUFamily].Insert(vmsize.Family)
 
-	setRequirementsEphemeralOSDiskSupported(requirements, sku, vmsize)
+	setRequirementsEphemeralOSDiskSupported(requirements, sku)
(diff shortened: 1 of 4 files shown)


Bryce-Soghigian (dismissed):
We probably should have had a test that catches this too

Bryce-Soghigian (approved):

Bryce-Soghigian (approved):

PR #1012, opened 2025-06-28
test: periodically run self-hosted e2es on main every 24 hours

Bryce-Soghigian:
**Description**
This PR runs these self hosted e2es every 24 hours on top of the merge and manual triggers. 
- schedule can only be triggered off of main and workflow files in main, so we can't test it, see github workflow docs: https://docs.github.com/en/actions/reference/events-that-trigger-workflows#schedule

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note

```

### DIFF
--- a/.github/workflows/e2e-matrix-trigger.yaml
+++ b/.github/workflows/e2e-matrix-trigger.yaml
@@ -1,7 +1,7 @@
 name: E2EMatrixTrigger
 on:
   schedule:
-    - cron: '0 6 * * *'
+    - cron: '0 11 * * *'
   workflow_dispatch:
     inputs:
       location:


.github/workflows/e2e-matrix-trigger.yaml
```diff
@@ -1,5 +1,7 @@
 name: E2EMatrixTrigger
 on:
+  schedule:
+    - cron: '0 6 * * *'
```
tallaxes: > "every 6 hours"
> cron: '0 6 * * *'

This is not every 6 hours, this is at 6:00 UTC. If you actually want every 6 hours would need to be */6. Though I am thinking once a day should be good enough (nighttime westcoast would be best)

Also, let's make sure failure notification works and gets triaged.

Bryce-Soghigian: 11pm PST seems good for me, but can lower to 7pm PST if you think that makes sense? 

Good catch, once a day is often enough.

tallaxes: I meant nighttime, as in maybe 3am PST, to minimize potential overlap with other E2E runs

tallaxes (changes requested):
Please configure for 3am PST, update title and description, and validate/document notification mechanism

tallaxes (approved):

PR #1014, opened 2025-06-29
fix: NIC error filter catches "ResourceNotFound", but not "NotFound"

comtalyst:
Fixes # 

**Description**
NIC not being found from `GET` returns `NotFound` rather than `ResourceNotFound`.
Seems like an issue for `azure-sdk-for-go-extensions`, if we have not been misusing it(?).
```
// IsNotFoundErr is used to determine if we are failing to find a resource within azure.
func IsNotFoundErr(err error) bool {
	azErr := IsResponseError(err)
	return azErr != nil && azErr.ErrorCode == ResourceNotFound
}
```

But `deleteVirtualMachineIfExists()` right below this code already takes this matter into our own hands.
The same problem exists for NIC, in which this the same solution could be applied.

* `make presubmit`

- [ ] Yes, issue opened: # 
- [X] No

**Release Note**

```release-note
* Fix an issue where NIC "NotFound" during deletion is considered an error
```

### DIFF
--- a/pkg/providers/instance/azureresourcemanagerutils.go
+++ b/pkg/providers/instance/azureresourcemanagerutils.go
@@ -107,7 +107,8 @@ func deleteNic(ctx context.Context, client NetworkInterfacesAPI, rg, nicName str
 func deleteNicIfExists(ctx context.Context, client NetworkInterfacesAPI, rg, nicName string) error {
 	_, err := client.Get(ctx, rg, nicName, nil)
 	if err != nil {
-		if sdkerrors.IsNotFoundErr(err) {
+		azErr := sdkerrors.IsResponseError(err)
+		if azErr != nil && (azErr.ErrorCode == "NotFound" || azErr.ErrorCode == "ResourceNotFound") {
 			return nil
 		}
 		return err


tallaxes (approved):
I am assuming this was actually observed. (In our previous discussion the - temporary -conclusion was that these are specific to each RP provider and cannot be generalized easily.) @matthchr

matthchr:
Best fix would be to just check HTTP 404 for these types of cases if we can.  I think we can here because `azerr` should have a `Code`?

pkg/providers/instance/azureresourcemanagerutils.go
```diff
@@ -107,7 +107,8 @@ func deleteNic(ctx context.Context, client NetworkInterfacesAPI, rg, nicName str
 func deleteNicIfExists(ctx context.Context, client NetworkInterfacesAPI, rg, nicName string) error {
 	_, err := client.Get(ctx, rg, nicName, nil)
 	if err != nil {
-		if sdkerrors.IsNotFoundErr(err) {
+		azErr := sdkerrors.IsResponseError(err)
+		if azErr != nil && (azErr.ErrorCode == "NotFound" || azErr.ErrorCode == "ResourceNotFound") {
```
Bryce-Soghigian, line 111: we should be using the azuresdkforgoextensions not found helper here which checks both

tallaxes: > we should be using the azuresdkforgoextensions not found helper here which checks both

It does not check both. And whether it should is complicated by the fact that this behavior differs by RP

tallaxes: The existing helper does not check both. We could introduce one, but that is complicated by the fact this behavior differs by RP

PR #1015, opened 2025-06-29
fix: make returning error on NIC deletion failure on cleanup optional

comtalyst:
Fixes # 

**Description**
We don't want to consider cloudprovider.Delete() a failure on cases where VM deletion succeeded, but not NIC deletion. NIC garbage collector will be responsible for cleaning up those cases anyway.

This will make `karpenter_cloudprovider_errors_total` more practical by excluding these "safe" errors.

* `make presubmit`

- [ ] Yes, issue opened: # 
- [X] No

**Release Note**

```release-note
NONE
```

### DIFF
--- a/pkg/providers/instance/instance.go
+++ b/pkg/providers/instance/instance.go
@@ -160,7 +160,7 @@ func (p *DefaultProvider) BeginCreate(
 	if err != nil {
 		// There may be orphan NICs (created before promise started)
 		// This err block is hit only for sync failures. Async (VM provisioning) failures will be returned by the vmPromise.Wait() function
-		if cleanupErr := p.cleanupAzureResources(ctx, GenerateResourceName(nodeClaim.Name)); cleanupErr != nil {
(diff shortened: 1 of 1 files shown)


tallaxes (approved):
The change makes sense to me. Have slight dislike of boolean parameter, but seems contained enough. Needs E2E, but the change is small/safe enough to watch E2E post merge.

comtalyst:
Considered alternatives are:
- Just not return an error (i.e., equivalent to that bool set to `false`) --- would result in "failed to cleanup resources for node claim" not being logged, not sure if we want that change
- Same as above but move the logging into the method --- not conforming to the pattern?
- Return 2 errors for "fatal" and "not fatal" --- increased complexity
- Having 2 different methods for true/false --- unnecessary code duplication
- Have cloudprovider.Delete implement its own rather than using the method --- unnecessary details exposure

With that said, thought I shouldn't invest too much time in this for known reasons.

tallaxes:
> Having 2 different methods for true/false --- unnecessary code duplication

Can also keep boolean and have two wrappers that make it easy to consume. Agree this is not critical.

PR #1021, opened 2025-07-02
chore: add fail-fast false to CI-TEST action

Bryce-Soghigian:
Fixes # 

**Description**
I constantly end up re-running ci-test due to flake, if it fails in one version it fails it for all tests. That is annoying and often makes deflaking this take a couple additional runs, lets not use fail-fast and try this for a bit. 

The tests are so flakey it was very easy to reproduce this 

image

This shows this change works! And decreases the number of tests we have to rerun :) 

*

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


tallaxes (approved):
Want to fix the flake instead? 🙂

Bryce-Soghigian:
> Want to fix the flake instead? 🙂 

:(

PR #1022, opened 2025-07-02
feat(fips): added Ubuntu enum to ImageFamily and FIPSMode field

lnguyen924:
Fixes # 

**Description**
Update the AKSNodeClass API to add an Ubuntu enum to the ImageFamily field and a FIPSMode field with validation to reject unsupported ImageFamily and FIPSMode combinations (Ubuntu2204 + FIPS)

* make presubmit
* unit tests to validate the CEL enforcement for ImageFamily + FIPSMode combinations

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
NONE
```


charts/karpenter-crd/templates/karpenter.azure.com_aksnodeclasses.yaml
```diff
@@ -341,10 +341,18 @@ spec:
...
+                  nodes
+                enum:
+                - FIPS
+                - Disabled
+                type: string
               imageFamily:
                 default: Ubuntu2204
                 description: ImageFamily is the image family that instances use.
                 enum:
+                - Ubuntu
```
Bryce-Soghigian, line 355: why Ubuntu and not Ubuntu2004?

pkg/apis/crds/karpenter.azure.com_aksnodeclasses.yaml
```diff
@@ -484,6 +492,10 @@ spec:
                 pattern: (?i)^\/subscriptions\/[^\/]+\/resourceGroups\/[a-zA-Z0-9_\-().]{0,89}[a-zA-Z0-9_\-()]\/providers\/Microsoft\.Network\/virtualNetworks\/[^\/]+\/subnets\/[^\/]+$
                 type: string
             type: object
+            x-kubernetes-validations:
+            - message: FIPS is not yet supported for Ubuntu2204
+              rule: 'has(self.imageFamily) && has(self.fipsMode) ?  !(self.imageFamily
+                == ''Ubuntu2204'' && self.fipsMode == ''FIPS'') : true'
```
Bryce-Soghigian: This is awesome, can we have some unit tests that validate the CEL enforcement for this?

charts/karpenter-crd/templates/karpenter.azure.com_aksnodeclasses.yaml
charliedmcb: `Ubuntu` is the generic family we are adding, similar to `AzureLinux` being able to either select for you `v2`, or `v3` depending upon other factors, this new image family will either select `2004` if you are on FIPS currently, or `2204` if not, and might shift its selection over time with `k8s` version as well.

pkg/apis/crds/karpenter.azure.com_aksnodeclasses.yaml
charliedmcb, line 498: are the double single quotes `''` needed? Or can we just use one quote/use a double quote?

pkg/apis/v1beta1/consts.go
```diff
@@ -0,0 +1,23 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1beta1
+
+const (
```
charliedmcb: Maybe add minor descriptions on this consts and what they mean when set?

pkg/apis/v1beta1/aksnodeclass.go
```diff
@@ -41,8 +42,12 @@ type AKSNodeClassSpec struct {
 	ImageID *string `json:"-"`
 	// ImageFamily is the image family that instances use.
 	// +kubebuilder:default=Ubuntu2204
```
charliedmcb, line 51: @matthchr, @tallaxes, can we possibly change this to default `Ubuntu` as the generic family?

Seems incorrect to lock ourselves into this current pattern of fixing our default to a version in the long term?

This should have no behavior difference besides what shows up in the CRD spec, unless the customer sets FIPS, since currently without FIPS enabled, the behavior will map the generic `Ubuntu` family to `2204` anyways.

- Could be a follow up separate from Linda project though.

pkg/apis/v1beta1/crd_validation_cel_test.go
```diff
@@ -94,6 +94,51 @@ var _ = Describe("CEL/Validation", func() {
...
+			if fipsMode != v1beta1.FIPSUnspecified {
+				nodeClass.Spec.FIPSMode = &fipsMode
+			}
+			if expected {
+				Expect(env.Client.Create(ctx, nodeClass)).To(Succeed())
+			} else {
+				Expect(env.Client.Create(ctx, nodeClass)).ToNot(Succeed())
+			}
+		},
+			Entry("generic Ubuntu when FIPSMode is explicitly Disabled should succeed", "Ubuntu", v1beta1.FIPSDisabled, true),
```
charliedmcb: nit: I'd use the const values for the image family that are specified in the `pkg/apis/v1beta1/labels.go` file.

At least for the specified cases.

If you want to add an unspecified const in there too, you could do that as well.

pkg/apis/v1beta1/crd_validation_cel_test.go
```diff
@@ -94,6 +94,51 @@ var _ = Describe("CEL/Validation", func() {
...
+			Entry("generic AzureLinux when FIPSMode is not explicitly set should succeed", "AzureLinux", v1beta1.FIPSUnspecified, true),
+			Entry("generic AzureLinux when FIPSMode is explicitly FIPS should succeed", "AzureLinux", v1beta1.FIPSEnabled, true),
+			Entry("unspecified ImageFamily (defaults to Ubuntu2204) when FIPSMode is explicitly Disabled should succeed", "", v1beta1.FIPSDisabled, true),
+			Entry("unspecified ImageFamily (defaults to Ubuntu2204) when FIPSMode is not explicitly set should succeed", "", v1beta1.FIPSUnspecified, true),
+			Entry("unspecified ImageFamily (defaults to Ubuntu2204) when FIPSMode is explicitly FIPS should fail", "", v1beta1.FIPSEnabled, false),
+			Entry("invalid ImageFamily when FIPSMode is explicitly Disabled should fail", "abc", v1beta1.FIPSDisabled, false),
+			Entry("invalid ImageFamily when FIPSMode is not explicitly set should fail", "abc", v1beta1.FIPSUnspecified, false),
+			Entry("invalid ImageFamily when FIPSMode is explicitly FIPS should fail", "abc", v1beta1.FIPSEnabled, false),
+			Entry("invalid FIPSMode when ImageFamily is explicitly Ubuntu should fail", "Ubuntu", "xyz", false),
```
charliedmcb: sorry, I lead you a little astray here. I think just one invalid case test for `FIPS` should cover it, as that's really testing the validation of the `fipsMode` field itself and is different than the CEL combinations between the two fields.

I might actually split the test case for the invalid one into its own `It` block, rather than being part of the table driven tests here.

charliedmcb: Also, for the validation on the ImageFamily invalid its kind of testing something different. I don't see it covered elsewhere in this file, so I do think its worth adding a simple `It` test case for that as well.

charliedmcb (commented):
Overall looks quite good. Just a couple suggestions/comments.

charts/karpenter-crd/templates/karpenter.azure.com_aksnodeclasses.yaml
```diff
@@ -484,6 +492,10 @@ spec:
                 pattern: (?i)^\/subscriptions\/[^\/]+\/resourceGroups\/[a-zA-Z0-9_\-().]{0,89}[a-zA-Z0-9_\-()]\/providers\/Microsoft\.Network\/virtualNetworks\/[^\/]+\/subnets\/[^\/]+$
                 type: string
             type: object
+            x-kubernetes-validations:
+            - message: FIPS is not yet supported for Ubuntu2204
+              rule: 'has(self.imageFamily) && has(self.fipsMode) ?  !(self.imageFamily
```
matthchr: extremely minor, nit: I think you have two spaces here, which I don't think you need:
```suggestion
              rule: 'has(self.imageFamily) && has(self.fipsMode) ? !(self.imageFamily
```

pkg/apis/v1beta1/aksnodeclass.go
matthchr: Assuming that there is no backend difference today, I agree we can probably make this change.

Let's do it in a different PR though?

pkg/apis/v1beta1/aksnodeclass.go
```diff
@@ -41,8 +42,12 @@ type AKSNodeClassSpec struct {
...
-	// +kubebuilder:validation:Enum:={Ubuntu2204,AzureLinux}
+	// +kubebuilder:validation:Enum:={Ubuntu,Ubuntu2204,AzureLinux}
 	ImageFamily *string `json:"imageFamily,omitempty"`
+	// FIPSMode controls FIPS compliance for the provisioned nodes
+	// +kubebuilder:validation:Enum:={FIPS,Disabled}
+	// +optional
+	FIPSMode *string `json:"fipsMode,omitempty"`
```
matthchr: Hah, I totally missed that even for above ImageFamily this is just a *string.

It _could_ be a string alias like:
```
type FIPSMode string

const (
	FIPSMode_FIPS = FIPSMode("FIPS")
	FIPSMode_Disabled = FIPSMode("Disabled")
)
```

It also probably could avoid a ptr-to-string (it's not really amazingly necessary in this case).

I see that `ImageFamily` is a ptr to string, (also not making use of type aliases), but there are other places in our API like kubelet `TopologyManagerPolicy` where we don't use *string we just used string.

I see that upstream Karpenter uses aliases for these cases at least sometimes:
https://github.com/kubernetes-sigs/karpenter/blob/218cca8f9a96046b356addb2d91cc42bf205e275/pkg/apis/v1/nodepool.go#L75C22-L75C42

Are we intentionally not? Why not?

matthchr: Ah, ptr is maybe required because of optional-ness

tallaxes: Aliases: Not intentional, most likely just following the pattern of `AMIFamily`, which uses just string: https://github.com/aws/karpenter-provider-aws/blob/db00ff92a8cd3179d61567d7b6688eb3169f12eb/pkg/apis/v1/ec2nodeclass.go#L74. 

Ptr vs string: I suspect the original intent was to make it "nillable" (distinguish between empty and not specified), though not sure if justified (or applied consistently). It looks like either can be used with "optional" and "default"?

tallaxes: > Ah, ptr is maybe required because of optional-ness

That's what I was thinking, but it looks like it is not required, e.g. `TopologyManagerPolicy` is also optional

charliedmcb: > Ah, ptr is maybe required because of optional-ness

This was my assumption. We want to leave it optional, with no defined defaulting, in case we want to make the defaulting reliant on other conditions later. However, if it can be optional without being a ptr, that seems fine. Any awareness of best practice here, since it does appear there's examples of both?

matthchr: I'm happy with *Foo given optionality. I _think_ it may be worth doing type alias for it though - its obviously not technically any different than raw string but gives more clarity for consumers in Go (not sure how many of those we'll ever have but still).

It's also what it seems upstream has done with the more recent enums at least.

charliedmcb: Discussed offline.

This refactor will go as follows:
```
type FIPSMode string

var (
  FIPSModeFIPS     = FIPSMode("FIPS")
  FIPSModeDisabled = FIPSMode("Disabled")
)
```
With:
```
...
	// FIPSMode controls FIPS compliance for the provisioned nodes
	// +kubebuilder:validation:Enum:={FIPS,Disabled}
	// +optional
	FIPSMode *FIPSMode `json:"fipsMode,omitempty"`
```

And drop the `const.go` file, since with the direct wrapping and declaration of the FIPSMode Enum values, we don't need it. Also, unspecified is actually `nil`, so can drop that const.

This refactor will happen in the other PR.

With that adjustment, this PR lgtm.

PR #1023, opened 2025-07-02
test: nodeclaim suite is a bit flakey, adding explicit retries to the standalone nodeclaim tests

Bryce-Soghigian:
Fixes # 

**Description**
Added FlakeAttempts to reduce the flake of nodeclaim suite.  Alternatively we can re-write thise, but putting that off for now to improve the test QOS as a signal.
<TODO: Trigger 10 runs without nodeclaim flake> 
1. https://github.com/Azure/karpenter-provider-azure/actions/runs/16036168968
2. https://github.com/Azure/karpenter-provider-azure/actions/runs/16036167798

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note

```


tallaxes (changes requested):
* CI fails, E2E fails - it does not compile
* Please put the effort into the actual deflaking rather than hiding flakes behind repeats

PR #1027, opened 2025-07-08
feat(fips): adding FIPS images to imagefamilies

lnguyen924:
Fixes # 

**Description**
Add FIPS images to image families and update image selection logic to support FIPS compliance

* unit tests to validate FIPS selection and expected images

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
NONE
```


pkg/providers/imagefamily/resolver.go
```diff
@@ -221,14 +224,22 @@ func prepareKubeletConfiguration(ctx context.Context, instanceType *cloudprovide
...
+		return imageFamily.FIPSImages()
+	}
 	return imageFamily.DefaultImages()
 }
 
-func getImageFamily(familyName *string, kubernetesVersion string, parameters *template.StaticParameters) ImageFamily {
+func getImageFamily(familyName *string, fipsMode *string, kubernetesVersion string, parameters *template.StaticParameters) ImageFamily {
 	switch lo.FromPtr(familyName) {
+	case v1beta1.UbuntuImageFamily:
+		if lo.FromPtr(fipsMode) == "FIPS" {
```
Bryce-Soghigian: can we have a constant for this?

pkg/providers/imagefamily/nodeimage_test.go
```diff
@@ -67,10 +67,14 @@ func getExpectedTestCIGImages(imageFamily string, version string, kubernetesVers
 	return nodeImages
 }
 
+// assuming FIPSMode == Disabled
+//
 //nolint:unparam // might always be using the same version in test, but could change in the future
 func getExpectedTestSIGImages(imageFamily string, version string, kubernetesVersion string) []imagefamily.NodeImage {
```
charliedmcb: Based on naming, I think `getExpectedTestFIPSImages` should actually be consolidated into `getExpectedTestSIGImages`, and take `fipsMode` as a parameter.

This is because fips images are a subcategory of SIG, so really it would make sense in my opinion to call `getExpectedTestSIGImages` and have it internally determine the difference between the fips case and not.

pkg/providers/imagefamily/nodeimage_test.go
```diff
@@ -171,39 +200,101 @@ var _ = Describe("NodeImageProvider tests", func() {
...
+
+			foundImages, err := nodeImageProvider.List(ctx, nodeClass)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(foundImages).To(ContainElements(getExpectedTestSIGImages(*nodeClass.Spec.ImageFamily, sigImageVersion, kubernetesVersion)))
+		})
+
+		//TODO: Modify when Ubuntu 22.04 with FIPS becomes available
+		It("should match expected images for FIPS Ubuntu2204 when FIPSMode is explicitly set to FIPS", func() {
+			nodeClass.Spec.ImageFamily = lo.ToPtr(v1beta1.Ubuntu2204ImageFamily)
+			nodeClass.Spec.FIPSMode = lo.ToPtr("FIPS")
```
charliedmcb: I think it might be worth adding a new `Context` in here for the FIPS tests, at least for the ones where FIPS is enabled.

If you do it just for the enabled cases you could do a `BeforeEach` which contains the `nodeClass.Spec.FIPSMode = lo.ToPtr("FIPS")`

If you do it more generically, I still think it could be nice to grouping of like tests, since its testing the sub cases of SIG images surrounding FIPS.

pkg/providers/imagefamily/nodeimage_test.go
```diff
@@ -171,39 +200,101 @@ var _ = Describe("NodeImageProvider tests", func() {
...
 			nodeClass.Spec.ImageFamily = lo.ToPtr(v1beta1.AzureLinuxImageFamily)
-			nodeClass.Status.KubernetesVersion = "1.31.0"
 
 			foundImages, err := nodeImageProvider.List(ctx, nodeClass)
 			Expect(err).ToNot(HaveOccurred())
-			Expect(foundImages).To(ContainElements(getExpectedTestSIGImages(*nodeClass.Spec.ImageFamily, sigImageVersion, "1.31.0")))
+			Expect(foundImages).To(ContainElements(getExpectedTestSIGImages(*nodeClass.Spec.ImageFamily, sigImageVersion, kubernetesVersion)))
 		})
 
-		It("should match expected images for AzureLinux with version >= 1.32", func() {
```
charliedmcb, line 199: Rather than modifying these tests, I would add new tests in addition to the existing ones.

Goes for the existing:
- `"should match expected images for Ubuntu2204"`
- `"should match expected images for AzureLinux, depending on the Kubernetes version"`
- `"should match expected images for AzureLinux with version < 1.32"`

pkg/providers/imagefamily/nodeimage_test.go
```diff
@@ -171,39 +200,101 @@ var _ = Describe("NodeImageProvider tests", func() {
...
 			foundImages, err := nodeImageProvider.List(ctx, nodeClass)
 			Expect(err).ToNot(HaveOccurred())
-			Expect(foundImages).To(ContainElements(getExpectedTestSIGImages(*nodeClass.Spec.ImageFamily, sigImageVersion, "1.32.0")))
+			Expect(foundImages).To(ContainElements(getExpectedTestFIPSImages(*nodeClass.Spec.ImageFamily, sigImageVersion, kubernetesVersion)))
 		})
+
+		DescribeTable("should match expected images",
+			func(imageFamily *string, fipsMode *string, version string, kubernetesVersion string) {
+				nodeClass.Spec.ImageFamily = imageFamily
+				effectiveFipsMode := ""
```
charliedmcb: I think it would be a bit simpler if you just have `fipsMode` be a string, instead of a `*string`.

Than you don't need `effectiveFipsMode`, and can do your if check on `fipsMode != nil` as `fipsMode != ""` instead.

charliedmcb (commented):
Initial review of just the testing file.

pkg/providers/imagefamily/azlinux.go
```diff
@@ -83,6 +85,34 @@ func (u AzureLinux) DefaultImages() []types.DefaultImageOutput {
...
+		{
+			PublicGalleryURL:     AKSAzureLinuxPublicGalleryURL,
+			GalleryResourceGroup: AKSAzureLinuxResourceGroup,
+			GalleryName:          AKSAzureLinuxGalleryName,
+			ImageDefinition:      AzureLinux2Gen2FIPSImageDefinition,
+			Requirements: scheduling.NewRequirements(
+				scheduling.NewRequirement(v1.LabelArchStable, v1.NodeSelectorOpIn, karpv1.ArchitectureAmd64),
+				scheduling.NewRequirement(v1beta1.LabelSKUHyperVGeneration, v1.NodeSelectorOpIn, v1beta1.HyperVGenerationV2),
+			),
+			Distro: "aks-azurelinux-v2-gen2-fips",
```
charliedmcb: Double checked all the FIPS distros looks good:
[AgentBaker/pkg/agent/datamodel/types.go at 7d14215035b51860de586510afd396f19d7005be · Azure/AgentBaker](https://github.com/Azure/AgentBaker/blob/7d14215035b51860de586510afd396f19d7005be/pkg/agent/datamodel/types.go#L145C1-L199C81)

pkg/providers/imagefamily/azlinux.go
```diff
@@ -31,9 +31,11 @@ import (
...
 
 const (
-	AzureLinuxGen2ImageDefinition    = "V2gen2"
-	AzureLinuxGen1ImageDefinition    = "V2"
-	AzureLinuxGen2ArmImageDefinition = "V2gen2arm64"
+	AzureLinuxGen2ImageDefinition      = "V2gen2"
+	AzureLinuxGen1ImageDefinition      = "V2"
+	AzureLinuxGen2ArmImageDefinition   = "V2gen2arm64"
+	AzureLinux2Gen2FIPSImageDefinition = "V2gen2fips"
+	AzureLinux2Gen1FIPSImageDefinition = "V2fips"
```
charliedmcb, line 39: Checked all the images vs the ListNodeImageVersions API results, and they look good:
https://github.com/Azure/karpenter-provider-azure/blob/deb71b4a03885a5294912a39fbcac0e9956150d4/pkg/fake/nodeimageversionsapi.go#L35

pkg/providers/imagefamily/nodeimage_test.go
```diff
@@ -68,15 +68,32 @@ func getExpectedTestCIGImages(imageFamily string, version string, kubernetesVers
...
+func getExpectedTestSIGImages(imageFamily string, fipsMode string, version string, kubernetesVersion string) []imagefamily.NodeImage {
 	var images []imagefamilytypes.DefaultImageOutput
-	if imageFamily == v1beta1.Ubuntu2204ImageFamily {
-		images = imagefamily.Ubuntu2204{}.DefaultImages()
-	} else if imageFamily == v1beta1.AzureLinuxImageFamily {
-		if imagefamily.UseAzureLinux3(kubernetesVersion) {
-			images = imagefamily.AzureLinux3{}.DefaultImages()
-		} else {
-			images = imagefamily.AzureLinux{}.DefaultImages()
+	if fipsMode == v1beta1.FIPSEnabled {
```
charliedmcb: nit: I think you could refactor this a bit, to have the check for `fipsMode` happen at the end,a nd have the internal if cases give you the image family, not the image list, and then simply call `FIPSImages` or `DefaultImages` afterwards based on fipsMode.

Beyond that you'd just need to have a check for `fipsMode` inside the `UbuntuImageFamily` branch for selecting the image family.

pkg/providers/imagefamily/ubuntu_2004.go
```diff
@@ -0,0 +1,130 @@
...
+func (u Ubuntu2004) Name() string {
+	return v1beta1.UbuntuImageFamily
+}
+
+func (u Ubuntu2004) DefaultImages() []types.DefaultImageOutput {
+	// Ubuntu2004 doesn't have default node images (only FIPS)
+	return []types.DefaultImageOutput{}
+}
+
+func (u Ubuntu2004) FIPSImages() []types.DefaultImageOutput {
```
charliedmcb: I'd add a note to the `FIPSImage` functions that they aren't supported in public galleries. However, I think its fine to leave the public gallery info there.

charliedmcb (commented):
will do another pass over the tests, but reviewed non-test files, and thing are looking good.

pkg/providers/imagefamily/nodeimage_test.go
```diff
@@ -171,39 +183,126 @@ var _ = Describe("NodeImageProvider tests", func() {
...
-		It("should match expected images for AzureLinux with version < 1.32", func() {
-			nodeClass.Spec.ImageFamily = lo.ToPtr(v1beta1.AzureLinuxImageFamily)
-			nodeClass.Status.KubernetesVersion = "1.31.0"
+		// current behavior for not setting FIPSMode is effectively setting it to Disabled
+		Context("List Default Images When FIPSMode Is Not Explicitly Set", func() {
+			BeforeEach(func() {
+				nodeClass.Spec.FIPSMode = nil
+			})
+
+			It("should match expected images for default Ubuntu, effectively Ubuntu2204", func() {
```
charliedmcb: nit: I'd drop the "default" and the reference to mapping to "Ubuntu2204" since we might change this over time, and wouldn't necessarily update the test naming:
```suggestion
			It("should match expected images for generic Ubuntu", func() {
```

pkg/providers/imagefamily/nodeimage_test.go
```diff
@@ -171,39 +183,126 @@ var _ = Describe("NodeImageProvider tests", func() {
...
+				nodeClass.Spec.ImageFamily = lo.ToPtr(v1beta1.UbuntuImageFamily)
+
+				foundImages, err := nodeImageProvider.List(ctx, nodeClass)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(foundImages).To(ContainElements(getExpectedTestSIGImages(*nodeClass.Spec.ImageFamily, v1beta1.FIPSUnspecified, sigImageVersion, kubernetesVersion)))
+			})
+
+			It("should match expected images for default Ubuntu2204", func() {
```
charliedmcb: nit: drop `default`
```suggestion
			It("should match expected images for Ubuntu2204", func() {
```

pkg/providers/imagefamily/nodeimage_test.go
```diff
@@ -171,39 +183,126 @@ var _ = Describe("NodeImageProvider tests", func() {
...
+		Context("List FIPS Images When FIPSMode Is Explicitly FIPS", func() {
+			BeforeEach(func() {
+				nodeClass.Spec.FIPSMode = lo.ToPtr(string(v1beta1.FIPSEnabled))
+			})
 
-		// This test changes depending on the Kubernetes version, in effect making the following version-specific tests unnecessary.
-		// They are still kept for clarity and to ensure that the behavior is explicitly tested.
-		It("should match expected images for AzureLinux, depending on the Kubernetes version", func() {
-			nodeClass.Spec.ImageFamily = lo.ToPtr(v1beta1.AzureLinuxImageFamily)
+			It("should match expected images for FIPS Ubuntu2004", func() {
```
charliedmcb: nit: naming I'd adjust as the selection of `Ubuntu2004` happens behind the scenes and might shift in the future.
```suggestion
			It("should match expected images for FIPS when using generic Ubuntu", func() {
```

pkg/providers/imagefamily/nodeimage_test.go
```diff
@@ -171,39 +183,126 @@ var _ = Describe("NodeImageProvider tests", func() {
...
-			nodeClass.Status.KubernetesVersion = "1.32.0"
+		Context("List Default Images When FIPSMode Is Explicitly Disabled", func() {
+			BeforeEach(func() {
+				nodeClass.Spec.FIPSMode = lo.ToPtr(string(v1beta1.FIPSDisabled))
+			})
 
-			foundImages, err := nodeImageProvider.List(ctx, nodeClass)
-			Expect(err).ToNot(HaveOccurred())
-			Expect(foundImages).To(ContainElements(getExpectedTestSIGImages(*nodeClass.Spec.ImageFamily, sigImageVersion, "1.32.0")))
+			It("should match expected images for default Ubuntu, effectively Ubuntu2204", func() {
```
charliedmcb: nit naming:
```suggestion
			It("should match expected images for generic Ubuntu", func() {
```

pkg/providers/imagefamily/nodeimage_test.go
```diff
@@ -171,39 +183,126 @@ var _ = Describe("NodeImageProvider tests", func() {
...
+				nodeClass.Spec.ImageFamily = lo.ToPtr(v1beta1.UbuntuImageFamily)
+
+				foundImages, err := nodeImageProvider.List(ctx, nodeClass)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(foundImages).To(ContainElements(getExpectedTestSIGImages(*nodeClass.Spec.ImageFamily, v1beta1.FIPSUnspecified, sigImageVersion, kubernetesVersion)))
```
charliedmcb: missed one update:
```suggestion
				Expect(foundImages).To(ContainElements(getExpectedTestSIGImages(*nodeClass.Spec.ImageFamily, v1beta1.FIPSDisabled, sigImageVersion, kubernetesVersion)))
```

pkg/providers/imagefamily/nodeimage_test.go
```diff
@@ -171,39 +183,126 @@ var _ = Describe("NodeImageProvider tests", func() {
...
+			})
+
+			It("should match expected images for default Ubuntu2204", func() {
```
charliedmcb: nit naming:
```suggestion
			It("should match expected images for Ubuntu2204", func() {
```

charliedmcb (commented):
Reviewed tests.

tests are looking good to me. 

Mostly nits on naming, and one location missed an update.

pkg/providers/imagefamily/azlinux3.go
```diff
@@ -111,6 +112,17 @@ func (u AzureLinux3) FIPSImages() []types.DefaultImageOutput {
 			),
 			Distro: "aks-azurelinux-v3-fips",
 		},
+		{
+			PublicGalleryURL:     AKSAzureLinuxPublicGalleryURL,
+			GalleryResourceGroup: AKSAzureLinuxResourceGroup,
+			GalleryName:          AKSAzureLinuxGalleryName,
+			ImageDefinition:      AzureLinux3Gen2Arm64FIPSImageDefinition,
+			Requirements: scheduling.NewRequirements(
+				scheduling.NewRequirement(v1.LabelArchStable, v1.NodeSelectorOpIn, karpv1.ArchitectureAmd64),
```
charliedmcb: This should be `arm` arch:
```suggestion
				scheduling.NewRequirement(v1.LabelArchStable, v1.NodeSelectorOpIn, karpv1.ArchitectureArm64),
```

pkg/providers/imagefamily/azlinux3.go
```diff
@@ -111,6 +112,17 @@ func (u AzureLinux3) FIPSImages() []types.DefaultImageOutput {
...
+				scheduling.NewRequirement(v1beta1.LabelSKUHyperVGeneration, v1.NodeSelectorOpIn, v1beta1.HyperVGenerationV2),
+			),
+			Distro: "aks-azurelinux-v3-arm64-gen2-fips",
```
charliedmcb: Distro looks good:
https://github.com/Azure/AgentBaker/blob/4f7102b0315c3ce0643c3f4db76c4685037ff4fe/pkg/agent/datamodel/types.go#L196

pkg/providers/imagefamily/resolver.go
```diff
@@ -197,9 +200,11 @@ func (r *defaultResolver) getStorageProfile(ctx context.Context, instanceType *c
 func mapToImageDistro(imageID string, imageFamily ImageFamily) (string, error) {
 	var imageInfo types.DefaultImageOutput
 	imageInfo.PopulateImageTraitsFromID(imageID)
-	for _, defaultImage := range imageFamily.DefaultImages() {
-		if defaultImage.ImageDefinition == imageInfo.ImageDefinition {
-			return defaultImage.Distro, nil
+	// checks both FIPS and default images
```
lnguyen924: I don't _think_ appending `FIPSImages()` and `DefaultImages()` will be an issue (?)

mapToImageDistro

pkg/controllers/nodeclass/status/images.go
```diff
@@ -112,6 +113,16 @@ func (r *NodeImageReconciler) Reconcile(ctx context.Context, nodeClass *v1beta1.
 	ctx = log.IntoContext(ctx, log.FromContext(ctx).WithName(nodeImageReconcilerName))
 	logger := log.FromContext(ctx)
 
+	// validate FIPS + useSIG
+	fipsMode := nodeClass.Spec.FIPSMode
+	useSIG := options.FromContext(ctx).UseSIG
+	if fipsMode != nil && *fipsMode == v1beta1.FIPSEnabled && !useSIG {
+		nodeClass.Status.Images = nil
+		nodeClass.StatusConditions().SetFalse(v1beta1.ConditionTypeImagesReady, "SIGRequiredForFIPS", "FIPS images require UseSIG to be enabled, but UseSIG is false")
```
charliedmcb: Not recommending any update, here, but considering if it makes sense to mention being unsupported for `self-hosted` here. However, that's implied logic, not direct.

Curious other reviewers opinions.

pkg/providers/imagefamily/nodeimage_test.go
charliedmcb, line 1: Lets fix the other cases of `ContainElements` in this file, while we're at it.

pkg/controllers/nodeclass/status/images.go
```diff
@@ -112,6 +113,16 @@ func (r *NodeImageReconciler) Reconcile(ctx context.Context, nodeClass *v1beta1.
 	ctx = log.IntoContext(ctx, log.FromContext(ctx).WithName(nodeImageReconcilerName))
 	logger := log.FromContext(ctx)
 
+	// validate FIPS + useSIG
+	fipsMode := nodeClass.Spec.FIPSMode
+	useSIG := options.FromContext(ctx).UseSIG
+	if fipsMode != nil && *fipsMode == v1beta1.FIPSEnabled && !useSIG {
```
matthchr: minor, style: I like to combine these nil-coalescing type checks with something like `lo.FromPtr(fipsMode)` because then you can just do:
```
if lo.FromPtr(fipsMode) == v1beta1.FIPSEnabled && !useSIG
```

pkg/controllers/nodeclass/status/images.go
```diff
@@ -112,6 +113,16 @@ func (r *NodeImageReconciler) Reconcile(ctx context.Context, nodeClass *v1beta1.
...
+		nodeClass.Status.Images = nil
+		nodeClass.StatusConditions().SetFalse(v1beta1.ConditionTypeImagesReady, "SIGRequiredForFIPS", "FIPS images require UseSIG to be enabled, but UseSIG is false")
+		logger.Info("FIPS images require SIG")
+		return reconcile.Result{RequeueAfter: 5 * time.Minute}, nil
```
matthchr: What's the argument for returning `reconcile.Result{RequeueAfter: 5 * time.Minute}, nil` here versus an empty result and an error?

pkg/controllers/nodeclass/status/images_test.go
```diff
@@ -166,6 +168,38 @@ var _ = Describe("NodeClass NodeImage Status Controller", func() {
...
+				nodeClass.Spec.FIPSMode = &fipsMode
+
+				// set ImageFamily to AzureLinux (to bypass unsupported FIPS on the default Ubuntu2204)
+				imageFamily := v1beta1.AzureLinuxImageFamily
+				nodeClass.Spec.ImageFamily = &imageFamily
+
+				_, err := imageReconciler.Reconcile(ctx, nodeClass)
+				Expect(err).ToNot(HaveOccurred())
+
+				condition := nodeClass.StatusConditions().Get(v1beta1.ConditionTypeImagesReady)
```
matthchr, line 197: The root ready condition should also be false, yes?

pkg/providers/imagefamily/resolver.go
```diff
@@ -197,9 +200,11 @@ func (r *defaultResolver) getStorageProfile(ctx context.Context, instanceType *c
...
+	allImages := append(imageFamily.DefaultImages(), imageFamily.FIPSImages()...)
```
matthchr: Are there conflicts with main here? In main I see that `DefaultImages` takes a `sig` parameter, so for the SIG support we didn't add another function that all providers have to implement instead we added a parameter and they dynamically include the SIG-only variants.

What's the argument for adding methods to the interface versus just adding a parameter to the existing interface method and then returning the results that way?

It seems to me that's cleaner because it:
1. Doesn't require implementers to change their interface impls quite as much:
2. It scales better if/when we have more filter parameters.

I can _sortof_ see the argument that if you need FIPS to always go "first" you can enforce it easier this way (versus providers that may put it last), but there are ways you could solve that problem if you really had to. In this case, I am not sure it's a problem because the ImageDefinition is always different for the FIPS images anyway so order doesn't actually matter? If order did matter though, the semantics could be that the parameter was a filter, so specifying FIPS gets you only fips images, or leaving it blank gets you the normal images (non fips) and then you call the one method twice: `DefaultImages(fips=false)` and `DefaultImages(fips=true)` - which results in the same code you have here but without needing a proliferation of methods on the providers?

matthchr: tl;dr: Why not just add a filter parameter to the method?

pkg/providers/imagefamily/resolver.go
```diff
@@ -221,14 +226,22 @@ func prepareKubeletConfiguration(ctx context.Context, instanceType *cloudprovide
 	return kubeletConfig
 }
 
-func getSupportedImages(familyName *string, kubernetesVersion string) []types.DefaultImageOutput {
+func getSupportedImages(familyName *string, fipsMode *string, kubernetesVersion string) []types.DefaultImageOutput {
 	// TODO: Options aren't used within DefaultImages, so safe to be using nil here. Refactor so we don't actually need to pass in Options for getting DefaultImage.
-	imageFamily := getImageFamily(familyName, kubernetesVersion, nil)
+	imageFamily := getImageFamily(familyName, fipsMode, kubernetesVersion, nil)
+	if lo.FromPtr(fipsMode) == v1beta1.FIPSEnabled {
```
matthchr: "just add a filter parameter" may fit better here too because you don't need to do the filtering here you delegate it to the individual image family provider:
```
return imageFamily.DefaultImages(fipsMode)
```

pkg/providers/imagefamily/ubuntu_2004.go
```diff
@@ -0,0 +1,131 @@
...
+		APIServerName:                  u.Options.APIServerName,
+		KubeletClientTLSBootstrapToken: u.Options.KubeletClientTLSBootstrapToken,
+		NetworkPlugin:                  u.Options.NetworkPlugin,
+		NetworkPolicy:                  u.Options.NetworkPolicy,
+		KubernetesVersion:              u.Options.KubernetesVersion,
+	}
+}
+
+// UserData returns the default userdata script for the image Family
+func (u Ubuntu2004) CustomScriptsNodeBootstrapping(kubeletConfig *bootstrap.KubeletConfiguration, taints []v1.Taint, startupTaints []v1.Taint, labels map[string]string, instanceType *cloudprovider.InstanceType, imageDistro string, storageProfile string, nodeBootstrappingClient types.NodeBootstrappingAPI) customscriptsbootstrap.Bootstrapper {
```
matthchr: minor, style: Long line, split across multiple lines (1 param per line)

pkg/providers/imagefamily/ubuntu_2004.go
```diff
@@ -0,0 +1,131 @@
...
+	return v1beta1.UbuntuImageFamily
+}
+
+func (u Ubuntu2004) DefaultImages() []types.DefaultImageOutput {
+	// Ubuntu2004 doesn't have default node images (only FIPS)
+	return []types.DefaultImageOutput{}
+}
+
+// FIPS images aren't supported in public galleries, only shared image galleries
+func (u Ubuntu2004) FIPSImages() []types.DefaultImageOutput {
```
matthchr: When you merge with main and the `DefaultImages` has a `sig` parameter, if you take the "do it all with one method" approach you can even have defense-in-depth here where you say:
```
if !fips {
   // return nothing
}

if !useSig {
    // return nothing
}

// return the fips images
```

pkg/providers/imagefamily/nodeimage.go
```diff
@@ -74,14 +74,21 @@ func NewProvider(versionsClient types.CommunityGalleryImageVersionsAPI, location
 
 // Returns the list of available NodeImages for the given AKSNodeClass sorted in priority ordering
 func (p *provider) List(ctx context.Context, nodeClass *v1beta1.AKSNodeClass) ([]NodeImage, error) {
+	// TODO: refactor to be part of construction, since this is a karpenter setting and won't change across the process.
+	useSIG := options.FromContext(ctx).UseSIG
+
+	// CIG has no FIPS images; FIPS images can only be accessed through SIG
+	// (this won't be an error since there just aren't any FIPS images for CIG)
+	if nodeClass.Spec.FIPSMode != nil && *nodeClass.Spec.FIPSMode == v1beta1.FIPSEnabled && !useSIG {
```
matthchr: `lo.FromPtr` trick again.

pkg/providers/imagefamily/nodeimage_test.go
```diff
@@ -171,60 +184,159 @@ var _ = Describe("NodeImageProvider tests", func() {
...
+			BeforeEach(func() {
+				nodeClass.Spec.FIPSMode = lo.ToPtr(v1beta1.FIPSEnabled)
+			})
+
+			It("should match expected images for FIPS when using generic Ubuntu", func() {
+				nodeClass.Spec.ImageFamily = lo.ToPtr(v1beta1.UbuntuImageFamily)
+				foundImages, err := nodeImageProvider.List(ctx, nodeClass)
+				Expect(err).ToNot(HaveOccurred())
+				expectedImages := getExpectedTestSIGImages(*nodeClass.Spec.ImageFamily, v1beta1.FIPSEnabled, sigImageVersion, kubernetesVersion)
+				// equal functions as a checker for both content (including length) and order of image sets
```
matthchr: minor, style: I don't think we need a comment for what Equal function does - IMO it's implied with two lists that the Equal function asserts the lists must be exactly equal (element-wise equality, incl length)

matthchr: Applies throughout.

pkg/controllers/nodeclass/status/images.go
charliedmcb: Chatted with Matthew about this offline.

We think what would be best here, is to return no result, and no error, so:
`return reconcile.Result{}, nil`

However, lets also update the logger line to make it clear we've experienced an error:
`logger.Info("FIPS images require SIG", "error", fmt.Errorf("FIPS images require UseSIG to be enabled, but UseSIG is false"))`

charliedmcb: Chatted with Matthew about this offline.

We think what would be best here, is to return no result, and no error, so:
`return reconcile.Result{}, nil`

However, lets also update the logger line to make it clear we've experienced an error:
`logger.Info("FIPS images require SIG", "error", fmt.Errorf("FIPS images require UseSIG to be enabled, but UseSIG is false"))`

pkg/providers/imagefamily/resolver.go
charliedmcb: I had initially suggested the seperate FIPSImages function. This was mostly based on surface area touch, and not doing external filtering, but as the DefaultImages is now accepting params, I think you're right that adding the fipsMode as a param makes sense.

pkg/controllers/nodeclass/status/images.go
charliedmcb: Discussed with Matthew and Alex. Let's add some context for where use is allowed.

```suggestion
		nodeClass.StatusConditions().SetFalse(v1beta1.ConditionTypeImagesReady, "SIGRequiredForFIPS", "FIPS images require UseSIG to be enabled, but UseSIG is false (note: UseSIG is only supported in AKS managed NAP)")
```

pkg/controllers/nodeclass/status/images_test.go
```diff
@@ -166,6 +169,40 @@ var _ = Describe("NodeClass NodeImage Status Controller", func() {
...
+
+				// set ImageFamily to AzureLinux (to bypass unsupported FIPS on the default Ubuntu2204)
+				imageFamily := v1beta1.AzureLinuxImageFamily
+				nodeClass.Spec.ImageFamily = &imageFamily
+
+				_, err := imageReconciler.Reconcile(ctx, nodeClass)
+				Expect(err).To(HaveOccurred())
+				Expect(err.Error()).To(Equal("FIPS images require UseSIG to be enabled, but UseSIG is false"))
+
+				condition := nodeClass.StatusConditions().Get(v1beta1.ConditionTypeImagesReady)
```
charliedmcb, line 197: Add an explicit check for:
`Expect(condition.IsFalse()).To(BeTrue())`

pkg/providers/imagefamily/azlinux.go
```diff
@@ -44,7 +47,36 @@ func (u AzureLinux) Name() string {
 	return v1beta1.AzureLinuxImageFamily
 }
 
-func (u AzureLinux) DefaultImages() []types.DefaultImageOutput {
+func (u AzureLinux) DefaultImages(fipsMode *v1beta1.FIPSMode) []types.DefaultImageOutput {
+	if lo.FromPtr(fipsMode) == v1beta1.FIPSModeFIPS {
+		// FIPS images aren't supported in public galleries, only shared image galleries
```
charliedmcb: hyper-nit.
```suggestion
		// Note: FIPS images aren't supported in public galleries, only shared image galleries
```

charliedmcb: Same with other locations.

charliedmcb (commented):
Assuming comments are taken care of, lgtm.

Probably need to re-run `make presubmit` though

pkg/controllers/nodeclass/status/images.go
charliedmcb: Also, update the log message

pkg/controllers/nodeclass/status/images.go
```diff
@@ -112,6 +113,16 @@ func (r *NodeImageReconciler) Reconcile(ctx context.Context, nodeClass *v1beta1.
 	ctx = log.IntoContext(ctx, log.FromContext(ctx).WithName(nodeImageReconcilerName))
 	logger := log.FromContext(ctx)
 
+	// validate FIPS + useSIG
+	fipsMode := nodeClass.Spec.FIPSMode
+	useSIG := options.FromContext(ctx).UseSIG
+	if lo.FromPtr(fipsMode) == v1beta1.FIPSModeFIPS && !useSIG {
+		nodeClass.Status.Images = nil
+		nodeClass.StatusConditions().SetFalse(v1beta1.ConditionTypeImagesReady, "SIGRequiredForFIPS", "FIPS images require UseSIG to be enabled, but UseSIG is false (note: UseSIG is only supported in AKS managed NAP)")
+		logger.Info("FIPS images require SIG", "error", fmt.Errorf("FIPS images require UseSIG to be enabled, but UseSIG is false"))
```
charliedmcb: nit: I'd update the error message with the same note as in the condition:
```suggestion
		logger.Info("FIPS images require SIG", "error", fmt.Errorf("FIPS images require UseSIG to be enabled, but UseSIG is false (note: UseSIG is only supported in AKS managed NAP)"))
```

charliedmcb (approved):
lgtm

PR #1031, opened 2025-07-10
feat: flag to enable/disable middleware logging

alimaazamat:
Contributes to #833 

**Description**
Azure SDK clients are interfacs that allow you to make API calls to Azure's REST APIs without the raw HTTP requests you use client objects with methods like `vmClient.Create(ctx, resourceGroup, vmName, vmParameters)`
Every client can have middleware attached that logs the API request, but this sometimes can be too much information for debugging.

1.
edit the _karpenter-values.yaml_ file and set `ENABLE_AZURE_SDK_LOGGING: false` or `true` in the controller environment variables, then run `make az-run` to redeploy.
OR
`export ENABLE_AZURE_SDK_LOGGING=true`/`false` and `make az-run` to redeploy
2.
Run `make az-klogs-pretty`
Shows the following type of logs when flag is true
```
{
  "time": "2025-07-09T22:12:45.804104299Z",
  "level": "INFO",
  "msg": "finished call",
  "source": "ApiRequestLog",
  "protocol": "REST",
  "method_type": "unary",
  "component": "client",
  "time_ms": 255,
  "method": "POST https://management.azure.com/providers/Microsoft.ResourceGraph/resources",
  "service": "management.azure.com",
  "url": "https://management.azure.com/providers/Microsoft.ResourceGraph/resources?api-version=2021-06-01-preview",
  "headers": {
    "x-ms-correlation-request-id": "c33202cf-77b6-431a-80bd-6f124ebdb497"
  },
  "error": "na",
  "code": 200
}
```
*

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/Makefile-az.mk
+++ b/Makefile-az.mk
@@ -1,5 +1,6 @@
 AZURE_LOCATION ?= westus2
 COMMON_NAME ?= karpenter
+ENABLE_AZURE_SDK_LOGGING ?= false
 ifeq ($(CODESPACES),true)
   AZURE_RESOURCE_GROUP ?= $(CODESPACE_NAME)
   AZURE_ACR_NAME ?= $(subst -,,$(CODESPACE_NAME))
@@ -139,6 +140,10 @@ az-rmrg: ## Destroy test ACR and AKS cluster by deleting the resource group (use
 az-configure-values:  ## Generate cluster-related values for Karpenter Helm chart
 	hack/deploy/configure-values.sh $(AZURE_CLUSTER_NAME) $(AZURE_RESOURCE_GROUP) $(KARPENTER_SERVICE_ACCOUNT_NAME) $(AZURE_KARPENTER_USER_ASSIGNED_IDENTITY_NAME)
 
+az-update-logging-flag: ## Update ENABLE_AZURE_SDK_LOGGING flag in karpenter-values.yaml based on environment variable
+	@echo "Setting ENABLE_AZURE_SDK_LOGGING to $(ENABLE_AZURE_SDK_LOGGING)"
+	yq -i '.controller.env[] |= select(.name == "ENABLE_AZURE_SDK_LOGGING").value = "$(ENABLE_AZURE_SDK_LOGGING)"' karpenter-values.yaml
+
 az-mkvmssflex: ## Create VMSS Flex (optional, only if creating VMs referencing this VMSS)
 	az vmss create --name $(AZURE_CLUSTER_NAME)-vmss --resource-group $(AZURE_RESOURCE_GROUP_MC) --location $(AZURE_LOCATION) \
 		--instance-count 0 --orchestration-mode Flexible --platform-fault-domain-count 1 --zones 1 2 3
@@ -189,7 +194,7 @@ az-build: ## Build the Karpenter controller and webhook images using skaffold bu
 az-creds: ## Get cluster credentials
 	az aks get-credentials --name $(AZURE_CLUSTER_NAME) --resource-group $(AZURE_RESOURCE_GROUP)
 
-az-run: ## Deploy the controller from the current state of your git repository into your ~/.kube/config cluster using skaffold run
+az-run: az-update-logging-flag ## Deploy the controller from the current state of your git repository into your ~/.kube/config cluster using skaffold run
 	az acr login -n $(AZURE_ACR_NAME)
 	skaffold run
 
@@ -403,7 +408,7 @@ az-pprof: ## Profile
 az-mkaks-user: az-mkrg ## Create compatible AKS cluster, the way we tell users to
 	hack/deploy/create-cluster.sh $(AZURE_CLUSTER_NAME) $(AZURE_RESOURCE_GROUP) "${KARPENTER_NAMESPACE}"
 
-az-helm-install-snapshot: az-configure-values ## Install Karpenter snapshot release
+az-helm-install-snapshot: az-configure-values az-update-logging-flag ## Install Karpenter snapshot release
 	$(eval SNAPSHOT_VERSION ?= $(shell git rev-parse HEAD)) # guess which, specify explicitly with SNAPSHOT_VERSION=...
 	helm upgrade --install karpenter oci://ksnap.azurecr.io/karpenter/snapshot/karpenter \
 		--version 0-$(SNAPSHOT_VERSION) \
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -102,7 +102,7 @@ func NewOperator(ctx context.Context, operator *operator.Operator) (context.Cont
 	azClient, err := instance.CreateAZClient(ctx, azConfig, cred)
 	lo.Must0(err, "creating Azure client")
 	if options.FromContext(ctx).VnetGUID == "" && options.FromContext(ctx).NetworkPluginMode == consts.NetworkPluginModeOverlay {
-		vnetGUID, err := getVnetGUID(cred, azConfig, options.FromContext(ctx).SubnetID)
+		vnetGUID, err := getVnetGUID(ctx, cred, azConfig, options.FromContext(ctx).SubnetID)
 		lo.Must0(err, "getting VNET GUID")
 		options.FromContext(ctx).VnetGUID = vnetGUID
 	}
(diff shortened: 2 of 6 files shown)


alimaazamat:
**Default Flag Value Conversation**: CC @tallaxes 
Different scenarios for when we would want the logging flag to be enabled/disabled by default:
1. NAP (Node Auto Provisioning) - Managed by AKS
    * Logs get processed by Msft's managed infra
    * Extra logging gets funneled to a different place automatically
    * _Enable_ by default for this
2. Self-hosted
    * Users probably want by default clean, minimal logs to start
    * _Disable_ by default for this
3. Development (we can override with skaffold.yaml)
    * Developers probably want by default clean, minimal logs to start
    * _Disable_ by default for this
    
Maybe for NAP we have a seperate karpenter values file that has enabled by default? and then otherwise we have disabled by default.

Or maybe we have a logic flow where we first check explicit setting -> check if USE_SIG = true then we are NAP then enable-> otherwise default false for development and self-hosted.

rakechill:
> Maybe for NAP we have a seperate karpenter values file that has enabled by default? and then otherwise we have disabled by default.
> 
> Or maybe we have a logic flow where we first check explicit setting -> check if USE_SIG = true then we are NAP then enable-> otherwise default false for development and self-hosted.

_I think your breakdown makes sense. We will need to be intentional on NAP side of things to make sure we enable by default. This can be added as part of the release it goes out in. Maybe update the release note to explicitly call out that it should be enabled by default for NAP?_

**edit: After further review, I think the release note is good as is. We should instead update the flag in `option.go` to be true by default to ensure NAP isn't required to add this new flag/env var to its deployment.**

Makefile-az.mk
```diff
@@ -141,6 +142,10 @@ az-rmrg: ## Destroy test ACR and AKS cluster by deleting the resource group (use
 az-configure-values:  ## Generate cluster-related values for Karpenter Helm chart
```
rakechill: Instead of creating a separate make command + having to update az-run + helm snapshot explicitly, can we instead update `az-configure-values` to also set this value? The hack script is already in charge of updating karpenter-values.yaml + this makes sense to extend it to also handle this case.

pkg/operator/operator.go
```diff
@@ -242,7 +242,8 @@ func getVnetGUID(creds azcore.TokenCredential, cfg *auth.Config, subnetID string
 		return "", err
 	}
 
-	opts := armopts.DefaultARMOpts(env.Cloud)
+	o := options.FromContext(ctx)
+	opts := armopts.DefaultARMOpts(env.Cloud, o.EnableAzureSDKLogging)
```
rakechill, line 246: This _should_ be compatible with our work to support non-public clouds, but it might be good to get feedback from @matthchr to be sure. He made a lot of updates to how our clients work, so not sure if there are any considerations there.

rakechill: nvm I see in armopts.go that you're just putting existing functionality behind a flag rather than adding something brand new. I don't think there are compatibility issues in that case.

pkg/operator/options/options.go
```diff
@@ -121,6 +122,7 @@ func (o *Options) AddFlags(fs *coreoptions.FlagSet) {
 	}
 	// See https://github.com/Azure/karpenter-provider-azure/issues/1042 for issue discussing improvements around this
 	fs.Var(additionalTagsFlag, "additional-tags", "Additional tags to apply to the resources in Azure. Format is key1=value1,key2=value2. These tags will be merged with the tags specified on the NodePool. In the case of a tag collision, the NodePool tag wins. These tags only apply to new nodes and do not trigger drift, which means that adding tags to this collection will not update existing nodes until drift triggers for some other reason.")
+	fs.BoolVar(&o.EnableAzureSDKLogging, "enable-azure-sdk-logging", env.WithDefaultBool("ENABLE_AZURE_SDK_LOGGING", false), "If set to true then Azure SDK middleware logging is enabled for debugging, logging all HTTP requests/responses to Azure APIs.")
```
rakechill: Let's actually approach this the opposite way to preserve our current behavior by default.

In our `karpenter-values-template` and our makefile, we already ensure that for OSS + dev usage, this is disabled by default.

However, if this flag is disabled by default here, then we'll require a NAP deployment change to preserve our current behavior. I think it's safer, then, to set this option to `true` by default if the env var is present so that NAP doesn't need to change to accommodate this.

Makefile-az.mk
alimaazamat: Done, thanks for this!

pkg/operator/options/options.go
alimaazamat: Done, thank you!

Makefile-az.mk
```diff
@@ -138,8 +139,9 @@ az-mkaks-savm: az-mkrg ## Create experimental cluster with standalone VMs (+ ACR
 az-rmrg: ## Destroy test ACR and AKS cluster by deleting the resource group (use with care!)
 	az group delete --name $(AZURE_RESOURCE_GROUP)
 
-az-configure-values:  ## Generate cluster-related values for Karpenter Helm chart
+az-configure-values:  ## Generate cluster-related values for Karpenter Helm chart and set middleware logging flag
 	hack/deploy/configure-values.sh $(AZURE_CLUSTER_NAME) $(AZURE_RESOURCE_GROUP) $(KARPENTER_SERVICE_ACCOUNT_NAME) $(AZURE_KARPENTER_USER_ASSIGNED_IDENTITY_NAME)
+	yq -i '.controller.env[] |= select(.name == "ENABLE_AZURE_SDK_LOGGING").value = "$(ENABLE_AZURE_SDK_LOGGING)"' karpenter-values.yaml
```
rakechill: Actually, I think if you look at the `configure-values.sh` script, it already sets a few variables in the chart. I was thinking we could pass in this env var along with the other args on line 143
$(ENABLE_AZURE_SDK_LOGGING)

And then in `hack/deploy/configure-values.sh`, we could update it to utilize your env var.

* update expected args line here: https://github.com/Azure/karpenter-provider-azure/blob/12c548e9345d1e1d63390b3d4a514536166d4528/hack/deploy/configure-values.sh#L7
* Set the script's env var equal to $5: https://github.com/Azure/karpenter-provider-azure/blob/12c548e9345d1e1d63390b3d4a514536166d4528/hack/deploy/configure-values.sh#L17
* ...

rakechill: And then I _think_ this line will do the copying into yaml for you: https://github.com/Azure/karpenter-provider-azure/blob/12c548e9345d1e1d63390b3d4a514536166d4528/hack/deploy/configure-values.sh#L79

Since the karpenter template yaml already has this env var in its list

alimaazamat: Ah I see that is way better you're right thanks! Then instructions changed so we should run `make az-configure-values` to regenerate `karpenter-values.yaml`

Makefile-az.mk
```diff
@@ -405,7 +406,7 @@ az-pprof: ## Profile
 az-mkaks-user: az-mkrg ## Create compatible AKS cluster, the way we tell users to
 	hack/deploy/create-cluster.sh $(AZURE_CLUSTER_NAME) $(AZURE_RESOURCE_GROUP) "${KARPENTER_NAMESPACE}"
 
-az-helm-install-snapshot: az-configure-values ## Install Karpenter snapshot release
+az-helm-install-snapshot: az-configure-values az-update-logging-flag ## Install Karpenter snapshot release
```
rakechill: still have this `az-update...` command. remove this change.

Makefile-az.mk
```diff
@@ -1,6 +1,7 @@
 AZURE_LOCATION ?= westus2
 AZURE_VM_SIZE ?= ""
 COMMON_NAME ?= karpenter
+ENABLE_AZURE_SDK_LOGGING ?= true
```
rakechill: let's have it be false here + a comment saying `// false to make it easier to debug`

alimaazamat: Wait I thought we said we want the flag enabled by default? So make file targets always going to be for developers?

rakechill: Ah sorry to go back and forth on that. Yes, makefile should still be true since we want the snapshots/releases to use NAP defaults. The karpenter-values-template being false by default is capturing dev/self-hosted expectations (and thus false).

alimaazamat: Ah I see so here is heirarchy:
1. setting env var
2. makefile default (false) which wouldn't be interacted with for NAP case
3. template default (false) which wouldn't be interacted with for NAP case
4. Go code default (true) options.go  which is what NAP relies on

Makefile-az.mk
alimaazamat: Ah thanks for this catch

Makefile-az.mk
rakechill: Sorry, that's why I resolved this comment thread actually. I think it was fine how it was/maybe slightly _better_ even. 

```
az-helm-install-snapshot: az-configure-values ## Install Karpenter snapshot release
```

I have to verify by testing on your branch, but this one also might possibly require it to be `TRUE` by default...
our release scripts are slightly different/might not use this directly, but I can't remember rn if they rely on this `configure-values` make command at all.

alimaazamat: What if for that we did `make az-helm-install-snapshot ENABLE_AZURE_SDK_LOGGING=true` just for that make target?

rakechill: let's just keep it to true by default in this makefile + we can update later if we find we prefer the opposite. main goal will be just to keep it adjustable + try to preserve current behavior

rakechill (approved):
lgtm, thanks!

PR #1038, opened 2025-07-16
fix: support standalone nodeclaims for lro async

Bryce-Soghigian:
Fixes # 

**Description**
Standalone nodeclaims should not follow the delete operation for LRO failures, as there is no nodepool to re-create them and reconcile them back. 
- unit tests
- e2e tests https://github.com/Azure/karpenter-provider-azure/actions/runs/16457200269

- [ ] Yes, issue opened: # 
- [x] No we don't have docs for standalone nodeclaim 

**Release Note**

```release-note

```


pkg/providers/instancetype/suite_test.go
```diff
@@ -541,6 +541,48 @@ var _ = Describe("InstanceType Provider", func() {
...
+			// Verify nodeclaim is preserved and status condition is set
+			Eventually(func(g Gomega) {
+				nodeClaims := ExpectNodeClaims(ctx, env.Client)
+				g.Expect(nodeClaims).To(HaveLen(1))
+				g.Expect(nodeClaims[0].DeletionTimestamp).To(BeNil())
+
+				// Verify the status condition was set properly
+				launchedCondition := nodeClaims[0].StatusConditions().Get(karpv1.ConditionTypeLaunched)
+				g.Expect(launchedCondition).ToNot(BeNil())
+				// ----------------
```
Bryce-Soghigian:
```suggestion
```

pkg/cloudprovider/cloudprovider.go
```diff
@@ -165,47 +165,58 @@ func (c *CloudProvider) Create(ctx context.Context, nodeClaim *karpv1.NodeClaim)
...
-			if err != nil {
-				log.FromContext(ctx).Error(err, "failed to delete nodeclaim, will wait for liveness TTL", "NodeClaim", nodeClaim.Name)
+		if c.isStandaloneNodeClaim(nodeClaim) {
+			// For standalone NodeClaims: preserve the NodeClaim, set status to failed, return error
+			freshNodeClaim := &karpv1.NodeClaim{}
+			if getErr := c.kubeClient.Get(ctx, client.ObjectKeyFromObject(nodeClaim), freshNodeClaim); getErr != nil {
+				log.FromContext(ctx).Error(getErr, "failed to get fresh nodeclaim for status update", "NodeClaim", nodeClaim.Name)
+				return
+			}
+			old := freshNodeClaim.DeepCopy()
```
matthchr: Is your intent here that karpenter core is going to pick this nodeclaim up and re-reconcile it?

Is that actually going to work? Have you tested it? I'm looking at the code and what I see is:
- We will initially return
- Karp code will save result of launchNodeClaim (which is really CloudProvider call) in [the cache](https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/controllers/nodeclaim/lifecycle/launch.go#L73)
- When we come back around, the condition is set which means that the Launch controller won't actually [do anything with it](https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/controllers/nodeclaim/lifecycle/launch.go#L48)

pkg/cloudprovider/cloudprovider.go
```diff
@@ -165,47 +165,58 @@ func (c *CloudProvider) Create(ctx context.Context, nodeClaim *karpv1.NodeClaim)
...
+			log.FromContext(ctx).Error(deleteErr, "failed to delete VM", "vmName", vmName)
 		}
 
-		if err = c.kubeClient.Delete(ctx, nodeClaim); err != nil {
-			err = client.IgnoreNotFound(err)
-			if err != nil {
-				log.FromContext(ctx).Error(err, "failed to delete nodeclaim, will wait for liveness TTL", "NodeClaim", nodeClaim.Name)
+		if c.isStandaloneNodeClaim(nodeClaim) {
+			// For standalone NodeClaims: preserve the NodeClaim, set status to failed, return error
+			freshNodeClaim := &karpv1.NodeClaim{}
```
matthchr: minor: You can probably just re-use `nodeClaim` rather than having a `freshNodeClaim`?

pkg/cloudprovider/cloudprovider.go
Bryce-Soghigian: lol that cache delete was added to core 11 hours ago https://github.com/kubernetes-sigs/karpenter/commit/fa7383179b3757a72e3fe6cbb95505979ba86013

matthchr: but even without that cache delete it won't work as desired because the "if condition not unknown" was still true, and when that happens we just return? You're not setting it to unknown you're setting it to false.

Bryce-Soghigian:
Ran the e2e tests, they all pass except for utilization, which is a known issue. 

I triggered 3 runs just to validate it eliminates the nodeclaim flake, it seems promising so far?

pkg/cloudprovider/cloudprovider.go
```diff
@@ -94,6 +94,48 @@ func New(
 }
 
 // Create a node given the constraints.
+func (c *CloudProvider) handleInstanceCreation(ctx context.Context, instancePromise *instance.VirtualMachinePromise, nodeClaim *karpv1.NodeClaim) error {
+	if c.isStandaloneNodeClaim(nodeClaim) {
+		// For standalone nodeclaims, wait synchronously
```
matthchr: Say why.

matthchr: "We must wait synchronously because..." (or similar)

pkg/cloudprovider/cloudprovider.go
```diff
@@ -94,6 +94,48 @@ func New(
...
+		err := instancePromise.Wait()
+		if err != nil {
+			// Clean up the VM
```
matthchr: This section is missing some metrics that I think we should have, and is possibly duplicate with

```
		c.recorder.Publish(cloudproviderevents.NodeClaimFailedToRegister(nodeClaim, err))
		log.FromContext(ctx).Error(err, "failed launching nodeclaim")

		// TODO: This won't clean up leaked NICs if the VM doesn't exist... intentional?
		vmName := lo.FromPtr(promise.VM.Name)
		err = c.instanceProvider.Delete(ctx, vmName)
		if cloudprovider.IgnoreNodeClaimNotFoundError(err) != nil {
			log.FromContext(ctx).Error(err, "failed to delete VM", "vmName", vmName)
		}
```

which is part of `waitOnPromise`.

Wondering if we should refactor the `waitOnPromise` error handling as well into some sort of:
```
func handleNodeClaimCreationError(err error, ..., removeNodeClaim bool) error
```

and then share that code for both paths?

pkg/cloudprovider/cloudprovider.go
Bryce-Soghigian: added a comment here.

pkg/cloudprovider/cloudprovider.go
Bryce-Soghigian: ok added that, agreeed there is a bit of shared logic there

pkg/cloudprovider/cloudprovider.go
```diff
@@ -232,6 +233,41 @@ func (c *CloudProvider) waitUntilLaunched(ctx context.Context, nodeClaim *karpv1
...
+		}
+		metrics.NodeClaimsDisruptedTotal.Inc(map[string]string{
+			metrics.ReasonLabel:       "async_provisioning",
+			metrics.NodePoolLabel:     nodeClaim.Labels[karpv1.NodePoolLabelKey],
+			metrics.CapacityTypeLabel: nodeClaim.Labels[karpv1.CapacityTypeLabelKey],
+		})
+	}
+
+	// For standalone node claims, return a CreateError
+	if !removeNodeClaim {
```
matthchr, line 265: minor, style: put it in an else?

pkg/cloudprovider/cloudprovider.go
```diff
@@ -232,6 +233,41 @@ func (c *CloudProvider) waitUntilLaunched(ctx context.Context, nodeClaim *karpv1
 	}
 }
 
+// handleNodeClaimCreationError handles common error processing for both standalone and async node claim creation failures
+func (c *CloudProvider) handleNodeClaimCreationError(ctx context.Context, err error, promise *instance.VirtualMachinePromise, nodeClaim *karpv1.NodeClaim, removeNodeClaim bool) error {
+	c.recorder.Publish(cloudproviderevents.NodeClaimFailedToRegister(nodeClaim, err))
```
matthchr, line 238: This is net-new, right?

Are there other events we should be emitting?

Do we have an issue tracking them (or should we?)

matthchr (approved):
LGTM, but CI is failing.

pkg/cloudprovider/cloudprovider.go
Bryce-Soghigian: its always existed in cloudproviderevents, but maybe other places we should be omitting them.  Probably worth creating an issue for it

MaximilianoUribe (approved):
LGTM

PR #1048, opened 2025-07-23
fix: temp drop support for AzLinux3 arm64 in CIG

tallaxes:
Fixes # 

**Description**

* `make presubmit`
* E2E (pending)

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/providers/imagefamily/nodeimage.go
+++ b/pkg/providers/imagefamily/nodeimage.go
@@ -79,9 +79,9 @@ func (p *provider) List(ctx context.Context, nodeClass *v1beta1.AKSNodeClass) ([
 		return []NodeImage{}, err
 	}
 
-	supportedImages := getSupportedImages(nodeClass.Spec.ImageFamily, kubernetesVersion)
 	// TODO: refactor to be part of construction, since this is a karpenter setting and won't change across the process.
 	useSIG := options.FromContext(ctx).UseSIG
+	supportedImages := getSupportedImages(nodeClass.Spec.ImageFamily, kubernetesVersion, useSIG)
 
 	key, err := p.cacheKey(
 		supportedImages,
--- a/test/suites/utilization/suite_test.go
+++ b/test/suites/utilization/suite_test.go
@@ -29,6 +29,7 @@ import (
 	"sigs.k8s.io/karpenter/pkg/test"
 
 	"github.com/Azure/karpenter-provider-azure/pkg/apis/v1beta1"
+	"github.com/Azure/karpenter-provider-azure/pkg/providers/imagefamily"
 	"github.com/Azure/karpenter-provider-azure/test/pkg/environment/azure"
 )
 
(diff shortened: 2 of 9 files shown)


pkg/providers/imagefamily/nodeimage.go
charliedmcb, line 1: If this is just a temporary fix/hack, might be easier to simply filter out the ImageDefinition == AzureLinux3Gen2ArmImageDefinition from consideration in `listCIG`, which is called in the else branch of 
`if useSIG {`

charliedmcb: The listCIG case already does a for loop over the supported images. Could just `continue`, if it matched the `ImageDefinition`?

tallaxes: * DefaultImages() are called in several other places (both in runtime and test) that would require similar treatment, and
* Not sure how long the CIG fix will take, so prefer a solid fix in code

Bryce-Soghigian: The fix probably involves deleting the community images and versions then recreating them, but with a different auth

charliedmcb: >DefaultImages() are called in several other places (both in runtime and test) that would require similar treatment

So, this part isn't entirely true.

The other non-test instances are still dependent upon the result produced by `listCIG`, meaning that extra images in those cases shouldn't be a concern.

However, if the fix is expected to take longer, I'm fine with something more substantial like this, to ensure any future developments keep the fix in place. Just was trying to make things as minor touch if a quick adjustment on our end, and fix for CIG would be coming.

pkg/providers/instancetype/suite_test.go
```diff
@@ -1582,6 +1582,9 @@ var _ = Describe("InstanceType Provider", func() {
 		DescribeTable("should select the right image for a given instance type",
 			func(instanceType string, imageFamily string, expectedImageDefinition string, expectedGalleryURL string) {
 				statusController := status.NewController(env.Client, azureEnv.KubernetesVersionProvider, azureEnv.ImageProvider, env.KubernetesInterface)
+				if expectUseAzureLinux3 && expectedImageDefinition == azureLinuxGen2ArmImageDefinition {
```
charliedmcb, line 1585: Maybe Drop a TODO, to adjust this later?

test/suites/utilization/suite_test.go
```diff
@@ -51,6 +52,9 @@ var _ = Describe("Utilization", func() {
 		ExpectProvisionPodPerNode(env.AZLinuxNodeClass, env.DefaultNodePool)
 	})
 	It("should provision one pod per node (AzureLinux, arm64)", func() {
+		if imagefamily.UseAzureLinux3(env.K8sVersion()) && env.InClusterController {
```
charliedmcb, line 55: InClusterController means self-hosted?

charliedmcb (approved):
Overall looks good. Left some minor comments/questions.

test/suites/utilization/suite_test.go
tallaxes: Yes

pkg/providers/imagefamily/nodeimage.go
tallaxes: > but with a different auth

I think you mean different architecture

pkg/providers/instancetype/suite_test.go
tallaxes: There is a comment in the initial filtering, and there are multiple relevant tests - all of which will break when we revert the original filtering, so we won't forget

PR #1059, opened 2025-07-25
test: add support for triggering e2e run for individual test suites

Bryce-Soghigian:
**Description**

This PR adds the ability to trigger individual E2E test suites directly via `workflow_dispatch` on `e2e.yaml`, enabling faster iteration when debugging specific test failures.

- **Faster debugging**: Run a single failing suite without waiting for all 14 suites
- **Simpler architecture**: Each E2E run uses `github.run_id` for unique resource naming, eliminating the need for hash generation and propagation
- **Backward compatible**: Matrix runs still work as before

**Changes**

**`e2e.yaml`**
- Added `workflow_dispatch` trigger with a dropdown to select individual suites
- Simplified resource naming to use `github.run_id` directly instead of accepting an external `hash` input

**`e2e-matrix.yaml`**
- Removed `initialize-generative-params` job since `e2e.yaml` now uses `github.run_id` internally
- Removed the `hash` parameter from the workflow call to `e2e.yaml`

### How was this change tested?

- [x] Triggered single suite (Subnet) via `workflow_dispatch` on `e2e.yaml` https://github.com/Azure/karpenter-provider-azure/actions/runs/20841815479
- [x] Triggered full E2E matrix via `e2e-matrix-trigger.yaml` https://github.com/Azure/karpenter-provider-azure/actions/runs/20841820639

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


tallaxes:
> Have you considered just adding workflow_dispatch trigger to e2e workflow?

@Bryce-Soghigian, I meant adding a new `workflow_dispatch` trigger to [e2e.yaml](https://github.com/Azure/karpenter-provider-azure/blob/main/.github/workflows/e2e.yaml), so that it can be run directly. Seems cleaner to me, making e2e workflow more useful, and leaving test matrix driver essentially as is. (You may need to migrate hash calculation into e2e.yaml - but that's a good refactoring.)

tallaxes:
@copilot can you pick this up? The intent and direction is good, just need to address the feedback

PR #1072, opened 2025-07-30
feat: use in-place update for tags rather than drift

matthchr:
**Description**
 * Update the drift hash version to exclude tags.
 * Add tags to consideration for in-place update.
 * Add tests for in-place update of tags.
 * Add E2E tests for in-place update of tags

* Unit tests
* E2E tests (new suite, InPlaceUpdate)
* Manual tests

- [x] Yes, issue opened: #TBD
- [ ] No

**Release Note**

```release-note
AKSNodeClass spec.tags field no longer triggers drift. Updates to tags will instead be patched onto existing VMs, Disks, NetworkInterfaces, etc in-place.
```


pkg/cloudprovider/cloudprovider.go
```diff
@@ -179,10 +181,19 @@ func (c *CloudProvider) Create(ctx context.Context, nodeClaim *karpv1.NodeClaim)
 	instanceType, _ := lo.Find(instanceTypes, func(i *cloudprovider.InstanceType) bool {
 		return i.Name == string(lo.FromPtr(instance.Properties.HardwareProfile.VMSize))
 	})
+
+	// TODO: Is this the right nodeclaim? We also have nc below, it's not 100% clear to me what the difference is
```
matthchr: Confirm this during review and remove the TODO

pkg/cloudprovider/cloudprovider.go
```diff
@@ -179,10 +181,19 @@ func (c *CloudProvider) Create(ctx context.Context, nodeClaim *karpv1.NodeClaim)
...
+	inPlaceUpdateHash, err := inplaceupdate.HashFromNodeClaim(options.FromContext(ctx), nodeClaim, nodeClass)
+	if err != nil {
+		return nil, fmt.Errorf("failed to calculate in place update hash, %w", err)
+	}
+
 	nc, err := c.instanceToNodeClaim(ctx, instance, instanceType)
 	nc.Annotations = lo.Assign(nc.Annotations, map[string]string{
 		v1beta1.AnnotationAKSNodeClassHash:        nodeClass.Hash(),
 		v1beta1.AnnotationAKSNodeClassHashVersion: v1beta1.AKSNodeClassHashVersion,
+		v1beta1.AnnotationInPlaceUpdateHash:       inPlaceUpdateHash,
```
matthchr, line 196: Possibly we should just patch this on from the inplace controller once the claim is launched?

pkg/controllers/nodeclaim/inplaceupdate/controller.go
```diff
@@ -129,32 +120,81 @@ func (c *Controller) Reconcile(ctx context.Context, nodeClaim *karpv1.NodeClaim)
...
-	options *options.Options,
-	// TODO: Can pass and consider NodeClaim and/or NodePool here if we need to in the future
-	currentVM *armcompute.VirtualMachine,
-) *armcompute.VirtualMachineUpdate {
-	// Determine the differences between the current state and the goal state
-	expectedIdentities := options.NodeIdentities
-	var currentIdentities []string
-	if currentVM.Identity != nil {
-		currentIdentities = lo.Keys(currentVM.Identity.UserAssignedIdentities)
+// TODO: duplicate from resolveNodeClassFromNodeClaim for CloudProvider
```
matthchr, line 121: Ideas on making this function more reusable? Or maybe in this case duplication is preferred for simplicity?

pkg/providers/instance/instance.go
```diff
@@ -81,6 +78,17 @@ const (
 	cseNameLinux   = "cse-agent-karpenter"
 )
 
+// TODO: Is there a better place to put this?
```
matthchr: Resolve the answer to this in reivew and remove comment

pkg/providers/launchtemplate/tags.go
```diff
@@ -0,0 +1,59 @@
...
+)
+
+// TODO: Would like to refactor this out of launchtemplate at some point
+// Tags returns the tags to be applied to a resource (VM, Disk, NIC, etc)
+func Tags(
+	options *options.Options,
+	nodeClass *v1beta1.AKSNodeClass,
+	nodeClaim *v1.NodeClaim,
+) map[string]*string {
+	// TODO: This may not quite work because we assign some additional labels during the creation
```
matthchr: This is actually fine now with how we have things, but in theory we could have ordering issues if we append additional labels in LaunchTemplate that we end up trying to read here (we won't see them).

Maybe just worth converting to a note rather than a TODO

pkg/cloudprovider/cloudprovider.go
tallaxes: The second nodeClaim - the result of instanceToNodeClaim - will have additional (status, labels, annotations etc.) fields populated, represents the actual NodeClaim to be returned by Create, and is probably a better logical target. Even though No, in this case it does not matter, both because the logic in quesiton does not depend on those addititinal fields, and and because the current implementation of HashFromNodeClaim does not even use NodeClaim. Maybe best to update the function and its signature to reflect that? Would eliminate the question :-)

tallaxes:
> Updates to tags will instead be patched onto existing VMs, Disks, NetworkInterfaces, etc in-place.

Not patching resources other than VMs - yet?

test/pkg/environment/common/utils.go
```diff
@@ -0,0 +1,32 @@
...
+
+// CoerceToObject converts a slice of objects that implement client.Object to a slice of client.Object
+// This is useful for passing multiple objects to functions that expect a slice of client.Object
+// and is required because Go slices are not covariant.
+func CoerceToObject[T client.Object](objs ...T) []client.Object {
+	coerced := make([]client.Object, len(objs))
+	for i, obj := range objs {
+		coerced[i] = obj
+	}
+	return coerced
```
tallaxes:
```golang
return lo.Map(objs, func(obj T, _ int) client.Object { return obj })
```
Would this work?

matthchr: Ahh yes I should have consider using `lo` for this.

pkg/controllers/nodeclaim/inplaceupdate/controller.go
tallaxes: Probably fine to duplicate. Or could exctract/reuse just the lookup part, with custom handling of not found ...

tallaxes (dismissed):
LGTM

matthchr:
Needs to run full E2E before merge

pkg/cloudprovider/cloudprovider.go
matthchr: Fixed this.

pkg/providers/instance/instance.go
matthchr: Leaving this here for now and removing this comment

pkg/providers/launchtemplate/tags.go
matthchr: Changed this to a note and rephrased slightly.

test/pkg/environment/common/utils.go
matthchr: Deleted this function entirely and swapped to using lo for this.

pkg/controllers/nodeclaim/inplaceupdate/controller.go
matthchr: Just going to leave as duplicate for now, but will leave the TODO.

pkg/apis/v1alpha2/aksnodeclass.go
```diff
@@ -49,7 +49,7 @@ type AKSNodeClassSpec struct {
 	// +kubebuilder:validation:XValidation:message="tags keys must not contain '\\'",rule="self.all(k, !k.contains('\\\\'))"
 	// +kubebuilder:validation:XValidation:message="tags values must be less than 256 characters",rule="self.all(k, size(self[k]) <= 256)"
 	// +optional
-	Tags map[string]string `json:"tags,omitempty"`
+	Tags map[string]string `json:"tags,omitempty" hash:"ignore"`
```
charliedmcb, line 52: Do we actually want to update v1alpha2? Or just `v1beta1? 

I was thinking we are only updating the newer API?

pkg/cloudprovider/cloudprovider.go
```diff
@@ -179,12 +181,21 @@ func (c *CloudProvider) Create(ctx context.Context, nodeClaim *karpv1.NodeClaim)
...
+		return nil, err
+	}
+	inPlaceUpdateHash, err := inplaceupdate.HashFromNodeClaim(options.FromContext(ctx), nc, nodeClass)
+	if err != nil {
+		return nil, fmt.Errorf("failed to calculate in place update hash, %w", err)
+	}
 	nc.Annotations = lo.Assign(nc.Annotations, map[string]string{
 		v1beta1.AnnotationAKSNodeClassHash:        nodeClass.Hash(),
 		v1beta1.AnnotationAKSNodeClassHashVersion: v1beta1.AKSNodeClassHashVersion,
+		v1beta1.AnnotationInPlaceUpdateHash:       inPlaceUpdateHash,
```
charliedmcb, line 196: Annotation already was defined, but we weren't actually adding it before?

charliedmcb: Ah, I see we were. You just refactored here.

pkg/providers/launchtemplate/tags.go
```diff
@@ -0,0 +1,65 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package launchtemplate
+
+import (
+	"strings"
+
+	v1 "sigs.k8s.io/karpenter/pkg/apis/v1"
```
charliedmcb: nit, not sure how we're standardizing on this import or not, but I've been liking k8s being the import of `v1`, and for karp vs using: `karpv1 "sigs.k8s.io/karpenter/pkg/apis/v1"`
```suggestion
	karpv1 "sigs.k8s.io/karpenter/pkg/apis/v1"
```

pkg/controllers/nodeclaim/inplaceupdate/controller.go
```diff
@@ -129,32 +120,81 @@ func (c *Controller) Reconcile(ctx context.Context, nodeClaim *karpv1.NodeClaim)
...
+
+func (c *Controller) shouldProcess(ctx context.Context, nodeClaim *karpv1.NodeClaim) (bool, reconcile.Result) {
+	if !nodeClaim.DeletionTimestamp.IsZero() {
+		return false, reconcile.Result{}
+	}
+
+	// If the node isn't registered yet, we need to wait until it is as otherwise all the resources we need to update may not exist yet
+	if !nodeClaim.StatusConditions().Get(karpv1.ConditionTypeRegistered).IsTrue() {
+		log.FromContext(ctx).V(1).Info("can't update yet as the claim is not registered")
+		return false, reconcile.Result{RequeueAfter: 20 * time.Second}
```
charliedmcb: 20 seconds seems faster than what we need?

If not registered yet, probably unlikely that its going to be getting an update that would require in place updating that quick? Same for `ProviderID == ""`

pkg/controllers/nodeclaim/inplaceupdate/controller.go
```diff
@@ -129,32 +120,81 @@ func (c *Controller) Reconcile(ctx context.Context, nodeClaim *karpv1.NodeClaim)
...
-	currentVM *armcompute.VirtualMachine,
-) *armcompute.VirtualMachineUpdate {
-	// Determine the differences between the current state and the goal state
-	expectedIdentities := options.NodeIdentities
-	var currentIdentities []string
-	if currentVM.Identity != nil {
-		currentIdentities = lo.Keys(currentVM.Identity.UserAssignedIdentities)
+// TODO: duplicate from resolveNodeClassFromNodeClaim for CloudProvider
+// resolveAKSNodeClass resolves the AKSNodeClass from the NodeClaim's NodeClassRef
+func (c *Controller) resolveAKSNodeClass(ctx context.Context, nodeClaim *karpv1.NodeClaim) (*v1beta1.AKSNodeClass, error) {
```
charliedmcb, line 123: Seems worth having as a reusable function in a utils package for us?

Similar code in cloudprovider.go:
https://github.com/Azure/karpenter-provider-azure/blob/3352b2d39ddaf6a11406f7beba861b5ba41e70f7/pkg/cloudprovider/cloudprovider.go#L395

Thinking something similar to the util package for core:
https://github.com/kubernetes-sigs/karpenter/blob/2613a66d85de881b40fb15603e073bf2910aedf1/pkg/utils/nodeclaim/nodeclaim.go#L208

Drop a TODO, if you don't do the refactor in this PR?

pkg/controllers/nodeclaim/inplaceupdate/controller.go
```diff
@@ -60,25 +63,27 @@ func NewController(
...
 	// nodePool, err := nodeclaimutil.Owner(ctx, c.kubeClient, nodeClaim)
 	// TODO: To look it up and use that as input to calculate the goal state as well
 
 	// Compare the expected hash with the actual hash
 	options := options.FromContext(ctx)
-	goalHash, err := HashFromNodeClaim(options, nodeClaim)
+	goalHash, err := HashFromNodeClaim(options, nodeClaim, nodeClass)
 	if err != nil {
 		return reconcile.Result{}, err
 	}
```
charliedmcb, line 80: nit for the log 3 lines below:
`log.FromContext(ctx).V(1).Info("comparing in place update hashes", "goalHash", goalHash, "actualHash", actualHash)`

Or actually contain the annotation name?

pkg/controllers/nodeclaim/inplaceupdate/controller.go
```diff
@@ -129,32 +120,81 @@ func (c *Controller) Reconcile(ctx context.Context, nodeClaim *karpv1.NodeClaim)
...
-	identity := instance.ConvertToVirtualMachineIdentity(toAdd)
+	if nodeClaim.Status.ProviderID == "" {
+		log.FromContext(ctx).V(1).Info("can't update yet as there's no provider ID")
+		return false, reconcile.Result{RequeueAfter: 20 * time.Second}
+	}
+
+	return true, reconcile.Result{}
+}
+
+func (c *Controller) getVM(ctx context.Context, nodeClaim *karpv1.NodeClaim) (*armcompute.VirtualMachine, error) {
```
charliedmcb, line 162: Feels like this could be another `utils` function?

pkg/providers/launchtemplate/tags.go
```diff
@@ -0,0 +1,65 @@
...
+) map[string]*string {
+	defaultTags := map[string]string{
+		KarpenterManagedTagKey: options.ClusterName,
+	}
+	// Note: Be careful depending on nodeClaim.Labels here, as we assign some additional labels during the creation
+	// of the static parameters for the launch template. Those labels haven't actually been applied to the nodeClaim yet,
+	// so if you try to use them here you will find they are missing.
+	// For now, we only depend on labels that are added to the nodeClaim itself.
+	if val, ok := nodeClaim.Labels[v1.NodePoolLabelKey]; ok {
+		defaultTags[NodePoolTagKey] = val
```
charliedmcb, line 52: This is a new tag being applied, or did I miss where this is being refactored from?

With this happening at the end of create, it seems a bit circular?
image

pkg/controllers/nodeclaim/inplaceupdate/controller.go
```diff
@@ -167,24 +207,117 @@ func (c *Controller) Register(_ context.Context, m manager.Manager) error {
...
-		// TODO: similar to https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/controllers/nodeclaim/disruption/controller.go#L214C3-L217C5
-		// TODO: if/when we need to monitor provisioner changes and flow updates on the NodePool down to the underlying VMs.
 		Complete(reconcile.AsReconciler(m.GetClient(), c))
 }
+
 func logVMPatch(ctx context.Context, update *armcompute.VirtualMachineUpdate) {
 	if log.FromContext(ctx).V(1).Enabled() {
 		rawStr := "<nil>"
 		if update != nil {
 			raw, _ := json.Marshal(update)
```
charliedmcb: This wouldn't possibly contain any sensitive data, right?

charliedmcb: Ah, I see you left a comment above about this.

pkg/providers/instance/instance.go
```diff
@@ -81,6 +78,18 @@ const (
...
 )
 
+// GetManagedExtensionNames gets the names of the VM extensions managed by Karpenter.
+// This is a set of 1 or 2 extensions (depending on provisionMode): aksIdentifyingExtension and (sometimes) cse.
+func GetManagedExtensionNames(provisionMode string) []string {
+	result := []string{
+		aksIdentifyingExtensionName,
+	}
+	if provisionMode == consts.ProvisionModeBootstrappingClient {
+		result = append(result, cseNameLinux) // TODO: Windows
```
charliedmcb, line 88: Assuming this is an unrelated fix you're bundling into here?

pkg/providers/launchtemplate/tags.go
charliedmcb: I see it was happening at VM create time in instance:
image

pkg/apis/v1alpha2/aksnodeclass.go
charliedmcb: Discussed offline. Makes sense to keep supporting both for now.

pkg/controllers/nodeclaim/inplaceupdate/controller.go
```diff
@@ -129,32 +120,81 @@ func (c *Controller) Reconcile(ctx context.Context, nodeClaim *karpv1.NodeClaim)
 	return reconcile.Result{}, nil
```
charliedmcb, line 117: Should we return a time here to trigger the controller again?

Will it periodically retrigger already, or just on the nodeclass = nodeclaim updates? 

Seems worth periodically checking if the tags themselves have been mutated?

pkg/providers/launchtemplate/tags.go
matthchr: Yes I refactored this partially because tag application and ordering was all over the place and it was hard to grok. Consolidating to one spot so it can be reused elsewhere (I probably could have broken just this change out into its own PR TBH).

pkg/controllers/nodeclaim/inplaceupdate/controller.go
matthchr: We aren't patching anything sensitive currently, unless you count tags as sensitive. I don't think tags are sensitive though given that they're on the CRD in plain-text (not in a secret or something).

pkg/providers/instance/instance.go
matthchr: No not really unrelated or a fix. This has always been the logic (we have 1 extension in scriptless mode, and 2 in bootstrapping client mode). This is just encoding that in a reusable way so that we can rely on it elsewhere (for example when we're patching extensions and/or in tests when we're expecting certain things to have been updated it knows what to look for).

pkg/controllers/nodeclaim/inplaceupdate/controller.go
matthchr: I think it's an interesting question. Part of me says yes and part of me says no. We don't seem to be 100% consistent in how we do this. I implemented this this way because _for the most part_ drift also seems the same - it doesn't examine the actual Azure resource for changes most of the time. Note that we compare hash to expectedHash, and only if they don't match do we go to Azure and patch it, so re-reconciling here isn't helpful because the hashes won't have changed.

If the user goes and changes the Azure state behind our back, it will not change hash or expectedHash because they're stored in an annotation on the k8s resource (or computed from the state of k8s resources). So just changing this to return a periodic time wouldn't be enough, we would also need to change how we determine "if we need to apply a patch".

I do think:
1. This PR is big enough as it is, so probably doesn't make sense to change here.
2. We should get @tallaxes thoughts on this as well.

pkg/controllers/nodeclaim/inplaceupdate/controller.go
matthchr: The controller will be logged as `nodeclaim.inplaceupdate`, so in theory it should already be clear that this is for in-place update, but I've updated the message to say `comparing in-place update hashes` just for more clarity as requested.

pkg/controllers/nodeclaim/inplaceupdate/controller.go
matthchr: I'm proxying Paul on this but I generally don't like `utils` packages (more ok with utils/nodeclaim) because they just become a big spaghetti mess of unrelated utilities that should've found better homes.

I'll refactor this in a smaller follow-up PR.

pkg/controllers/nodeclaim/inplaceupdate/controller.go
matthchr: Agreed. I've left a TODO and will send a follow up PR

pkg/cloudprovider/cloudprovider.go
matthchr: For now I think doing this is fine. (It's obviously what we do for Drift too)

pkg/controllers/nodeclaim/inplaceupdate/controller.go
charliedmcb: > If the user goes and changes the Azure state behind our back, it will not change hash or expectedHash

Very good point! 

Still does feel like we should have something for checking on actual state, and reconciling, but a discussion beyond this PR.

pkg/controllers/nodeclaim/inplaceupdate/controller.go
matthchr: Yeah you have a point. I've changed to 60s

I've actually reordered things to check the hash first, then check shouldProcess as well, so we shouldn't have as many new nodeClaims that don't need patches waiting in the queue. It'll only get requeued if we need to patch it after it joins, which should be uncommon.

pkg/providers/launchtemplate/tags.go
matthchr: Fixed.

test/suites/drift/suite_test.go
```diff
@@ -547,12 +546,11 @@ var _ = Describe("Drift", func() {
...
 		})
-		// Updating `nodeClass.Spec.Tags` would normally trigger drift on all nodeclaims using the
+
+		// Updating `nodeClass.Spec.MaxPods` would normally trigger drift on all nodeclaims using the
 		// nodeclass. However, the aksnodeclass-hash-version does not match the controller hash version, so we will see that
 		// none of the nodeclaims will be drifted and all nodeclaims will have an updated `aksnodeclass-hash` and `aksnodeclass-hash-version` annotation
-		nodeClass.Spec.Tags = lo.Assign(nodeClass.Spec.Tags, map[string]string{
-			"test-key": "test-value",
-		})
+		nodeClass.Spec.MaxPods = lo.ToPtr[int32](10)
```
charliedmcb, line 555: nit: maybe set this to an initial different value explicitly within this test before doing the first 
`env.ExpectCreated(dep, nodeClass, nodePool)`
Just to reinforce the explicate change?

matthchr: Done

pkg/controllers/nodeclaim/inplaceupdate/controller.go
tallaxes: Conceptually / in the ideal world - Yes, reconciliation between desired and actual state better know what actual state is. And there is the question of whether Karpenter-requested tags represent all desired tags, or a subset (and there could be more on the resource). But fetching actual state - especially on many cloud resources, and frequently/periodically - is potentially expensive, so likely best deferred until we have a good reason / actual use case.

Having said that, if it turns out that always replacing the tags on the resource (that's what we are doing, right?) is unaaceptable for some reason - e.g. interferes with policy-controlled tags, or some other use case where for whatever reason tag has to be on resource but having Karpenter tag is not acceptable (I don't know, dynamically calculated value or something?) - then we may need to revisit this.

matthchr:
Confirmed that the consolidation failure is a known flake. Rest I think are OK but running one more E2E just to make sure.

matthchr:
Ok final E2E pass is: https://github.com/Azure/karpenter-provider-azure/actions/runs/16686117906

Failing tests:
  [FAIL] Drift NodePool Drift [It] NodeRequirements
  /home/runner/work/karpenter-provider-azure/karpenter-provider-azure/test/suites/drift/suite_test.go:414
  [FAIL] Drift [It] should update the aksnodeclass-hash annotation on the aksnodeclass and nodeclaim when the aksnodeclass's aksnodeclass-hash-version annotation does not match the controller hash version
  /home/runner/work/karpenter-provider-azure/karpenter-provider-azure/test/suites/drift/suite_test.go:530
  [FAIL] StandaloneNodeClaim [It] should create a standard NodeClaim based on resource requests
  /home/runner/work/karpenter-provider-azure/karpenter-provider-azure/test/suites/nodeclaim/nodeclaim_test.go:79

I ran all 3 of these locally and all 2/3 passed. The one that didn't was:

> [FAIL] Drift [It] should update the aksnodeclass-hash annotation on the aksnodeclass and nodeclaim when the aksnodeclass's aksnodeclass-hash-version annotation does not match the controller hash version

This is because I made a change to that test at @charliedmcb suggestion, which broke it. I've pushed a fix for that (re-ran it locally after fix and it passed).

tallaxes (approved):

PR #1079, opened 2025-08-01
chore: readme should reference azure/karpenter-provider-azure not azure/karpenter in curl commands

### LINKED ISSUE #966: Some fixes to do on the Readme
I tried the self hosted procedure described in the Readme.

Apart from the fact that I used an AKS created previously ad hoc via Terraform, I had to correct two things regarding the Karpenter deployment.

1. The following URL doesn't works form me 
`curl -sO https://raw.githubusercontent.com/Azure/karpenter/v${KARPENTER_VERSION}/karpenter-values-template.yaml `
so I used 
`curl -sO https://raw.githubusercontent.com/Azure/karpenter-provider-azure/refs/tags/v${KARPENTER_VERSION}/karpenter-values-template.yaml`

2. In the example of creating the NodePool, v1beta is indicated, but in reality by installing the latest version of helm chart, 1.4.0, v1 is provided.

3. It may also be useful to specify that it may be necessary, depending on the case, to perform role assignments not only with scope on the resource group of the nodes.

I hope these pointers based on my experience can help.

Bryce-Soghigian:
Fixes #966

**Description**
repo was renamed to karpenter-provider-azure, we need to update the readme to reflect that

 ran curl command locally reproduced the issue and then fixed with changing repo name to the right reference.

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/designs/national-cloud-support.md
+++ b/designs/national-cloud-support.md
@@ -1,0 +1,122 @@
+# Supporting non-public clouds
+
+## Overview
+
+This document discusses how Karpenter will support non-public clouds.
+
+There are two classes of non-public cloud
+
+1. Known clouds (Mooncake, Fairfax, etc).
+2. Nonstandard/Azure-Stack like clouds.
+
+## Current state
+
+* Karpenter accepts an [`ARM_CLOUD` env variable](https://github.com/Azure/karpenter-provider-azure/blob/main/pkg/auth/config.go#L84) but it is not used consistently.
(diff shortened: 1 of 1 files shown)


charliedmcb (approved):
lgtm

tallaxes (approved):
> ran curl command locally reproduced the issue and then fixed with changing repo name to the right reference.

Did you try actual install? Would be good to validate everything still works (and update if it does not)

tallaxes:
@Bryce-Soghigian, did you try actual install?

PR #1084, opened 2025-08-04
fix: bug with CloudProvider list where it never used correct tags key

matthchr:
It used the wrong tag key so would never get the InstanceType details

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
NONE
```

### DIFF
--- a/pkg/cloudprovider/cloudprovider.go
+++ b/pkg/cloudprovider/cloudprovider.go
@@ -468,7 +468,7 @@ func (c *CloudProvider) resolveInstanceTypeFromInstance(ctx context.Context, ins
 }
 
 func (c *CloudProvider) resolveNodePoolFromInstance(ctx context.Context, instance *armcompute.VirtualMachine) (*karpv1.NodePool, error) {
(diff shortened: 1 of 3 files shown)


matthchr:
Looks like some test issue(s) possibly? I will look into it.

matthchr:
Test issue was a set setup issue. We were doing:
```
vm.Properties = &armcompute.VirtualMachineProperties{
    TimeCreated: lo.ToPtr(time.Now().Add(-time.Minute * 10)),
}
```

this overwrites the whole Properties field. Now that we've fixed instance list, the operator was panicing because the VM it got back from Azure was impossibly shaped (missing `HardwareProfile.VMSize` field).

I thought about hardening the production code against such VM shapes, but it doesn't seem practically possible for Azure to really return such a shape at the moment and there are quite a few places in the code where we made the assumption that the VM returned by Azure looks "sane", so I just fixed the test.

charliedmcb (approved):
lgtm.

Agreed offline, will monitor e2e test run in main as an additional testing sanity check.

PR #1092, opened 2025-08-08
feat: support stateless cni in 1.34+

Bryce-Soghigian:
Fixes # 

**Description**
1.34 introduces a new pattern with how we target daemonsets for different CNI modes.  This adds a pr to add labels for the new feature
- make test
- e2e run: https://github.com/Azure/karpenter-provider-azure/actions/runs/16953512983

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/providers/instancetype/suite_test.go
+++ b/pkg/providers/instancetype/suite_test.go
@@ -246,6 +246,34 @@ var _ = Describe("InstanceType Provider", func() {
 				ContainSubstring("kubernetes.azure.com/azure-cni-overlay=true"),
 			))
 		})
+		It("should include stateless CNI label for kubernetes 1.34+", func() {
+			// Set kubernetes version to 1.34.0
+			nodeClass.Status.KubernetesVersion = "1.34.0"
+			ExpectApplied(ctx, env.Client, nodePool, nodeClass)
+			pod := coretest.UnschedulablePod()
+			ExpectProvisioned(ctx, env.Client, cluster, cloudProvider, coreProvisioner, pod)
+			ExpectScheduled(ctx, env.Client, pod)
+
+			decodedString := ExpectDecodedCustomData(azureEnv)
+			Expect(decodedString).To(SatisfyAll(
+				ContainSubstring("kubernetes.azure.com/azure-cni-overlay=true"),
+				ContainSubstring("kubernetes.azure.com/network-stateless-cni=true"),
+			))
+		})
+		It("should not include stateless CNI label for kubernetes < 1.34", func() {
+			// Set kubernetes version to 1.33.0
+			nodeClass.Status.KubernetesVersion = "1.33.0"
+			ExpectApplied(ctx, env.Client, nodePool, nodeClass)
+			pod := coretest.UnschedulablePod()
+			ExpectProvisioned(ctx, env.Client, cluster, cloudProvider, coreProvisioner, pod)
+			ExpectScheduled(ctx, env.Client, pod)
+
+			decodedString := ExpectDecodedCustomData(azureEnv)
+			Expect(decodedString).To(SatisfyAll(
+				ContainSubstring("kubernetes.azure.com/azure-cni-overlay=true"),
+				Not(ContainSubstring("kubernetes.azure.com/network-stateless-cni")),
+			))
+		})
 		It("should use the subnet specified in the nodeclass", func() {
 			nodeClass.Spec.VNETSubnetID = lo.ToPtr("/subscriptions/12345678-1234-1234-1234-123456789012/resourceGroups/sillygeese/providers/Microsoft.Network/virtualNetworks/karpenter/subnets/nodeclassSubnet")
 			ExpectApplied(ctx, env.Client, nodePool, nodeClass)
--- a/pkg/providers/instancetype/suite_test.go
+++ b/pkg/providers/instancetype/suite_test.go
@@ -246,7 +246,7 @@ var _ = Describe("InstanceType Provider", func() {
 				ContainSubstring("kubernetes.azure.com/azure-cni-overlay=true"),
 			))
 		})
-		It("should include stateless CNI label for kubernetes 1.34+", func() {
+		It("should include stateless CNI label for kubernetes 1.34+ set to true", func() {
 			// Set kubernetes version to 1.34.0
 			nodeClass.Status.KubernetesVersion = "1.34.0"
 			ExpectApplied(ctx, env.Client, nodePool, nodeClass)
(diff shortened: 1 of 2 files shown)


pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -140,7 +142,11 @@ func (p *Provider) getStaticParameters(
 
 	if isAzureCNIOverlay(ctx) {
 		// TODO: make conditional on pod subnet
-		vnetLabels, err := p.getVnetInfoLabels(subnetID)
+		kubernetesVersion, err := nodeClass.GetKubernetesVersion()
```
Bryce-Soghigian, line 146: @charliedmcb is this going to always equal the kubelet minor version of the node we are trying to launch?

Bryce-Soghigian: discussed offline, should be fine

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -243,5 +249,14 @@ func (p *Provider) getVnetInfoLabels(subnetID string) (map[string]string, error)
 		azureCNIOverlayLabel: strconv.FormatBool(true),
 		podNetworkTypeLabel:  consts.NetworkPluginModeOverlay,
 	}
+
+	// Add stateless CNI label for Kubernetes 1.34+
```
paulgmiller: I don't think we'll ever change this. Evan Baker will just disable his ds instead.) would be nice if we could point to code it matches but oh well can't do that with labels either.

pkg/providers/instancetype/suite_test.go
```diff
@@ -246,6 +246,34 @@ var _ = Describe("InstanceType Provider", func() {
...
+			nodeClass.Status.KubernetesVersion = "1.33.0"
+			ExpectApplied(ctx, env.Client, nodePool, nodeClass)
+			pod := coretest.UnschedulablePod()
+			ExpectProvisioned(ctx, env.Client, cluster, cloudProvider, coreProvisioner, pod)
+			ExpectScheduled(ctx, env.Client, pod)
+
+			decodedString := ExpectDecodedCustomData(azureEnv)
+			Expect(decodedString).To(SatisfyAll(
+				ContainSubstring("kubernetes.azure.com/azure-cni-overlay=true"),
+				Not(ContainSubstring("kubernetes.azure.com/network-stateless-cni")),
```
paulgmiller: I went with always populatign it with false in my pr.

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -243,5 +249,14 @@ func (p *Provider) getVnetInfoLabels(subnetID string) (map[string]string, error)
...
+	if kubernetesVersion != "" {
+		parsedVersion, err := semver.Parse(kubernetesVersion)
```
paulgmiller: eating this error. if its a bad semver.

Bryce-Soghigian: fair

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -243,5 +249,14 @@ func (p *Provider) getVnetInfoLabels(subnetID string) (map[string]string, error)
...
+		if err == nil && parsedVersion.Minor >= 34 {
+			vnetLabels[networkStatelessCNILabel] = "true"
```
paulgmiller: I went and always created this label setting it to false in < 34
https://github.com/Azure/karpenter-provider-azure/pull/1092/files#r2274701432

paulgmiller (dismissed):
just one point about always creating the label.

pkg/providers/instancetype/suite_test.go
Bryce-Soghigian: Ill keep parity with that, seems reasonable to do, and when we switch over to machine api we don't want label inconsistency

pkg/providers/launchtemplate/launchtemplate.go
charliedmcb: Yea, this lgtm

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -243,5 +249,9 @@ func (p *Provider) getVnetInfoLabels(subnetID string) (map[string]string, error)
 		azureCNIOverlayLabel: strconv.FormatBool(true),
 		podNetworkTypeLabel:  consts.NetworkPluginModeOverlay,
 	}
+
+	parsedVersion, _ := semver.Parse(kubernetesVersion)
```
charliedmcb: I think you need light modification/safety here.

See `UseAzureLinux3`
https://github.com/Azure/karpenter-provider-azure/blob/38368736f1304b6156443367bcc9ced16cdcd006/pkg/providers/imagefamily/utils.go#L25
```
	// Parse version, stripping any 'v' prefix if present
	version, err := semver.ParseTolerant(strings.TrimPrefix(kubernetesVersion, "v"))
	if err != nil {
		// If we can't parse the version, default to AzureLinux (false)
		return false
	}
```

At least stripping the `v` I think is required. I think we prefix with `v`, and semver treats that as an error.

Alex recently added the `ParseTolerant`, instead of `Parse` as well.

charliedmcb: Looked into it further, and the prefix should be trimmed by the k8s provider, and set as such without it by controller. However, I still think its at least work checking the potential error, rather than discarding it `_`, and as such, might as well be defensive with the same parsing as `UseAzureLinux3`

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -243,5 +249,9 @@ func (p *Provider) getVnetInfoLabels(subnetID string) (map[string]string, error)
...
+	vnetLabels[networkStatelessCNILabel] = lo.Ternary(parsedVersion.Minor >= 34, "true", "false")
```
charliedmcb: Again, maybe best following the style on version comparison from `UseAzureLinux3`?
https://github.com/Azure/karpenter-provider-azure/blob/38368736f1304b6156443367bcc9ced16cdcd006/pkg/providers/imagefamily/utils.go#L32

In this case it would be:
```suggestion
	vnetLabels[networkStatelessCNILabel] = lo.Ternary(parsedVersion.GE(semver.Version{Major: 1, Minor: 34}), "true", "false")
```

pkg/providers/instancetype/suite_test.go
```diff
@@ -246,6 +246,32 @@ var _ = Describe("InstanceType Provider", func() {
...
+			nodeClass.Status.KubernetesVersion = "1.33.0"
+			ExpectApplied(ctx, env.Client, nodePool, nodeClass)
+			pod := coretest.UnschedulablePod()
+			ExpectProvisioned(ctx, env.Client, cluster, cloudProvider, coreProvisioner, pod)
+			ExpectScheduled(ctx, env.Client, pod)
+			decodedString := ExpectDecodedCustomData(azureEnv)
+			Expect(decodedString).To(SatisfyAll(
+				ContainSubstring("kubernetes.azure.com/network-stateless-cni=false"),
+			))
+
```
charliedmcb, line 273: nit
```suggestion
```

charliedmcb (commented):
overall, lgtm. 

Would like to at minimum have the error handled in parsing. Have some more comments on  possibly aligning k8s version handling with `UseAzureLinux3` as well.

charliedmcb (approved):
Forgot to approve

pkg/providers/launchtemplate/launchtemplate.go
Bryce-Soghigian: yeah i originally had the error handling that if k8s version parsing failed we would return an error, i can add this back and use ParseTolerant

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -250,7 +251,11 @@ func (p *Provider) getVnetInfoLabels(subnetID string, kubernetesVersion string)
 		podNetworkTypeLabel:  consts.NetworkPluginModeOverlay,
 	}
 
-	parsedVersion, _ := semver.Parse(kubernetesVersion)
+	parsedVersion, err := semver.ParseTolerant(strings.TrimPrefix(kubernetesVersion, "v"))
+	// Sanity Check: in production we should always have a k8s version set
+	if err != nil {
+		return map[string]string{}, err
```
charliedmcb, line 257: nit: should this be a empty map, or nil.

Just differs from the first error return of the function.

match?
```suggestion
		return nil, err
```

PR #1093, opened 2025-08-08
chore: Small nodeclaim utils refactor

matthchr:
**Release Note**

```release-note
NONE
```

### DIFF
--- a/pkg/cloudprovider/drift.go
+++ b/pkg/cloudprovider/drift.go
@@ -30,11 +30,11 @@ import (
 	"github.com/Azure/karpenter-provider-azure/pkg/apis/v1beta1"
 	"github.com/Azure/karpenter-provider-azure/pkg/operator/options"
 	"github.com/Azure/karpenter-provider-azure/pkg/providers/instance"
-	"github.com/Azure/karpenter-provider-azure/pkg/utils"
+	nodeclaimutils "github.com/Azure/karpenter-provider-azure/pkg/utils/nodeclaim"
 
 	karpv1 "sigs.k8s.io/karpenter/pkg/apis/v1"
 	"sigs.k8s.io/karpenter/pkg/cloudprovider"
-	nodeclaimutils "sigs.k8s.io/karpenter/pkg/utils/nodeclaim"
+	karpnodeclaimutils "sigs.k8s.io/karpenter/pkg/utils/nodeclaim"
 )
 
 const (
@@ -138,7 +138,7 @@ func (c *CloudProvider) isImageVersionDrifted(
 ) (cloudprovider.DriftReason, error) {
 	logger := log.FromContext(ctx)
 
-	id, err := utils.GetVMName(nodeClaim.Status.ProviderID)
+	id, err := nodeclaimutils.GetVMName(nodeClaim.Status.ProviderID)
 	if err != nil {
 		// TODO (charliedmcb): Do we need to handle vm not found here before its provisioned?
 		//     I don't think we can get to Drift, until after ProviderID is set, so this should be fine/impossible.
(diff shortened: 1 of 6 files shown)


pkg/controllers/nodeclaim/inplaceupdate/controller.go
```diff
@@ -34,13 +33,13 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	karpv1 "sigs.k8s.io/karpenter/pkg/apis/v1"
 	"sigs.k8s.io/karpenter/pkg/operator/injection"
-	nodeclaimutils "sigs.k8s.io/karpenter/pkg/utils/nodeclaim"
+	karpnodeclaimutils "sigs.k8s.io/karpenter/pkg/utils/nodeclaim"
```
Bryce-Soghigian: maybe corenodeclaimutils?

tallaxes (dismissed):

pkg/utils/errors.go
```diff
@@ -0,0 +1,31 @@
...
+package utils
+
+import (
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+// NewTerminatingResourceError returns a NotFound error for handling by
```
charliedmcb: `handling by`...?

pkg/utils/nodeclaim/nodeclaim.go
```diff
@@ -0,0 +1,87 @@
...
+		return nil, fmt.Errorf("nodeClaim %s does not have a nodeClassRef", nodeClaim.Name)
+	}
+
+	nodeClass := &v1beta1.AKSNodeClass{}
+	if err := kubeClient.Get(ctx, types.NamespacedName{Name: nodeClaim.Spec.NodeClassRef.Name}, nodeClass); err != nil {
+		return nil, fmt.Errorf("getting AKSNodeClass %s: %w", nodeClaim.Spec.NodeClassRef.Name, err)
+	}
+
+	// For the purposes of in-place updates, we treat deleting NodeClasses as an error
+	// This error should ideally be transient and the NodeClaim itself will be deleted
```
charliedmcb: Maybe add this note to the function itself? 

Or move this `if` block to the callee location in the `in-place` update controller, if we're expecting other's won't want to treat it the same way?

Or parametrize treating deletion as an error?

It just feels wrong having a comment about the `in-place` controller (one callee) in the internals of the utils func.

pkg/utils/nodeclaim/nodeclaim.go
```diff
@@ -0,0 +1,87 @@
...
+	if !nodeClass.DeletionTimestamp.IsZero() {
+		return nil, utils.NewTerminatingResourceError(schema.GroupResource{Group: apis.Group, Resource: "aksnodeclasses"}, nodeClass.Name)
+	}
+
+	return nodeClass, nil
+}
+
+// TODO: Could go onto instanceProvider?
```
charliedmcb, line 55: Does feel like it could, but still happy with this adjustment.

If you want to move it to `instanceProvider` happy to review.

pkg/utils/nodeclaim/nodeclaim.go
```diff
@@ -0,0 +1,87 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("getting azure VM %s: %w", vmName, err)
+	}
+
+	return vm, nil
+}
+
+// GetVMName parses the provider ID stored on the node to get the vmName
+// associated with a node
+func GetVMName(providerID string) (string, error) {
```
charliedmcb, line 73: Why not remove the dead code copy of this func in utils.go?

pkg/cloudprovider/drift.go
```diff
@@ -30,11 +30,11 @@ import (
 	"github.com/Azure/karpenter-provider-azure/pkg/apis/v1beta1"
 	"github.com/Azure/karpenter-provider-azure/pkg/operator/options"
 	"github.com/Azure/karpenter-provider-azure/pkg/providers/instance"
-	"github.com/Azure/karpenter-provider-azure/pkg/utils"
+	nodeclaimutils "github.com/Azure/karpenter-provider-azure/pkg/utils/nodeclaim"
 
 	karpv1 "sigs.k8s.io/karpenter/pkg/apis/v1"
 	"sigs.k8s.io/karpenter/pkg/cloudprovider"
-	nodeclaimutils "sigs.k8s.io/karpenter/pkg/utils/nodeclaim"
+	karpnodeclaimutils "sigs.k8s.io/karpenter/pkg/utils/nodeclaim"
```
charliedmcb: consistency:
```suggestion
	corenodeclaimutils "sigs.k8s.io/karpenter/pkg/utils/nodeclaim"
```

pkg/utils/nodeclaim/nodeclaim.go
charliedmcb: I see in the cloudprovider implementation that got removed in place of this too, it handled it the same way, so maybe just make a generalized comment, rather than mentioning `in-place` controller?

charliedmcb (approved):
very minor comments/suggestions.

Overall, lgtm.

pkg/utils/errors.go
matthchr: Fixed

pkg/utils/nodeclaim/nodeclaim.go
matthchr: I think keeping it here at least for now makes sense to keep the interface of instanceProvider clean. Will keep as-is for now but leave the TODO in case in the future it makes sense to move.

pkg/utils/nodeclaim/nodeclaim.go
matthchr: Doh, I meant to but missed it. Great catch.

pkg/utils/nodeclaim/nodeclaim.go
matthchr: Yeah I don't think we want to parameterize this, but agree this comment as-is doesn't make as much sense anymore.

Have moved the comment to the function scope and cleaned up wording to make it clear that it applies to all calls to this function.

PR #1094, opened 2025-08-08
feat: support non-public clouds

matthchr:
* Use ARM_CLOUD parameter in all clients.
* Support new AZURE_ENVIRONMENT_FILEPATH parameter which points to custom cloud configuration (AzureStackCloud-like).
* Remove SIGAccessTokenScope in favor of the ARM_CLOUD or AZURE_ENVIRONMENT_FILEPATH sourced token scope.
* Update pricing provider to not run in non-public clouds. It just returns cached/static data for those clouds as the pricing API isn't supported there.

* Unit tests
* E2E test: https://github.com/Azure/karpenter-provider-azure/actions/runs/17223153444
* Manual test in AzureChinaCloud - with selfhosted we can't actually provision VMs because there are no community galleries in non-public clouds, but we got to the point where we tried to provision a VM.

Other E2E tests:
- https://github.com/Azure/karpenter-provider-azure/actions/runs/17219071963
- https://github.com/Azure/karpenter-provider-azure/actions/runs/17219068721

These have some flakes

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
Add support for non-public clouds in NAP mode via the ARM_CLOUD and AZURE_ENVIRONMENT_FILEPATH parameters
```

### DIFF
--- a/pkg/auth/environment_test.go
+++ b/pkg/auth/environment_test.go
@@ -62,22 +62,22 @@ func TestReadEnvironmentFromFile(t *testing.T) {
 			wantErr:     false,
 		},
 		{
-			name: "empty filepath",
+			name: "empty path",
 			setupFile: func(t *testing.T) string {
 				return ""
 			},
 			expectedEnv:    nil,
 			wantErr:        true,
-			expectedErrMsg: "filepath is empty",
+			expectedErrMsg: "path is empty",
 		},
 		{
-			name: "relative filepath",
+			name: "relative path",
 			setupFile: func(t *testing.T) string {
 				return "relative/path.json"
 			},
 			expectedEnv:    nil,
 			wantErr:        true,
-			expectedErrMsg: "filepath must be absolute",
+			expectedErrMsg: "path must be absolute",
 		},
 		{
 			name: "non-existent file",


pkg/auth/environment.go
```diff
@@ -0,0 +1,177 @@
...
+				Endpoint: env.ResourceManagerEndpoint,
+				Audience: env.TokenAudience,
+			},
+		},
+	}
+
+	// Add graph endpoint if available (missing from CloudProvider mapping)
+	// Note: Using custom service name since MicrosoftGraph might not be available in older SDK versions
+	if env.GraphEndpoint != "" {
+		config.Services["MicrosoftGraph"] = cloud.ServiceConfiguration{ // TODO(matthchr): Confirm this service is accurate
```
matthchr: I dont think this is right.

matthchr: Turns out we dont need this.

matthchr:
Need to run E2E tests and do some manual testing in both public and non-public clouds to make sure it works.

pkg/auth/environment.go
```diff
@@ -0,0 +1,165 @@
...
+		return fmt.Errorf("active directory endpoint is required")
+	}
+	if env.TokenAudience == "" {
+		return fmt.Errorf("token audience is required")
+	}
+	return nil
+}
+
+// mapTrack1ToTrack2Environment converts azclient.Environment (Track1 format) to cloud.Configuration (Track2 format)
+func mapTrack1ToTrack2Environment(env *azclient.Environment) (cloud.Configuration, error) {
```
matthchr, line 87: I had originally written that we shouldn't depend on [OverrideAzureCloudConfigFromEnv](https://github.com/kubernetes-sigs/cloud-provider-azure/blob/master/pkg/azclient/cloud.go#L106) for the following reasons:
  * The logic there would need to be changed to support the graph URI and any other URIs Karpenter ends up needing. This is not hard to upstream but it's also a small amount of code to
    duplicate and   being able to evolve separately from CloudProvider is probably worth it here.
  * It requires the `ARM_CLOUD` set to `AZURESTACKCLOUD`. We could also require this for consistency with other tools (CAS) but from a purely technical standpoint it's not correct.
  * As written, `OverrideAzureCloudConfigFromEnv` allows for partial overrides which seems risky at best.

The first is actually not true (I'll correct the design doc as part of this PR in a future iteration). I do still feel the other two are correct. Just bringing this up here though as _technically_ we could share code, though I do still think the way `OverrideAzureCloudConfigFromEnv` is factored makes it harder than ideal.

pkg/auth/config.go
```diff
@@ -86,16 +87,26 @@ func (cfg *Config) Build() error {
...
+	// Default is AzurePublicCloud if not set
+	if cfg.Cloud == "" && cfg.AzureEnvironmentFilepath == "" {
+		cfg.Cloud = "AzurePublicCloud"
+	}
+
 	return nil
 }
 
 func (cfg *Config) Validate() error {
+	// Validate that ARM_CLOUD and AZURE_ENVIRONMENT_FILEPATH are not both set
```
matthchr, line 92: Just confirming that as proposed in design, we dont require AZURESTACKCLOUD for the ARM_CLOUD (which is different than CAS). Open to being convinced it's better to require it to be set for consistency sake.

rakechill: Did we talk to anyone about _why_ CAS sets it to that? I re-read the design and generally agree + it seems cleaner to have distinct implementations for known vs. not known clouds

pkg/auth/environment.go
rakechill: > Just bringing this up here though as technically we could share code


Do you mean just calling the function that cloud provider has + wrapping in a struct of our making w/ some additional fields (e.g. graph endpoint)?

pkg/auth/util.go
```diff
@@ -19,10 +19,17 @@ package auth
...
 
 func GetUserAgentExtension() string {
 	// Note: do not change "karpenter-aks/" prefix, some infra depends on it
 	return fmt.Sprintf("karpenter-aks/v%s", operator.Version)
 }
+
+// TokenScope returns the token scope for the Azure environment, such as "https://management.azure.com/.default" (for public cloud)
+func TokenScope(cloudCfg cloud.Configuration) string {
+	rm := cloudCfg.Services[cloud.ResourceManager]
+	return fmt.Sprintf("%s/.default", rm.Audience)
```
rakechill, line 34: Would love to see an e2e run using a snapshot just to confirm that the audience for public cloud is structured like this. I guess the deployment will need to have `ARM_CLOUD` set in that case.

pkg/operator/operator.go
```diff
@@ -224,7 +223,13 @@ func getCABundle(restConfig *rest.Config) (*string, error) {
 }
 
 func getVnetGUID(creds azcore.TokenCredential, cfg *auth.Config, subnetID string) (string, error) {
-	opts := armopts.DefaultArmOpts()
+	// TODO: We should possibly just put the vnet client on azclient and use it here
```
rakechill: little confused by this todo comment here. can you clarify?

pkg/providers/instance/azure_client.go
```diff
@@ -183,7 +186,8 @@ func NewAZClient(ctx context.Context, cfg *auth.Config, env *azclient.Environmen
 		}
 	}
 
-	return NewAZClientFromAPI(virtualMachinesClient,
```
rakechill: thank you for making style fixes like these!

pkg/test/environment.go
```diff
@@ -107,7 +108,7 @@ func NewRegionalEnvironment(ctx context.Context, env *coretest.Environment, regi
 	var auxiliaryTokenServer *fake.AuxiliaryTokenServer
 	if testOptions.UseSIG {
 		auxiliaryTokenServer = fake.NewAuxiliaryTokenServer("test-token", time.Now().Add(1*time.Hour), time.Now().Add(5*time.Minute))
-		auxTokenPolicy = auth.NewAuxiliaryTokenPolicy(auxiliaryTokenServer, testOptions.SIGAccessTokenServerURL, testOptions.SIGAccessTokenScope)
+		auxTokenPolicy = auth.NewAuxiliaryTokenPolicy(auxiliaryTokenServer, testOptions.SIGAccessTokenServerURL, auth.TokenScope(cloud.AzurePublic))
```
rakechill: maybe use the test env's cloud instead of manually passing in AzurePublic?

pkg/providers/instance/azure_client.go
matthchr: There's actually a linter/tool we can run that will just fix them all automatically (or at least most of them - import ordering, this one, a few others). I have a local set of changes to try to turn it on but just haven't gotten time to revisit it.

pkg/auth/environment.go
matthchr: I just mean that [OverrideAzureCloudConfigFromEnv](https://github.com/kubernetes-sigs/cloud-provider-azure/blob/master/pkg/azclient/cloud.go#L106) does _basically_ what this function does. It does it in a slightly different way, but doing very similar things.

My point here is just confirming we all agree to not use it.

pkg/auth/util.go
matthchr: We default to PublicCloud if nothing is set, so we don't need any deployment changes to test that (though I can make them in a standalone so if we did it wouldn't be a blocker).

Do you mean AKS E2E here? What's the difference between that and the E2E in this repo? presumably, if E2E for selfhosted passes here, we're good, since this token scope is also used for a bunch of selfhosted scenarios?

It does get some additional coverage for NAP-only scenarios - is that what you're getting at? Or selfhosted E2E works too?

pkg/test/environment.go
matthchr: Where is that defined?

`coretest.Environment` doesn't have a cloud field (it's from upstream Karpenter core so that makes sense).
We're constructing the Azure-specific env in this method and (as far as I can tell) there's no notion of cloud defined there either -- which does make _some_ sense because all the APIs are fake anyway so there's not a lot of need to differentiate one cloud from another.

ARM_CLOUD (and friends) aren't defined in `testOptions` because we read them in a different spot. I do think we should probably just combine how we read those env variables and how we read other options, but that felt a bit out of scope for this PR.

pkg/operator/operator.go
matthchr: Expanded to:
```
	// TODO: Current the VNET client isn't used anywhere but this method. As such, it is not
	// held on azclient like the other clients.
	// We should possibly just put the vnet client on azclient, and then pass azclient in here, rather than
	// constructing the VNET client here separate from all the other Azure clients.
```

Is that clearer?

pkg/auth/config.go
matthchr: I'll ask around.

pkg/test/environment.go
matthchr: I've changed this somewhat to just default to AzurePublic. I don't actually store that on the returned Environment at the moment because there isn't any need to, but it's used in a few places during test initialization which I _think_ is fine for now.

pkg/auth/config.go
matthchr: My assumption is that when the Track1 SDK was written, AzureStackCloud was the only "custom" cloud (not MC/FF), and so it was called that, and then other non-public clouds came along and the same logic was just borrowed for that as well, and CloudProvider/CAS/etc also just used this track1 SDK standard.
 
Then Track2 SDKs came along and did it totally differently, but it didn't matter because at that point CloudProvider was committed, which trickled up to CAS/others as well.

I spoke with Jack as well and he seemed to agree that the above was his understanding/assumption to.

pkg/auth/util.go
rakechill: I did mean AKS E2E. The idea being that we want to see that the deployment having these new vars set would help validate that we won't break public cloud. Accepting the defaults doesn't show us as much (but we still would see that client inits + endpoints + scopes still work)

pkg/test/environment.go
rakechill: Ah I think I got confused for a second and thought it was in the test options...looks fine to me.

pkg/auth/environment.go
rakechill: I'd say I'm neutral on having the logic duplicated here vs. calling OverrideAzureCloudConfigFromEnv within this func + additionally setting the graph endpoint given that we'd have to use AZURESTACKCLOUD iiuc.

pkg/auth/util.go
matthchr: I don't think there's actually any difference between AKS E2E and selfhosted E2E in this case because I default to public cloud [here](https://github.com/Azure/karpenter-provider-azure/pull/1094/files#diff-a055f299d1537638822e01418b4740b015220e0d6ea6abc51691b41b22ca650cR97). AKS E2E will test the deployment paths (to make sure we're passing AzurePublic there), but selfhosted also defaults to that so ideally it should be the same?

pkg/test/environment.go
```diff
@@ -104,19 +104,25 @@ func NewEnvironmentNonZonal(ctx context.Context, env *coretest.Environment) *Env
 func NewRegionalEnvironment(ctx context.Context, env *coretest.Environment, region string, nonZonal bool) *Environment {
 	testOptions := options.FromContext(ctx)
 
+	azureEnv, err := auth.EnvironmentFromName("AzurePublicCloud")
+	if err != nil {
+		// This really shouldn't happen
+		panic(err)
```
matthchr: lo.Must here?

pkg/auth/config.go
```diff
@@ -67,35 +67,33 @@ func BuildAzureConfig() (*Config, error) {
...
-}
-
 func (cfg *Config) Build() error {
 	// May require more than this behind the scenes: https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azidentity/README.md#defaultazurecredential
 	cfg.Cloud = strings.TrimSpace(os.Getenv("ARM_CLOUD"))
 	cfg.Location = strings.TrimSpace(os.Getenv("LOCATION"))
 	cfg.ResourceGroup = strings.TrimSpace(os.Getenv("ARM_RESOURCE_GROUP"))
 	cfg.TenantID = strings.TrimSpace(os.Getenv("ARM_TENANT_ID"))
 	cfg.SubscriptionID = strings.TrimSpace(os.Getenv("ARM_SUBSCRIPTION_ID"))
+	cfg.AzureEnvironmentFilepath = strings.TrimSpace(os.Getenv("AZURE_ENVIRONMENT_FILEPATH"))
```
Bryce-Soghigian, line 77: There is an argument to be made to move this CFG -> Options. 

on one hand its nice to have the separation of options + CFG, as most of these are only used for karpenter, then separately only used for azure auth. 


but for cases like MC_Resource_Group, and location, they belong as karpenter options too. Any thoughts on the two patterns for specifying options from env vars?

pkg/auth/environment.go
```diff
@@ -0,0 +1,156 @@
...
+// readEnvironmentFromFile reads environment configuration from a JSON file
+func readEnvironmentFromFile(filepath string) (*azclient.Environment, error) {
+	if filepath == "" {
+		return nil, fmt.Errorf("filepath is empty")
+	}
+	if !strings.HasPrefix(filepath, "/") {
+		return nil, fmt.Errorf("filepath must be absolute: %s", filepath)
+	}
+
+	// Read file content
```
Bryce-Soghigian:
```suggestion
```

pkg/auth/environment.go
```diff
@@ -0,0 +1,156 @@
...
+	content, err := os.ReadFile(filepath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read environment file %s: %w", filepath, err)
+	}
+
+	// Parse JSON content into azclient.Environment
```
Bryce-Soghigian:
```suggestion
```

pkg/auth/config.go
matthchr: Yeah, I had the same thought/question about this config versus options.

Options is a bit of a behemoth, but I think I agree that this should probably move there. I think it makes sense to do it separately though.

rakechill: It feels a little awkward, though, that cfg.Cloud and opts.AzureEnvironmentFilePath could control which env we resolve to eventually. Or are we suggesting the ENTIRE cfg struct become part of options?

matthchr: At least I was suggesting that just do away with `cfg` entirely (at least the env parsing part of it) and read al those opts from opts

We could still have `auth.Config` for the actual struct itself, but it would be configured via opts rather than doing its own os.getenv.

matthchr:
Ran 2 E2E's, neither was clean but also neither totally failed (which is what I would've expected if there was some issue with this PR in public cloud):
- https://github.com/Azure/karpenter-provider-azure/actions/runs/17219071963
- https://github.com/Azure/karpenter-provider-azure/actions/runs/17219068721

Between the two of them, these tests failed both times:
- [FAIL] Drift NodePool Drift [It] NodeRequirements

Both flaked in consolidation but on different tests.

another run: https://github.com/Azure/karpenter-provider-azure/actions/runs/17223153444

Drift didn't flake in this one, and the consolidation test suite failure seems to be due to runner shutdown or process termination not an actual test failure.

pkg/providers/pricing/suite_test.go
```diff
@@ -34,39 +34,36 @@ import (
...
+var mainCtx context.Context
 var ctx context.Context
 var stop context.CancelFunc
 
 var fakePricingAPI *fake.PricingAPI
 var env *auth.Environment
 
 func TestAzure(t *testing.T) {
-	ctx = TestContextWithLogger(t)
+	mainCtx = TestContextWithLogger(t)
```
rakechill, line 45: Why do we need two contexts here?

rakechill (approved):
Left one more comment for my own understanding, but this looks good to me after reviewing the e2e runs + looking over pricing + sku changes again

pkg/providers/pricing/suite_test.go
matthchr: Ah good question. I should put this in a comment.

So, I added two new tests:
- `should poll pricing data in public clouds`
- `should not poll pricing data in non-public clouds`

These actually launch the pricing provider (the async/pollign portion), which up-til now we had not done.

Launching the gorourtine for polling means I should stop it. Stopping it requires cancelling the context. If you cancel the root context then all subsequent tests will fail because that context is cancelled (note that mainCtx, which was called `ctx` before is NOT per Ginkgo test, but one global one for all the tests).

So we need a per-Gingko-test context we can cancel, while leaving the main context active the whole time.

I'll add a comment here explaining some of this.

matthchr: Added a comment:
```
	// Create and use ctx in the tests below rather than mainCtx because some of
	// the tests start the pricing poller. Stopping the poller requires cancelling the context
	// but if we cancel the main context, it will break tests that run afterwards.
	// We still need the mainCtx though because it attaches the test logger which we cannot do
	// in BeforeEach.
```

rakechill (dismissed):
lgtm, thanks!

pkg/auth/config.go
Bryce-Soghigian: Yeah agreed, auth config can still hold the values for creating a client but it would be populated by options rather than having two parts that read in from env

pkg/auth/config.go
```diff
@@ -97,3 +109,12 @@ func (cfg *Config) Validate() error {
 
 	return nil
 }
+
+func (cfg *Config) String() string {
+	json, err := json.Marshal(cfg)
+	if err != nil {
+		return "couldn't marshal Config JSON"
+	}
```
Bryce-Soghigian: https://pkg.go.dev/encoding/json#hdr-Migrating_to_v2 should we migrate to v2? I recently read an article on is. 

Separately, what errors can this return? Would it make sense to do something like 

fmt.Sprintf("couldn't marshal Config JSON, failed with : %s, err.Error())

This way we see the actual error type?

pkg/auth/environment.go
Bryce-Soghigian: Rather than creating this function, should we just migrate skewer off of track 1? https://github.com/Azure/skewer/pull/23 has an implementation for that, we just need to review it.

pkg/auth/config.go
matthchr: In our case v1 vs v2 is probably not _that_ big a deal, but yeah we should probably migrate eventually.

Will fix the error return, agreed.

pkg/auth/config_test.go
```diff
@@ -65,30 +68,77 @@ func TestBuildAzureConfig(t *testing.T) {
...
 		},
+		{
+			name:     "both ARM_CLOUD and AZURE_ENVIRONMENT_FILEPATH set",
+			expected: nil,
+			wantErr:  true,
+			env: map[string]string{
+				"ARM_RESOURCE_GROUP":         "my-rg",
+				"ARM_SUBSCRIPTION_ID":        "12345",
+				"AZURE_SUBNET_ID":            "12345",
+				"AZURE_SUBNET_NAME":          "my-subnet",
```
Bryce-Soghigian, line 79: If we ever move this to options, this will not work, same with subscripton id

pkg/auth/environment.go
```diff
@@ -0,0 +1,153 @@
...
+}
+
+// readEnvironmentFromFile reads environment configuration from a JSON file
+func readEnvironmentFromFile(filepath string) (*azclient.Environment, error) {
+	if filepath == "" {
+		return nil, fmt.Errorf("filepath is empty")
+	}
+	if !strings.HasPrefix(filepath, "/") {
+		return nil, fmt.Errorf("filepath must be absolute: %s", filepath)
+	}
```
Bryce-Soghigian: Are there existing go filepath helpers for this already?

pkg/auth/environment.go
```diff
@@ -0,0 +1,153 @@
...
+	// Cloud is the track 2 representation of the Azure environment
+	Cloud cloud.Configuration
+}
+
+// readEnvironmentFromFile reads environment configuration from a JSON file
+func readEnvironmentFromFile(filepath string) (*azclient.Environment, error) {
+	if filepath == "" {
+		return nil, fmt.Errorf("filepath is empty")
+	}
+	if !strings.HasPrefix(filepath, "/") {
```
Bryce-Soghigian:
```suggestion
	if !filepath.IsAbs(filepath)  {
```

pkg/auth/environment.go
```diff
@@ -0,0 +1,153 @@
...
+type Environment struct {
+	// Environment is the track 1 representation of the Azure environment
+	Environment *azclient.Environment
+	// Cloud is the track 2 representation of the Azure environment
+	Cloud cloud.Configuration
+}
+
+// readEnvironmentFromFile reads environment configuration from a JSON file
+func readEnvironmentFromFile(filepath string) (*azclient.Environment, error) {
+	if filepath == "" {
```
Bryce-Soghigian: would we use filepath clean as well? cleanPath := filepath.Clean(filepath)

pkg/auth/environment.go
matthchr: As discussed, we need this function regardless because this is the file-format that CloudProvider expects (and is what AKS writes to the nodes).

So we have to parse this file format. It happens to be the Track1 format but at this point you can think of it as the "CloudProvider format" if it makes you feel better (it makes me feel a bit better).

There is no supported Track2 file format so this one being "the one" is OK, IMO.

LMK if I should improve comments here... maybe I will just to make this much clearer.

pkg/auth/config_test.go
matthchr: Just because they are validated, yea? So we can't use dummy values they have to at least look real?

pkg/auth/environment.go
```diff
@@ -0,0 +1,153 @@
...
+		return nil, fmt.Errorf("filepath is empty")
+	}
+	if !strings.HasPrefix(filepath, "/") {
+		return nil, fmt.Errorf("filepath must be absolute: %s", filepath)
+	}
+
+	content, err := os.ReadFile(filepath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read environment file %s: %w", filepath, err)
+	}
```
Bryce-Soghigian: nit: should we use defer file.Close() with file, err := os.Open(filepath)? 

I could be convinced either way, os.ReadFile() is more readable but there might be some helpers for file validation on the file object

pkg/auth/environment.go
```diff
@@ -0,0 +1,153 @@
...
+
+	var env azclient.Environment
+	if err := json.Unmarshal(content, &env); err != nil {
+		return nil, fmt.Errorf("failed to parse environment file %s as JSON: %w", filepath, err)
+	}
+
+	if err := validateEnvironment(&env); err != nil {
```
Bryce-Soghigian: if !strings.HasSuffix(strings.ToLower(filepath), ".json") {
    return nil, fmt.Errorf("expected JSON file, got: %s", filepath)
}

Should we validate the file we have is json? There might be a better way to right this validation with file stub object.

pkg/auth/environment.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+	if env.ActiveDirectoryEndpoint == "" {
+		return fmt.Errorf("active directory endpoint is required")
+	}
+	if env.TokenAudience == "" {
+		return fmt.Errorf("token audience is required")
+	}
+	return nil
+}
+
```
Bryce-Soghigian, line 80: nit: We could add something like https://github.com/Azure/karpenter-provider-azure/blob/main/pkg/auth/config.go#L87 if we find we are validating a lot of fields on azure environment.

pkg/auth/environment.go
```diff
@@ -0,0 +1,153 @@
...
+		Cloud:       cloudConfig,
+	}, nil
+}
+
+// ResolveCloudEnvironment resolves the cloud environment using the following precedence:
+// 1. File-based environment (AZURE_ENVIRONMENT_FILEPATH)
+// 2. Known cloud names (ARM_CLOUD)
+// 3. Default (Azure Public Cloud)
+func ResolveCloudEnvironment(cfg *Config) (*Environment, error) {
+	// 1. Try file-based environment first (highest precedence)
```
Bryce-Soghigian, line 142: cfg is a pointer, it could be nil, are we confident from other parts of the code that its already defensively checked? Since ithis a public method for the auth package, maybe just validating 

if cfg == nil {
return nil, err
}

pkg/auth/environment.go
```diff
@@ -0,0 +1,153 @@
...
+	if cfg.AzureEnvironmentFilepath != "" {
+		env, err := readEnvironmentFromFile(cfg.AzureEnvironmentFilepath)
+		if err != nil {
+			return nil, fmt.Errorf("failed to read environment from file: %w", err)
+		}
+
+		cloudConfig, err := mapTrack1ToTrack2Environment(env)
+		if err != nil {
+			return nil, fmt.Errorf("failed to map environment to Track2 format: %w", err)
+		}
```
Bryce-Soghigian, line 152: I am wondering if we have a more standard way to do this , or if some of this shared auth code should be put in the azure-sdk-for-go-extensions, so we solve this problem once rather than resolving for each component.

pkg/auth/environment.go
```diff
@@ -0,0 +1,153 @@
...
+		}, nil
+	}
+
+	// 2. Try known cloud names (ARM_CLOUD)
+	if cfg.Cloud != "" {
+		return EnvironmentFromName(cfg.Cloud)
+	}
+
+	// 3. Default to Azure Public Cloud
+	return EnvironmentFromName("AzurePublicCloud")
```
Bryce-Soghigian, line 167: Thoughts on Adding the defaulting into a function itself? Where it defaults back into azure cloud? Or do you prefer the separation here?

pkg/providers/pricing/client/pricingapi.go
```diff
@@ -24,26 +24,35 @@ import (
...
+	return &pricingAPI{cloud: cloud}
 }
 
 func (papi *pricingAPI) GetProductsPricePages(_ context.Context, filters []*Filter, pageHandler func(output *ProductsPricePage)) error {
 	nextURL := pricingURL
 
+	if papi.cloud.Services[cloud.ResourceManager].Endpoint != cloud.AzurePublic.Services[cloud.ResourceManager].Endpoint {
+		// If the cloud is not Azure Public, the pricing API isn't supported and we return an error
+		return fmt.Errorf("pricing API is not supported in non-public clouds")
+	}
```
Bryce-Soghigian, line 54: should we return an error or use some fallback pregenerated pricing data?

pkg/providers/pricing/client/pricingapi.go
```diff
@@ -24,26 +24,35 @@ import (
 	"net/http"
 	"net/url"
 	"strings"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/cloud"
 )
 
 const (
 	apiVersion = "2023-01-01-preview"
```
Bryce-Soghigian, line 32: is this api still in preview? should we re-evaluate this? Edit: Reading through the docs looks like we still need the preview version for filtering we do

pkg/providers/pricing/suite_test.go
```diff
@@ -26,46 +26,52 @@ import (
...
 })
 
 var _ = BeforeEach(func() {
+	// Create and use ctx in the tests below rather than mainCtx because some of
+	// the tests start the pricing poller. Stopping the poller requires canceling the context
+	// but if we cancel the main context, it will break tests that run afterwards.
+	// We still need the mainCtx because it attaches the test logger which we cannot do
+	// in BeforeEach.
+	ctx, stop = context.WithCancel(mainCtx)
 	fakePricingAPI.Reset()
```
Bryce-Soghigian, line 64: Nice to see this fixed!

pkg/providers/pricing/client/pricingapi.go
```diff
@@ -24,26 +24,35 @@ import (
...
-func New() PricingAPI {
-	return &pricingAPI{}
+func New(cloud cloud.Configuration) PricingAPI {
+	return &pricingAPI{cloud: cloud}
 }
 
 func (papi *pricingAPI) GetProductsPricePages(_ context.Context, filters []*Filter, pageHandler func(output *ProductsPricePage)) error {
 	nextURL := pricingURL
 
+	if papi.cloud.Services[cloud.ResourceManager].Endpoint != cloud.AzurePublic.Services[cloud.ResourceManager].Endpoint {
```
Bryce-Soghigian, line 51: papi? Just realized we call the pricing api papi

Bryce-Soghigian (dismissed):
Approved with Suggestions

pkg/auth/environment.go
matthchr: I've updated the documentation here to make this clearer.

pkg/auth/config.go
matthchr: Not going to do this in this PR. We can pursue this in a followup PR

pkg/auth/config.go
matthchr: Fixed

pkg/auth/environment.go
matthchr: I don't think it's really necessary. `filepath.Clean` doesn't actually change the path presented, and we will either load or it won't. I guess the main thing it would do is change (slightly) the error text presented to the user? But in this case, it feels just as correct to present to the user the exact path they presented to us (un-cleaned), as it should be more obvious to them where that path came from.

pkg/auth/environment.go
matthchr: What sorts of things are you thinking we would want to validate?

My thinking is: We can either read the file bytes or we can't. If we can't, `os.ReadFile` should return us a reasonable user-facing error that explains what the issue is (no access, file doesn't exist, etc). We don't need to do that checking ourselves because all we care about is reading the file. The validations are useful for good error messages for the user but that should be taken care of by `os.ReadFile` anyway - so seems best to just use `os.ReadFile`?

pkg/auth/environment.go
matthchr: json extension doesn't mean json formatted, lack of json extension doesn't mean not json formatted.

It's not uncommon in unix to pass around files without an extension at all, so IMO it doesn't make sense to have this additional validation. We care about the bytes in the file that's it - not the name.

pkg/auth/environment.go
matthchr: Agreed - but don't think 3 is a lot so this seems fine as-is?

pkg/auth/environment.go
matthchr: Fixed

pkg/auth/environment.go
matthchr: There is some code that does something similar to this in CloudProvider, but it does it in a way that isn't great (IMO).
I talked about this some in [the design](https://github.com/Azure/karpenter-provider-azure/blob/main/designs/national-cloud-support.md)

I'll move a little snippet of that to a comment here as well to make that clearer, and leave a TODO that we could think about moving it upstream to either CloudProvider or azure-sdk-for-go-extensions). It would be nice to fix cloudprovider but the parts I like least about the cloudProvider impl it might be breaking to change/fix...

pkg/providers/pricing/client/pricingapi.go
matthchr: The pricing _provider_ already falls back to pregenerated pricing data. I think that's the right level for that fallback - the actual API should probably just map to the raw API.

pkg/providers/pricing/client/pricingapi.go
matthchr: Yes it is preview AFAIK.

pkg/auth/environment.go
matthchr: Not sure I understand what you're talking about.

We actually do already have defaulting in a function in `auth.Config` (see `Default`), which defaults this. This line shouldn't ever be hit in production because of that, but just have it as a sensible fallback anyway because it seemed better than erroring here.

pkg/auth/config_test.go
rakechill: yep that's what he's referring to + good callout

rakechill (approved):

PR #1102, opened 2025-08-13
[DO NOT MERGE] PoC: feat: AKS machine API integration

comtalyst:
DO NOT MERGE. PoC only.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,7 +11,7 @@ require (
 	github.com/Azure/azure-sdk-for-go/sdk/azidentity v1.10.1
 	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/compute/armcompute v1.0.0
 	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/compute/armcompute/v5 v5.7.0
-	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/containerservice/armcontainerservice/v4 v4.8.0
+	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/containerservice/armcontainerservice/v7 v7.3.0-beta.1
 	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/network/armnetwork v1.1.0
 	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/resourcegraph/armresourcegraph v0.9.0
 	github.com/Azure/go-autorest/autorest v0.11.30
--- a/pkg/apis/v1beta1/labels.go
+++ b/pkg/apis/v1beta1/labels.go
@@ -99,6 +99,7 @@ var (
 
 	AnnotationAKSNodeClassHash        = apis.Group + "/aksnodeclass-hash"
 	AnnotationAKSNodeClassHashVersion = apis.Group + "/aksnodeclass-hash-version"
+	AnnotationAKSMachineResourceID    = apis.Group + "/aks-machine-resource-id" // resource ID of the associated AKS machine
 )
 
 const (
--- a/pkg/cloudprovider/cloudprovider.go
+++ b/pkg/cloudprovider/cloudprovider.go
@@ -38,9 +38,11 @@ import (
 
 	// nolint SA1019 - deprecated package
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/compute/armcompute/v5"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/containerservice/armcontainerservice/v7"
 
 	"github.com/Azure/karpenter-provider-azure/pkg/apis"
 	"github.com/Azure/karpenter-provider-azure/pkg/apis/v1beta1"
+	"github.com/Azure/karpenter-provider-azure/pkg/consts"
 	"github.com/Azure/karpenter-provider-azure/pkg/controllers/nodeclaim/inplaceupdate"
 	"github.com/Azure/karpenter-provider-azure/pkg/operator/options"
 
@@ -49,6 +51,7 @@ import (
 	cloudproviderevents "github.com/Azure/karpenter-provider-azure/pkg/cloudprovider/events"
 	"github.com/Azure/karpenter-provider-azure/pkg/providers/imagefamily"
 	"github.com/Azure/karpenter-provider-azure/pkg/providers/instance"
+	"github.com/Azure/karpenter-provider-azure/pkg/providers/instance/offerings"
 	"github.com/Azure/karpenter-provider-azure/pkg/providers/instancetype"
 	"github.com/Azure/karpenter-provider-azure/pkg/providers/launchtemplate"
 	"github.com/Azure/karpenter-provider-azure/pkg/utils"
@@ -74,7 +77,8 @@ var _ cloudprovider.CloudProvider = (*CloudProvider)(nil)
 
 type CloudProvider struct {
 	instanceTypeProvider instancetype.Provider
-	instanceProvider     instance.Provider
+	vmInstanceProvider   instance.VMProvider // Note that even when provision mode does not create with VM instance provider, it is still being used to handle existing VM instances.
+	aksMachineProvider   instance.AKSMachineProvider
 	kubeClient           client.Client
 	imageProvider        imagefamily.NodeImageProvider
 	recorder             events.Recorder
@@ -82,42 +86,20 @@ type CloudProvider struct {
 
 func New(
 	instanceTypeProvider instancetype.Provider,
-	instanceProvider instance.Provider,
+	vmInstanceProvider instance.VMProvider,
+	aksMachineProvider instance.AKSMachineProvider,
 	recorder events.Recorder,
 	kubeClient client.Client,
 	imageProvider imagefamily.NodeImageProvider,
 ) *CloudProvider {
 	return &CloudProvider{
 		instanceTypeProvider: instanceTypeProvider,
-		instanceProvider:     instanceProvider,
+		vmInstanceProvider:   vmInstanceProvider,
+		aksMachineProvider:   aksMachineProvider,
 		kubeClient:           kubeClient,
 		imageProvider:        imageProvider,
 		recorder:             recorder,
 	}
-}
-
-// Create a node given the constraints.
-func (c *CloudProvider) handleInstanceCreation(ctx context.Context, instancePromise *instance.VirtualMachinePromise, nodeClaim *karpv1.NodeClaim) error {
-	if c.isStandaloneNodeClaim(nodeClaim) {
-		// processStandaloneNodeClaimDeletion:
-		//   Standalone NodeClaims aren’t re-queued for reconciliation in the provision_trigger controller,
-		//   so we delete them synchronously. After marking Launched=true,
-		//   their status can’t be reverted to false once the delete completes due to how core caches nodeclaims in
-		// 	 the lanch controller. This ensures we retry continuously until we hit the registration TTL
-		err := instancePromise.Wait()
-		if err != nil {
-			return c.handleNodeClaimCreationError(ctx, err, instancePromise, nodeClaim, false)
-		}
-	} else {
-		// For NodePool-managed nodeclaims, launch a single goroutine to poll the returned promise.
-		// Note that we could store the LRO details on the NodeClaim, but we don't bother today because Karpenter
-		// crashes should be rare, and even in the case of a crash, as long as the node comes up successfully there's
-		// no issue. If the node doesn't come up successfully in that case, the node and the linked claim will
-		// be garbage collected after the TTL, but the cause of the nodes issue will be lost, as the LRO URL was
-		// only held in memory.
-		go c.waitOnPromise(ctx, instancePromise, nodeClaim)
-	}
-	return nil
 }
 
 func (c *CloudProvider) validateNodeClass(nodeClass *v1beta1.AKSNodeClass) error {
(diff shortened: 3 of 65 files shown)


pkg/utils/image.go
```diff
@@ -0,0 +1,56 @@
...
+
+// WARNING: not supporting CIG images yet. Awaiting AKS machine API support.
+func GetAKSMachineNodeImageVersionFromImageID(imageID string) (string, error) {
+	if strings.HasPrefix(imageID, "/CommunityGalleries") {
+		// Awaiting AKS machine API support
+		return "", fmt.Errorf("CIG images are not supported yet for AKS machines, consider not using PROVISION_MODE=aksmachineapi: %s", imageID)
+	} else {
+		return GetAKSMachineNodeImageVersionFromSIGImageID(imageID)
+	}
+}
```
Bryce-Soghigian, line 33: ideally we would have some image parsing helpers for the different places we parse and manage images, similar to the VnetSubnetIDComponents helper

Bryce-Soghigian (commented):
👏 wow you turned that around fast!!

pkg/providers/instance/aksmachineinstance.go
```diff
@@ -0,0 +1,580 @@
...
+}
+
+func (p *AKSMachinePromise) Wait() error {
+	return p.waitFunc()
+}
+func (p *AKSMachinePromise) GetInstanceName() string {
+	return p.AKSMachineName
+}
+
+type AKSMachineProvider interface {
```
matthchr, line 105: Can we not have a single provider interface and two impls, rather than 2 interfaces?

matthchr: This interface looks very similar except it's lacking the NIC methods - but you could make the argument that the NIC methods should be on its own provider anyway probably?

I guess the main issue then is that we've got two different parameter types because of `armcompute.VirtualMachine` as a response type/input type compared to `armcontainerservice.Machine`?

Still think it may be worth looking into?

go.mod
```diff
@@ -11,7 +11,7 @@ require (
 	github.com/Azure/azure-sdk-for-go/sdk/azidentity v1.10.1
 	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/compute/armcompute v1.0.0
 	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/compute/armcompute/v5 v5.7.0
-	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/containerservice/armcontainerservice/v4 v4.8.0
+	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/containerservice/armcontainerservice/v7 v7.3.0-beta.1
```
matthchr, line 14: Should be split out into a separate PR probably that just does this?

pkg/apis/v1beta1/labels.go
```diff
@@ -99,6 +99,7 @@ var (
 
 	AnnotationAKSNodeClassHash        = apis.Group + "/aksnodeclass-hash"
 	AnnotationAKSNodeClassHashVersion = apis.Group + "/aksnodeclass-hash-version"
+	AnnotationAKSMachineResourceID    = apis.Group + "/aks-machine-resource-id" // resource ID of the associated AKS machine
```
matthchr, line 102: Do we need this duplicated in the v1alpha package too? (Probably not but just checking)

pkg/providers/instance/aksmachineinstance.go
matthchr: Ok I looked more and I understand - there isn't a ton of value in the above atm because we have both coexisting anyway.

pkg/cloudprovider/cloudprovider.go
```diff
@@ -169,56 +151,158 @@ func (c *CloudProvider) Create(ctx context.Context, nodeClaim *karpv1.NodeClaim)
 	if len(instanceTypes) == 0 {
 		return nil, cloudprovider.NewInsufficientCapacityError(fmt.Errorf("all requested instance types were unavailable during launch"))
 	}
-	instancePromise, err := c.instanceProvider.BeginCreate(ctx, nodeClass, nodeClaim, instanceTypes)
+
+	// Choose provider based on provision mode
+	if options.FromContext(ctx).ProvisionMode == consts.ProvisionModeAKSMachineAPI {
```
matthchr, line 156: Wondering if we want an interface that abstracts over at least this part so that we don't need this `options.FromContext(ctx).ProvisionMode == consts.ProvisionModeAKSMachineAPI` everywhere.

pkg/providers/instance/aksmachineinstance.go
```diff
@@ -0,0 +1,580 @@
...
+	"github.com/Azure/karpenter-provider-azure/pkg/providers/instancetype"
+	"github.com/Azure/karpenter-provider-azure/pkg/utils"
+)
+
+var (
+	NodePoolTagKey = strings.ReplaceAll(karpv1.NodePoolLabelKey, "/", "_")
+)
+
+// Intended for lifecycle handling on the higher abstractions.
+type InstancePromise interface {
```
matthchr: minor: consider pulling into separate file rather than with machine, since it's used for both?

pkg/cloudprovider/cloudprovider.go
```diff
@@ -169,56 +151,158 @@ func (c *CloudProvider) Create(ctx context.Context, nodeClaim *karpv1.NodeClaim)
...
 		return nil, cloudprovider.NewCreateError(fmt.Errorf("creating instance failed, %w", err), CreateInstanceFailedReason, truncateMessage(err.Error()))
 	}
 
-	if err = c.handleInstanceCreation(ctx, instancePromise, nodeClaim); err != nil {
+	if err := c.handleInstancePromise(ctx, vmPromise, nodeClaim); err != nil {
 		return nil, err
 	}
 
-	instance := instancePromise.VM
+	vm := vmPromise.VM // This may not be thoroughly populated
```
matthchr: minor: Comment needs clarity? What does this mean?

pkg/cloudprovider/cloudprovider.go
```diff
@@ -169,56 +151,158 @@ func (c *CloudProvider) Create(ctx context.Context, nodeClaim *karpv1.NodeClaim)
...
 	})
+	newNodeClaim, err := c.vmInstanceToNodeClaim(ctx, vm, instanceType)
+	if err != nil {
+		return nil, err
+	}
+	inPlaceUpdateHash, err := inplaceupdate.HashFromNodeClaim(options.FromContext(ctx), newNodeClaim, nodeClass)
+	if err != nil {
+		return nil, fmt.Errorf("failed to calculate in place update hash, %w", err)
+	}
+	newNodeClaim.Annotations = lo.Assign(newNodeClaim.Annotations, map[string]string{
```
matthchr: minor: code is duplicated with below - possibly extract to func

pkg/controllers/nodeclaim/garbagecollection/instance_garbagecollection.go
```diff
@@ -40,29 +40,29 @@ import (
 	corecloudprovider "sigs.k8s.io/karpenter/pkg/cloudprovider"
 )
 
-type VirtualMachine struct {
+type CloudProviderInstances struct {
```
matthchr, line 43: minor: Good to split this renaming stuff out into it's own PR when you're ready to promote from draft as well

pkg/controllers/nodeclaim/garbagecollection/suite_test.go
```diff
@@ -306,14 +308,190 @@ var _ = Describe("VirtualMachine Garbage Collection", func() {
...
+			ExpectSingletonReconciled(ctx, cloudProviderInstancesGCController)
 			_, err = cloudProvider.Get(ctx, providerID)
 			Expect(err).To(HaveOccurred())
 			Expect(corecloudprovider.IsNodeClaimNotFoundError(err)).To(BeTrue())
 
 			ExpectNotFound(ctx, env.Client, node)
 		})
 	})
+
+	var _ = Context("AKS Machine instances", func() {
```
matthchr: Probably makes sense to split these tests into different file. Applies to all the machine vs vm unit tests.

pkg/controllers/nodeclaim/inplaceupdate/controller.go
```diff
@@ -158,6 +176,31 @@ func (c *Controller) applyPatch(
...
+) error {
+	update := CalculateAKSMachinePatch(options, nodeClaim, nodeClass, aksMachine)
+	// This is safe only as long as we're not updating fields which we consider secret.
+	// If we do/are, we need to redact them.
+	logAKSMachinePatch(ctx, update)
+
+	// Apply the update, if one is needed
+	if update != nil {
+		// Given AKS machine support PUT, but not PATCH, the AKS machine object will be patched directly.
+		err := c.aksMachineInstanceProvider.Update(ctx, aksMachineName, *aksMachine)
```
matthchr: May need to check etag here, if machiune supports etag?

If it doesn't maybe leave a TODO here and raise it with Xu? GET + modify + PUT isn't concurrency-safe without an etag

pkg/controllers/nodeclaim/inplaceupdate/controller.go
```diff
@@ -36,24 +36,28 @@ import (
...
 	kubeClient client.Client,
-	instanceProvider instance.Provider,
+	vmInstanceProvider instance.VMProvider,
+	aksMachineInstanceProvider instance.AKSMachineProvider,
 ) *Controller {
 	return &Controller{
-		kubeClient:       kubeClient,
-		instanceProvider: instanceProvider,
+		kubeClient:                 kubeClient,
+		vmInstanceProvider:         vmInstanceProvider,
```
matthchr, line 59: I wonder if it might be better if we made a higher level provider that held both of these.

That would:
- Save you having to pass 2 everywhere (easier to manage them when we always treat them as a unit on their own struct w/ 2 fields)
- Give you a place to write higher level constructs like "GetAndPatch" - though to do this you would need to also abstract the vm type and the machine type which might be annoying, so doing this part may not be worth it. I still thing putting them on a struct together and passing that everywhere probably is though?

pkg/controllers/nodeclaim/inplaceupdate/patch.go
```diff
@@ -133,3 +179,50 @@ func patchTags(
 	update.Tags = expectedTags
 	return true
 }
+
+func patchAKSMachineTags(
```
matthchr, line 184: Does AKS RP inject tags onto machine in addition to what we put?

I assume no...? But you could see the argument for it doing so since AKS RP does for cluster -> VMSS tags for example. Do tags trigger drift?

pkg/controllers/nodeclaim/inplaceupdate/suite_test.go
```diff
@@ -178,15 +182,369 @@ var _ = Describe("Unit tests", func() {
...
+			}
+			hash2, err := inplaceupdate.HashFromNodeClaim(options, aksMachineNodeClaim, nodeClass)
+			Expect(err).ToNot(HaveOccurred())
+
+			// Should be different because AKS machines don't include identities in hash
+			Expect(hash1).ToNot(Equal(hash2))
+		})
+	})
+
+	Context("calculateAKSMachinePatch", func() {
```
matthchr, line 223: again IMO split to different file

pkg/cloudprovider/cloudprovider.go
```diff
@@ -169,56 +151,158 @@ func (c *CloudProvider) Create(ctx context.Context, nodeClaim *karpv1.NodeClaim)
...
+	}
+
+	// Handle the promise
+	if err := c.handleInstancePromise(ctx, aksMachinePromise, nodeClaim); err != nil {
 		return nil, err
 	}
-	inPlaceUpdateHash, err := inplaceupdate.HashFromNodeClaim(options.FromContext(ctx), nc, nodeClass)
+
+	// Convert the AKS machine to a NodeClaim
+	newNodeClaim, err := instance.BuildNodeClaimFromAKSMachineTemplate(
```
matthchr, line 201: Talked with @tallaxes and @Bryce-Soghigian in the past about getting rid of the "template" construct entirely because it doesn't really seem to get us much and adds (IMO) a bunch of book keeping. Doesn't make sense for us to have it like it does for AWS because in AWS there really is a template resource but here we do not have one.

It may make sense to think a bit about this and for this template stuff do what we would like to refactor to for VMs for machines right out of the gate. I don't know exactly what the formulation of that is but possibly we can discuss/design it pretty quickly.

pkg/cloudprovider/cloudprovider.go
```diff
@@ -169,56 +151,158 @@ func (c *CloudProvider) Create(ctx context.Context, nodeClaim *karpv1.NodeClaim)
...
 		v1beta1.AnnotationAKSNodeClassHash:        nodeClass.Hash(),
 		v1beta1.AnnotationAKSNodeClassHashVersion: v1beta1.AKSNodeClassHashVersion,
 		v1beta1.AnnotationInPlaceUpdateHash:       inPlaceUpdateHash,
 	})
-	return nc, nil
+
+	return newNodeClaim, nil
 }
 
-func (c *CloudProvider) waitOnPromise(ctx context.Context, promise *instance.VirtualMachinePromise, nodeClaim *karpv1.NodeClaim) {
```
matthchr, line 202: Not totally following why dropping this helper makes sense

pkg/cloudprovider/drift.go
```diff
@@ -43,21 +45,40 @@ const (
...
-		c.isSubnetDrifted,
+
+	if nodeClaim == nil || nodeClaim.Status.ProviderID == "" {
+		return "", fmt.Errorf("nodeclaim %s is missing provider ID", nodeClaim.Name)
+	}
+
+	var checks []func(ctx context.Context, nodeClaim *karpv1.NodeClaim, nodeClass *v1beta1.AKSNodeClass) (cloudprovider.DriftReason, error)
+	if _, isAKSMachine := instance.GetAKSMachineNameFromNodeClaim(nodeClaim); isAKSMachine {
+		// XPMT: (topic) What to drift?
+		checks = []func(ctx context.Context, nodeClaim *karpv1.NodeClaim, nodeClass *v1beta1.AKSNodeClass) (cloudprovider.DriftReason, error){
```
matthchr, line 65: Isn't drift easier when RP just tells us if stuff is drifted? I thought machine API was supposed to handle a bunch of this.

pkg/cloudprovider/suite_test.go
```diff
@@ -391,4 +2578,263 @@ var _ = Describe("CloudProvider", func() {
...
+				allNodeClaims, err := cloudProvider.List(ctx)
+				Expect(err).To(HaveOccurred())
+				Expect(allNodeClaims).To(BeEmpty())
+
+				// Clear the error for cleanup
+				azureEnv.AKSMachinesAPI.AKSMachineNewListPagerBehavior.Error.Set(nil)
+			})
+		})
+
+		Context("AKS Machines Pool Management", func() {
```
matthchr: Test file is huge, should probably be split up into sub-files?

pkg/operator/options/options.go
```diff
@@ -90,6 +90,7 @@ type Options struct {
 	SIGSubscriptionID          string            `json:"sigSubscriptionId,omitempty"`
 	NodeResourceGroup          string            `json:"nodeResourceGroup,omitempty"`
 	AdditionalTags             map[string]string `json:"additionalTags,omitempty"`
+	AKSMachinesPoolName        string            `json:"AKSMachinesPoolName,omitempty"` // The name of the agent pool for the AKS machine API, assuming that all machines belong to the same agent pool. Only used on AKS machine API provision mode.
```
matthchr: your json casing is wrong here.

pkg/operator/options/options.go
```diff
@@ -106,11 +107,11 @@ func (o *Options) AddFlags(fs *coreoptions.FlagSet) {
...
 	fs.StringVar(&o.NodeResourceGroup, "node-resource-group", env.WithDefaultString("AZURE_NODE_RESOURCE_GROUP", ""), "[REQUIRED] the resource group created and managed by AKS where the nodes live")
 	fs.StringVar(&o.KubeletIdentityClientID, "kubelet-identity-client-id", env.WithDefaultString("KUBELET_IDENTITY_CLIENT_ID", ""), "The client ID of the kubelet identity.")
-	fs.BoolVar(&o.UseSIG, "use-sig", env.WithDefaultBool("USE_SIG", false), "If set to true karpenter will use the AKS managed shared image galleries and the node image versions api. If set to false karpenter will use community image galleries. Only a subset of image features will be available in the community image galleries and this flag is only for the managed node provisioning addon.")
+	fs.BoolVar(&o.UseSIG, "use-sig", env.WithDefaultBool("USE_SIG", true), "If set to true karpenter will use the AKS managed shared image galleries and the node image versions api. If set to false karpenter will use community image galleries. Only a subset of image features will be available in the community image galleries and this flag is only for the managed node provisioning addon.")
```
matthchr: Some of these changes dont seem right for selfhosted - I assume just for testing?

pkg/providers/imagefamily/resolver.go
```diff
@@ -95,25 +100,25 @@ func NewDefaultResolver(_ client.Client, imageProvider *provider, instanceTypePr
 	}
 }
 
-// Resolve fills in dynamic launch template parameters
+// Resolve fills in dynamic launch template parameters.
+// The name "imageFamilyResolver.Resolve()" is potentially misleading here.
+// Suggestion: refactor would help, but this won't be used by PROVISION_MODE=aksmachineapi anyway. May not be worth it.
+// ATTENTION!!!: changes here may NOT be effective on AKS machine nodes (ProvisionModeAKSMachineAPI); See aksmachineinstance.go/aksmachineinstancehelpers.go.
```
matthchr, line 109: Agree this is sorta confusing. Also not sure if refactor is worth it, but splitting these specific changes into a smaller PR and/or looking more closely at just these bits may give us a better idea.

pkg/providers/instance/aksmachineinstance.go
```diff
@@ -0,0 +1,580 @@
...
+) (*AKSMachinePromise, error) {
+	// Reuse the existing AKS machine if it exists, and skip the creation.
+	// Note: currently, we do not support different offerings requirements for the NodeClaim with the same name that attempted creation recently. The same applies with VM-based provisioning.
+	// This supported case is often a result of Karpenter restarting while the node is being created, in which the NodeClaim to create stays the same after restart.
+	resp, err := p.azClient.aksMachinesClient.Get(ctx, p.clusterResourceGroup, p.clusterName, p.aksMachinesPoolName, aksMachineName, nil)
+	if err == nil {
+		// Existing AKS machine found, reuse it.
+
+		// Reconstruct properties from existing AKS machine instance.
+		if len(resp.Machine.Zones) == 0 || resp.Machine.Zones[0] == nil {
```
matthchr: minor: could split this validation into a validate helper function

matthchr: This seems strange/wrong to me possibly?

What if machine exists but wrong configuration? Returning success for beginCreate seems incorrect then? Plus API should be idempotent if we PUT w/ same configuration again, so why do we need to do this GET check at all?

pkg/providers/instance/aksmachineinstance.go
```diff
@@ -0,0 +1,580 @@
...
+	header.Set("OSImageVersion", osImageVersion)                    // E.g., "2022.10.03"
+	ctxWithHeader := context.WithValue(ctx, "vmimageid", vmImageID) // This line is really just for testing (see fake/aksmachinesapi.go). Azure-sdk-for-go is restrictive in extracting the header out.
+	ctxWithHeader = policy.WithHTTPHeader(ctxWithHeader, header)
+
+	poller, err := p.azClient.aksMachinesClient.BeginCreateOrUpdate(ctxWithHeader, p.clusterResourceGroup, p.clusterName, p.aksMachinesPoolName, aksMachineName, *aksMachineTemplate, nil)
+	if err != nil {
+		return nil, fmt.Errorf("aksMachine.BeginCreateOrUpdate for AKS machine %q failed: %w", aksMachineName, err)
+	}
+
+	// Get once after begin create to retrieve VMResourceID.
```
matthchr, line 429: minor: Worth a TODO that we can probably extract this from response to BeginCreateOrUpdate anbd about second GET if we need to

pkg/providers/instance/aksmachineinstance.go
```diff
@@ -0,0 +1,580 @@
...
+	if gotAKSMachine.Properties.ResourceID == nil {
+		return nil, fmt.Errorf("failed to get AKS machine instance %q once after begin creation: AKS machine VM resource ID is nil", aksMachineName)
+	}
+	if gotAKSMachine.Properties.Status == nil || gotAKSMachine.Properties.Status.CreationTimestamp == nil {
+		return nil, fmt.Errorf("failed to get AKS machine instance %q once after begin creation: AKS machine creation timestamp is nil", aksMachineName)
+	}
+	if gotAKSMachine.Properties.NodeImageVersion == nil {
+		return nil, fmt.Errorf("failed to get AKS machine instance %q once after begin creation: AKS machine node image version is nil", aksMachineName)
+	}
+	if gotAKSMachine.Properties.ProvisioningState != nil && lo.FromPtr(gotAKSMachine.Properties.ProvisioningState) == "Failed" {
```
matthchr, line 440: Shoul;d have consts for terminal ProvStates - I assume SDK already defines them - use them

matthchr (commented):
Didnt go through every file in depth but left some general comments. I think I got at least most of the really important files.

pkg/operator/options/options.go
Bryce-Soghigian: use-sig in a sense makes sense to be set with machine api, because machine api is using shared image gallery images.

pkg/providers/instance/aksmachineinstance.go
comtalyst: > Can we not have a single provider interface and two impls, rather than 2 interfaces?

The current approach is that they don't share an interface already

> I guess the main issue then is that we've got two different parameter types because of armcompute.VirtualMachine as a response type/input type compared to armcontainerservice.Machine?

That and NIC is the main issues, and yes, they will coexist.

pkg/apis/v1beta1/labels.go
comtalyst: Don't think so?

pkg/controllers/nodeclaim/garbagecollection/instance_garbagecollection.go
comtalyst: Yep that's the plan

pkg/cloudprovider/drift.go
comtalyst: That is `isMachineDriftActionAvailable`, handling RP/cluster-controlled fields. We barely have any as of today. 
We still need to handle drift on Karpenter-controlled fields regardless of that.

pkg/cloudprovider/cloudprovider.go
comtalyst: More like merging with `handleInstanceCreation()`. Primarily due to less distinguishable (IMO) responsibility/abstraction of `handleInstanceCreation()` (it seems to be a thin layer than can be included in "handling async phase" umbrella rather than a separate function with additional cross-function assumption on `handleNodeClaimCreationError()`), plus some complications around VM/Machine-specific logic.

pkg/cloudprovider/cloudprovider.go
comtalyst: I think there is a misunderstanding here, although I probably have made naming confusing.

The "template" you meant probably refers to `launchtemplate`, right? That concept is not being used here for the reason you mentioned.
Now, this "template" here is in fact, the Machine struct. The concept "template" is there to distinguish between Karpenter-populated struct and the struct we retrieve from the API, which is "nourished" with status fields etc. In fact, the same pattern exists in VM instance provider even without considering `launchtemplate`, just that it wasn't named "template".

As for why there is this distinguishment, it is to make clear that this struct may not be thoroughly populated, preventing engineers/code from making assumptions on the validity of some fields, especially status fields. For example, if something on the other side of [this](https://github.com/Azure/karpenter-provider-azure/blob/d0207f537e79d6f3bcf0a5ecaa348714eb57e0cc/pkg/providers/instance/instance.go#L707-L712) were to extend. Thoughts?

For the naming, if `launchtemplate` never exists, I thought "template" could have been a perfect name here... Hopefully there is a better one coming.

pkg/cloudprovider/suite_test.go
comtalyst: This file is pending rework (can ignore reviewing this file for now).

pkg/operator/options/options.go
comtalyst: Need to be careful when replacing strings...

pkg/operator/options/options.go
comtalyst: This is temporary for testing. Changing the defaults should be a separate consideration--as long as testing is still convenient.

pkg/providers/instance/aksmachineinstance.go
comtalyst: > Plus API should be idempotent if we PUT w/ same configuration again

This check occurs before Karpenter select SKU/priority/zone. They are not necessarily deterministic, and immutable in Machine API. PUT again would fail if they change, even though both are correct per NodeClaim requirements (which gives the list of applicable ones than the selected one). 

> so why do we need to do this GET check at all?

Unless we handle that PUT failure instead (which is more complicated, as it is post-selection), we would waste a machine, which will be in numbers in the case where Karpenter restarts during mass provisioning, hurting performance. That was the case in past scale tests.

> What if machine exists but wrong configuration? Returning success for beginCreate seems incorrect then?

If the configuration is wrong, then drift would be detected just like other machines?
The same issue and resolution exist on the race condition between configuration changes and ongoing provisioning.

pkg/providers/instance/aksmachineinstance.go
comtalyst: > assume SDK already defines them

Well, that's not true. Unless my search has been broken...

pkg/controllers/nodeclaim/inplaceupdate/patch.go
comtalyst: No, at least not onto the machine object. But yes, makes sense for RP to do so on VM/VMSS/abstracted resources.

pkg/cloudprovider/cloudprovider.go
comtalyst: Same concern with [the thread below](https://github.com/Azure/karpenter-provider-azure/pull/1102#discussion_r2329262722). But will add more clarification.

pkg/controllers/nodeclaim/inplaceupdate/controller.go
comtalyst: Great catch. Implemented.

pkg/cloudprovider/cloudprovider.go
comtalyst: That's actually the only one. `Delete`/`IsDrifted`/`Get`/ is per nature of the given NodeClaim. `List` will list both regardless.

But, to the point that "decision + calling" should be abstracted, I don't think there is much value from it:
image

- cloudprovider.go: it is already one use for each, to the intuitively matching method
- drift.go: each type needs different handling, so better make it clear
- inplaceupdate/controller.go: see another thread

pkg/controllers/nodeclaim/inplaceupdate/controller.go
comtalyst: I don't think so, primarily due to the fact that they are not similar enough, either generally or specific to this module.

> Give you a place to write higher level constructs like "GetAndPatch" - though to do this you would need to also abstract the vm type and the machine type which might be annoying, so doing this part may not be worth it. 

Yes, plus etag as a result of PATCH vs PUT.

> I still thing putting them on a struct together and passing that everywhere probably is though?

There is not much reuse of each instance operation (one get, one update each), so that logic, plus determining if an instance is machine or VM, has to be somewhere once anyway. I think there is not much value from shared abstraction compared to the more clarity/directness we are getting.

PR #1105, opened 2025-08-15
feat: add zone provider

### LINKED ISSUE #826: Karpenter determines whether a region is zonal from a manually-maintained list, resulting in delayed updates and maintenance burden

Follow-up of #731. There should be a better way than doing https://github.com/Azure/karpenter-provider-azure/pull/825 every time. Maybe a script to generate the list, if not a GitHub automation. 

Having to maintain the list manually means updates could be missed or not delivered on time, not to mention the waste of brainpower on an uninnovative chore.

No

### Additional Context

This would benefit both NAP and self-hosted Karpenter.

### Attachments

_No response_

### Community Note

matthchr:
Calls the list subscriptions API and dynamically determines if a region is zonal or not

Fixes #826.

**Description**

* Unit tests
* E2E tests: https://github.com/Azure/karpenter-provider-azure/actions/runs/17003024324

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
source region zonal information from API, with fallback to hardcoded list if API is unavailable.
```

### DIFF
--- a/hack/codegen.sh
+++ b/hack/codegen.sh
@@ -18,6 +18,17 @@ pricing() {
   checkForUpdates "${GIT_DIFF}" "${NO_UPDATE}" "${SUBJECT} beside timestamps since last update" "${GENERATED_FILE}"
 }
 
+locationsgen() {
+  GENERATED_FILE="pkg/fake/locations.json"
+  NO_UPDATE=$' pkg/fake/locations.json | 2 +-\n 1 file changed, 1 insertion(+), 1 deletion(-)'
+  SUBJECT="Locations"
+
+  go run hack/code/locations_gen/main.go -- "${GENERATED_FILE}"
+
+  GIT_DIFF=$(git diff --stat "${GENERATED_FILE}")
+  checkForUpdates "${GIT_DIFF}" "${NO_UPDATE}" "${SUBJECT} beside timestamps since last update" "${GENERATED_FILE}"
+}
+
 skugen() {
   location=${1:-eastus}
 
@@ -41,9 +52,9 @@ skugen() {
 
 
 skugen-all() {
-  SUBSCRIPTION_ID=$(az account show --query 'id' --output tsv)
-  export SUBSCRIPTION_ID
-  if [ -z "${SUBSCRIPTION_ID}" ]; then
+  AZURE_SUBSCRIPTION_ID=$(az account show --query 'id' --output tsv)
+  export AZURE_SUBSCRIPTION_ID
+  if [ -z "${AZURE_SUBSCRIPTION_ID}" ]; then
     echo "No subscription is set. Please login and set a subscription."
     exit 1
   fi
@@ -95,4 +106,5 @@ fi
 
 # Run all the codegen scripts
 pricing
+locationsgen
 skugen-all
--- a/pkg/fake/subscriptionsapi.go
+++ b/pkg/fake/subscriptionsapi.go
@@ -18,9 +18,9 @@ package fake
 
 import (
 	"context"
+	_ "embed"
 	"encoding/json"
 	"fmt"
-	"os"
 	"sort"
 	"sync"
 
@@ -30,6 +30,9 @@ import (
 	"github.com/samber/lo"
 )
 
+//go:embed locations.json
+var fakeLocationsJSON string
+
 type ListLocationsInput struct {
 	SubscriptionID string
 	Options        *armsubscriptions.ClientListLocationsOptions
@@ -38,7 +41,7 @@ type ListLocationsInput struct {
 type SubscriptionsAPIBehavior struct {
 	NewListLocationsPagerBehavior MockedFunction[ListLocationsInput, armsubscriptions.ClientListLocationsResponse]
 	Locations                     sync.Map
-	FakeDataFilepath              string
+	UseFakeData                   bool
 }
 
 var _ zone.SubscriptionsAPI = &SubscriptionsAPI{}
@@ -52,7 +55,7 @@ func NewSubscriptionsAPI() (*SubscriptionsAPI, error) {
 		SubscriptionsAPIBehavior: SubscriptionsAPIBehavior{
 			NewListLocationsPagerBehavior: MockedFunction[ListLocationsInput, armsubscriptions.ClientListLocationsResponse]{},
 			Locations:                     sync.Map{},
-			FakeDataFilepath:              "pkg/fake/locations.json",
+			UseFakeData:                   true,
 		},
 	}
 
@@ -69,7 +72,7 @@ func (api *SubscriptionsAPI) Reset() {
 		api.Locations.Delete(k)
 		return true
 	})
-	if api.FakeDataFilepath != "" {
+	if api.UseFakeData {
 		err := loadLocationsFromFile(api)
 		if err != nil {
 			panic(err) // Not ideal, but shouldn't happen
@@ -120,13 +123,10 @@ func (api *SubscriptionsAPI) NewListLocationsPager(
 }
 
 func loadLocationsFromFile(api *SubscriptionsAPI) error {
-	data, err := os.ReadFile(api.FakeDataFilepath)
-	if err != nil {
-		return fmt.Errorf("failed to read locations from file %s: %w", api.FakeDataFilepath, err)
-	}
+	data := []byte(fakeLocationsJSON)
 
 	var locations []*armsubscriptions.Location
-	err = json.Unmarshal(data, &locations)
+	err := json.Unmarshal(data, &locations)
 	if err != nil {
 		return fmt.Errorf("failed to unmarshal locations JSON: %w", err)
 	}
--- a/pkg/fake/subscriptionsapi.go
+++ b/pkg/fake/subscriptionsapi.go
@@ -18,6 +18,9 @@ package fake
 
 import (
 	"context"
+	"encoding/json"
+	"fmt"
+	"os"
 	"sort"
 	"sync"
 
@@ -35,6 +38,7 @@ type ListLocationsInput struct {
 type SubscriptionsAPIBehavior struct {
 	NewListLocationsPagerBehavior MockedFunction[ListLocationsInput, armsubscriptions.ClientListLocationsResponse]
 	Locations                     sync.Map
+	FakeDataFilepath              string
 }
 
 var _ zone.SubscriptionsAPI = &SubscriptionsAPI{}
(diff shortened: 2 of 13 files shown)


matthchr:
There is stuff at utils/zones.go that I could also move into the zone provider... I almost did but it felt a bit awkward so left them where they are. LMK if you disagree and think it should move.

pkg/fake/subscriptionsapi.go
```diff
@@ -0,0 +1,94 @@
...
+		SubscriptionID: subscriptionID,
+		Options:        options,
+	}
+
+	pagingHandler := runtime.PagingHandler[armsubscriptions.ClientListLocationsResponse]{
+		More: func(page armsubscriptions.ClientListLocationsResponse) bool {
+			return false // TODO: It might be ideal if we had a MockPager which sometimes simulated multiple pages of results to ensure we handle that correctly
+		},
+		Fetcher: func(ctx context.Context, _ *armsubscriptions.ClientListLocationsResponse) (armsubscriptions.ClientListLocationsResponse, error) {
+			return api.NewListLocationsPagerBehavior.Invoke(input, func(input *ListLocationsInput) (armsubscriptions.ClientListLocationsResponse, error) {
```
matthchr, line 97: This is a bit weird but I wanted a way to check if we'd called the Fetcher and this seemed the best way with what we ahve.

pkg/providers/zone/zone.go
```diff
@@ -0,0 +1,157 @@
...
+	"fmt"
+	"sync"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/resources/armsubscriptions"
+	"github.com/samber/lo"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+const maxFailures = 10
```
matthchr: Should have a comment here probably?

Also maybe worth highlighting this behavior which is here to avoid us spamming the API forever. Argument could be made some sort of backoff might be better... maybe I should update to do that instead it does seem better.

pkg/providers/zone/zone.go
```diff
@@ -0,0 +1,157 @@
...
+		} else {
+			p.hasLoaded = true
+		}
+	}
+
+	return p.zoneSupport[region] // if cache doesn't have our region, assume no zone support
+}
+
+// loadFromAzure discovers zone support by calling Azure Subscriptions API
+func (p *Provider) loadFromAzure(ctx context.Context) error {
```
matthchr, line 109: We could try to load this from Azure on another thread right from the start... which might be better?

tallaxes: When is this provider called first? If it is in the path of scale up - then loading on startup (and I am not even sure it needs to be on a different thread) would be better, as it would shave a bit of latency from the first scaleup. If it is not in the path of scaleup (e.g. maybe loaded when instance types are computed?) - then it does not matter, I think ...

matthchr: Discussed, agreed how it is now is best. Left a comment with the reasoning in this method.

pkg/providers/zone/zone.go
```diff
@@ -0,0 +1,198 @@
...
+}
+
+// Provider handles zone support detection for Azure regions
+type Provider struct {
+	subscriptionsAPI SubscriptionsAPI
+	subscriptionID   string
+	clock            Clock
+
+	// Cached zone support data - maps region name to zone support boolean
+	zoneSupport map[string]bool
```
tallaxes, line 60: Any reason to keep the whole mapping, rather than just the region we are running in? AKS clusters are inherently single region, right?

matthchr: Right now yes but I think there's discussion of changes there.
It also matched the signature of the old function and wasn't hard to just track all the regions (I think it's like ~100 in public but even if it's 300 it's not a lot of memory since we're not holding on to everything)
This way it allows the instance provider to continue passing `p.region` and it'll work for any value of p.region.

TBH open to restricting it, it just felt like it might be _more_ work to restrict it than not.

pkg/providers/instancetype/instancetypes.go
```diff
@@ -179,13 +189,13 @@ func (p *DefaultProvider) Get(ctx context.Context, nodeClass *v1beta1.AKSNodeCla
...
 // The strings have to match Zone labels that will be placed on Node
-func instanceTypeZones(sku *skewer.SKU, region string) sets.Set[string] {
+func (p *DefaultProvider) instanceTypeZones(ctx context.Context, sku *skewer.SKU, region string) sets.Set[string] {
 	// skewer returns numerical zones, like "1" (as keys in the map);
 	// prefix each zone with "<region>-", to have them match the labels placed on Node (e.g. "westus2-1")
 	// Note this data comes from LocationInfo, then skewer is used to get the SKU info
 	// If an offering is non-zonal, the availability zones will be empty.
 	skuZones := lo.Keys(sku.AvailabilityZones(region))
-	if hasZonalSupport(region) && len(skuZones) > 0 {
+	if p.zoneProvider.SupportsZones(ctx, region) && len(skuZones) > 0 {
```
matthchr: Realized (just now) that this check actually seems useless because when is skewer going to return a nonzero set of zones in a region that isn't zonal?

Is that even possible?

I could just rip all this stuff out entirely... though there might be value in keeping it for possible futures where we do stuff with physical zones, since we have access to that information now that we're calling this API.

Thoughts?

tallaxes: Ok, on a closer look, it does indeed appear we don't need to know if the region is zonal anymore. It had been deemed needed to distinguish empty "supported zones list" (as returned by `sku.AvailabilityZones`, which filters out restricted zones) in non-zonal regions (= "non-zonal SKU") and zonal regions (= no offering). This latter interpretation was wrong, it means "non-zonal SKU in zonal region". We [fixed this](https://github.com/Azure/karpenter-provider-azure/pull/410), but overlooked the fact that we no longer need to know if the region is zonal: I think _we can always interpret empty zone list (as returned by `sku.AvailabiliytZones`, which filters out restricted zones) as "non-zonal offering"_. The only other cases where empty zone list can be returned (after filtering out restricted zones), also have overall location restriction, these are already filtered out earlier and won't be observed.

matthchr: As discussed, I will remove this usage here but keep the zone provider and tests (it will be unused). We may want it in the future.

matthchr: ok, I've removed the zone provider usage from instance provider for now.

pkg/providers/instancetype/instancetypes.go
```diff
@@ -179,13 +185,13 @@ func (p *DefaultProvider) Get(ctx context.Context, nodeClass *v1beta1.AKSNodeCla
 
 // instanceTypeZones generates the set of all supported zones for a given SKU
 // The strings have to match Zone labels that will be placed on Node
-func instanceTypeZones(sku *skewer.SKU, region string) sets.Set[string] {
+func (p *DefaultProvider) instanceTypeZones(sku *skewer.SKU, region string) sets.Set[string] {
```
Bryce-Soghigian: shouldn't need to pass in (region string) if you are using the provider, you can do p.region

pkg/providers/zone/zone.go
```diff
@@ -0,0 +1,200 @@
...
+			return fmt.Errorf("failed to list Azure locations: %w", err)
+		}
+
+		for _, location := range page.Value {
+			locationName := lo.FromPtr(location.Name)
+			if locationName == "" {
+				continue
+			}
+
+			result[locationName] = len(location.AvailabilityZoneMappings) > 0
```
Bryce-Soghigian, line 127: in the fake for this, can we generate the data it ends up getting back for the ListLocationsPager just to validate manually a bit this is always mapping correctly? Doesn't have to be this pr but good to look at.

pkg/providers/zone/zone.go
```diff
@@ -0,0 +1,200 @@
...
+	"centralindia":     true,
+	"japaneast":        true,
+	"koreacentral":     true,
+	"southeastasia":    true,
+	"eastasia":         true,
+	"chinanorth3":      true,
+	"indonesiacentral": true,
+	"japanwest":        true,
+	"newzealandnorth":  true,
+}
```
Bryce-Soghigian: Q: Are the zones in this list accurate in the sense that all of them match what we get back from the programatic way of doing things?

pkg/fake/subscriptionsapi.go
```diff
@@ -0,0 +1,94 @@
...
+func (api *SubscriptionsAPI) NewListLocationsPager(
+	subscriptionID string,
+	options *armsubscriptions.ClientListLocationsOptions,
+) *runtime.Pager[armsubscriptions.ClientListLocationsResponse] {
+	input := &ListLocationsInput{
+		SubscriptionID: subscriptionID,
+		Options:        options,
+	}
+
+	pagingHandler := runtime.PagingHandler[armsubscriptions.ClientListLocationsResponse]{
```
Bryce-Soghigian, line 92: this seems like data we could have codegen'd like the other providers that do mostly reading from things that don't change a ton.

pkg/providers/instancetype/instancetypes.go
matthchr: Fixed.

pkg/providers/zone/zone.go
matthchr: I can't confirm chinanorth3 and usgovvirginia. The rest was mostly right but there were a few issues:
(From copilot):
```
2. Regions from your list that do NOT exist in the JSON file:
❌ These regions were not found in the JSON file at all:

centraluseuap (found but NO availability zones)
usgovvirginia (not found)
chinanorth3 (not found)
3. Additional regions with availability zones NOT in your list:
🆕 You missed these regions that DO have availability zones:

malaysiawest - has availability zones but wasn't in your list
austriaeast - has availability zones but wasn't in your list
Summary
centraluseuap from your list exists but does NOT have availability zones
usgovvirginia and chinanorth3 don't exist in this JSON file (might be in different Azure clouds)
You missed malaysiawest and austriaeast which do support availability zones
```

I've fixed the above.

pkg/providers/zone/zone.go
matthchr: I still need some of the infrastructure here (counting calls, etc) for the zone provider caching logic, but I added loading the data from the real API and storing it in a JSON file.

pkg/fake/subscriptionsapi.go
matthchr: I need the ability to track calls and simulate errors still though. I looked and I didn't see a fake data package that provided both. I've taken a shot at adding sourcing this data from a JSON file.

Bryce-Soghigian (approved):

hack/codegen.sh
```diff
@@ -41,9 +52,9 @@ skugen() {
 
 
 skugen-all() {
-  SUBSCRIPTION_ID=$(az account show --query 'id' --output tsv)
-  export SUBSCRIPTION_ID
-  if [ -z "${SUBSCRIPTION_ID}" ]; then
+  AZURE_SUBSCRIPTION_ID=$(az account show --query 'id' --output tsv)
```
matthchr, line 55: Swapped this to `AZURE_SUBSCRIPTION_ID` because it's what we require in Karpenter setup and it felt reasonable to be consistent.

Bryce-Soghigian (approved):

PR #1118, opened 2025-08-22
chore: clean up skuclient

matthchr:
Now that we're using the Track2 SDK for authentication, we don't need to periodically refresh the client.

We know this is true because other clients using the same credential do not hit issues with token refresh. Additionally if we examine the ServicePrincipal credential we can see that it does support refresh:
[ClientSecretCredential](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azidentity/client_secret_credential.go#L73) calls [GetToken()](https://github.com/Azure/azure-sdk-for-go/blob/f0871996d2bc4f9ebb97fb94cb00285f8df8437d/sdk/azidentity/confidential_client.go#L104) which calls [AcquireTokenSilent](https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/main/apps/confidential/confidential.go#L589) which handles refresh/expiry at [this line](https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/main/apps/internal/base/base.go#L354)

* Unit tests
* End to end tests: https://github.com/Azure/karpenter-provider-azure/actions/runs/17162604216
* Manual test w/ pod older than 24h

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
NONE
```

### DIFF
--- a/pkg/providers/instance/azure_client.go
+++ b/pkg/providers/instance/azure_client.go
@@ -37,6 +37,7 @@ import (
 	"github.com/Azure/karpenter-provider-azure/pkg/providers/loadbalancer"
 	"github.com/Azure/karpenter-provider-azure/pkg/providers/networksecuritygroup"
 	"github.com/Azure/karpenter-provider-azure/pkg/providers/zone"
+	"github.com/Azure/skewer"
 
 	armopts "github.com/Azure/karpenter-provider-azure/pkg/utils/opts"
 )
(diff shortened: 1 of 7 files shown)


pkg/providers/instance/azure_client.go
```diff
@@ -193,7 +194,7 @@ func NewAZClient(ctx context.Context, cfg *auth.Config, env *azclient.Environmen
 
 	// TODO: this one is not enabled for rate limiting / throttling ...
 	// TODO Move this over to track 2 when skewer is migrated
```
charliedmcb, line 196: Remove this `TODO` comment?
```suggestion
```

pkg/providers/instance/azure_client.go
```diff
@@ -80,7 +81,7 @@ type AZClient struct {
 	ImageVersionsClient     imagefamilytypes.CommunityGalleryImageVersionsAPI
 	NodeBootstrappingClient imagefamilytypes.NodeBootstrappingAPI
 	// SKU CLIENT is still using track 1 because skewer does not support the track 2 path. We need to refactor this once skewer supports track 2
```
charliedmcb, line 83: Remove comment?
```suggestion
```

charliedmcb (approved):
assuming tests pass, lgtm.

pkg/providers/instance/azure_client.go
matthchr: The SDK itself is still track1 (just using track2 auth), so I think this comment actually still applies

pkg/providers/instance/azure_client.go
matthchr: As above, the actual SDK is still track1, so I think this comment still applies.

matthchr:
FWIW my manual test with pod >24h old also worked just fine w/ this change.

PR #1132, opened 2025-08-30
chore(logging): add logging const pkg with beginning structure and examples

charliedmcb:
Fixes # 

**Description**
Follow up to this PR:
https://github.com/Azure/karpenter-provider-azure/pull/1129

As discussed with @matthchr creating a logging package to contain common logging const values for reuse, and insurance of reused values cross logs 

- E2E Runs:
    - https://github.com/Azure/karpenter-provider-azure/actions/runs/17337118093
    - https://github.com/Azure/karpenter-provider-azure/actions/runs/17337160358

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
NONE
```

### DIFF
--- a/pkg/logging/keys.go
+++ b/pkg/logging/keys.go
@@ -20,13 +20,13 @@ const (
 	// MARK: Common Keys
 	// > Note: we should only define common log keys here, that we want to ensure consistency between logging occurrences
 	// > No need to define every adhoc key.
-	// ---
+
 	ImageID = "imageID"
 
 	// MARK: Upstream Log Keys
 	// > In the cases where a key is defined in upstream logs we match their casing regardless of if it aligns with our general casing pattern here.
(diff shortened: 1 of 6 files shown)


pkg/logging/keys.go
```diff
@@ -0,0 +1,33 @@
...
+*/
+
+package logging
+
+const (
+	// MARK: Common Keys
+	// > Note: we should only define common log keys here, that we want to ensure consistency between logging occurrences
+	// > No need to define every adhoc key.
+
+	ImageID = "imageID"
```
charliedmcb, line 24: Should these be suffixed with `Key` so `ImageIDKey` in this case?

suffixing with `Key` feels like more correct naming, but gets away from the const directly mapping its name to the underlaying string,

charliedmcb: Going without suffix

PR #1135, opened 2025-09-04
DRAFT: Support for Disk Encryption Sets w BYOK

Bryce-Soghigian:
Fixes # 

**Description**

*

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


Makefile-az.mk
```diff
@@ -85,7 +90,7 @@ az-mkaks-cniv1: az-mkacr ## Create test AKS cluster (with --network-plugin azure
 az-mkaks-cilium: az-mkacr ## Create test AKS cluster (with --network-dataplane cilium, --network-plugin azure, and --network-plugin-mode overlay)
 	az aks create          --name $(AZURE_CLUSTER_NAME) --resource-group $(AZURE_RESOURCE_GROUP) --attach-acr $(AZURE_ACR_NAME) \
 		--enable-managed-identity --node-count 3 --generate-ssh-keys -o none --network-dataplane cilium --network-plugin azure --network-plugin-mode overlay \
-		--enable-oidc-issuer --enable-workload-identity
+		--enable-oidc-issuer --enable-workload-identity --node-provisioning-mode Auto
```
Bryce-Soghigian:
```suggestion
		--enable-oidc-issuer --enable-workload-identity
```

Makefile-az.mk
```diff
@@ -1,4 +1,4 @@
-AZURE_LOCATION ?= westus2
+AZURE_LOCATION ?= westus2
```
matthchr: Excess space?

Makefile-az.mk
matthchr: ??

Makefile-az.mk
Bryce-Soghigian:
```suggestion
AZURE_LOCATION ?= westus2
```

.github/actions/e2e/create-cluster/action.yaml
```diff
@@ -29,6 +29,9 @@ inputs:
   location:
     description: "The azure location to run the e2e test in"
     default: "eastus"
+  suite:
+    description: "Name of the test suite (used to determine cluster configuration)"
+    required: true
```
Bryce-Soghigian:
```suggestion
```

.github/actions/e2e/create-cluster/action.yaml
```diff
@@ -57,7 +60,9 @@ runs:
         AZURE_RESOURCE_GROUP: ${{ inputs.resource_group }}
         AZURE_ACR_NAME: ${{ inputs.acr_name }} 
         AZURE_LOCATION: ${{ inputs.location }}
-      run: make az-mkaks-cilium
+      run: |
+        echo "Creating default Cilium cluster for suite: ${{ inputs.suite }}"
+        make az-mkaks-cilium
```
Bryce-Soghigian:
```suggestion
      run: make az-mkaks-cilium
```

.github/workflows/e2e.yaml
```diff
@@ -157,6 +158,16 @@ jobs:
...
+      - name: set BYOK environment variable (byok suite only)
+        if: inputs.suite == 'byok'
+        shell: bash
+        env:
+          AZURE_CLUSTER_NAME: ${{ env.CLUSTER_NAME }}
+          AZURE_RESOURCE_GROUP: ${{ env.RG_NAME }}
+        run: |
+          DES_ID=$(make az-show-des-id)
+          echo "AZURE_DISK_ENCRYPTION_SET_ID=$DES_ID" >> "$GITHUB_ENV"
+          echo "Set AZURE_DISK_ENCRYPTION_SET_ID=$DES_ID"
```
Bryce-Soghigian:
```suggestion
```

.github/workflows/e2e.yaml
```diff
@@ -143,6 +143,7 @@ jobs:
           acr_name: ${{ env.ACR_NAME }}
           git_ref: ${{ inputs.git_ref }}
           location: ${{ inputs.location }}
+          suite: ${{ inputs.suite }}
```
Bryce-Soghigian:
```suggestion
```

designs/disk-encryption.md
```diff
@@ -0,0 +1,338 @@
...
+
+
+## RBAC: Azure RBAC Requirements 
+1. DES Access: The DES is the azure resource that links the AKS disks to the Key Vault Key. To allow AKS to use CMKs, the user must grant the Cluster Identity permission to use the Disk Encryption Set. 
+2. Key Vault Access: The managed identity associated with AKS must be able to access the keys within the key vault, this allows AKS to wrap and unwrap the data encryption keys used for disk encryption with the master key
+
+| Principal | Role Assignment | Scope | 
+| - | - | - | 
+| Cluster Identity | Disk Encryption Set Reader | diskEncryptionSetID | 
+| Cluster Identity | Key Vault Crypto Service Encryption User | Key Vault Resource |
```
Bryce-Soghigian, line 324: TODO: Update this list based on e2e test run

.github/workflows/e2e-matrix.yaml
```diff
@@ -47,7 +47,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        suite: [Scheduling, Utilization, GPU, Drift, InPlaceUpdate, Integration, NodeClaim, Consolidation, Chaos, ACR, KubernetesUpgrade, Spot, Subnet]
+        suite: [BYOK]
+        # suite: [ACR, BYOK, Chaos, Consolidation, Drift, GPU, InPlaceUpdate, Integration, KubernetesUpgrade, NodeClaim, Scheduling, Spot, Subnet, Utilization]
```
Bryce-Soghigian:
```suggestion
        suite: [ACR, BYOK, Chaos, Consolidation, Drift, GPU, InPlaceUpdate, Integration, KubernetesUpgrade, NodeClaim, Scheduling, Spot, Subnet, Utilization]
```

Makefile-az.mk
```diff
@@ -22,6 +22,7 @@ KARPENTER_FEDERATED_IDENTITY_CREDENTIAL_NAME ?= KARPENTER_FID
 CUSTOM_VNET_NAME ?= $(AZURE_CLUSTER_NAME)-vnet
 CUSTOM_SUBNET_NAME ?= nodesubnet
 
+
```
matthchr: ? Excess newline?

Makefile-az.mk
```diff
@@ -120,6 +121,7 @@ az-mkaks-custom-vnet: az-mkacr az-mkvnet az-mksubnet ## Create test AKS cluster
 	az aks get-credentials --name $(AZURE_CLUSTER_NAME) --resource-group $(AZURE_RESOURCE_GROUP) --overwrite-existing
 	skaffold config set default-repo $(AZURE_ACR_NAME).$(AZURE_ACR_SUFFIX)/karpenter
 
+
```
matthchr: ? Excess newline?

designs/disk-encryption.md
```diff
@@ -0,0 +1,338 @@
+# Disk Encryption Sets with Bring Your Own Key (BYOK) Support in Karpenter
+
+## Overview
+Azure automatically encrypts all data in managed disks at rest. By default, this encryption uses Microsoft-managed keys. However, some customers require additional control over encryption keys for compliance or security requirements.
+
+Azure Disk Encryption Sets enable customers to bring their own keys (BYOK) for encryption at rest, applying to both OS and data disks. This design document outlines the implementation of BYOK support in Karpenter for Azure. 
+
+## Goals
+- **BYOK Support**: Enable customers to use their own encryption keys for AKS OS and data disks
+- **Mutability Coverage**: AKS is considering making these fields mutable karpenter needs to account for that
```
matthchr:
```suggestion
- **Mutability Coverage**: AKS is considering making these fields mutable Karpenter needs to account for that
```

designs/disk-encryption.md
```diff
@@ -0,0 +1,338 @@
...
+- **Bootstrapping Mode Compatibility**: Support BYOK across different provisioning modes
+
+## Non-Goals
+- **Encryption at Host**: [Host-based encryption](https://learn.microsoft.com/en-us/azure/aks/enable-host-encryption) is out of scope
+- **Automatic Key Rotation Detection**: No automatic VM replacement on key expiration. Customers accept the [existing AKS limitations](https://learn.microsoft.com/en-us/azure/aks/azure-disk-customer-managed-keys#limitations) 
+
+## Implementation Requirements
+Disk encryption with customer-managed keys must be configured at AKS cluster creation time using the `--node-osdisk-diskencryptionset-id` parameter today. This constraint shapes our implementation approach.
+
+## Phased Delivery Plan
```
matthchr, line 23: This section feels more like an implementation plan and less like a design?

I assume written for AI? I don't mind having these checked in but not sold it is right to do it in the design document.

designs/disk-encryption.md
```diff
@@ -0,0 +1,338 @@
...
+
+Key behaviors:
+- AKSNodeClass DiskEncryptionSetID is mutable with drift detection (Phase 2)
+- Cluster-level changes only affect nodes without AKSNodeClass overrides
+- **Complete drift detection implemented in Phase 2** (both AKSNodeClass and cluster-level)
+
+## AKSNodeClass Mutability Design Decision
+
+### Phased Approach to Mutability
+We will adopt a phased approach to DiskEncryptionSetID mutability:
```
matthchr, line 72: minor, style, feel free to ignore: IMO designs are cleaner when they read:
1. The plan (where are we going)
2. How are we going to get there.

So rather than talking about the phases in the conceptual portion you maybe label/number things and only bring up phases at the end.

The reason I at least like this is that it keeps the conceptual vision shorter/simpler, because you don't have to talk about "this part is gonna be in phase 1" and "this part is gonna be in phase 2" - less cognitive load for readers as well because they're just focusing on "where are we going" and don't also have to keep track of phases while they're doing that.

Then, once you've laid out the north star, you can talk about how we're going to get from where we are to there.

designs/disk-encryption.md
```diff
@@ -0,0 +1,338 @@
...
+### Phase 2 Mutability Decision
+
+In Phase 2, we need to decide whether to make AKSNodeClass.DiskEncryptionSetID mutable. This decision has significant implications:
+
+**Important Distinction:**
+- Key rotation within same DES: No node replacement needed (handled by Azure automatically)
+- Changing DES ID: Requires node replacement (what we're discussing here)
+
+#### Option 1: Keep Field Immutable
+We could add a CEL validation rule to constrain karpenter from allowing these fields to mutable, until AKS also adds support for the field being mutable.
```
matthchr, line 89:
```suggestion
We could add a CEL validation rule to keep this field immutable until AKS also adds support for mutating this field at the cluster scope.
```

designs/disk-encryption.md
```diff
@@ -0,0 +1,338 @@
...
+```
+
+**Pros:**
+- Seamless DES migration experience
+- Aligns with Karpenter's drift philosophy
+- Automatic migration to new encryption sets
+- No manual NodePool management
+
+**Cons:**
+- We don't know if this would be a problem with any customers having inconsistent encryption keys in the pool
```
matthchr, line 137: Can you expand on this? I assume you mean that there's a period of time where some nodes are on the old key and some are on the new key?

designs/disk-encryption.md
```diff
@@ -0,0 +1,338 @@
...
+            ID: to.StringPtr(provider.diskEncryptionSetID),
+        },
+    }
+}
+```
+
+#### AKSMachineAPI Mode
+The AKSMachineAPI automatically inherits the `--node-osdisk-diskencryptionset-id` value from the managed cluster object. The Machine API and AgentPool APIs default this value and explicitly prevent mutation after initial configuration.
+
+In Phase 3, the Machine API will need to:
```
matthchr, line 180: Does it actually need to? Is it mutable on VMs?

The Machine API allowing this field to be mutable would be weird if it weren't mutable on VMs and we just need to create a new VM anyway? What would setting it even do - just make RP trigger the "isDrifted" bit?

designs/disk-encryption.md
```diff
@@ -0,0 +1,338 @@
...
+- Machine API inherits correctly from cluster settings
+
+### Validation Updates
+
+**Phase 1**: Remove validation blocking customers from enabling nap when they have disk encryption set id enabled. 
+**Phase 3**: Allow PUT operations on cluster DiskEncryptionSetID when NAP is enabled
+
+
+## RBAC: Azure RBAC Requirements 
+1. DES Access: The DES is the azure resource that links the AKS disks to the Key Vault Key. To allow AKS to use CMKs, the user must grant the Cluster Identity permission to use the Disk Encryption Set.
```
matthchr:
```suggestion
1. DES Access: The DES is the Azure resource that links the AKS disks to the Key Vault Key. To allow AKS to use CMKs, the user must grant the Cluster Identity permission to use the Disk Encryption Set. 
```

pkg/operator/options/options.go
```diff
@@ -112,6 +113,7 @@ func (o *Options) AddFlags(fs *coreoptions.FlagSet) {
 	fs.BoolVar(&o.UseSIG, "use-sig", env.WithDefaultBool("USE_SIG", false), "If set to true karpenter will use the AKS managed shared image galleries and the node image versions api. If set to false karpenter will use community image galleries. Only a subset of image features will be available in the community image galleries and this flag is only for the managed node provisioning addon.")
 	fs.StringVar(&o.SIGAccessTokenServerURL, "sig-access-token-server-url", env.WithDefaultString("SIG_ACCESS_TOKEN_SERVER_URL", ""), "The URL for the SIG access token server. Only used for AKS managed karpenter. UseSIG must be set tot true for this to take effect.")
 	fs.StringVar(&o.SIGSubscriptionID, "sig-subscription-id", env.WithDefaultString("SIG_SUBSCRIPTION_ID", ""), "The subscription ID of the shared image gallery.")
+	fs.StringVar(&o.DiskEncryptionSetID, "node-osdisk-diskencryptionset-id", env.WithDefaultString("NODE_OSDISK_DISKENCRYPTIONSET_ID", ""), "The ARM resource ID of the disk encryption set to use for customer-managed key (BYOK) encryption.")
```
matthchr, line 116: Is this actually only used for the os disk? You mention Data disks in your design (including PVCs/etc). Assuming this is also used for those it feels wrong to call the parameter `node-osdisk-...`?

matthchr, line 116: minor: Missing validation that this is a valid ARM ID?

I thought we had similar validation for other fields?

pkg/providers/instance/instance.go
```diff
@@ -555,6 +560,17 @@ func setVMPropertiesOSDiskType(vmProperties *armcompute.VirtualMachineProperties
 	}
 }
 
+func setVMPropertiesOSDiskEncryption(vmProperties *armcompute.VirtualMachineProperties, diskEncryptionSetID string) {
+	if diskEncryptionSetID != "" {
+		if vmProperties.StorageProfile.OSDisk.ManagedDisk == nil {
```
matthchr, line 575: Is managedDisk the default?

Does setting this property ever opt us into managed disk when we actually weren't using one previously?

test/pkg/environment/azure/byok.go
```diff
@@ -0,0 +1,262 @@
...
+//   - Azure clients are configured with retry logic for 403 (Forbidden) errors caused by RBAC propagation delays
+//   - Azure clients also retry on 400 (Bad Request) errors for PrincipalNotFound when identities haven't replicated
+//
+// Returns the DES resource ID for use in VM disk encryption configuration.
+func (env *Environment) CreateKeyVaultAndDiskEncryptionSet(ctx context.Context) string {
+	keyVaultName := fmt.Sprintf("karpentertest%d", time.Now().Unix())
+	keyName := "test-key"
+	desName := fmt.Sprintf("karpenter-test-des-%d", time.Now().Unix())
+
+	cred, err := azidentity.NewDefaultAzureCredential(nil)
```
matthchr: Maybe `env` should just have `Cred` set+exported - or at least `cred` (unexported?)

See test/pkg/environment/azure/environment.go which already does this once. Feels like we should share credential objects and not duplicate instantiation of them.

test/pkg/environment/azure/identity.go
```diff
@@ -0,0 +1,94 @@
...
+
+func (env *Environment) GetTenantID(ctx context.Context) string {
+	cluster, err := env.managedClusterClient.Get(ctx, env.ClusterResourceGroup, env.ClusterName, nil)
+	Expect(err).ToNot(HaveOccurred())
+	Expect(cluster.Identity).ToNot(BeNil())
+	Expect(cluster.Identity.PrincipalID).ToNot(BeNil())
+	return lo.FromPtr(cluster.Identity.TenantID)
+}
+
+func (env *Environment) GetClusterIdentityPrincipalID(ctx context.Context) string {
```
matthchr: minor: Consider `GetClusterIdentity` instead, and then re-use in `GetTenantID` impl? Caller can do `identity.PrincipalID` easily and that gives them access to other things too if they want (like tenant ID).

test/pkg/environment/azure/identity.go
```diff
@@ -0,0 +1,94 @@
...
+	Expect(err).ToNot(HaveOccurred())
+	Expect(cluster.Identity).ToNot(BeNil())
+	Expect(cluster.Identity.PrincipalID).ToNot(BeNil())
+	return lo.FromPtr(cluster.Identity.PrincipalID)
+}
+
+func (env *Environment) GetKarpenterWorkloadIdentity(ctx context.Context) string {
+	karpenterMSIName := "karpentermsi" // matches AZURE_KARPENTER_USER_ASSIGNED_IDENTITY_NAME
+
+	cred, err := azidentity.NewDefaultAzureCredential(nil)
```
matthchr: Another dupe of this call - avoid?

test/pkg/environment/azure/identity.go
```diff
@@ -0,0 +1,94 @@
...
+	Expect(err).ToNot(HaveOccurred())
+
+	identity, err := msiClient.Get(ctx, env.ClusterResourceGroup, karpenterMSIName, nil)
+	Expect(err).ToNot(HaveOccurred())
+
+	return lo.FromPtr(identity.Properties.PrincipalID)
+}
+
+// getCurrentUserPrincipalID gets the principal ID of the current authenticated user
+func (env *Environment) GetCurrentUserPrincipalID(ctx context.Context, cred *azidentity.DefaultAzureCredential) string {
```
matthchr:
```suggestion
func (env *Environment) GetCurrentUserPrincipalID(ctx context.Context, cred azcore.TokenCredential) string {
```

Maybe?

test/pkg/environment/azure/identity.go
```diff
@@ -0,0 +1,94 @@
...
+		fmt.Printf("Warning: Could not extract claims from token\n")
+		return ""
+	}
+
+	// Extract the oid (object ID) claim which is the principal ID
+	if oid, ok := claims["oid"].(string); ok {
+		return oid
+	}
+
+	fmt.Printf("Warning: Could not find oid claim in token\n")
```
matthchr: Feels like this method should just return an error?

Or do `Expect` asserts?

test/pkg/environment/azure/byok.go
```diff
@@ -0,0 +1,262 @@
...
+		return fmt.Errorf("failed to assign Administrator role to test user: %w", err)
+	}
+
+	// RBAC propagation is handled by retry logic in the clients (403 errors are retried)
+
+	return nil
+}
+
+// createKeyVaultKey creates a key in the Key Vault
+func (env *Environment) createKeyVaultKey(ctx context.Context, keyVaultName, keyName string, cred *azidentity.DefaultAzureCredential) (*azkeys.KeyBundle, error) {
```
matthchr: use TokenCredential interface not defaultCredential everywhere.

test/pkg/environment/azure/byok.go
```diff
@@ -0,0 +1,262 @@
...
+//   - Assigns Key Vault Crypto Officer and Administrator roles to the test user (for key creation)
+//   - Assigns Reader role to the Karpenter workload identity for DES access (so it can create VMs)
+//   - Assigns Key Vault Crypto Service Encryption User role to the DES managed identity (so DES can access keys)
+//
+// 3. Retry Logic:
+//   - Azure clients are configured with retry logic for 403 (Forbidden) errors caused by RBAC propagation delays
+//   - Azure clients also retry on 400 (Bad Request) errors for PrincipalNotFound when identities haven't replicated
+//
+// Returns the DES resource ID for use in VM disk encryption configuration.
+func (env *Environment) CreateKeyVaultAndDiskEncryptionSet(ctx context.Context) string {
```
matthchr: Is this really a primitive we want/need on env?

Feels like possibly instead of being on env it should just live in the BYOK suite and be built out of the (would need to be exported) primitive functions for making a KV and/or DES?

test/pkg/environment/azure/byok.go
```diff
@@ -0,0 +1,262 @@
...
+	err = env.assignDiskEncryptionSetRBAC(ctx, lo.FromPtr(des.ID), karpenterIdentity)
+	Expect(err).ToNot(HaveOccurred())
+
+	err = env.updateKeyVaultAccessForDES(ctx, keyVaultName, desIdentity)
+	Expect(err).ToNot(HaveOccurred())
+
+	return lo.FromPtr(des.ID)
+}
+
+// createKeyVault creates an Azure Key Vault with proper access policies
```
matthchr: minor: There are no access policies? So that part of this comment seems wrong?

test/pkg/environment/azure/byok.go
```diff
@@ -0,0 +1,262 @@
...
+				MaxRetries: 15,
+				RetryDelay: time.Second * 5,
+				StatusCodes: []int{
+					http.StatusForbidden, // RBAC assignments haven't propagated to Key Vault yet
+				},
+			},
+		},
+	}
+
+	keyClient, err := azkeys.NewClient(fmt.Sprintf("https://%s.vault.azure.net/", keyVaultName), cred, keyClientOptions)
```
matthchr: IIRC the KV should have this endpoint on the GET response - you could pass the whole KV object in and might be able to just read it off there?

test/pkg/environment/azure/default_aksnodeclass.yaml
```diff
@@ -3,4 +3,4 @@ kind: AKSNodeClass
 metadata:
   name: default
 spec:
-  imageFamily: Ubuntu2204
+  imageFamily: Ubuntu
```
matthchr: Is this a holdover from one of your other PRs?

Maybe split into a separate smaller PR?
Feel free to ignore if that's annoying to do.

test/pkg/environment/azure/rbac_manager.go
```diff
@@ -0,0 +1,106 @@
...
+)
+
+type RBACManager struct {
+	subscriptionID string
+	client         *armauthorization.RoleAssignmentsClient
+}
+
+// NewRBACManager builds a client with DefaultAzureCredential.
+func NewRBACManager(subscriptionID string) (*RBACManager, error) {
+	cred, err := azidentity.NewDefaultAzureCredential(nil)
```
matthchr: Should take this as a parameter (see other SDKs)

test/pkg/environment/azure/rbac_manager.go
```diff
@@ -0,0 +1,106 @@
...
+	if err != nil {
+		return nil, err
+	}
+	c, err := armauthorization.NewRoleAssignmentsClient(subscriptionID, cred, &arm.ClientOptions{
```
matthchr: This retry configuration is duplicated in at least 2 places (maybe 3 if `azkeys.ClientOptions` is an alias). Share it (i.e. put it in a function that constructs it and re-use the function)

designs/disk-encryption.md
Bryce-Soghigian: Yes, this was written for AI, but also I believe it helps give context as to why phase 3 + 4 are additive new workstreams.

pkg/operator/options/options.go
Bryce-Soghigian: Yes, data disks are configured in the storage class instead though. Its outside of the lifecycle of karpenter after further research. 


So node-osdisk makes sense because this is for configuring os disk not for configuring data disk.

test/pkg/environment/azure/byok.go
Bryce-Soghigian: yes, this is wrong now, we used to use access policies in earlier versions of the pr but i later switched to rbac

test/pkg/environment/azure/identity.go
Bryce-Soghigian: ACK was actually wondering the same thing

test/pkg/environment/azure/rbac_manager.go
Bryce-Soghigian: azkeys.ClientOptions isn't an alias to arm.ClientOptions :( its an alias for azcore.ClientOptions. 

Made a reusable function for other cases, removed the case for rbac manager, it probably doesn't need to retry assignment, it shouldn't get 403 on assignment anymore since we removed some of the retry stuff from the clientside princpal retries and added principalType.

designs/disk-encryption.md
Bryce-Soghigian: Yes, thats exactly what I mean, some nodes will have old and others will have new.

test/pkg/environment/azure/byok.go
Bryce-Soghigian: Fair point, i moved it over, just felt weird because every single function call passed in ENV, maybe a separate struct for creating the DES and i can put it in the BYOK suite?

pkg/providers/instance/instance.go
Bryce-Soghigian: image
It looks like for the ephemeral case it still bootstraps with ephmeral disk so long as the DiskCache options are set to ephemeral

designs/disk-encryption.md
Bryce-Soghigian: We would set it on the create request. As it is today, there is not field in machineapi put for this field. Its readonly. I cannot specify that all new vms are created with a different DES. 

We would mark vms with an old DES as drifted and fully replace them.

test/pkg/environment/azure/byok.go
Bryce-Soghigian: I exported cred, that fixes alot of usecases actually, fanastic call out

pkg/operator/options/options.go
Bryce-Soghigian: Added some validation for it

pkg/operator/options/options_validation.go
```diff
@@ -194,3 +195,60 @@ func isValidURL(u string) bool {
...
+	return o.validateDiskEncryptionSetValues(parts)
+}
+
+func (o *Options) parseDiskEncryptionSetID() ([]string, error) {
+	if !strings.HasPrefix(o.DiskEncryptionSetID, "/") {
+		return nil, fmt.Errorf("disk-encryption-set-id is invalid: must start with /subscriptions/, got %s", o.DiskEncryptionSetID)
+	}
+
+	parts := strings.Split(o.DiskEncryptionSetID, "/")
+	if len(parts) != 9 {
```
matthchr, line 222: Might be cleaner to validate this with a regexp rather than by hand

test/suites/byok/suite_test.go
```diff
@@ -126,3 +136,235 @@ var _ = Describe("BYOK", func() {
...
+	Expect(err).ToNot(HaveOccurred())
+
+	err = updateKeyVaultAccessForDES(ctx, env, keyVaultName, desIdentity)
+	Expect(err).ToNot(HaveOccurred())
+
+	return lo.FromPtr(des.ID)
+}
+
+// createKeyVault creates an Azure Key Vault
+func createKeyVault(ctx context.Context, env *azure.Environment, keyVaultName, clusterTenant string) (*armkeyvault.Vault, error) {
```
matthchr, line 210: Ah sorry my previous comment wasn't clear maybe.

I was just saying that _one method_ `CreateKeyVaultAndDiskEncryptionSet` doesn't fit on env.
I'm fine with `createKeyVault` and `assignKeyVaultRBAC` being exported and on env. They feel reusable for other tests than this one.

even `createDiskEncryptionSet` could probably be on env (who knows maybe we'll re-use it?)

`CreateKeyVaultAndDiskEncryptionSet` though felt more like "this is specific setup for a specific test" and so was less reusable.

designs/disk-encryption.md
matthchr: So it doesn't need to be mutable in the machine API it just needs to be settable on create, and then Karpenter can deal with diff computation?

pkg/providers/instance/instance.go
matthchr: May be worth leaving a note here stating that or something similar?

test/pkg/environment/azure/environment.go
```diff
@@ -105,11 +110,28 @@ func NewEnvironment(t *testing.T) *Environment {
...
+	azureEnv.DiskEncryptionSetClient = lo.Must(armcompute.NewDiskEncryptionSetsClient(azureEnv.SubscriptionID, cred, byokRetryOptions))
+	azureEnv.RBACManager = lo.Must(NewRBACManager(azureEnv.SubscriptionID, cred))
+	return azureEnv
+}
 
-	// Retry options for BYOK-related clients that may encounter RBAC propagation delays
-	// RBAC assignments can take time to propagate, resulting in 403 Forbidden errors
-	// With 15 retries at 5 second intervals = 75 seconds total retry time
-	byokRetryOptions := &arm.ClientOptions{
+func (env *Environment) GetDefaultCredential() azcore.TokenCredential {
```
matthchr, line 126: minor, style: I don't think we care that this is from DefaultCredential. It's just the credential. I would just call it `GetCredential`?

matthchr (approved):
Approved with some comments.

pkg/operator/options/options_validation.go
Bryce-Soghigian: Based on previous outages with regex, I prefer to stick away from it.

test/suites/byok/suite_test.go
Bryce-Soghigian: to be honest most keyvault scenarios we have will probably be able to be part of this suite. For now we can keep it here and refactor if we see motion in that direction.

dennisbalsam99:
hi team, was going through the BYOK design doc and came across this MR, does this also cover the point in phase 1 of 

`Remove NAP validation blocking DiskEncryptionSetID`, checked the test pipeline and saw cluster was created without
`--node-os-disk-encryptionset-id`

We are trying to enable Karpenter in NAP mode, but we are getting blocked on enabling NAP when the DiskEncryptionSetID is set, which is required for us due to policy

wdarko1:
Hi @dennisbalsam99 the NAP validation is still blocked pending some additional testing, which we are working to resolve and update soon.

PR #1100, opened 2025-08-12
test: we should wait longer on the vm extension condition we can get …

Bryce-Soghigian:
…into a situation where the node is up but the extensions aren't finished installing yet see: https://github.com/Azure/karpenter-provider-azure/actions/runs/16848701021/job/47731853226

[CREATED/UPDATED 2025-08-09T11:19:18Z] POD default/handrift-4-zjybrbp7ai provisionable=false phase=Running nodename=aks-harepattern-2-p6rhcvizm4-sj4zk owner=[]v1.OwnerReference(nil) [storkmad-5-c0pig7y1kl restarts=0]
  [FAILED] Expected
      <[]string | len:1, cap:1>: [AKSLinuxExtension]
  to contain elements
      <[]string | len:1, cap:1>: [
          computeAksLinuxBilling,
      ]
  the missing elements were
      <[]string | len:1, cap:1>: [
          computeAksLinuxBilling,
      ]
  In [It] at: /home/runner/work/karpenter-provider-azure/karpenter-provider-azure/test/suites/nodeclaim/vm_extension_test.go:44 @ 08/09/25 11:19:19.254
  < Exit [It] should install all valid vm extensions - /home/runner/work/karpenter-provider-azure/karpenter-provider-azure/test/suites/nodeclaim/vm_extension_test.go:28 @ 08/09/25 11:19:19.254 (1m48.067s)

Fixes # 

**Description**

**How was this change tested?**

*

**Does this change impact docs?**
- [ ] Yes, PR includes docs updates
- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

PR #1068, opened 2025-07-29
feat(fips): generic Ubuntu family and FIPS support for SIG

lnguyen924:
Fixes # 

**Description**

Enable support for FIPS images via a new and optional FIPSMode field and added a generic Ubuntu ImageFamily in AKSNodeClass

**How was this change tested?**

* Unit tests
* Manual tests

**Does this change impact docs?**
- [ ] Yes, PR includes docs updates
- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
New ImageFamily supported (Ubuntu); maps to Ubuntu 20.04 or Ubuntu 22.04 based on FIPSMode
New AKSNodeClass.Spec field (fipsMode); values of Disabled and FIPS for determining FIPS enablement
```

PR #1147, opened 2025-09-08
feat: Support Ubuntu2404 and have it be the default version for generic Ubuntu Image Family starting on 1.34

### LINKED ISSUE #1136: Stuck with an old kernel and poor NFS performance

Expand access to the latest kernel versions:

- Ubuntu 24.04 AKS images
- Azure Linux images on more instance types especially the latest GPU instances (A100/H100)
- Custom images

In our benchmarking, we have noticed 4-6X lower max bandwidth for NFS mounts using Linux kernels version 5.15 vs 6.6 or 6.11. From observation, on the older kernels the software interrupt load  (`%soft` from `mpstat`) during a `fio` benchmark of the NFS mount seems to concentrate all on a single core bounding the NFS throughput by the processing power of this core. In the later versions the interrupt load gets distributed across multiple cores allowing the client to reach significantly higher throughput levels.

With Karpenter, the latest image from the default image family `Ubuntu2204` comes with the kernel version 5.15:
https://github.com/Azure/AKS/blob/master/vhd-notes/aks-ubuntu/AKSUbuntu-2204/202508.20.0.txt#L909

Switching to the `AzureLinux` family improves performance significantly as it comes with kernel version 6.6, but this image family has limited support across instance types and does not support the modern GPU instances with the A100 or H100 cards which we need:
https://github.com/Azure/karpenter-provider-azure/blob/38ea60bd9def2f1b81a8b519eaa9f6f47ffaf99f/pkg/utils/supported-gpus.yaml#L96-L110

The latest Ubuntu 24.04 AKS images come with kernel version 6.8, but this family is not supported by karpenter on Azure yet:
https://github.com/Azure/AKS/blob/master/vhd-notes/aks-ubuntu/AKSUbuntu-2404/202508.20.0.txt#L942

As a final resort, in browsing through the issues I did not find a way to use a custom image.

No, not able to work around the issue.

### Additional Context

_No response_

### Attachments

_No response_

### Community Note

Bryce-Soghigian:
Fixes #1136

**Description**
This PR introduces support for Ubuntu2204, as well as defaulting based on k8s version or FIPS setting for the generic `Ubuntu` image family. 

Conditionally, if the user is on 1.34+, with the Ubuntu generic family specifed, we default to using 2404, otherwise we use 2204. 

Also introduced support for 1.34 envtest.

**Follow Up Work** 
- Split Utilization into smaller scale runs that test bootstrap rather than bin pack
- Runtime Validation for minimum K8s Version per ImageFamily

- GPU E2E tests validating Ubuntu2404, and the Defaulting for `Ubuntu` as a generic family work
- utilization e2e testing bootstrap
- ✅(Minus nodeclaim, flakey separately [fixed here](https://github.com/Azure/karpenter-provider-azure/pull/1100)) https://github.com/Azure/karpenter-provider-azure/actions/runs/17604580348

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/providers/imagefamily/nodeimage_test.go
+++ b/pkg/providers/imagefamily/nodeimage_test.go
@@ -64,7 +64,7 @@ func renderExpectedCIGNodeImages(
 func renderExpectedSIGNodeImages(
 	fam imagefamily.ImageFamily,
 	fips *v1beta1.FIPSMode,
-	version, sigSub string,
+	version, sigSub string, //nolint:unparam
 ) []imagefamily.NodeImage {
 	defaultImages := fam.DefaultImages(true, fips)
 	out := make([]imagefamily.NodeImage, 0, len(defaultImages))


pkg/apis/v1alpha2/crd_validation_cel_test.go
```diff
@@ -145,12 +145,16 @@ var _ = Describe("CEL/Validation", func() {
...
+			Entry("Ubuntu2404 when FIPSMode is explicitly Disabled should succeed", v1alpha2.Ubuntu2404ImageFamily, &v1alpha2.FIPSModeDisabled, true),
+			Entry("Ubuntu2404 when FIPSMode is not explicitly set should succeed", v1alpha2.Ubuntu2404ImageFamily, nil, true),
+			//TODO: Modify when Ubuntu 24.04 with FIPS becomes available
+			Entry("Ubuntu2404 when FIPSMode is explicitly FIPS should fail", v1alpha2.Ubuntu2404ImageFamily, &v1alpha2.FIPSModeFIPS, false),
 			Entry("generic AzureLinux when FIPSMode is explicitly Disabled should succeed", v1alpha2.AzureLinuxImageFamily, &v1alpha2.FIPSModeDisabled, true),
 			Entry("generic AzureLinux when FIPSMode is not explicitly set should succeed", v1alpha2.AzureLinuxImageFamily, nil, true),
 			Entry("generic AzureLinux when FIPSMode is explicitly FIPS should succeed", v1alpha2.AzureLinuxImageFamily, &v1alpha2.FIPSModeFIPS, true),
 			Entry("unspecified ImageFamily (defaults to Ubuntu2204) when FIPSMode is explicitly Disabled should succeed", "", &v1alpha2.FIPSModeDisabled, true),
 			Entry("unspecified ImageFamily (defaults to Ubuntu2204) when FIPSMode is not explicitly set should succeed", "", nil, true),
```
Bryce-Soghigian:
```suggestion
			Entry("unspecified ImageFamily (defaults to Ubuntu2404) when FIPSMode is explicitly Disabled should succeed", "", &v1alpha2.FIPSModeDisabled, true),
			Entry("unspecified ImageFamily (defaults to Ubuntu2404) when FIPSMode is not explicitly set should succeed", "", nil, true),
```

pkg/providers/imagefamily/resolver.go
```diff
@@ -231,25 +231,36 @@ func prepareKubeletConfiguration(ctx context.Context, instanceType *cloudprovide
...
 			return &AzureLinux3{Options: parameters}
 		}
 		return &AzureLinux{Options: parameters}
 	default:
+		if lo.FromPtr(fipsMode) == v1beta1.FIPSModeFIPS {
+			return &Ubuntu2004{Options: parameters}
+		}
+		if UseUbuntu2404(kubernetesVersion) {
+			return &Ubuntu2404{Options: parameters}
+		}
```
Bryce-Soghigian: Default should be an impossible state to hit, since the cel validation will default a field on there. But still added this case here.

pkg/providers/imagefamily/nodeimage_test.go
```diff
@@ -70,22 +70,7 @@ func getExpectedTestCIGImages(imageFamily string, fipsMode *v1beta1.FIPSMode, ve
...
-	} else if imageFamily == v1beta1.AzureLinuxImageFamily {
-		if imagefamily.UseAzureLinux3(kubernetesVersion) {
-			actualImageFamily = &imagefamily.AzureLinux3{}
-		} else {
-			actualImageFamily = &imagefamily.AzureLinux{}
-		}
-	}
+	actualImageFamily := imagefamily.GetImageFamily(lo.ToPtr(imageFamily), fipsMode, kubernetesVersion, nil)
 	images = actualImageFamily.DefaultImages(true, fipsMode)
 	nodeImages := []imagefamily.NodeImage{}
```
Bryce-Soghigian: we should come up with a better way to generate expected test images.

charts/karpenter-crd/templates/karpenter.azure.com_aksnodeclasses.yaml
```diff
@@ -68,11 +68,12 @@ spec:
                 - Disabled
                 type: string
               imageFamily:
-                default: Ubuntu2204
+                default: Ubuntu
```
charliedmcb, line 71: Think this is good, but want to double check with @tallaxes

pkg/providers/imagefamily/customscriptsbootstrap/provisionclientbootstrap_test.go
charliedmcb, line 1: For this there are also blocks of tests for the individual image families that Robin added.

Top block of this diff is an example of Linda adding it for 2004:
[pkg/providers/imagefamily/customscriptsbootstrap/provisionclientbootstrap_test.go](https://github.com/Azure/karpenter-provider-azure/pull/1068/files#diff-545684fd5e8da05e55e82a9cab941b77414a89773cd924e3a1791d77a4fe2d5f)

pkg/providers/imagefamily/resolver.go
```diff
@@ -231,27 +231,37 @@ func prepareKubeletConfiguration(ctx context.Context, instanceType *cloudprovide
...
 			return &AzureLinux3{Options: parameters}
 		}
 		return &AzureLinux{Options: parameters}
 	default:
-		return &Ubuntu2204{Options: parameters}
+		return defaultUbuntu(fipsMode, kubernetesVersion, parameters)
+	}
+}
+
+func defaultUbuntu(fipsMode *v1beta1.FIPSMode, kubernetesVersion string, parameters *template.StaticParameters) ImageFamily {
```
charliedmcb, line 256: nit (we just happen to use generic as default):
```suggestion
func genericUbuntu(fipsMode *v1beta1.FIPSMode, kubernetesVersion string, parameters *template.StaticParameters) ImageFamily {
```

pkg/providers/imagefamily/utils.go
```diff
@@ -31,3 +31,14 @@ func UseAzureLinux3(kubernetesVersion string) bool {
 	}
 	return version.GE(semver.Version{Major: 1, Minor: 32})
 }
+
+// UseUbuntu2404 is when AKS starts defaulting support for Ubuntu2404
+func UseUbuntu2404(kubernetesVersion string) bool {
```
charliedmcb, line 36: These are tested indirectly, and I don't think we actually have a direct test for `UseAzureLinux3`, but you want to add one?

pkg/providers/instancetype/instancetypes.go
```diff
@@ -257,10 +257,10 @@ func (p *DefaultProvider) isInstanceTypeSupportedByImageFamily(skuName, imageFam
 	if !(utils.IsNvidiaEnabledSKU(skuName) || utils.IsMarinerEnabledGPUSKU(skuName)) {
 		return true
 	}
-	switch imageFamily {
-	case v1beta1.Ubuntu2204ImageFamily:
+	switch {
```
charliedmcb, line 260: fix?
```suggestion
	switch imageFamily {
```

pkg/providers/instancetype/instancetypes.go
```diff
@@ -257,10 +257,10 @@ func (p *DefaultProvider) isInstanceTypeSupportedByImageFamily(skuName, imageFam
...
+	case v1beta1.UbuntuFamilies.Has(imageFamily):
 		return utils.IsNvidiaEnabledSKU(skuName)
```
charliedmcb, line 262: Just sanity checking, should this apply to 2004 too?

test/suites/gpu/suite_test.go
```diff
@@ -104,7 +104,22 @@ var _ = Describe("GPU", func() {
 			env.ExpectCreatedNodeCount("==", int(*deployment.Spec.Replicas))
 		},
 		Entry("should provision one GPU Node and one GPU Pod (AzureLinux)", env.AZLinuxNodeClass()),
-		Entry("should provision one GPU Node and one GPU Pod (Ubuntu2204)", env.DefaultAKSNodeClass()),
+		Entry("should provision one GPU Node and one GPU Pod (Ubuntu)", func() *v1beta1.AKSNodeClass { // This ensures the case statement for GPU Filtering covers the generic Ubuntu Image family
```
charliedmcb, line 107: Hit 2004 too, by setting Generic + FIPS, but only for the `!env.InClusterController`

test/suites/gpu/suite_test.go
```diff
@@ -104,7 +104,22 @@ var _ = Describe("GPU", func() {
 			env.ExpectCreatedNodeCount("==", int(*deployment.Spec.Replicas))
 		},
 		Entry("should provision one GPU Node and one GPU Pod (AzureLinux)", env.AZLinuxNodeClass()),
-		Entry("should provision one GPU Node and one GPU Pod (Ubuntu2204)", env.DefaultAKSNodeClass()),
```
charliedmcb, line 107: Do wonder if we should update `DefaultAKSNodeClass` to return a AKSNodeClass with generic Ubuntu

test/suites/utilization/fips_test.go
```diff
@@ -92,6 +92,34 @@ var _ = Describe("FIPS", func() {
 			imageRef := expectNodeUsesFIPS(node)
 			expectNodeClassHasExpectedImages(nodeClass, imageRef, true) // true = expect FIPS images
 		})
+
+		It("should not provision FIPS-enabled Ubuntu2204 nodes (not supported)", func() {
```
charliedmcb: I don't think either of these tests are needed. 

We already have the CEL validation, and since both of these should fail at the apply stage, and never actually run anything, I don't think they need a E2E test, and can just be covered in unit testing?

test/suites/utilization/fips_test.go
```diff
@@ -112,6 +140,32 @@ var _ = Describe("FIPS", func() {
 			imageRef := expectNodeDoesNotUseFIPS(node)
 			expectNodeClassHasExpectedImages(nodeClass, imageRef, false) // false = expect non-FIPS images
 		})
+
+		It("should provision Ubuntu2404 nodes with FIPS disabled", func() {
```
charliedmcb: I don't think these are needed either. 

They also don't hurt imo, beyond increasing runtime.

I just wanting one `FIPSModeDisabled` case, since the other provisioning handles it just not set.

I'm fine either way if these are added or not, but don't think they are needed if we are worried about bloating eventual runtime

test/suites/utilization/suite_test.go
```diff
@@ -55,21 +56,55 @@ var _ = AfterEach(func() { env.Cleanup() })
...
+			}, env.ArmNodepool)
+		})
+		It("should provision one pod per node (Ubuntu2204, amd64)", Label(v1beta1.Ubuntu2204ImageFamily), func() {
+			ExpectProvisionPodPerNode(func() *v1beta1.AKSNodeClass {
+				nodeClass := env.DefaultAKSNodeClass()
+				nodeClass.Spec.ImageFamily = lo.ToPtr(v1beta1.Ubuntu2204ImageFamily)
+				return nodeClass
+			}, env.DefaultNodePool)
+		})
+		It("should provision one pod per node (Ubuntu2204, arm64)", Label(v1beta1.Ubuntu2204ImageFamily), func() {
```
charliedmcb: Generic is already covering 2204. Need to update the underlaying AKsNodeClass that gets setup for DefaultAKSNodeClass if you want the coverage:
https://github.com/Azure/karpenter-provider-azure/blob/22e4311351c0548367732690df7847253d879a17/pkg/test/nodeclass.go#L56
, or you need to update the cases:
- `should provision one pod per node (Ubuntu, amd64)`
- `should provision one pod per node (Ubuntu, arm64)`
to set generic Ubuntu

pkg/providers/imagefamily/ubuntu_2404_test.go
```diff
@@ -0,0 +1,99 @@
...
+	"github.com/Azure/karpenter-provider-azure/pkg/apis/v1beta1"
+	"github.com/Azure/karpenter-provider-azure/pkg/providers/imagefamily"
+	"github.com/Azure/karpenter-provider-azure/pkg/providers/imagefamily/customscriptsbootstrap"
+	template "github.com/Azure/karpenter-provider-azure/pkg/providers/launchtemplate/parameters"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestUbuntu2404_Name(t *testing.T) {
+	ubuntu := &imagefamily.Ubuntu2404{
+		Options: &template.StaticParameters{},
```
charliedmcb, line 31: nit: shouldn't need. matches other cases

pkg/providers/imagefamily/ubuntu_2404_test.go
```diff
@@ -0,0 +1,99 @@
...
+	t.Run("should return empty images for FIPS mode with SIG (not yet supported)", func(t *testing.T) {
+		fipsMode := v1beta1.FIPSModeFIPS
+		images := ubuntu.DefaultImages(true, &fipsMode)
+		assert.Empty(t, images)
+	})
+}
+
+func TestUbuntu2404_CustomScriptsNodeBootstrapping(t *testing.T) {
+	ubuntu := &imagefamily.Ubuntu2404{
+		Options: &template.StaticParameters{
```
charliedmcb, line 70: These AI-gen?

Its fine, but I really dislike havings our consts we use for things like SubscriptionID not be a UUID, and also I've run into issue where they don't match the sub id used in the mocks, etc. 

This is an isolated instance I'm not concerned about it, but a styling nit I keep seeing in our tests

charliedmcb: The other cases actually do a very similar thing. Don't worry about changing.

pkg/providers/imagefamily/ubuntu_2404_test.go
```diff
@@ -0,0 +1,99 @@
...
+			SubnetID:                       "/subscriptions/test/resourceGroups/test/providers/Microsoft.Network/virtualNetworks/test/subnets/test",
+		},
+	}
+
+	bootstrapper := ubuntu.CustomScriptsNodeBootstrapping(
+		nil, nil, nil, nil, nil, "test-distro", "Standard_LRS", nil, nil,
+	)
+	provisionBootstrapper, ok := bootstrapper.(customscriptsbootstrap.ProvisionClientBootstrap)
+	assert.True(t, ok, "Expected ProvisionClientBootstrap type")
+	assert.Equal(t, customscriptsbootstrap.ImageFamilyOSSKUUbuntu2404, provisionBootstrapper.OSSKU, "ImageFamily field must be set to prevent unsupported image family errors")
```
charliedmcb, line 95: It feels a little redundant, but if you're doing this test, and asserting, kind of worth asserting on all of the fields?

pkg/providers/imagefamily/ubuntu_2404_test.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+	assert.Equal(t, v1beta1.Ubuntu2404ImageFamily, ubuntu.Name())
+}
+
+func TestUbuntu2404_DefaultImages(t *testing.T) {
+	ubuntu := &imagefamily.Ubuntu2404{
+		Options: &template.StaticParameters{},
+	}
+
+	t.Run("should return correct default images", func(t *testing.T) {
```
charliedmcb, line 41: nit:
```suggestion
	t.Run("should return correct default images for cig without fips", func(t *testing.T) {
```

pkg/providers/instancetype/instancetypes.go
Bryce-Soghigian: switch imageFamily {
case v1beta1.Ubuntu2204ImageFamily:
case v1beta1.UbuntuFamilies.Has(imageFamily):
 
 
 The second case would be invalid because the return type of v1beta1.UbuntuFamilies.Has() returns a bool and not a string. 
 
 "All case clauses must be constant expressions or values of the same type of expr"

pkg/providers/imagefamily/ubuntu_2404_test.go
Bryce-Soghigian: agreed, but all of our other tests do similar things, i think its ok in cases like this.

test/suites/gpu/suite_test.go
Bryce-Soghigian: Fips Support isn't really the goal of this pr, and we already have some provisioning tests for fips in fips_test.go

```
	Context("FIPS Provisioning", func() {
		BeforeEach(func() {
			if env.InClusterController {
				Skip("FIPS tests require SIG access - skipping in self-hosted mode")
			}
		})

		It("should provision FIPS-enabled Ubuntu nodes", func() {
			nodeClass.Spec.ImageFamily = lo.ToPtr(v1beta1.UbuntuImageFamily)
			nodeClass.Spec.FIPSMode = lo.ToPtr(v1beta1.FIPSModeFIPS)

			pod := test.Pod()
			env.ExpectCreated(nodeClass, nodePool, pod)
			env.EventuallyExpectHealthy(pod)
			node := env.ExpectCreatedNodeCount("==", 1)[0]

			imageRef := expectNodeUsesFIPS(node)
			expectNodeClassHasExpectedImages(nodeClass, imageRef, true) // true = expect FIPS images
		})

		It("should provision FIPS-enabled AzureLinux nodes", func() {
			nodeClass.Spec.FIPSMode = lo.ToPtr(v1beta1.FIPSModeFIPS)

			pod := test.Pod()
			env.ExpectCreated(nodeClass, nodePool, pod)
			env.EventuallyExpectHealthy(pod)
			node := env.ExpectCreatedNodeCount("==", 1)[0]

			imageRef := expectNodeUsesFIPS(node)
			expectNodeClassHasExpectedImages(nodeClass, imageRef, true) // true = expect FIPS images
		})
	})
```

test/suites/utilization/suite_test.go
Bryce-Soghigian: Good point, that default in pkg/test/nodeclass.go we set the Ubuntu2204ImageFamily. 

We might need to update the codebase to use generic ubuntu by default.

pkg/providers/instancetype/instancetypes.go
Bryce-Soghigian: https://github.com/Azure/karpenter-provider-azure/blob/159be31162861b0f4557940de93b89ac94ab2417/pkg/providers/instancetype/instancetypes.go#L160

We pass in imageFamily from the AKSNodeClass. 2004 will not be an image family in the AKSNodeClass API.

test/suites/gpu/suite_test.go
Bryce-Soghigian: Probably, lets create a followup item for that.

test/suites/gpu/suite_test.go
Bryce-Soghigian: Fips GPU support isn't somethihng I want to test in this PR as this pr is more about adding 2404, but maybe can be a followup work item. Trying to scope down the work.

pkg/providers/imagefamily/utils.go
```diff
@@ -31,3 +31,14 @@ func UseAzureLinux3(kubernetesVersion string) bool {
...
+	// Parse version, stripping any 'v' prefix if present
+	version, err := semver.ParseTolerant(strings.TrimPrefix(kubernetesVersion, "v"))
+	if err != nil {
+		// If we can't parse the version, default to AzureLinux (false)
+		return false
```
tallaxes:
```suggestion
		// If we can't parse the version, default to Ubuntu2204 (false)
		return false
```

pkg/providers/imagefamily/resolver.go
```diff
@@ -231,27 +231,36 @@ func prepareKubeletConfiguration(ctx context.Context, instanceType *cloudprovide
...
 
-func getImageFamily(familyName *string, fipsMode *v1beta1.FIPSMode, kubernetesVersion string, parameters *template.StaticParameters) ImageFamily {
+func GetImageFamily(familyName *string, fipsMode *v1beta1.FIPSMode, kubernetesVersion string, parameters *template.StaticParameters) ImageFamily {
 	switch lo.FromPtr(familyName) {
 	case v1beta1.UbuntuImageFamily:
-		if lo.FromPtr(fipsMode) == v1beta1.FIPSModeFIPS {
-			return &Ubuntu2004{Options: parameters}
-		}
-		return &Ubuntu2204{Options: parameters}
+		return defaultUbuntu(fipsMode, kubernetesVersion, parameters)
```
tallaxes: You can just "fallthrough" here (and I would move this closer to default). 

Could also just delete this case and let default handle it, but keeping a clear indidation of how it is handled seems better

charts/karpenter-crd/templates/karpenter.azure.com_aksnodeclasses.yaml
tallaxes: Technically it is a breaking change. But the impact should be very contained. And if we do change it, the sooner the better (especially prior to 1.34), so yes let's do it.

tallaxes: Please also update the examples to use the generic Ubuntu, and we may want to update the default NAP nodepools as well.

pkg/providers/imagefamily/nodeimage_test.go
```diff
@@ -310,6 +294,89 @@ var _ = Describe("NodeImageProvider tests", func() {
...
+				// Should use Ubuntu2204 for K8s < 1.34
+				expectedImages := []imagefamily.NodeImage{}
+				ubuntu2204Images := imagefamily.Ubuntu2204{}.DefaultImages(true, nil)
+				for _, image := range ubuntu2204Images {
+					expectedImages = append(expectedImages, imagefamily.NodeImage{
+						ID:           fmt.Sprintf("/subscriptions/%s/resourceGroups/%s/providers/Microsoft.Compute/galleries/%s/images/%s/versions/%s", sigSubscription, image.GalleryResourceGroup, image.GalleryName, image.ImageDefinition, sigImageVersion),
+						Requirements: image.Requirements,
+					})
+				}
+				Expect(foundImages).To(Equal(expectedImages))
```
tallaxes: Is there any reason why the pattern used everywhere else getExpectedTestCIG/SIGImages would not be applicable here?

pkg/providers/imagefamily/nodeimage_test.go
tallaxes: There are tradeoffs around reusing (or not) the production code for generating test data. I would generally suggest separating the two, if at all possible, even at the expense of reproducing similar logic - otherwise there is a growing risk of a bug slipping through tests. But: whatever choice you make here, plase be consistent. If you are changing how getExpectedSIGImages work - also change how getExpectedTestCIGImages work.

pkg/providers/imagefamily/nodeimage_test.go
```diff
@@ -310,6 +294,89 @@ var _ = Describe("NodeImageProvider tests", func() {
 			Entry("for default AzureLinux with version >= 1.32 when FIPSMode is not explicitly set", lo.ToPtr(v1beta1.AzureLinuxImageFamily), nil, sigImageVersion, "1.32.0"),
 			Entry("for FIPS AzureLinux with version >= 1.32 when FIPSMode is explicitly set to FIPS", lo.ToPtr(v1beta1.AzureLinuxImageFamily), &v1beta1.FIPSModeFIPS, sigImageVersion, "1.32.0"),
 		)
+
+		Context("Ubuntu default image selection based on Kubernetes version", func() {
+			It("should select Ubuntu2204 for generic Ubuntu when K8s < 1.34", func() {
```
tallaxes, line 426: Please consider following the pattern used elsewhere of also keeping a single version-generic test (look for "This test changes depending on the Kubernetes version" comment)

pkg/providers/imagefamily/customscriptsbootstrap/provisionclientbootstrap.go
```diff
@@ -151,6 +152,8 @@ func (p *ProvisionClientBootstrap) ConstructProvisionValues(ctx context.Context)
 		provisionProfile.OsSku = to.Ptr(models.OSSKUUbuntu)
 	case ImageFamilyOSSKUUbuntu2204:
 		provisionProfile.OsSku = to.Ptr(models.OSSKUUbuntu)
+	case ImageFamilyOSSKUUbuntu2404:
+		provisionProfile.OsSku = to.Ptr(models.OSSKUUbuntu)
 	case ImageFamilyOSSKUAzureLinux2:
 		provisionProfile.OsSku = to.Ptr(models.OSSKUAzureLinux)
 	case ImageFamilyOSSKUAzureLinux3:
```
tallaxes: This logic - essentialy forced generalization of whatever image family is chosen - seems wrong. Can we take a closer look? If the models is just outdated - should it be refreshed? @comtalyst

pkg/providers/imagefamily/nodeimage_test.go
Bryce-Soghigian: I feel that whole set of images is ripe for refactoring. Generating a list of expected images was fine when we had 2 families, but as we get more that part gets more messy. I am thinking it might make more sense to default like this for each image family. 

Thoughts? I can think through this a bit more to see if there is a better way to do expectedImages.

pkg/providers/imagefamily/customscriptsbootstrap/provisionclientbootstrap.go
Bryce-Soghigian: image

Bryce-Soghigian: Looks like they have updated the list to accept Ubuntu2204.

pkg/providers/imagefamily/nodeimage_test.go
tallaxes: If by "like this" you mean replicating the above code (with minor changes) whenever it is needed in test code - I don't see how this is an improvement. Frankly I don't see a major problem with the current approach (getExpectedTestCIG/SIGImages, implemented one way or another) - and note that the reusable helpers have more complexity budget; the important part is making resilent tests easy to author. But sure, open to suggestion for real improvement (probably best isolated to its own PR?) Until then let's use the existing one-line pattern?

pkg/providers/imagefamily/customscriptsbootstrap/provisionclientbootstrap.go
tallaxes: Yes, but let's not use CLI as the source of truth. There are more constants available in the API/~~internally~~ - including a distinct one for AzureLinux3, for example.

pkg/providers/imagefamily/nodeimage_test.go
Bryce-Soghigian: When I said "like this" I realize thats a lazy way to think. I spent some time thinking through the 3 separate ways we can solve the expectations in tests. 

Approach A: Explicitly declare the family we expect in each test.
Approach B: Dynamically resolve the family using the GetImageFamily code, and share the function implementation for CIG + SIG 
Approach C: Use the existing solution


Goals: 
- Single Solution for getting expected images, that works for both CIG + SIG
- We shouldn't have to update the Test handler whenever we have a new ImageFamily

Approach A(Preferred): 
```go
func renderExpectedNodeImages(
	fam imagefamily.ImageFamily,
	useSIG bool,
	fips *v1beta1.FIPSMode,
	version, sigSub string,
) []imagefamily.NodeImage {
	defaultImages := fam.DefaultImages(useSIG, fips)
	out := make([]imagefamily.NodeImage, 0, len(defaultImages))
	for _, img := range defaultImages {
		id := lo.Ternary(useSIG,
			imagefamily.BuildImageIDSIG(sigSub, img.GalleryResourceGroup, img.GalleryName, img.ImageDefinition, version),
			imagefamily.BuildImageIDCIG(img.PublicGalleryURL, img.ImageDefinition, version),
		)
		out = append(out, imagefamily.NodeImage{ID: id, Requirements: img.Requirements})
	}
	return out
}
```

## Usage
```go
It("should match expected images for Ubuntu2204", func() {
    expectedImages := renderExpectedNodeImages(&imagefamily.Ubuntu2204{}, false, nodeClass.Spec.FIPSMode, cigImageVersion, "")
    Expect(foundImages).To(Equal(expectedImages))
})
```


### Pros 
- It follows the damp principle, Descriptive + Meaningful Phrases, we are clearly declaring what we expect, and don't dynamically resolve expectations. My biggest ick with approach b + c, is they dynamically make assumptions. If there is a bug in that dynamic resolution, the tests pass or fail. Given there will still have to be some dynamic input  for passing in which we expect per k8s version, but overall is more declarative.  

### Cons
- Requires explicitly declaring the image family struct in each test `&imagefamily.Ubuntu2204{}`, which could be seen as verbose
- Less dry if multiple tests use the same image family configuration, youre repeating the instantiation(meh)
- Doesn't automatically handle the image family resolution based on FIPS mode or Kubernetes version, but its good to make the caller think through these things 

Approach B: Fully Resolve Image Family, don't repeat yourself 

```go
type ImageExpectationsConfig struct {
	ImageFamily       string
	FIPSMode          *v1beta1.FIPSMode
	Version           string
	KubernetesVersion string
	UseSIG            bool
}

func NewImageExpectationsConfig(nodeClass *v1beta1.AKSNodeClass, version, kubernetesVersion string, useSIG bool) ImageExpectationsConfig {
	return ImageExpectationsConfig{
		ImageFamily:       lo.FromPtr(nodeClass.Spec.ImageFamily),
		FIPSMode:          nodeClass.Spec.FIPSMode,
		Version:           version,
		KubernetesVersion: kubernetesVersion,
		UseSIG:            useSIG,
	}
}

func (i *ImageExpectationsConfig) SIGSubscriptionID() string {
	return "10945678-1234-1234-1234-123456789012"
}

func (i *ImageExpectationsConfig) GetExpectedImages() []imagefamily.NodeImage {
	expectedFamily := imagefamily.GetImageFamily(lo.ToPtr(i.ImageFamily), i.FIPSMode, i.KubernetesVersion, nil)
	images := expectedFamily.DefaultImages(i.UseSIG, i.FIPSMode)
	return i.generateNodeImages(images)
}

func (i *ImageExpectationsConfig) generateNodeImages(images []imagefamilytypes.DefaultImageOutput) []imagefamily.NodeImage {
	nodeImages := make([]imagefamily.NodeImage, 0, len(images))
	for _, image := range images {
		id := lo.Ternary(i.UseSIG,
			imagefamily.BuildImageIDSIG(
				i.SIGSubscriptionID(),
				image.GalleryResourceGroup,
				image.GalleryName,
				image.ImageDefinition,
				i.Version,
			),
			imagefamily.BuildImageIDCIG(
				image.PublicGalleryURL,
				image.ImageDefinition,
				i.Version,
			),
		)
		nodeImages = append(nodeImages, imagefamily.NodeImage{
			ID:           id,
			Requirements: image.Requirements,
		})
	}
	return nodeImages
}
```


## Usage 


```go
It("should match expected images for Ubuntu2404", func() {
    nodeClass.Spec.ImageFamily = lo.ToPtr(v1beta1.Ubuntu2404ImageFamily)

    config := NewImageExpectationsConfig(nodeClass, cigImageVersion, kubernetesVersion, false)
    Expect(foundImages).To(Equal(config.GetExpectedImages()))
})
```


### Pros
- More DRY centralizes image family resolution logic in one place
- Automatically handles image family selection based on all relevant parameters
- Reusable config object can be shared across multiple assertions in the same test lets people know the inputs at least

### Cons
- More complex abstraction that could hide bugs in the resolution logic
- If GetImageFamily has a bug, tests will still pass theyre testing against dynamically resolved expectations
- Less explicit about what's being tested the actual image family is hidden behind the resolution logic, ideally when you write a test you know what all of the expected output is 
- Harder to debug when tests fail since you need to trace through the resolution logic
- Tests are less self documenting

Approach C: Original Approach
```go
//nolint:unparam // might always be using the same version in test, but could change in the future
func getExpectedTestSIGImages(imageFamily string, fipsMode *v1beta1.FIPSMode, version string, kubernetesVersion string) []imagefamily.NodeImage {
	var images []imagefamilytypes.DefaultImageOutput
	var actualImageFamily imagefamily.ImageFamily
	if imageFamily == v1beta1.UbuntuImageFamily {
		if lo.FromPtr(fipsMode) == v1beta1.FIPSModeFIPS {
			actualImageFamily = &imagefamily.Ubuntu2004{}
		} else {
			actualImageFamily = &imagefamily.Ubuntu2204{}
		}
	} else if imageFamily == v1beta1.Ubuntu2204ImageFamily {
		actualImageFamily = &imagefamily.Ubuntu2204{}
	} else if imageFamily == v1beta1.AzureLinuxImageFamily {
		if imagefamily.UseAzureLinux3(kubernetesVersion) {
			actualImageFamily = &imagefamily.AzureLinux3{}
		} else {
			actualImageFamily = &imagefamily.AzureLinux{}
		}
	}
	images = actualImageFamily.DefaultImages(true, fipsMode)
	nodeImages := []imagefamily.NodeImage{}
	for _, image := range images {
		nodeImages = append(nodeImages, imagefamily.NodeImage{
			ID:           fmt.Sprintf("/subscriptions/%s/resourceGroups/%s/providers/Microsoft.Compute/galleries/%s/images/%s/versions/%s", sigSubscription, image.GalleryResourceGroup, image.GalleryName, image.ImageDefinition, version),
			Requirements: image.Requirements,
		})
	}
	return nodeImages
}


func getExpectedTestCIGImages(imageFamily string, fipsMode *v1beta1.FIPSMode, version string, kubernetesVersion string) []imagefamily.NodeImage {
	var images []imagefamilytypes.DefaultImageOutput
	if imageFamily == v1beta1.Ubuntu2204ImageFamily {
		images = imagefamily.Ubuntu2204{}.DefaultImages(false, fipsMode)
	} else if imageFamily == v1beta1.AzureLinuxImageFamily {
		if imagefamily.UseAzureLinux3(kubernetesVersion) {
			images = imagefamily.AzureLinux3{}.DefaultImages(false, fipsMode)
		} else {
			images = imagefamily.AzureLinux{}.DefaultImages(false, fipsMode)
		}
	}
	nodeImages := []imagefamily.NodeImage{}
	for _, image := range images {
		nodeImages = append(nodeImages, imagefamily.NodeImage{
			ID:           fmt.Sprintf("/CommunityGalleries/%s/images/%s/versions/%s", image.PublicGalleryURL, image.ImageDefinition, version),
			Requirements: image.Requirements,
		})
	}
	return nodeImages
}
```


```go
		It("should match expected images for Ubuntu2204", func() {
			expectedImages := getExpectedTestCIGImages(*nodeClass.Spec.ImageFamily, nodeClass.Spec.FIPSMode, cigImageVersion, kubernetesVersion)
			Expect(foundImages).To(Equal(expectedImages))
		})
```


### Pros
- Was the existing working solution, no code changes needed ig   
- All logic is contained within the test. There are no external dependencies on GetImageFamily like in approach B

## Cons 
- Violates dry principle heavily, logic is duplicated between CIG and SIG functions in mulitple ways
- Harder to maintain as new image families are added and must be updated everywhere 
- The if else chains make the code harder to read and more prone to errors
- Branching logic for each imageFamily we add duplicate logic to GetImageFamily essentially

Bryce-Soghigian: Answered: https://github.com/Azure/karpenter-provider-azure/pull/1147#discussion_r2345344720 here

pkg/providers/imagefamily/customscriptsbootstrap/provisionclientbootstrap.go
comtalyst: @tallaxes This is what we discussed earlier: the ability to specify specific OSSKU is new. Ubuntu2204 is just ready, but AzureLinux3 is not yet.

> There are more constants available in the API/internally - including a distinct one for AzureLinux3, for example.

The philosophy behind the stability of GET NodeBootstrapping is to "always pair the usage with AKS", as it has been borrowing the data model/constants rather than being a formalized API. So, if AKS is not using the new constant, I don't think we should. 
Not being available in the public API/used by AKS main path could mean it is not stable. Dev process can result in the constants being added in that data model anytime, and engineers could assume "AKS not using this" means "no one (incl. Karpenter) will use this". 

In the end it is case-by-case, and test coverage enforce the legitimacy of our change. But for this one, I would rather get the green light from the owners. Machine API will also change this to be on public API. I don't think we should invest much more in this now. 
However, I think Bryce's addition should work(?). From my understanding, this is how AKS supports Ubuntu2404 (unless it has been using the internal constants behind the scenes, in which, we should be good to pair then).

comtalyst: + @charliedmcb was working with Linda/Node team on this area recently

pkg/providers/imagefamily/utils.go
Bryce-Soghigian: Testing indirectly is fine now

pkg/providers/imagefamily/customscriptsbootstrap/provisionclientbootstrap.go
tallaxes: @comtalyst +1 on only using what's available in the API, I was only pushing against relying on CLI as an indication of what's available. FWIW, I was under impression that both Ubuntu-specific enums and AzureLinux3 are available in the API (the latter may not yet be in the CLI?)

pkg/providers/imagefamily/nodeimage.go
```diff
@@ -37,8 +37,8 @@ const (
 	ImageExpirationInterval    = time.Hour * 24 * 3
 	ImageCacheCleaningInterval = time.Hour * 1
 
-	sharedImageGalleryImageIDFormat = "/subscriptions/%s/resourceGroups/%s/providers/Microsoft.Compute/galleries/%s/images/%s/versions/%s"
-	communityImageIDFormat          = "/CommunityGalleries/%s/images/%s/versions/%s"
+	SharedImageGalleryImageIDFormat = "/subscriptions/%s/resourceGroups/%s/providers/Microsoft.Compute/galleries/%s/images/%s/versions/%s"
+	CommunityImageIDFormat          = "/CommunityGalleries/%s/images/%s/versions/%s"
 )
```
tallaxes, line 42: These don't need to be exported

pkg/providers/imagefamily/resolver.go
```diff
@@ -231,27 +231,36 @@ func prepareKubeletConfiguration(ctx context.Context, instanceType *cloudprovide
...
 	case v1beta1.AzureLinuxImageFamily:
 		if UseAzureLinux3(kubernetesVersion) {
 			return &AzureLinux3{Options: parameters}
 		}
 		return &AzureLinux{Options: parameters}
+	case v1beta1.UbuntuImageFamily:
+		fallthrough
 	default:
-		return &Ubuntu2204{Options: parameters}
+		return defaultUbuntu(fipsMode, kubernetesVersion, parameters)
```
tallaxes, line 252: minor: can just inline the Ubuntu logic here (same as AzureLinux logic is inlined above)

pkg/providers/imagefamily/nodeimage_test.go
```diff
@@ -260,7 +269,16 @@ var _ = Describe("NodeImageProvider tests", func() {
...
 				foundImages, err := nodeImageProvider.List(ctx, nodeClass)
 				Expect(err).ToNot(HaveOccurred())
-				expectedImages := getExpectedTestSIGImages(*nodeClass.Spec.ImageFamily, nodeClass.Spec.FIPSMode, sigImageVersion, kubernetesVersion)
+
+				// Parse version to determine which Ubuntu version to expect
+				version, err := semver.ParseTolerant(strings.TrimPrefix(kubernetesVersion, "v"))
+				Expect(err).ToNot(HaveOccurred())
+
+				// Generic Ubuntu defaults to Ubuntu2404 for K8s >= 1.34, Ubuntu2204 otherwise
+				fam := lo.Ternary(version.Minor >= 34,
```
tallaxes: Why not use `UseUbuntu2404` helper? (I can understand an argument for using separate logic in test - but there are several uses of these helpers elsewhere in these tests, let's be consistent)

pkg/providers/imagefamily/nodeimage_test.go
tallaxes: * I do see some value in making image family expectation explicit. Would prefer to use something other than interfaces for this, but if we don't define constants for each (yet?) that's probably Ok.
* I would suggest going back to separate CIG/SIG helpers (with the new implementation). IMO this is better than a loose boolean arg - and you won't have to pass in dummy subscription into CIG version. (You can still share logic between the helpers, if needed; the ergonomics of tests themselves is more important)
* (minor) There is a mix of either using helpers like UseAzureLinux3 or explicit logic, and either using if statement of lo.Ternary. Please be consistent.

tallaxes (approved):
LGTM. Left minor feedback. Still not convinced re correctness of OSSKU logic but will defer to @comtalyst on that.

> Please also take a look at whether we can validate Ubuntu2404 use in Kubernetes upgrade E2E (we do this for AzureLinux)

Did you have a chance to take a look?

pkg/providers/imagefamily/nodeimage_test.go
Bryce-Soghigian: This was something i was debating myself.


Should we 

A: Use the helpers directly, and for each one assert that the k8s versions and defaulting works as we expect 
or 
B: Use k8s versions + lo.Ternary() directly, UseUbuntu2404 will be used in the real logic, but this tests that logic is properly honored. 

Also a third consideration, this is fine and easy to do now on if else, since we only have 2 cases
(2204 -> 2404) that are expected. 

But if we add 2604, or the previous 1804, then this ternary assignment will need to be refactored. 
I am leaning towards a k8s version distinction with if else, because thats more extensible when we have greater than 2 versions, and doesn't rely on the direct production code helpers.

pkg/providers/imagefamily/nodeimage.go
tallaxes: Not fixing?

pkg/providers/imagefamily/nodeimage_test.go
```diff
@@ -46,55 +47,32 @@ const (
...
-	nodeImages := []imagefamily.NodeImage{}
-	for _, image := range images {
-		nodeImages = append(nodeImages, imagefamily.NodeImage{
-			ID:           fmt.Sprintf("/subscriptions/%s/resourceGroups/%s/providers/Microsoft.Compute/galleries/%s/images/%s/versions/%s", sigSubscription, image.GalleryResourceGroup, image.GalleryName, image.ImageDefinition, version),
-			Requirements: image.Requirements,
-		})
+func renderExpectedSIGNodeImages(
+	fam imagefamily.ImageFamily,
+	fips *v1beta1.FIPSMode,
+	version, sigSub string, //nolint:unparam
```
tallaxes, line 67: I would just remove the unused/constant params, until they are actually needed. Again, the ergonomics / readability of tests themselves is what's primary IMO ...

tallaxes (approved):

pkg/providers/imagefamily/nodeimage_test.go
Bryce-Soghigian: will followup on this in followup pr for a couple of other cleanup things.

pkg/providers/imagefamily/nodeimage.go
Bryce-Soghigian: followign up on this after this pr

Issue #684, opened 2025-02-07
be able to enable "Encryption at host" if supported by sku


flbla:
### Tell us about your request

the nodes in this Node Pool can have host encryption enabled

### Tell us about the problem you're trying to solve. What are you trying to do, and why is it hard?

I want my OS disk to be encrypted

### Are you currently working around this issue?

i don't use karpenter...

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

raiviskrumins:
We are also struggling with this issue. nodes are not being created because of Policy, which requires `securityProfile.encryptionAtHost = true`
```
  "policyRule": {
      "if": {
        "anyOf": [
          {
            "allOf": [
              {
                "field": "type",
                "equals": "Microsoft.Compute/virtualMachines"
              },
              {
                "field": "Microsoft.Compute/virtualMachines/securityProfile.encryptionAtHost",
                "notEquals": "true"
              }
            ]
          },
          {
            "allOf": [
              {
                "field": "type",
                "equals": "Microsoft.Compute/virtualMachineScaleSets"
              },
              {
                "field": "Microsoft.Compute/virtualMachineScaleSets/virtualMachineProfile.securityProfile.encryptionAtHost",
                "notEquals": "true"
              }
            ]
          }
        ]
      },
      "then": {
        "effect": "[parameters('effect')]"
      }
    }
```
nodeclaim creation fails with:
```
apiVersion: karpenter.sh/v1
kind: NodeClaim
metadata:
  annotations:
    karpenter.sh/nodepool-hash: "3044138309914671815"
    karpenter.sh/nodepool-hash-version: v3
    karpenter.sh/stored-version-migrated: "true"
  creationTimestamp: "2025-04-24T07:08:14Z"
  finalizers:
  - karpenter.sh/termination
  generateName: spot-
  generation: 1
  labels:
    karpenter.sh/nodepool: spot
    kubernetes.azure.com/scalesetpriority: spot
  name: spot-xc8bf
  ownerReferences:
  - apiVersion: karpenter.sh/v1
    blockOwnerDeletion: true
    kind: NodePool
    name: spot
[...]

PR #1151, opened 2025-09-12
feat: support encryption at host

Bryce-Soghigian:
Fixes https://github.com/Azure/karpenter-provider-azure/issues/684

**Description**
This pr adds support for AKS's encryption on host https://learn.microsoft.com/en-us/azure/aks/enable-host-encryption, and filtering to filter out instance types that don't match the trait. 

- https://github.com/Azure/karpenter-provider-azure/actions/runs/17679683611/job/50250684713 Integration Focus test of our new scenario
- local make az-all + scale up 
- make test

- Yes, we need to update the docs
**Release Note**

```release-note
Add support for encryption at host
```

### DIFF
--- a/pkg/providers/instance/instance.go
+++ b/pkg/providers/instance/instance.go
@@ -126,6 +126,7 @@ type DefaultProvider struct {
 	unavailableOfferings         *cache.UnavailableOfferings
 	provisionMode                string
 	responseErrorHandlers        []responseErrorHandler
+	diskEncryptionSetID          string
 
 	vmListQuery, nicListQuery string
 }
@@ -141,6 +142,7 @@ func NewDefaultProvider(
 	resourceGroup string,
 	subscriptionID string,
 	provisionMode string,
+	diskEncryptionSetID string,
 ) *DefaultProvider {
 	return &DefaultProvider{
 		azClient:                     azClient,
@@ -154,6 +156,7 @@ func NewDefaultProvider(
 		unavailableOfferings:         offeringsCache,
 		provisionMode:                provisionMode,
 		responseErrorHandlers:        defaultResponseErrorHandlers(),
+		diskEncryptionSetID:          diskEncryptionSetID,
 
 		vmListQuery:  GetVMListQueryBuilder(resourceGroup).String(),
 		nicListQuery: GetNICListQueryBuilder(resourceGroup).String(),
@@ -468,19 +471,20 @@ func (p *DefaultProvider) createNetworkInterface(ctx context.Context, opts *crea
 
 // createVMOptions contains all the parameters needed to create a VM
 type createVMOptions struct {
-	VMName             string
-	NicReference       string
-	Zone               string
-	CapacityType       string
-	Location           string
-	SSHPublicKey       string
-	LinuxAdminUsername string
-	NodeIdentities     []string
-	NodeClass          *v1beta1.AKSNodeClass
-	LaunchTemplate     *launchtemplate.Template
-	InstanceType       *corecloudprovider.InstanceType
-	ProvisionMode      string
-	UseSIG             bool
+	VMName              string
+	NicReference        string
+	Zone                string
+	CapacityType        string
+	Location            string
+	SSHPublicKey        string
+	LinuxAdminUsername  string
+	NodeIdentities      []string
+	NodeClass           *v1beta1.AKSNodeClass
+	LaunchTemplate      *launchtemplate.Template
+	InstanceType        *corecloudprovider.InstanceType
+	ProvisionMode       string
+	UseSIG              bool
+	DiskEncryptionSetID string
 }
 
 // newVMObject creates a new armcompute.VirtualMachine from the provided options
@@ -542,6 +546,7 @@ func newVMObject(opts *createVMOptions) *armcompute.VirtualMachine {
 		Tags:  opts.LaunchTemplate.Tags,
 	}
 	setVMPropertiesOSDiskType(vm.Properties, opts.LaunchTemplate)
+	setVMPropertiesOSDiskEncryption(vm.Properties, opts.DiskEncryptionSetID)
 	setImageReference(vm.Properties, opts.LaunchTemplate.ImageID, opts.UseSIG)
 	setVMPropertiesBillingProfile(vm.Properties, opts.CapacityType)
 	setVMPropertiesSecurityProfile(vm.Properties, opts.NodeClass)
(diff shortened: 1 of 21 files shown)


charts/karpenter-crd/templates/karpenter.azure.com_aksnodeclasses.yaml
```diff
@@ -60,6 +60,12 @@ spec:
               AKSNodeClassSpec is the top level specification for the AKS Karpenter Provider.
               This will contain configuration necessary to launch instances in AKS.
             properties:
+              encryptionAtHost:
+                description: |-
+                  EncryptionAtHost enables encryption at host for the provisioned nodes
```
matthchr: minor: Give users some intro to the links.

"For more details, see: " or similar.

pkg/apis/v1beta1/aksnodeclass.go
```diff
@@ -81,6 +81,11 @@ type AKSNodeClassSpec struct {
 	// +kubebuilder:validation:Maximum:=250
 	// +optional
 	MaxPods *int32 `json:"maxPods,omitempty"`
+	// EncryptionAtHost enables encryption at host for the provisioned nodes
+	// https://learn.microsoft.com/en-us/azure/aks/enable-host-encryption
+	// https://learn.microsoft.com/en-us/azure/virtual-machines/disk-encryption#encryption-at-host---end-to-end-encryption-for-your-vm-data
+	// +optional
+	EncryptionAtHost *bool `json:"encryptionAtHost,omitempty"`
```
matthchr, line 96: VM puts this in `SecurityProfile`. I don't know that we want to repeat the "...Profile" pattern, but even AgentPool has a `securityProfile` section (though they don't put this there).

I suppose you could've made the argument that FIPSMode should have gone there too...

Curious @tallaxes thoughts on if we should have a `security` field and put this stuff in it to keep a bit more organized versus putting all at the top level.

pkg/providers/instancetype/instancetypes.go
```diff
@@ -267,6 +271,24 @@ func (p *DefaultProvider) isInstanceTypeSupportedByImageFamily(skuName, imageFam
...
+	return p.supportsEncryptionAtHost(sku)
+}
+
+// supportsEncryptionAtHost checks if the SKU supports encryption at host
+func (p *DefaultProvider) supportsEncryptionAtHost(sku *skewer.SKU) bool {
+	value, err := sku.GetCapabilityString("EncryptionAtHostSupported")
+	if err != nil {
+		return false
+	}
+	return value == "True"
```
matthchr: Is it always cased this way?

Should we be case-insensitive just in case?

pkg/apis/v1beta1/aksnodeclass.go
matthchr: Actually if/when we add a field for DiskEncryptionSetID we also may want that to go in security too... so feels possibly worth having.

pkg/providers/instancetype/instancetypes.go
Bryce-Soghigian: We could use strings.EqualFold just in case, I did a quick CLI List + Grep but only found UpperCase true for this field though. 


Probably good to code defensively.

pkg/apis/v1beta1/aksnodeclass.go
Bryce-Soghigian: AWS has `blockDeviceMappings` in the EC2NodeClass which specifies some of the storage properties. 

```
spec:
  blockDeviceMappings:
    - deviceName: /dev/xvda
      ebs:
        volumeSize: 100Gi
        volumeType: gp3
        iops: 10000
        encrypted: true
        kmsKeyID: "1234abcd-12ab-34cd-56ef-1234567890ab"
        deleteOnTermination: true
        throughput: 125
        snapshotID: snap-0123456789
        volumeInitializationRate: 100
  ```

File: pkg/providers/instance/instance.go [hunk 4]
```diff
@@ -588,6 +589,15 @@ func setVMPropertiesBillingProfile(vmProperties *armcompute.VirtualMachineProper
...
 
+func setVMPropertiesSecurityProfile(vmProperties *armcompute.VirtualMachineProperties, nodeClass *v1beta1.AKSNodeClass) {
+	if nodeClass.Spec.EncryptionAtHost != nil {
+		if vmProperties.SecurityProfile == nil {
+			vmProperties.SecurityProfile = &armcompute.SecurityProfile{}
+		}
+		vmProperties.SecurityProfile.EncryptionAtHost = nodeClass.Spec.EncryptionAtHost
+	}
+}
+
```
comtalyst, line 616: In which field that Machine API supports this? I saw `EnableEncryptionAtHost` in AP profile, but not yet machine. Is that been discussed with their side?

charts/karpenter-crd/templates/karpenter.azure.com_aksnodeclasses.yaml
Bryce-Soghigian: I was also starting to wonder, since we added these nodeclass docs, if for each field we should a documentation section for it: https://learn.microsoft.com/en-us/azure/aks/node-autoprovision-aksnodeclass#azure-resource-tags for example and reference it in the cr

charts/karpenter-crd/templates/karpenter.azure.com_aksnodeclasses.yaml
```diff
@@ -60,6 +60,12 @@ spec:
               AKSNodeClassSpec is the top level specification for the AKS Karpenter Provider.
               This will contain configuration necessary to launch instances in AKS.
             properties:
+              encryptionAtHost:
```
matthchr: Need to regenerate templates or something?

Missing security field.

pkg/providers/instancetype/instancetypes.go
```diff
@@ -267,6 +271,24 @@ func (p *DefaultProvider) isInstanceTypeSupportedByImageFamily(skuName, imageFam
...
+	if !nodeClass.GetEncryptionAtHost() {
+		return true
+	}
+	// If EncryptionAtHost is enabled, only include instance types that support it
+	return p.supportsEncryptionAtHost(sku)
+}
+
+// supportsEncryptionAtHost checks if the SKU supports encryption at host
+func (p *DefaultProvider) supportsEncryptionAtHost(sku *skewer.SKU) bool {
+	value, err := sku.GetCapabilityString("EncryptionAtHostSupported")
```
matthchr, line 285: Doesn't skewer bake in a bunch of these helpers?

Should this be there eventually?

Bryce-Soghigian: yes the expectation is to move this into skewer(eventually)

PR #1155, opened 2025-09-16
fix: Don't block PATCH VM if billing extension is not installed

matthchr:
There are currently cases where a node can join the cluster successfully but not have the billing extension installed. This blocks identity propagation and tags propagation.

We tolerate a missing billing extension since this is currently possible.

* Unit test

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
NONE
```

### DIFF
--- a/pkg/controllers/nodeclaim/inplaceupdate/controller.go
+++ b/pkg/controllers/nodeclaim/inplaceupdate/controller.go
@@ -112,6 +112,7 @@ func (c *Controller) Reconcile(ctx context.Context, nodeClaim *karpv1.NodeClaim)
 	if err != nil {
 		return reconcile.Result{}, client.IgnoreNotFound(err)
 	}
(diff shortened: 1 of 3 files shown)


Bryce-Soghigian:
should the garbage collection controller remove a vm if its there but doesn't have the expected extensions? 

Or are we ok with just doing this and waiting for the machine api?

tallaxes (approved):

matthchr:
Holding off on merging this til I can get E2E test run internally passing

charliedmcb (approved):
lgtm

matthchr:
> should the garbage collection controller remove a vm if its there but doesn't have the expected extensions?
> Or are we ok with just doing this and waiting for the machine api?

Probably just wait for machine API? Don't think it makes sense to fix this in this PR certainly, and given how close machine API is and how long we've already lived with this behavior it seems OK to continue living with it for a little longer?

PR #1158, opened 2025-09-16
feat: a script to add AKS machines pool to a cluster

comtalyst:
Fixes # 

**Description**

Add a script to add AKS machines pool to a cluster, being used directly in development. Currently not being used by any other command yet.

To be a part of [Machine API integration](https://github.com/Azure/karpenter-provider-azure/pull/1102).

* Manually

- [ ] Yes, issue opened: # 
- [X] No

**Release Note**

```release-note

```

### DIFF
--- a/hack/deploy/add-aks-machines-pool.sh
+++ b/hack/deploy/add-aks-machines-pool.sh
@@ -1,0 +1,64 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+if [ "$#" -ne 4 ]; then
+    echo "Usage: $0 <subscription-id> <resource-group> <cluster-name> <nodepool-name>"
+    echo "This script adds a nodepool with 'machines' mode to an existing AKS cluster using Azure REST API."
+    echo "Example:"
+    echo "  $0 00000000-0000-0000-0000-000000000000 some-rg some-cluster somepool"
+    exit 1
+fi
+
+AZURE_SUBSCRIPTION_ID=$1
+RESOURCE_GROUP=$2
+CLUSTER_NAME=$3
+NODEPOOL_NAME=$4
+
+echo "Adding nodepool '$NODEPOOL_NAME' with machines mode to cluster '$CLUSTER_NAME' in resource group '$RESOURCE_GROUP'..."
+
+# Get access token
+echo "Getting access token..."
+ACCESS_TOKEN=$(az account get-access-token --query accessToken --output tsv)
+
+# REST API endpoint
+API_VERSION="2025-07-01"
+URL="https://management.azure.com/subscriptions/${AZURE_SUBSCRIPTION_ID}/resourceGroups/${RESOURCE_GROUP}/providers/Microsoft.ContainerService/managedClusters/${CLUSTER_NAME}/agentPools/${NODEPOOL_NAME}?api-version=${API_VERSION}"
+
+# Request body with machines mode
+REQUEST_BODY=$(cat <<EOF
+{
+  "properties": {
+    "mode": "Machines",
+  }
+}
+EOF
+)
+
+echo "Making REST API call..."
+echo "URL: $URL"
+echo "Request Body:"
+echo "$REQUEST_BODY"
+
+# Make the REST API call
+RESPONSE=$(curl -X PUT \
+  -H "Authorization: Bearer $ACCESS_TOKEN" \
+  -H "Content-Type: application/json" \
+  -d "$REQUEST_BODY" \
+  "$URL" \
+  -w "\n%{http_code}" \
+  -s)
+
+# Extract HTTP status code from response
+HTTP_CODE=$(echo "$RESPONSE" | tail -n1)
+RESPONSE_BODY=$(echo "$RESPONSE" | head -n -1)
+
+echo "HTTP Status Code: $HTTP_CODE"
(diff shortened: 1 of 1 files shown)


Bryce-Soghigian (commented):
is there a reason to do this outside of the makefile.az pattern we have been doing for all other setup steps?

comtalyst:
> is there a reason to do this outside of the makefile.az pattern we have been doing for all other setup steps?

This piece is just a quick dev tool. Integration with Makefile is more coupled to the existence of Machine API integration itself in the current plan (see https://github.com/Azure/karpenter-provider-azure/pull/1102). It would be in a different PR.

matthchr:
Even the makefile sometimes shells out to individual scripts, no?

hack/deploy/add-aks-machines-pool.sh
```diff
@@ -0,0 +1,64 @@
...
+NODEPOOL_NAME=$4
+
+echo "Adding nodepool '$NODEPOOL_NAME' with machines mode to cluster '$CLUSTER_NAME' in resource group '$RESOURCE_GROUP'..."
+
+# Get access token
+echo "Getting access token..."
+ACCESS_TOKEN=$(az account get-access-token --query accessToken --output tsv)
+
+# REST API endpoint
+API_VERSION="2025-07-01"
```
matthchr, line 21: Do we not have to use a preview API version?

comtalyst:
> Even the makefile sometimes shells out to individual scripts, no?

Yes. Though it is more that Makefile is a higher-level interface, and I think we should try to keep it "clean"--to reserve it for more-production-ready things. The list of files in this folder, on the other hand, is at lower level, so adding this less-relevant-to-production-Karpenter tool wouldn't hurt.

See https://github.com/Azure/karpenter-provider-azure/pull/1197

hack/deploy/add-aks-machines-pool.sh
comtalyst: Not necessary for this one (adding machines AgentPool). Adding machines is a different story.

matthchr: Oh good point.

hack/deploy/add-aks-machines-pool.sh
```diff
@@ -0,0 +1,64 @@
...
+URL="https://management.azure.com/subscriptions/${AZURE_SUBSCRIPTION_ID}/resourceGroups/${RESOURCE_GROUP}/providers/Microsoft.ContainerService/managedClusters/${CLUSTER_NAME}/agentPools/${NODEPOOL_NAME}?api-version=${API_VERSION}"
```
matthchr, line 22: Is it not possible to just use the vanilla `az cli` here?

Is the problem that it has a bunch of validations on AP shape that don't apply to Machines AP?

If that's the case, maybe leave a comment here explaining why you're doing it this way?

hack/deploy/add-aks-machines-pool.sh
```diff
@@ -0,0 +1,64 @@
...
+EOF
+)
+
+echo "Making REST API call..."
+echo "URL: $URL"
+echo "Request Body:"
+echo "$REQUEST_BODY"
+
+# Make the REST API call
+RESPONSE=$(curl -X PUT \
```
matthchr: TBH might be better to use `az rest` rather than this?

Something like `az rest --method put --url $URL --body $REQUEST_BODY` would probably do it?

It does the token management for you so you don't need to do this above:
`ACCESS_TOKEN=$(az account get-access-token --query accessToken --output tsv)`

wdyt?

hack/deploy/add-aks-machines-pool.sh
```diff
@@ -0,0 +1,64 @@
...
+  -H "Authorization: Bearer $ACCESS_TOKEN" \
+  -H "Content-Type: application/json" \
+  -d "$REQUEST_BODY" \
+  "$URL" \
+  -w "\n%{http_code}" \
+  -s)
+
+# Extract HTTP status code from response
+HTTP_CODE=$(echo "$RESPONSE" | tail -n1)
```
matthchr: Do we need to poll until AP in stable ProvState?

I assume we do?

hack/deploy/add-aks-machines-pool.sh
comtalyst: This API specifically (mode=Machine) is not available in CLI yet. Soon it will, just not at this stage.
Added a comment.

hack/deploy/add-aks-machines-pool.sh
comtalyst: Claude...

hack/deploy/add-aks-machines-pool.sh
comtalyst: Not really. The pool with machine mode is a blank structure.

matthchr: It still has a ProvState, doesn't it?

matthchr (approved):

hack/deploy/add-aks-machines-pool.sh
comtalyst: Yes but I don't think it will do anything that takes time. Can revisit if that's problematic.

PR #1162, opened 2025-09-17
refactor: rename generic "instance" to "VM" from their natural ties to VM instances

comtalyst:
Fixes # 

**Description**

The concept of "instance" is on the higher abstraction level than "vm" (`armcompute.VirtualMachine`), which is its only available implementation as of today. However, many places in the code have been using the word "instance" while making assumptions on VM-specific details. This could lead to confusion and bad practices if there will be more implementations of an "instance" in the future. In fact, [one is right around the corner](https://github.com/Azure/karpenter-provider-azure/pull/1102).

This PR renames VM-specific "instance" variables to denote its VM nature.

*

- [ ] Yes, issue opened: # 
- [X] No

**Release Note**

```release-note

```

### DIFF
--- a/Makefile-az.mk
+++ b/Makefile-az.mk
@@ -1,6 +1,7 @@
 AZURE_LOCATION ?= westus2
 AZURE_VM_SIZE ?= ""
 COMMON_NAME ?= karpenter
+ENABLE_AZURE_SDK_LOGGING ?= true
 ifeq ($(CODESPACES),true)
   AZURE_RESOURCE_GROUP ?= $(CODESPACE_NAME)
   AZURE_ACR_NAME ?= $(subst -,,$(CODESPACE_NAME))
@@ -138,8 +139,8 @@ az-mkaks-savm: az-mkrg ## Create experimental cluster with standalone VMs (+ ACR
 az-rmrg: ## Destroy test ACR and AKS cluster by deleting the resource group (use with care!)
 	az group delete --name $(AZURE_RESOURCE_GROUP)
 
-az-configure-values:  ## Generate cluster-related values for Karpenter Helm chart
-	hack/deploy/configure-values.sh $(AZURE_CLUSTER_NAME) $(AZURE_RESOURCE_GROUP) $(KARPENTER_SERVICE_ACCOUNT_NAME) $(AZURE_KARPENTER_USER_ASSIGNED_IDENTITY_NAME)
+az-configure-values:  ## Generate cluster-related values for Karpenter Helm chart and set middleware logging flag
+	hack/deploy/configure-values.sh $(AZURE_CLUSTER_NAME) $(AZURE_RESOURCE_GROUP) $(KARPENTER_SERVICE_ACCOUNT_NAME) $(AZURE_KARPENTER_USER_ASSIGNED_IDENTITY_NAME) $(ENABLE_AZURE_SDK_LOGGING)
 
 az-mkvmssflex: ## Create VMSS Flex (optional, only if creating VMs referencing this VMSS)
 	az vmss create --name $(AZURE_CLUSTER_NAME)-vmss --resource-group $(AZURE_RESOURCE_GROUP_MC) --location $(AZURE_LOCATION) \
--- a/hack/deploy/configure-values.sh
+++ b/hack/deploy/configure-values.sh
@@ -3,8 +3,8 @@ set -euo pipefail
 # This script interrogates the AKS cluster and Azure resources to generate
 # the karpenter-values.yaml file using the karpenter-values-template.yaml file as a template.
 
-if [ "$#" -ne 4 ]; then
-    echo "Usage: $0 <cluster-name> <resource-group> <karpenter-service-account-name> <karpenter-user-assigned-identity-name>"
+if [ "$#" -ne 5 ]; then
+    echo "Usage: $0 <cluster-name> <resource-group> <karpenter-service-account-name> <karpenter-user-assigned-identity-name> <enable-azure-sdk-logging>"
     exit 1
 fi
 
@@ -14,6 +14,7 @@ CLUSTER_NAME=$1
 AZURE_RESOURCE_GROUP=$2
 KARPENTER_SERVICE_ACCOUNT_NAME=$3
 AZURE_KARPENTER_USER_ASSIGNED_IDENTITY_NAME=$4
+ENABLE_AZURE_SDK_LOGGING=$5
 
 # Optional values through env vars:
 LOG_LEVEL=${LOG_LEVEL:-"info"}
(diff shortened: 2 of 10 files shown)


pkg/providers/instance/instance_test.go
```diff
@@ -17,90 +17,13 @@ limitations under the License.
...
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/network/armnetwork"
-	"github.com/Azure/karpenter-provider-azure/pkg/cache"
 	"github.com/stretchr/testify/assert"
-	corev1 "k8s.io/api/core/v1"
-	karpv1 "sigs.k8s.io/karpenter/pkg/apis/v1"
-	"sigs.k8s.io/karpenter/pkg/cloudprovider"
-	"sigs.k8s.io/karpenter/pkg/scheduling"
 )
 
-func TestGetPriorityCapacityAndInstanceType(t *testing.T) {
```
comtalyst, line 33: moved to vminstance_test.go

pkg/cloudprovider/cloudprovider.go
```diff
@@ -169,21 +169,22 @@ func (c *CloudProvider) Create(ctx context.Context, nodeClaim *karpv1.NodeClaim)
...
 		return nil, cloudprovider.NewCreateError(fmt.Errorf("creating instance failed, %w", err), CreateInstanceFailedReason, truncateMessage(err.Error()))
 	}
 
-	if err = c.handleInstanceCreation(ctx, instancePromise, nodeClaim); err != nil {
+	if err = c.handleVMInstanceCreation(ctx, vmPromise, nodeClaim); err != nil {
 		return nil, err
 	}
 
-	instance := instancePromise.VM
+	vm := vmPromise.VM // This is best-effort populated by Karpenter to be used to create the VM server-side. Not all fields are guaranteed to be populated, especially status fields.
```
matthchr, line 181: What exactly do status fields mean in the context of the CRP API?

pkg/controllers/nodeclaim/garbagecollection/nic_garbagecollection.go
```diff
@@ -45,20 +45,20 @@ const (
 )
 
 type NetworkInterface struct {
-	kubeClient       client.Client
-	instanceProvider instance.Provider
+	kubeClient         client.Client
+	vmInstanceProvider instance.VMProvider
```
matthchr, line 49: Is this struct only every going to take the VM provider?
I _think_ we don't do NIC GC for machines?

If no, I wonder if we want to do:
type NICLister interface {
    ListNics(...) ...
}

and then pass this a combinedProvider that does it for both and satisfies that interface?
Even if this doesn't apply here, it may apply elsewhere?

pkg/providers/instance/resperrorhandlers.go
```diff
@@ -31,11 +31,11 @@ import (
 type responseErrorHandler struct {
 	name                string
 	matchError          func(error) bool
-	handleResponseError func(ctx context.Context, provider *DefaultProvider, sku *skewer.SKU, instanceType *corecloudprovider.InstanceType, zone, capacityType string, responseError error) error
+	handleResponseError func(ctx context.Context, provider *DefaultVMProvider, sku *skewer.SKU, instanceType *corecloudprovider.InstanceType, zone, capacityType string, responseError error) error
```
matthchr, line 34: minor, style: `DefaultVMProvider` is a weird name... default-ness is (IMO) a property of usage (i.e. used by default), not a property of the type itself. It might be cleaner to just call this `VMProvider`.

With that said, this is very minor so you can feel free to ignore.

pkg/providers/instance/resperrorhandlers.go
```diff
@@ -31,11 +31,11 @@ import (
 type responseErrorHandler struct {
```
matthchr, line 31: Are we going to put `providers/instance/...` into `providers/instance/vm` and `providers/instance/machine`?

It might make sense to do so?

If we do, then calling this `DefaultVMProvider` may be stuttering and we could just call it/use it as `vm.DefaultProvider` and `machine.DefaultProvider` (or `machine.Provider` + `vm.Provider` if you prefer my renaming comment below).

pkg/providers/instance/vminstance.go
```diff
@@ -100,7 +100,7 @@ type VirtualMachinePromise struct {
 	Wait func() error
 }
 
-type Provider interface {
+type VMProvider interface {
```
matthchr, line 103: Same question about package names changing for cleaner separation.

matthchr: applies throughout

pkg/cloudprovider/cloudprovider.go
comtalyst: Read-only fields E.g., ID. They are naturally not populated before we send create request. Although, we did best-effortly did that after just so we can use this struct without calling a `GET` (or hack into the client).

pkg/controllers/nodeclaim/garbagecollection/nic_garbagecollection.go
comtalyst: Note "VM instance" shall means "VM-powered instance", not "VM". So it includes NIC behind the abstraction. Not sure if there is a better name though.

We don't do NIC GC for machines. And NIC GC currently call instance provider to list NIC. I _think_ it is simpler to just keep that pattern, calling only VM instance provider, while NIC GC holding on the knowledge of which kind of instance we do NIC GC on. 

The current pattern follow the idea that these two instance providers are different even on the interface assumptions. So, it should be the consumer that acknowledges. 
Alternatively, if we want the knowledge of "we don't do NIC GC for machines" to fall into instance provider instead, the interface should be `ListNicsForGC` instead of `ListNics`, which IMO is weird if we want to return nothingness for machine provider just because we don't want to garbage collect it. Though, I don't think this pattern is worth pursuing right now.

pkg/providers/instance/resperrorhandlers.go
comtalyst: I will ignore for now. Might follow-up later.

pkg/providers/instance/resperrorhandlers.go
comtalyst: > Are we going to put providers/instance/... into providers/instance/vm and providers/instance/machine?

Not currently, and not sure if it is worth it as these providers are "of the same category/purpose". But might give it some more thought.

pkg/providers/instance/vminstance.go
comtalyst: (this is inevitable per above comment, but will give it more thought and might follow-up)

pkg/providers/instance/resperrorhandlers.go
matthchr: I kinda think it would make things cleaner.

comtalyst: (discussed offline--filed a ticket for potential clean-up, example concerns include import cycles/exports)

PR #1165, opened 2025-09-17
refactor: more instance-generic async provisioning

comtalyst:
Fixes # 

**Description**

In cloudprovider module, async provisioning handling logic unncessarily assume an instance is a VM instance. With the new instance kind coming up in [Machine API integration](https://github.com/Azure/karpenter-provider-azure/pull/1102), this refactor will lead to more code reusability between instances, as well as more clear layering regarding standalone NodeClaim handling.

* Manual runs on provisioning failures with [Machine API integration](https://github.com/Azure/karpenter-provider-azure/pull/1102)
* TODO(?)

- [ ] Yes, issue opened: # 
- [X] No

**Release Note**

```release-note

```


pkg/providers/instance/instancepromise.go
```diff
@@ -0,0 +1,31 @@
...
+*/
+
+package instance
+
+import (
+	"context"
+)
+
+// Intended for lifecycle handling on the higher abstractions.
+type InstancePromise interface {
```
matthchr: minor, style: `instance.InstancePromise` is somewhat stuttering -- maybe just `Promise`?

pkg/providers/instance/vminstance.go
```diff
@@ -96,8 +96,24 @@ func GetManagedExtensionNames(provisionMode string) []string {
...
+	WaitFunc func() error
+
+	providerRef VMProvider
+}
+
+func (p *VirtualMachinePromise) Cleanup(ctx context.Context) error {
+	// This won't clean up leaked NICs if the VM doesn't exist... intentional?
+	// From Delete(): "Leftover network interfaces (if any) will be cleaned by by GC controller"
+	// Still, we could try to DeleteNic()?
+	return p.providerRef.Delete(ctx, lo.FromPtr(p.VM.Name))
```
matthchr, line 109: This seems _fine_ to me, but a bit weird?

Usage is in `cloudprovider.go`, where we have the `instanceProvider`, and all this is doing is calling a method on the `instanceProvider` which cloud provider already has access to.

Is this here just because in the future when there are multiple instance providers, you _could_ pass the provider used down to `handleInstancePromise` but it's more you have to pass in to that method and it was easier to put it here since this structure already existed?

matthchr: Ah, and the providers don't share an interface which is slightly annoying so you're trying to abstract over it here which seems reasonable. Though you could also just do this:
```
type deleter interface {
    Delete(context.Context, string) error
}
```

(since even though the whole interface isn't the same the delete method should be I think?) and then pass that to `handleInstancePromise`.

Overall I am OK with this as-is but just wanted to check if you considered the other options and if I understood why you did it this way. Assuming I understood correctly above, I do agree it seems probably cleaner than that alternative (though only by a bit)

pkg/providers/instance/vminstance.go
```diff
@@ -96,8 +96,24 @@ func GetManagedExtensionNames(provisionMode string) []string {
 type Resource = map[string]interface{}
 
 type VirtualMachinePromise struct {
-	VM   *armcompute.VirtualMachine
-	Wait func() error
+	VM       *armcompute.VirtualMachine
+	WaitFunc func() error
```
matthchr, line 100: minor, style: You could make this `wait` instead and it wouldn't collide w/ the function - but I'm fine with `WaitFunc` as well.

pkg/cloudprovider/cloudprovider.go
```diff
@@ -183,37 +164,84 @@ func (c *CloudProvider) Create(ctx context.Context, nodeClaim *karpv1.NodeClaim)
...
 		}
-	}()
+	}
+	// For NodePool-managed nodeclaims, launch a single goroutine to poll the returned promise.
+	// Note that we could store the LRO details on the NodeClaim, but we don't bother today because Karpenter
+	// crashes should be rare, and even in the case of a crash, as long as the node comes up successfully there's
+	// no issue. If the node doesn't come up successfully in that case, the node and the linked claim will
+	// be garbage collected after the TTL, but the cause of the nodes issue will be lost, as the LRO URL was
+	// only held in memory.
+	// go c.waitOnInstancePromise(ctx, instancePromise, nodeClaim, false)
```
matthchr: Why get rid of `go c.waitOnInstancePromise(ctx, instancePromise, nodeClaim, false)` in favor of doing it all inline here?

It seemed cleaner the other way?

comtalyst: (From the old code) In `handleVMInstanceCreation`, `waitOnPromise` is similar to the case for standalone NodeClaim with small differences (panic catching, the boolean in the end). I think they should have the same layering, given the standalone NodeClaim case already rely on the implementation of `waitOnPromise`.

One way is to have another boolean for `waitOnPromise` indicating whether it is standalone NodeClaim. But that means the knowledge of standalone NodeClaim handling will be in more places which are tightly coupled to each other, and each function have to support two modes. 
With that, I think it is "cleaner" for all those knowledge to be in the same function, here. This is also given that the function is yet to be too complicated and too many duplicated codes between two modes.

pkg/providers/instance/instancepromise.go
comtalyst: The trade-off will be that it could be overgeneralized within the module where `instance.` does not exist. IMO it is better to stutter than overgeneralize(?).

pkg/providers/instance/vminstance.go
comtalyst: I think `WaitFunc()` is more clear.

pkg/providers/instance/vminstance.go
comtalyst: Yes, the two providers do not share interface. 

Between `deleter` and generic `InstancePromise` interface (currently), I think the `InstancePromise` already serves the purpose of the `deleter`? Then each instance kind could implement that like in this code.

pkg/providers/instance/instancepromise.go
matthchr: It's not actually really used in this package at all from what I can see? AFAIK all usages of this type at least in this PR are `instance.InstancePromise` so everywhere this type is actually used, it repeats itself.

General pattern in Go is to avoid stuttering AFAIK: https://google.github.io/styleguide/go/decisions#repetition so I think in general err on that side and only repeat yourself if you really have to to maintain clarity, which at least IMO you do not have to here.

pkg/providers/instance/vminstance.go
matthchr: Fine with the above. Closing this.

pkg/cloudprovider/cloudprovider.go
```diff
@@ -183,37 +164,84 @@ func (c *CloudProvider) Create(ctx context.Context, nodeClaim *karpv1.NodeClaim)
...
-			err := fmt.Errorf("%v", r)
-			log.FromContext(ctx).Error(err, "panic during waitOnPromise")
+// handleInstancePromise handles the instance promise, primarily deciding on sync/async provisioning.
+func (c *CloudProvider) handleInstancePromise(ctx context.Context, instancePromise instance.InstancePromise, nodeClaim *karpv1.NodeClaim) error {
+	if isNodeClaimStandalone(nodeClaim) {
+		// Standalone NodeClaims aren't re-queued for reconciliation in the provision_trigger controller,
+		// so we delete them synchronously. After marking Launched=true,
+		// their status can't be reverted to false once the delete completes due to how core caches nodeclaims in
+		// the launch controller. This ensures we retry continuously until we hit the registration TTL
+		// return c.waitOnInstancePromise(ctx, instancePromise, nodeClaim, true)
```
matthchr: minor: Remove commented code.

pkg/cloudprovider/cloudprovider.go
matthchr: remove commented code here?

matthchr: I'm not advocating to pass a boolean to instancePromise.Wait() -- agree handling it at the cloudprovider level makes more sense.

My point is, this:
```
go func() {
    <some big inline function>
}
```
is IMO harder to read than `go <localfunc>()`. I see what you're getting at with moving things to the same level, and I'm ambivalent on `handleNodeClaimCreationError` handling removing the NodeClaim via `removeNodeClaim bool` versus the caller doing that after calling `handleNodeClaimCreationError`, but I am not sold that putting things on the same level is adding clarity when it also comes with putting a big inlined function in here.

If you want things on the same level IMO:
```
handleInstancePromise(...) {
    if standaloneNodeClaim {
        c.handleStandaloneNodeClaimPromise(...)
    } else {
        go c.handlePromise()
    }
}
```

or some variant of the above (modulo names) is clearer. Then handlePromise and handleStandaloneNode are still on the same level (each one is handling their one case) and the top level handleInstancePromise is just routing w/ no logic. As you have it, it's routing and also logic. Before, it was _basically_ just routing, as the standalone case was a single function call ti instancePromise.Wait() + error popagation.

So you've taken a function that was purely a switch case routing and made it more complex (== increased cyclomatic complexity), including both routing and logic.

Not going to block the PR on this but I don't think it's clearer I think it's less clear, personally.

matthchr (approved):
Approved with some more nits

pkg/cloudprovider/cloudprovider.go
comtalyst: My concerns on the routing-logic separation pattern you suggest, and even the idea of optimizing for cyclomatic complexity minimization, would be:
- The separation is not complete as `go` (logic of the decision to handle them sync/async) is still in the routing function
- `handleStandaloneNodeClaimPromise()` still pairs with `handlePromise()` in a part of implementation/usage assumptions. Separating them further would hurt that knowledge of parity
- More micro-functions might look cleaner in code, but, from the big picture, it takes more to see this layering (e.g., the fact that `handleStandaloneNodeClaimPromise()` is one layer below `handleInstancePromise()`).

In the end it is a trade-off though.

matthchr: > The separation is not complete as go (logic of the decision to handle them sync/async) is still in the routing function

I still don't see this as a compelling argument to avoid pushing them both down a level to make the functions smaller and more self-contained, but not blocking on this atm.

PR #1166, opened 2025-09-17
refactor: split offerings handling into a sub-module

comtalyst:
Fixes # 

**Description**

The processes of instance provisioning and SKU/Priority/zone selection/handling should be clearly distinguishable from one another. However, current code does not seem to reflect that well enough. This concern is also previously noted in https://github.com/Azure/karpenter-provider-azure/pull/791. 
Naturally, it has been an obstacle to efficient improvements on each, especially for upcoming [Machine API integration](https://github.com/Azure/karpenter-provider-azure/pull/1102).

This PR attempts to separate the latter into its own sub-module, with the logic intact.

* Acceptance tests in instancetype/suite_test.go, targeting expected provisioning failures and handling
* E2E: https://github.com/Azure/karpenter-provider-azure/actions/runs/17817499782 + https://github.com/Azure/karpenter-provider-azure/actions/runs/17793168031

- [ ] Yes, issue opened: # 
- [X] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/providers/instance/vminstance.go
+++ b/pkg/providers/instance/vminstance.go
@@ -81,8 +81,24 @@ func GetManagedExtensionNames(provisionMode string) []string {
 type Resource = map[string]interface{}
 
 type VirtualMachinePromise struct {
-	VM   *armcompute.VirtualMachine
-	Wait func() error
+	VM       *armcompute.VirtualMachine
+	WaitFunc func() error
+
+	providerRef VMProvider
+}
+
+func (p *VirtualMachinePromise) Cleanup(ctx context.Context) error {
+	// This won't clean up leaked NICs if the VM doesn't exist... intentional?
+	// From Delete(): "Leftover network interfaces (if any) will be cleaned by by GC controller"
+	// Still, we could try to DeleteNic()?
+	return p.providerRef.Delete(ctx, lo.FromPtr(p.VM.Name))
+}
+
+func (p *VirtualMachinePromise) Wait() error {
+	return p.WaitFunc()
+}
+func (p *VirtualMachinePromise) GetInstanceName() string {
+	return lo.FromPtr(p.VM.Name)
 }
 
 type VMProvider interface {
@@ -738,7 +754,8 @@ func (p *DefaultVMProvider) beginLaunchInstance(
 	result.VM.Properties.TimeCreated = lo.ToPtr(time.Now())
 
 	return &VirtualMachinePromise{
-		Wait: func() error {
+		providerRef: p,
+		WaitFunc: func() error {
 			if result.Poller == nil {
 				// Poller is nil means the VM existed already and we're done.
 				// TODO: if the VM doesn't have extensions this will still happen and we will have to
--- a/pkg/cloudprovider/cloudprovider.go
+++ b/pkg/cloudprovider/cloudprovider.go
@@ -97,30 +97,6 @@ func New(
 	}
 }
 
-// Create a node given the constraints.
-func (c *CloudProvider) handleVMInstanceCreation(ctx context.Context, vmPromise *instance.VirtualMachinePromise, nodeClaim *karpv1.NodeClaim) error {
-	if c.isStandaloneNodeClaim(nodeClaim) {
-		// processStandaloneNodeClaimDeletion:
-		//   Standalone NodeClaims aren’t re-queued for reconciliation in the provision_trigger controller,
-		//   so we delete them synchronously. After marking Launched=true,
-		//   their status can’t be reverted to false once the delete completes due to how core caches nodeclaims in
-		// 	 the lanch controller. This ensures we retry continuously until we hit the registration TTL
-		err := vmPromise.Wait()
-		if err != nil {
-			return c.handleNodeClaimCreationError(ctx, err, vmPromise, nodeClaim, false)
-		}
-	} else {
-		// For NodePool-managed nodeclaims, launch a single goroutine to poll the returned promise.
-		// Note that we could store the LRO details on the NodeClaim, but we don't bother today because Karpenter
-		// crashes should be rare, and even in the case of a crash, as long as the node comes up successfully there's
-		// no issue. If the node doesn't come up successfully in that case, the node and the linked claim will
-		// be garbage collected after the TTL, but the cause of the nodes issue will be lost, as the LRO URL was
-		// only held in memory.
-		go c.waitOnPromise(ctx, vmPromise, nodeClaim)
-	}
-	return nil
-}
-
 func (c *CloudProvider) validateNodeClass(nodeClass *v1beta1.AKSNodeClass) error {
 	nodeClassReady := nodeClass.StatusConditions().Get(status.ConditionReady)
 	if nodeClassReady.IsFalse() {
@@ -170,12 +146,17 @@ func (c *CloudProvider) Create(ctx context.Context, nodeClaim *karpv1.NodeClaim)
 	if len(instanceTypes) == 0 {
 		return nil, cloudprovider.NewInsufficientCapacityError(fmt.Errorf("all requested instance types were unavailable during launch"))
 	}
+
+	return c.createVMInstance(ctx, nodeClass, nodeClaim, instanceTypes)
+}
+
+func (c *CloudProvider) createVMInstance(ctx context.Context, nodeClass *v1beta1.AKSNodeClass, nodeClaim *karpv1.NodeClaim, instanceTypes []*cloudprovider.InstanceType) (*karpv1.NodeClaim, error) {
 	vmPromise, err := c.vmInstanceProvider.BeginCreate(ctx, nodeClass, nodeClaim, instanceTypes)
 	if err != nil {
 		return nil, cloudprovider.NewCreateError(fmt.Errorf("creating instance failed, %w", err), CreateInstanceFailedReason, truncateMessage(err.Error()))
 	}
 
-	if err = c.handleVMInstanceCreation(ctx, vmPromise, nodeClaim); err != nil {
+	if err := c.handleInstancePromise(ctx, vmPromise, nodeClaim); err != nil {
 		return nil, err
 	}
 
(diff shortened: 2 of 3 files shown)


pkg/providers/instance/offerings/commonerrorhandlers.go
```diff
@@ -148,48 +113,48 @@ func handleSKUNotAvailableError(ctx context.Context, provider *DefaultVMProvider
...
 	// InsufficientCapacityError is appropriate here because trying any other instance type will not help
 	return corecloudprovider.NewInsufficientCapacityError(
 		fmt.Errorf(
 			"regional %s vCPU quota limit for subscription has been reached. To scale beyond this limit, please review the quota increase process here: https://learn.microsoft.com/en-us/azure/quotas/regional-quota-requests",
 			capacityType))
 }
 
-func cpuLimitIsZero(err error) bool {
-	return strings.Contains(err.Error(), "Current Limit: 0")
+func cpuLimitIsZero(errorMessage string) bool {
```
matthchr, line 158: Why this change?

pkg/providers/instance/resperrorhandlers.go
```diff
@@ -14,102 +14,67 @@ See the License for the specific language governing permissions and
...
 
-func handleLowPriorityQuotaError(ctx context.Context, provider *DefaultVMProvider, sku *skewer.SKU, instanceType *corecloudprovider.InstanceType, zone, capacityType string, err error) error {
+func handleLowPriorityQuotaError(ctx context.Context, unavailableOfferings *cache.UnavailableOfferings, sku *skewer.SKU, instanceType *corecloudprovider.InstanceType, zone, capacityType, errorCode, errorMessage string) error {
 	// Mark in cache that spot quota has been reached for this subscription
-	provider.unavailableOfferings.MarkSpotUnavailableWithTTL(ctx, SubscriptionQuotaReachedTTL)
+	unavailableOfferings.MarkSpotUnavailableWithTTL(ctx, SubscriptionQuotaReachedTTL)
 	return fmt.Errorf("this subscription has reached the regional vCPU quota for spot (LowPriorityQuota). To scale beyond this limit, please review the quota increase process here: https://docs.microsoft.com/en-us/azure/azure-portal/supportability/low-priority-quota")
 }
 
-func handleSKUFamilyQuotaError(ctx context.Context, provider *DefaultVMProvider, sku *skewer.SKU, instanceType *corecloudprovider.InstanceType, zone, capacityType string, err error) error {
```
matthchr, line 112: Can we not accept `err` anymore?

Why not? the SDK error type is shared among CRP and AKS, isn't it?

pkg/providers/instance/offerings/commonerrorhandlers.go
```diff
@@ -14,102 +14,67 @@ See the License for the specific language governing permissions and
...
-			matchError:          sdkerrors.RegionalQuotaHasBeenReached,
-			handleResponseError: handleRegionalQuotaError,
-		},
+		unavailableOfferings.MarkUnavailableWithTTL(ctx, reason, instanceType.Name, zone, karpv1.CapacityTypeOnDemand, ttl)
+		unavailableOfferings.MarkUnavailableWithTTL(ctx, reason, instanceType.Name, zone, karpv1.CapacityTypeSpot, ttl)
 	}
 }
 
-func handleLowPriorityQuotaError(ctx context.Context, provider *DefaultVMProvider, sku *skewer.SKU, instanceType *corecloudprovider.InstanceType, zone, capacityType string, err error) error {
+func handleLowPriorityQuotaError(ctx context.Context, unavailableOfferings *cache.UnavailableOfferings, sku *skewer.SKU, instanceType *corecloudprovider.InstanceType, zone, capacityType, errorCode, errorMessage string) error {
```
matthchr, line 71: Why can't we accept error anymore?

The SDK type is shared among CRP and AKS, isn't it?

Could we just pass the actual SDK error type here, rather than code/message as strings?
edit: applies throughout

pkg/providers/instance/offerings/commonerrorhandlers.go
```diff
@@ -14,102 +14,67 @@ See the License for the specific language governing permissions and
...
+	OverconstrainedAllocationFailureReason      = "OverconstrainedAllocationFailure"
+	SKUNotAvailableReason                       = "SKUNotAvailable"
+
+	SubscriptionQuotaReachedTTL = 1 * time.Hour
+	AllocationFailureTTL        = 1 * time.Hour
+	SKUNotAvailableSpotTTL      = 1 * time.Hour
+	SKUNotAvailableOnDemandTTL  = 23 * time.Hour
+)
+
+type commonErrorHandle func(ctx context.Context, unavailableOfferings *cache.UnavailableOfferings, sku *skewer.SKU, instanceType *corecloudprovider.InstanceType, zone, capacityType, errorCode, errorMessage string) error
```
matthchr: minor:

```suggestion
type errorHandler func(ctx context.Context, unavailableOfferings *cache.UnavailableOfferings, sku *skewer.SKU, instanceType *corecloudprovider.InstanceType, zone, capacityType, errorCode, errorMessage string) error
```

maybe?

pkg/providers/instance/offerings/commonerrorhandlers_test.go
```diff
@@ -0,0 +1,197 @@
...
+import (
+	"context"
+	"testing"
+
+	"github.com/Azure/azure-sdk-for-go/profiles/latest/compute/mgmt/compute"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
+	"github.com/Azure/karpenter-provider-azure/pkg/apis/v1beta1"
+	"github.com/Azure/karpenter-provider-azure/pkg/cache"
+	"github.com/Azure/skewer"
+	"github.com/stretchr/testify/assert"
```
matthchr, line 28: pet peeve of mine, but we should use a single assertion library. I'm fine with some tests being Ginkgo and some not (I prefer not but at least for E2E tests I see the value) - but IMO we should just use Ginkgo for asserts in both.

Otherwise you get a mix of assertion styles and it becomes more confusing (IMO) to read the code and the test output.

I know we use testify some elsewhere, but only in 16 files compared to 69 files for Gomega (many of those are also Ginkgo, but not all).

Let's unify on Gomega for assertions regardless of Ginkgo or not?

matthchr: Ahh you didn't actually change this (just copied it)

pkg/providers/instance/vminstance.go
```diff
@@ -52,26 +49,14 @@ import (
...
-	ZonalAllocationFailureReason                = "ZonalAllocationFailure"
-	OverconstrainedZonalAllocationFailureReason = "OverconstrainedZonalAllocationFailure"
-	OverconstrainedAllocationFailureReason      = "OverconstrainedAllocationFailure"
-	SKUNotAvailableReason                       = "SKUNotAvailable"
-
-	SubscriptionQuotaReachedTTL = 1 * time.Hour
-	AllocationFailureTTL        = 1 * time.Hour
-	SKUNotAvailableSpotTTL      = 1 * time.Hour
-	SKUNotAvailableOnDemandTTL  = 23 * time.Hour
+	KarpCapacityTypeToVMPriority = map[string]armcompute.VirtualMachinePriorityTypes{
```
matthchr, line 52: Why add `Karp` prefix here and below?

pkg/providers/instance/offerings/resperrorhandlers.go
```diff
@@ -0,0 +1,94 @@
...
+	"github.com/Azure/skewer"
+	corecloudprovider "sigs.k8s.io/karpenter/pkg/cloudprovider"
+)
+
+type responseErrorHandler struct {
+	match  func(error) bool
+	handle commonErrorHandle
+}
+
+type ResponseErrorHandling struct {
```
matthchr: minor, style: Consider just:

```suggestion
type ResponseErrorHandler struct {
```

pkg/providers/instance/offerings/resperrorhandlers.go
```diff
@@ -0,0 +1,94 @@
...
+
+func (h *ResponseErrorHandling) extractErrorCodeAndMessage(err error) (string, string) {
+	var respErr *azcore.ResponseError
+	if errors.As(err, &respErr) {
+		return respErr.ErrorCode, respErr.Error()
+	}
+	return "", err.Error()
+}
+
+func (h *ResponseErrorHandling) HandleResponseError(ctx context.Context, sku *skewer.SKU, instanceType *corecloudprovider.InstanceType, zone, capacityType string, responseError error) error {
```
matthchr: minor
```suggestion
func (h *ResponseErrorHandling) HandleError(ctx context.Context, sku *skewer.SKU, instanceType *corecloudprovider.InstanceType, zone, capacityType string, responseError error) error {
```

Or even just Handle

pkg/providers/instance/offerings/offerings.go
```diff
@@ -0,0 +1,130 @@
...
+}
+
+func getOfferingZone(offering *corecloudprovider.Offering) string {
+	return offering.Requirements.Get(v1.LabelTopologyZone).Any()
+}
+
+// GetAllSingleValuedRequirementLabels converts instanceType.Requirements to labels
+// Like   instanceType.Requirements.Labels() it uses single-valued requirements
+// Unlike instanceType.Requirements.Labels() it does not filter out restricted Node labels
+func GetAllSingleValuedRequirementLabels(instanceType *corecloudprovider.InstanceType) map[string]string {
```
matthchr, line 110: Is offerings really the right place for this method?

pkg/providers/instance/offerings/resperrorhandlers.go
```diff
@@ -0,0 +1,94 @@
...
+		if handler.match(responseError) {
+			errorCode, errorMessage := h.extractErrorCodeAndMessage(responseError)
+			return handler.handle(ctx, h.UnavailableOfferings, sku, instanceType, zone, capacityType, errorCode, errorMessage)
+		}
+	}
+
+	return nil
+}
+
+func DefaultResponseErrorHandlers() []responseErrorHandler {
```
matthchr: I think you could do away with this being exported and just make a `NewResponseErrorHandler(cache *cache.UnavailableOfferings) ResponseErrorHandling ` method that constructed the `ResponseErrorHandling` type and filled out the handlers.

Saves the caller needing to "know" to put these there.

pkg/providers/instance/vminstance.go
```diff
@@ -738,8 +723,12 @@ func (p *DefaultVMProvider) beginLaunchInstance(
 		if skuErr != nil {
 			return nil, fmt.Errorf("failed to get instance type %q: %w", instanceType.Name, err)
 		}
-		azErr := p.handleResponseErrors(ctx, sku, instanceType, zone, capacityType, err)
-		return nil, azErr
+		handledError := p.errorHandling.HandleResponseError(ctx, sku, instanceType, zone, capacityType, err)
+		if handledError != nil {
+			// Handled in caches, but might also contain core-expected errors (e.g., InsufficientCapacityError)
```
matthchr: Comment is a bit confusing: What does "handled in caches" mean?

pkg/providers/instance/vminstance.go
```diff
@@ -766,8 +755,12 @@ func (p *DefaultVMProvider) beginLaunchInstance(
 				if skuErr != nil {
 					return fmt.Errorf("failed to get instance type %q: %w", instanceType.Name, err)
 				}
-				azErr := p.handleResponseErrors(ctx, sku, instanceType, zone, capacityType, err)
-				return azErr
+				handledError := p.errorHandling.HandleResponseError(ctx, sku, instanceType, zone, capacityType, err)
+				if handledError != nil {
+					// Handled in caches, but might also contain core-expected errors (e.g., InsufficientCapacityError)
```
matthchr: Similar comment on this one too.

pkg/providers/instance/offerings/commonerrorhandlers_test.go
comtalyst: Yea I just copied it. 
For the idea in general, I am half-agree actually. Two styles in one codebase is clearly confusing. Although, if we limit to just two, it might be _okay_. The value I see in that is that Gomega (at least in the ones we have) are "simpler" compared to Ginkgo. But given Ginkgo is more "necessary" for E2E, Ginkgo for everything will resolve the two styles concern, but that hurts the simplicity IMO.
Although, that means we have to distinguish E2E and traditional unit tests all the time. Sometimes the line becomes blurry, and categorization may just make things inefficient.

But I think this sounds like a larger discussion outside of this scope.

pkg/providers/instance/vminstance.go
comtalyst: To denote that this `CapacityType` is in Karpenter's format (in contrast to `VMPriority`), from `karpv1`.

pkg/providers/instance/vminstance.go
comtalyst: Attempted to improve. Is it more clear?

pkg/providers/instance/vminstance.go
comtalyst: (will keep the discussion in another comment above)

pkg/providers/instance/offerings/resperrorhandlers.go
comtalyst: Not sure if that will be confusable with `responseErrorHandler` right above.
Though, I did rename both. Wdyt?

pkg/providers/instance/offerings/offerings.go
comtalyst: I think so? Requirement labels are from instance types, which fall under the scope of offerings.

pkg/providers/instance/offerings/resperrorhandlers.go
comtalyst: The caller will be the same though, just calling `NewResponseErrorHandler` instead of `DefaultResponseErrorHandlers`. Neither requires the caller to provide handlers, and both still rely on the caller knowing that the handlers are compatible. But, don't think where will be other than "Default" anytime soon. And `NewResponseErrorHandler` is cleaner.

pkg/providers/instance/offerings/commonerrorhandlers.go
comtalyst: See https://github.com/Azure/karpenter-provider-azure/pull/1196 for Machine API's addition. It is not an `error` type implementation.

pkg/providers/instance/resperrorhandlers.go
comtalyst: Not for the "preserved CRP error" returned by Machine API. This error is not being reflected in the SDK polling error. See https://github.com/Azure/karpenter-provider-azure/blob/575faa339f2b5de312976c87ddedb45e0a8a50d5/pkg/providers/instance/aksmachineinstance.go#L458

pkg/providers/instance/offerings/commonerrorhandlers.go
comtalyst: (will be discussed in other comments)

pkg/providers/instance/offerings/commonerrorhandlers.go
comtalyst: Changed to `errorHandle`. I think it still shouldn't be "Handl**er**" though, as it can be confused with `ResponseErrorHandler` and such, which is on a different layer. An alternative to this is `errorHandlerFunc`.

matthchr: IMO it's OK to have 2 different handlers (with different names, obviously). With that said if you feel strongly: `errorHandlerFunc` is IMO better than `errorHandle`. My ranking is:
`errorHandler`, `errorHandlerFunc`, ..., `errorHandle`.

pkg/providers/instance/offerings/commonerrorhandlers_test.go
matthchr: My point is more about assertion libraries than test formats, tohugh I agree both are worth talking about.

I'm fine with:
1. Gomega (asserts) in go-test
2. Gomega (asserts) in Ginkgo (test format)

I'm less fine with
1. Gomega (asserts) in go-test
2. testify (asserts) in go-test
3. Gomega (asserts) in Ginkgo

So IMO it's not 1 vs 2, it's 2 vs 3.

matthchr: agree you dont need to fix here though.

pkg/providers/instance/resperrorhandlers.go
matthchr: Hmm this is unfortunate -- why aren't we getting proxied CRP errors from the LRO? What advantage is there at the machine API level of AKS RP wrapping the errors.

From a user perspective, if I'm using this API to make my hot new autoscaler or whatever... isn't this pattern harder than the standard SDK pattern?

pkg/providers/instance/offerings/resperrorhandlers.go
matthchr: lgtm

pkg/providers/instance/offerings/offerings.go
matthchr: Happy with tthis for now

pkg/providers/instance/vminstance.go
matthchr: Yes, thanks

matthchr (approved):
Approved.

My main issue here remaining is the fact that the error shape from machine API cannot be used like the CRP error shape, we have to go get the error off the entity itself.

I think that should be fixed (or else at least I want to understand better why it has to be that way), but that is more a machine API behavior problem and less an issue with this Karpenter code.

LGTM.

PR #1169, opened 2025-09-18
feat(images): add expired image upgrade logic to nodeclass status controller

charliedmcb:
Fixes # 

**Description**
Add 90-day expiration logic for Linux images to comply with AKS requirements. Images older than 90 days are automatically upgraded to the latest available version.

*

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/controllers/nodeclass/status/images.go
+++ b/pkg/controllers/nodeclass/status/images.go
@@ -104,6 +104,9 @@ func (r *NodeImageReconciler) Register(_ context.Context, m manager.Manager) err
 //   - 5. Handles update cases when customer changes image family, SIG usage, or other means of image selectors
 //   - 6. Handles softly adding newest image version of any newly supported SKUs by Karpenter
 //
+// Scenario C: Upgrade expired images
+//   - 7. Upgrades Linux images that are more than 90 days old to comply with AKS requirements
+//
 // Note: While we'd currently only need to store a SKU -> version mapping in the status for avilaible Images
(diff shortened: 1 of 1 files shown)


pkg/controllers/nodeclass/status/images.go
```diff
@@ -160,7 +165,9 @@ func (r *NodeImageReconciler) Reconcile(ctx context.Context, nodeClass *v1beta1.
 	}
 	if !shouldUpdate {
 		// Scenario B: Calculate any partial update based on image selectors, or newly supports SKUs
-		goalImages = overrideAnyGoalStateVersionsWithExisting(nodeClass, goalImages)
+		updatedImages := overrideAnyGoalStateVersionsWithExisting(nodeClass, goalImages)
+		// Scenario C: Upgrade any images that are more than 90 days old (AKS compliance requirement)
+		goalImages = upgradeAnyExpiredImages(ctx, updatedImages, goalImages)
```
charliedmcb, line 170: instead of expiring the images, we could set the images not ready with reason `ImagesExpired` to prevent provisioning, and only bump them out of this particular unready state when a K8s upgrade occurs, or a maintenance window opens.

Something like:
```
nodeClass.StatusConditions().SetFalse(v1beta1.ConditionTypeImagesReady, "ImagesExpired", "Images are older than 90 days, and have expired. Must open a maintenance window or preform a Kubernetes upgrade")
```

PR #1172, opened 2025-09-19
docs: warning comments for VM vs AKS machine code paths

comtalyst:
Fixes # 

**Description**

Within AKS machine API provision mode from [Machine API integration](https://github.com/Azure/karpenter-provider-azure/pull/1102), the existing VM provisioning code path will not be used, primarily due to different assumptions, and the new direction on the codebase (e.g., split provisioning/image handling concerns). 
This PR adds comments on the old code path for the engineers to be advised that their changes may not reflect in the new provision mode.
Refactor of the existing code path to adhere to the new philosophy is currently not being invested on, and not worth it in general, IMO.

*

- [ ] Yes, issue opened: # 
- [X] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/providers/imagefamily/customscriptsbootstrap/provisionclientbootstrap.go
+++ b/pkg/providers/imagefamily/customscriptsbootstrap/provisionclientbootstrap.go
@@ -94,6 +94,8 @@ func (p ProvisionClientBootstrap) GetCustomDataAndCSE(ctx context.Context) (stri
 }
 
 // nolint: gocyclo
+// ATTENTION!!!: changes here may NOT be effective on AKS machine nodes (ProvisionModeAKSMachineAPI); See aksmachineinstance.go/aksmachineinstancehelpers.go.
+// Refactoring for code unification is not being invested immediately.
 func (p *ProvisionClientBootstrap) ConstructProvisionValues(ctx context.Context) (*models.ProvisionValues, error) {
 	if p.IsWindows {
 		// TODO(Windows)
--- a/pkg/providers/imagefamily/resolver.go
+++ b/pkg/providers/imagefamily/resolver.go
@@ -56,6 +56,8 @@ type Resolver interface {
 var _ Resolver = &defaultResolver{}
 
 // defaultResolver is able to fill-in dynamic launch template parameters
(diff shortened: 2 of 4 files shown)


pkg/providers/imagefamily/customscriptsbootstrap/provisionclientbootstrap.go
```diff
@@ -94,6 +94,8 @@ func (p ProvisionClientBootstrap) GetCustomDataAndCSE(ctx context.Context) (stri
 }
 
 // nolint: gocyclo
+// ATTENTION!!!: changes here may NOT be effective on AKS machine nodes (ProvisionModeAKSMachineAPI); See aksmachineinstance.go/aksmachineinstancehelpers.go.
```
Bryce-Soghigian, line 97: DO you have a standard you are following for these comments? https://stackoverflow.com/questions/34372532/choosing-between-attention-note-remark-todo-and-warning-in-doxygen 


You also have suggestion, etc. 

I was wondering if you were following a specific standard or if this comment style is one you developed yourself. Attention is in some standards so thats what made me curious

comtalyst: No, I did not follow that standard, and it sounds like a good idea to try to do so as much as practical. Though, with some caveats/questions:
- What could be good use of these standards apart from general cleanliness, specifically to this repo?
  - Not sure if doc generation brings good value in this case though
- `Attention` will grab less attention than `ATTENTION!!!`, at least IMO
  - I think that's an okay trade-off. But that also means I hope us all pay attention to these kind of comments.
- `Suggestion` is something I developed myself: it is similar to, but justifiably not as strong as `TODO`
  - But perhaps not harmful to the standard if we have one extra? Can also have `Note` prefix, idk

tallaxes: FWIW, we (and Karpenter core) use "WARNING:" (or "Warning:") in a couple of places, and it is [not uncommon](https://github.com/search?q=lang%3AGo+%22%2F%2F+WARNING%3A%22&type=code) in other Go codebases; would be my vote.

tallaxes (approved):
LGTM, maybe a bit of overkill in that many places. A definitive statement (rather than "may not be effective") ~~may~~ would be better. Would also love to see a quick overall design doc that summarizes which parts are used in which bootstrap mechanism (maybe already exists somewhere?)

PR #1173, opened 2025-09-19
feat: exclude AKS machine VMs/NICs from respective ARG list queries

comtalyst:
Fixes # 

**Description**

With the new abstraction in [Machine API integration](https://github.com/Azure/karpenter-provider-azure/pull/1102), their VMs and NICs will be included in current ARG queries given they may have Karpenter NodePool tags like the ones from VM instances. That is not desirable, as it is AKS machine instance that will takes care of that, especially not by accessing the VM/NIC instances directly, which breaks the new machine abstraction per design.

This PR modifies the ARG query to exclude these instances from the queries, under the assumption that no valid instances have this tag as of today.
Note that there is no validation in CEL to prevent users from setting system tags (incl. NodePool) as of today. This shall be a follow-up.

* TODO

- [ ] Yes, issue opened: # 
- [X] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/providers/launchtemplate/tags.go
+++ b/pkg/providers/launchtemplate/tags.go
@@ -31,7 +31,8 @@ const (
 )
 
 var (
-	NodePoolTagKey = strings.ReplaceAll(karpv1.NodePoolLabelKey, "/", "_")
+	NodePoolTagKey                     = strings.ReplaceAll(karpv1.NodePoolLabelKey, "/", "_")
+	KarpenterAKSMachineNodeClaimTagKey = "karpenter.azure.com_aksmachine_nodeclaim"
 )
 
 // TODO: Would like to refactor this out of launchtemplate at some point
--- a/pkg/fake/azureresourcegraphapi.go
+++ b/pkg/fake/azureresourcegraphapi.go
@@ -85,7 +85,8 @@ func (c *AzureResourceGraphAPI) getResourceList(query string) []interface{} {
 	switch query {
 	case c.vmListQuery:
 		vmList := lo.Filter(c.loadVMObjects(), func(vm armcompute.VirtualMachine, _ int) bool {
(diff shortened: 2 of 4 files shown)


pkg/providers/launchtemplate/tags.go
```diff
@@ -27,7 +27,8 @@ import (
 )
 
 const (
-	KarpenterManagedTagKey = "karpenter.azure.com_cluster"
+	KarpenterManagedTagKey             = "karpenter.azure.com_cluster"
+	KarpenterAKSMachineNodeClaimTagKey = "karpenter.azure.com_aksmachine_nodeclaim"
```
tallaxes, line 31: Do you have a good justification for including "nodeclaim" as part of the tag key?

comtalyst: Yes. Full docs will be uploaded later.

# Retrieving NodeClaim name from instance

## Facts/concerns

- Current state:
  - NodeClaim name is VM name
  - On `Get()`/`List()`, NodeClaim name is populated into NodeClaim in-memory struct, field `Name`, however:
    - Core just ignore it upon return, and likely to stay that way, given core defines the name in the first place
    - Nothing is provider has really been using that... So it is technically useless now
- Machine API:
  - Per option C in "Node naming crisis", we need a new way of retrieving NodeClaim name

## Decisions:
  - **Like option B in "Creation timestamp existence", it will be stored as a resource tag**
  - Always less than 256 characters, so, good to go
  - However, it then inherits the same weakness: user can tamper with it
    - Despite that, this one is repairable by in-place update controller, as it reconciles based on actual NodeClaim object
    - Given no current usage, the gap is not significant, for now, and unlikely to be

comtalyst: Re: test: they are in a different PR, will be discussed separately

PR #1175, opened 2025-09-19
fix: incorrect GenerateNodeClaimName (from VM name) implementation

comtalyst:
Fixes # 

**Description**

Sounds like this little function has always been incorrect? Although, doesn't look like this (`NodeClaim.Name` populated in `instanceToNodeClaim(...)`) is being used anywhere, including the core.

*

- [ ] Yes, issue opened: # 
- [X] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/cloudprovider/cloudprovider.go
+++ b/pkg/cloudprovider/cloudprovider.go
@@ -506,7 +506,7 @@ func (c *CloudProvider) instanceToNodeClaim(ctx context.Context, vm *armcompute.
 }
 
 func GenerateNodeClaimName(vmName string) string {
-	return strings.TrimLeft("aks-", vmName)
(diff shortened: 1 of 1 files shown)


pkg/cloudprovider/cloudprovider.go
```diff
@@ -506,7 +506,7 @@ func (c *CloudProvider) instanceToNodeClaim(ctx context.Context, vm *armcompute.
 }
 
 func GenerateNodeClaimName(vmName string) string {
```
matthchr: Can you add a unit test too?

comtalyst:
> LGTM, though would much prefer to drop population of NodeClaim.Name altogether if it is indeed unused.

might be done separately

PR #1176, opened 2025-09-19
docs: add comments on instance garbage collection reasonings and suggestions

comtalyst:
Fixes # 

**Description**

Add some comments on instance garbage collection. See the changes for details.

*

- [ ] Yes, issue opened: # 
- [X] No

**Release Note**

```release-note

```


pkg/controllers/nodeclaim/garbagecollection/instance_garbagecollection.go
```diff
@@ -81,9 +81,18 @@ func (c *VirtualMachine) Reconcile(ctx context.Context) (reconcile.Result, error
 	})...)
 	errs := make([]error, len(retrieved))
 	workqueue.ParallelizeUntil(ctx, 100, len(managedRetrieved), func(i int) {
+		// managedRetrieved, although represented as NodeClaim, is actually actually populated by provider.List() rather than sourcing from the cluster.
```
matthchr: I think that Karpenter upstream does this a lot - for example the entire existence of `cloudProvider.List` and its implementation is coercing VMs into NodeClaims, rather than sourcing this data from the NodeClaim.

I think calling this out is good but I am not sure it makes sense to ask "should we change" since I think the whole `CloudProvider` interface is somewhat predicated on this being "the way it's done". It has confused me slightly a few times too.

pkg/controllers/nodeclaim/garbagecollection/instance_garbagecollection.go
```diff
@@ -81,9 +81,18 @@ func (c *VirtualMachine) Reconcile(ctx context.Context) (reconcile.Result, error
...
+		// In the implementation, CreationTimestamp in this context is also different: representing instance's (not NodeClaim's) creation time.
+		// Suggestion: this "borrowing" pattern and its inconsistency is not intuitive..., should reconsider this implementation?
 		if !resolvedProviderIDs.Has(managedRetrieved[i].Status.ProviderID) &&
+			// Garbage collect if the instance has been around for more than 5 minutes, yet still no matching (per ProviderID) NodeClaim.
+			// Note that the "match" occurs after cloudprovider.Create() returns and ProviderID is populated as a result.
+			// Although, the intention of garbage collection is to clear instances with missing/deleted NodeClaim.
+			// This 5m is more of a grace period for newly-created instances that have yet to populate NodeClaim after.
 			time.Since(managedRetrieved[i].CreationTimestamp.Time) > time.Minute*5 {
 			errs[i] = c.garbageCollect(ctx, managedRetrieved[i], nodeList)
+			// In the case that CreationTimestamp is irretrievable (epoch), grace period will effectively be disabled.
```
matthchr: Not sure I am following this bit.

pkg/controllers/nodeclaim/garbagecollection/instance_garbagecollection.go
comtalyst: I would take `Suggestion` comments as "ideas" than "pending actions" (that's for `TODO`) as it would need more research on its practicality. Thoughts?

But I added the note of the idea of "following the core". That's the helpful part of the research.

pkg/controllers/nodeclaim/garbagecollection/instance_garbagecollection.go
comtalyst: Made some changes. How is it now?

pkg/controllers/nodeclaim/garbagecollection/instance_garbagecollection.go
tallaxes:
```suggestion
		// managedRetrieved, although represented as NodeClaim, is actually populated by provider.List() rather than sourcing from the cluster.
```

tallaxes: Before we discuss the content of the comment, can I ask you to update the code following AWS provider version ([current](https://github.com/aws/karpenter-provider-aws/blob/da1a095948e2c95fdae671687d259e4fa993e2f2/pkg/controllers/nodeclaim/garbagecollection/controller.go#L55]), or [this PR](https://github.com/aws/karpenter-provider-aws/commit/7304f30ee4f00a3a7d1d1a84106660e140b63da6#diff-09fe69a347cbc26e9c97b75d9315667ae160f39a3283c3b02090c563b8f1daf4)), and then we can check if the comment is still needed - or possibly could be revised a bit. Variable (re-)names in that update - `cloudNodeClaims`/`clusterNodeClaims` - reflect the intent much better.

Note that that PR overall makes the binary operate only on managed resources (those associated with CP-declared NodeClass), which affects GC as well. My strong preference would be to add full support for this to our provider - it will be important in some contexts. But, if this is too much, this PR can adopt better naming just for clarity of logic and flow, skipping the filtering part.

comtalyst: Updated naming of variables. I think the filter for managed NodeClaims only should be done separately.

pkg/controllers/nodeclaim/garbagecollection/instance_garbagecollection.go
tallaxes: Not sure CreationTimestamp will ever be empty, but the updated comments makes it clear this is what you mean

PR #1179, opened 2025-09-19
test: fix incorrect expectation for managed cluster tag

comtalyst:
Fixes # 

**Description**

The tag before the change seems incorrect and could not be find in the codebase(?). This PR should fix to the intended one.

Although, this has no direct effect to validation correctness, as this function is being used to create placeholder instances only, for now.

*

- [ ] Yes, issue opened: # 
- [X] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/test/utils.go
+++ b/pkg/test/utils.go
@@ -36,7 +36,7 @@ func RandomName(prefix string) string {
 
 func ManagedTags(nodepoolName string) map[string]*string {
 	return map[string]*string{
-		"karpenter.sh_cluster":  lo.ToPtr("test-cluster"),
-		"karpenter.sh_nodepool": lo.ToPtr(nodepoolName),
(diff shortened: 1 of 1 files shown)


pkg/test/utils.go
```diff
@@ -36,7 +36,7 @@ func RandomName(prefix string) string {
 
 func ManagedTags(nodepoolName string) map[string]*string {
 	return map[string]*string{
-		"karpenter.sh_cluster":  lo.ToPtr("test-cluster"),
-		"karpenter.sh_nodepool": lo.ToPtr(nodepoolName),
+		"karpenter.azure.com_cluster": lo.ToPtr("test-cluster"),
```
tallaxes, line 39: Given opportunities for mistakes here, maybe using an existing constant would be even better?

comtalyst: Will be considered separately. An argument against it is if we want the test to be sanity check, thus not being swayed by code modifications.

PR #1180, opened 2025-09-19
refactor: image utility functions for AKS machine instance

comtalyst:
Fixes # 

**Description**

With [Machine API integration](https://github.com/Azure/karpenter-provider-azure/pull/1102), this utility will be useful cross-module, for instance provisioning and drift handling. It has no uses at the moment, but not changing existing code paths.

*

- [ ] Yes, issue opened: # 
- [X] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/utils/image.go
+++ b/pkg/utils/image.go
@@ -1,0 +1,56 @@
+/*
+Portions Copyright (c) Microsoft Corporation.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package utils
+
+import (
+	"fmt"
+	"regexp"
+	"strings"
+)
+
+// WARNING: not supporting CIG images yet.
+func GetAKSMachineNodeImageVersionFromImageID(imageID string) (string, error) {
+	if strings.HasPrefix(imageID, "/CommunityGalleries") {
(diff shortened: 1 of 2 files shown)


pkg/utils/image.go
```diff
@@ -0,0 +1,56 @@
...
+		return "", fmt.Errorf("CIG images are not supported yet for AKS machines, consider not using PROVISION_MODE=aksmachineapi: %s", imageID)
+	} else {
+		return GetAKSMachineNodeImageVersionFromSIGImageID(imageID)
+	}
+}
+
+// Convert from "/subscriptions/10945678-1234-1234-1234-123456789012/resourceGroups/AKS-Ubuntu/providers/Microsoft.Compute/galleries/AKSUbuntu/images/2204gen2containerd/versions/2022.10.03"
+// to "AKSUbuntu-2204gen2containerd-2022.10.03".
+func GetAKSMachineNodeImageVersionFromSIGImageID(imageID string) (string, error) {
+	matches := regexp.MustCompile(`(?i)/subscriptions/(\S+)/resourceGroups/(\S+)/providers/Microsoft.Compute/galleries/(\S+)/images/(\S+)/versions/(\S+)`).FindStringSubmatch(imageID)
```
tallaxes: Best extract precompiled regex into package-level var; no need to compute every time

pkg/utils/image.go
```diff
@@ -0,0 +1,56 @@
...
+	if matches == nil {
+		return "", fmt.Errorf("incorrect SIG image ID id=%s", imageID)
+	}
+
+	// SubscriptionID := matches[1]
+	// ResourceGroup := matches[2]
+	Gallery := matches[3]
+	Definition := matches[4]
+	Version := matches[5]
```
tallaxes: Maybe lowercase variables?

tallaxes (changes requested):

pkg/utils/image.go
comtalyst: Changed. Note that this util function is taken from a different codebase. But don't think it needs to be the same on this aspect, and not sure why they did this...

tallaxes (approved):

PR #1183, opened 2025-09-24
 feat: AKS machine API interface and clients

comtalyst:
Fixes # 

**Description**

This PR contains API interfaces and clients for [Machine API integration](https://github.com/Azure/karpenter-provider-azure/pull/1102).
The added option/environment variable will allow the client creation failure to be tolerated, and will have further uses in the upcoming changes. Specifically, to determine whether Karpenter should continue to manage AKS machines in other provision modes.

*

- [ ] Yes, issue opened: # 
- [X] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/providers/instance/offerings/errordetailhandlers.go
+++ b/pkg/providers/instance/offerings/errordetailhandlers.go
@@ -1,0 +1,105 @@
+/*
+Portions Copyright (c) Microsoft Corporation.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package offerings
+
+import (
+	"context"
+
+	sdkerrors "github.com/Azure/azure-sdk-for-go-extensions/pkg/errors"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/containerservice/armcontainerservice/v8"
+	"github.com/Azure/karpenter-provider-azure/pkg/cache"
+	"github.com/Azure/skewer"
+	corecloudprovider "sigs.k8s.io/karpenter/pkg/cloudprovider"
+)
+
+type errorDetailHandlerEntry struct {
+	match  func(errorDetail armcontainerservice.ErrorDetail) bool
+	handle errorHandle
+}
+
+type ErrorDetailHandler struct {
+	UnavailableOfferings *cache.UnavailableOfferings
+	HandlerEntries       []errorDetailHandlerEntry
+}
+
+// Comparing to ResponseErrorHandler, this is handling same errors, but for a different error data model.
+// HandlerEntries should generally be kept in sync.
+//
+// This exists because AKS machine data model returns this armcontainerservice.ErrorDetail (AKS-specific) instead of azcore.ResponseError (SDK-wide). They have no common interface.
+// VM instance is still using ResponseErrorHandler.
+// Ideally, if AKS machine data model returns azcore.ResponseError, we don't need this split at all.
+func NewErrorDetailHandler(unavailableOfferings *cache.UnavailableOfferings) *ErrorDetailHandler {
+	return &ErrorDetailHandler{
+		UnavailableOfferings: unavailableOfferings,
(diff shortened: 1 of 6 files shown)


pkg/operator/options/options.go
```diff
@@ -92,6 +92,7 @@ type Options struct {
 	AdditionalTags             map[string]string `json:"additionalTags,omitempty"`
 	EnableAzureSDKLogging      bool              `json:"enableAzureSDKLogging,omitempty"` // Controls whether Azure SDK middleware logging is enabled
 	DiskEncryptionSetID        string            `json:"diskEncryptionSetId,omitempty"`
+	ManageExistingAKSMachines  bool              `json:"manageExistingAKSMachines,omitempty"` // If set to true, existing AKS machines created with PROVISION_MODE=aksmachineapi will be managed even with other provision modes. This option does not have any effect if PROVISION_MODE=aksmachineapi, as it will behave as if this option is set to true.
```
tallaxes: Consider moving comment above the field? Long lines are hard to read ...

pkg/test/options.go
```diff
@@ -88,5 +89,6 @@ func Options(overrides ...OptionsFields) *azoptions.Options {
 		AdditionalTags:                 options.AdditionalTags,
 		DiskEncryptionSetID:            lo.FromPtrOr(options.DiskEncryptionSetID, ""),
 		DNSServiceIP:                   lo.FromPtrOr(options.ClusterDNSServiceIP, ""),
+		ManageExistingAKSMachines:      lo.FromPtrOr(options.ManageExistingAKSMachines, true),
```
tallaxes: Different default than production code? Because this mode is suitable for test?

pkg/providers/instance/azure_client.go
```diff
@@ -199,10 +219,34 @@ func NewAZClient(ctx context.Context, cfg *auth.Config, env *auth.Environment, c
 			return nil, err
 		}
 	}
+	if o.ManageExistingAKSMachines {
```
tallaxes, line 228: I don't (yet) fully understand the purpose of this flag, and - probably as a consequence - the intent behind "dry" clients. Will need to discuss first to be able to review intelligently ...

pkg/test/options.go
comtalyst: Looks like leftover from dev process. Removed.

pkg/providers/instance/azure_client.go
comtalyst: Discussed offline on the intent.

For the suspected issue of this not including the case where `ProvisionMode` is `aksmachineapi`, this is intentional at that is being added in a different PR ([this part](https://github.com/Azure/karpenter-provider-azure/pull/1197/files#r2395563283)). This PR is independent from the new provision mode. Thus, the scope of this code is for existing AKS machines only, regardless of how they come to existence.

For how to implement the design of "blocking all contact with AKS machine API", as discussed offline on the intent, the currently planned approach is to use both no-client and blocking if statements in [AKS machine instance provider](https://github.com/Azure/karpenter-provider-azure/pull/1197/files#diff-039f8834b9d330dc4127ef63753481daeb74c3a2a8c04de209d9b01ab9652b53). Some alternatives:
- (A) Just use no-clients. No need for if statements.
  - Should work correctly
  - Harder to unit-test in the current infrastructure, as we substitute fake clients, and not testing the client selection in azure_client.go
    - Although, the same problem exists in the current approach. Just with lower risk from two approaches combined.
- (B) Just use blocking if statements in [AKS machine instance provider](https://github.com/Azure/karpenter-provider-azure/pull/1197/files#diff-039f8834b9d330dc4127ef63753481daeb74c3a2a8c04de209d9b01ab9652b53). No need for no-clients.
  - Should work correctly
  - If we still skip creating client, given the current struct of clients, those fields will be left unintuitively(?) `nil` (in contrast to A/current approach, where the client exists and ready to be called)
    - Need to rely on thorough testing for other parts to accept this (i.e., verify that it won't attempt to use the client, or at least handle nil properly).
      - Though this is likely cause no real issue initially as there are no other users, yet.
  - If we instead always create client, then there is more risk of the clients being used elsewhere (in which our design doesn't want that), or even failure to create client when it is not planned to be used
  - Violate some designed(?) abstractions: AKS machine instance provider may don't want to know what `ProvisionMode` it is on.
- (C) Don't create AKS machine instance provider. Use if statements in places that consume/expect instance provider.
  - Less risk/cycles spent on creating what we wouldn't use (AKS machine instance provider)
  - Same problem with (B), but worse: it is already in multiple places (e.g., `cloudprovider`, `garbagecollection`, `inplaceupdate`) and more likely to grow.
    - _May_ be less bad with Matthew's idea of a common abstraction for instance providers (that we discussed), though that's a larger effort, with a known concern being differences in the interface expectations itself.

Can discuss further here or in the real Machine API PR (perhaps this PR split may not be _that_ intuitive...)

PR #1185, opened 2025-09-24
refactor: designate VM instance-specific acceptance tests

comtalyst:
Fixes # 

**Description**

Upcoming [Machine API integration](https://github.com/Azure/karpenter-provider-azure/pull/1102) will introduce a kind of instance with different assumptions than VM instances, which is the only supported type as of today. 
Most of the existing acceptance tests are not immediately compatible with this new assumption. 
This PR will designate those tests as tests specific to VM instances, so that Machine API integration changes can add new ones with clear distinction.

A considered alternative was to refactor the tests to be more generic. Although, given that each test has different degree of incompatibility, it is estimated to not be worth it.

*

- [ ] Yes, issue opened: # 
- [X] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/controllers/nodeclaim/garbagecollection/instance_garbagecollection.go
+++ b/pkg/controllers/nodeclaim/garbagecollection/instance_garbagecollection.go
@@ -60,30 +60,36 @@ func (c *VirtualMachine) Reconcile(ctx context.Context) (reconcile.Result, error
 	// We LIST VMs on the CloudProvider BEFORE we grab NodeClaims/Nodes on the cluster so that we make sure that, if
 	// LISTing instances takes a long time, our information is more updated by the time we get to nodeclaim and Node LIST
 	// This works since our CloudProvider instances are deleted based on whether the NodeClaim exists or not, not vice-versa
-	retrieved, err := c.cloudProvider.List(ctx)
+	cloudNodeClaims, err := c.cloudProvider.List(ctx)
 	if err != nil {
 		return reconcile.Result{}, fmt.Errorf("listing cloudprovider VMs, %w", err)
 	}
 
-	managedRetrieved := lo.Filter(retrieved, func(nc *karpv1.NodeClaim, _ int) bool {
+	cloudNodeClaims = lo.Filter(cloudNodeClaims, func(nc *karpv1.NodeClaim, _ int) bool {
 		return nc.DeletionTimestamp.IsZero()
 	})
-	nodeClaimList := &karpv1.NodeClaimList{}
-	if err = c.kubeClient.List(ctx, nodeClaimList); err != nil {
+	clusterNodeClaims := &karpv1.NodeClaimList{}
+	if err = c.kubeClient.List(ctx, clusterNodeClaims); err != nil {
 		return reconcile.Result{}, err
 	}
 	nodeList := &v1.NodeList{}
 	if err := c.kubeClient.List(ctx, nodeList); err != nil {
 		return reconcile.Result{}, err
 	}
-	resolvedProviderIDs := sets.New[string](lo.FilterMap(nodeClaimList.Items, func(n karpv1.NodeClaim, _ int) (string, bool) {
+	clusterProviderIDs := sets.New[string](lo.FilterMap(clusterNodeClaims.Items, func(n karpv1.NodeClaim, _ int) (string, bool) {
 		return n.Status.ProviderID, n.Status.ProviderID != ""
 	})...)
-	errs := make([]error, len(retrieved))
-	workqueue.ParallelizeUntil(ctx, 100, len(managedRetrieved), func(i int) {
-		if !resolvedProviderIDs.Has(managedRetrieved[i].Status.ProviderID) &&
-			time.Since(managedRetrieved[i].CreationTimestamp.Time) > time.Minute*5 {
-			errs[i] = c.garbageCollect(ctx, managedRetrieved[i], nodeList)
+	errs := make([]error, len(cloudNodeClaims))
+	workqueue.ParallelizeUntil(ctx, 100, len(cloudNodeClaims), func(i int) {
+		// Garbage collect if the cloud instance has been around for more than 5 minutes, yet still no matching (per ProviderID) cluster NodeClaim.
+		// Note that the "match" occurs after cloudprovider.Create() returns and cluster NodeClaim ProviderID is populated as a result.
+		// Although, the intention of garbage collection is to clear instances with missing/deleted NodeClaim.
+		// This 5m is more of a grace period for newly-created instances that have yet to populate NodeClaim after.
+		if !clusterProviderIDs.Has(cloudNodeClaims[i].Status.ProviderID) &&
+			time.Since(cloudNodeClaims[i].CreationTimestamp.Time) > time.Minute*5 {
+			errs[i] = c.garbageCollect(ctx, cloudNodeClaims[i], nodeList)
+			// In the case that CreationTimestamp is irretrievable (technically, when CreationTimestamp = 0 = epoch), grace period will effectively be disabled.
+			// Which could be dangerous if the instance is legitimately awaiting NodeClaim population.
 		}
 	})
 	if err = multierr.Combine(errs...); err != nil {


pkg/controllers/nodeclaim/garbagecollection/suite_test.go
```diff
@@ -83,6 +85,7 @@ var _ = BeforeSuite(func() {
 	azureEnv = test.NewEnvironment(ctx, env)
 	cloudProvider = cloudprovider.New(azureEnv.InstanceTypesProvider, azureEnv.VMInstanceProvider, events.NewRecorder(&record.FakeRecorder{}), env.Client, azureEnv.ImageProvider)
 	virtualMachineGCController = garbagecollection.NewVirtualMachine(env.Client, cloudProvider)
+	inPlaceUpdateController = inplaceupdate.NewController(env.Client, azureEnv.VMInstanceProvider)
```
tallaxes, line 88: Unused?

pkg/controllers/nodeclaim/garbagecollection/suite_test.go
```diff
@@ -129,274 +132,285 @@ var _ = Describe("VirtualMachine Garbage Collection", func() {
...
-		})
-		It("should delete many instances if they all don't have NodeClaim owners", func() {
-			// Generate 100 instances that have different vmIDs
-			var ids []string
-			var vmName string
-			for i := 0; i < 100; i++ {
-				pod := coretest.UnschedulablePod()
+	// Attention: tests under "VM instances" are not applicable to AKS machine instances, created with ProvisionModeAKSMachineAPI.
+	// Due to different assumptions, not all tests can be shared. Add tests for AKS machine instances in a different Context/file.
+	// If VM instances are no longer supported, their code/tests will be replaced with AKS Machine instances.
```
tallaxes, line 137: It would be helpful to briefly describe here what it is about GC that is not applicable to ProvisionModeAKSMachineAPI cases

pkg/controllers/nodeclaim/inplaceupdate/suite_test.go
```diff
@@ -509,323 +512,363 @@ var _ = Describe("Unit tests", func() {
...
-			ID:   lo.ToPtr(fake.MakeVMExtensionID(azureEnv.AzureResourceGraphAPI.ResourceGroup, vmName, "cse-agent-karpenter")),
-			Name: lo.ToPtr("cse-agent-karpenter"),
-			Tags: map[string]*string{
-				"karpenter.azure.com_cluster": lo.ToPtr("test-cluster"),
-			},
-		}
+	// Attention: tests under "VM instances" are not applicable to AKS machine instances, created with ProvisionModeAKSMachineAPI.
+	// Due to different assumptions, not all tests can be shared. Add tests for AKS machine instances in a different Context/file.
+	// If VM instances are no longer supported, their code/tests will be replaced with AKS Machine instances.
+	Context("VM instances", func() {
```
tallaxes, line 526: It would be helpful to briefly describe what it is about in-place update behavior, basic tests, tag tests etc. that is not applicable to machine instances

pkg/controllers/nodeclaim/inplaceupdate/suite_test.go
```diff
@@ -509,323 +512,363 @@ var _ = Describe("Unit tests", func() {
...
+				result, err := inPlaceUpdateController.Reconcile(ctx, nodeClaim)
 
-			updatedVM, err := azureEnv.VMInstanceProvider.Get(ctx, vmName)
-			Expect(err).ToNot(HaveOccurred())
+				Expect(err).To(HaveOccurred())
+				Expect(result).To(Equal(reconcile.Result{}))
+			})
 
-			Expect(updatedVM).To(Equal(vm)) // No change expected
+			It("should handle VM not found gracefully", func() {
```
tallaxes, line 636: These are new, right?

pkg/controllers/nodeclaim/inplaceupdate/suite_test.go
```diff
@@ -509,323 +512,363 @@ var _ = Describe("Unit tests", func() {
...
+					})
+				// Expect the identities to remain unchanged
+				Expect(updatedVM.Identity).To(BeNil())
+
+				nodeClaim = ExpectExists(ctx, env.Client, nodeClaim)
+				Expect(nodeClaim.Annotations).To(HaveKey(v1beta1.AnnotationInPlaceUpdateHash))
+				Expect(nodeClaim.Annotations[v1beta1.AnnotationInPlaceUpdateHash]).ToNot(BeEmpty())
+			})
+
+			It("should propagate update to tags from NodeClass with standard tags", func() {
```
tallaxes: What was the problem with DescribeTable here?

pkg/cloudprovider/suite_test.go
```diff
@@ -126,269 +111,112 @@ var _ = BeforeEach(func() {
...
+	// If ProvisionMode = AKSScriptless is no longer supported, their code/tests will be replaced with ProvisionMode = AKSMachineAPI.
+	Context("ProvisionMode = AKSScriptless", func() {
 		BeforeEach(func() {
-			instanceType := "Standard_D2_v2"
-			ExpectApplied(ctx, env.Client, nodePool, nodeClass)
-			pod = coretest.UnschedulablePod(coretest.PodOptions{
-				NodeSelector: map[string]string{v1.LabelInstanceTypeStable: instanceType},
+			testOptions = test.Options(test.OptionsFields{
+				ProvisionMode: lo.ToPtr(consts.ProvisionModeAKSScriptless),
 			})
```
tallaxes, line 171: Ok, so this is the proposed mechanism for executing provision mode-specific tests in their respective provision mode? I don't see this applied in other tests ... Also, this set of tests was presumably working for bootstrapping client as well?

pkg/cloudprovider/suite_drift_test.go
```diff
@@ -0,0 +1,299 @@
...
+)
+
+var _ = Describe("CloudProvider", func() {
+	// Attention: tests under "ProvisionMode = AKSScriptless" are not applicable to ProvisionMode = AKSMachineAPI option.
+	// Due to different assumptions, not all tests can be shared. Add tests for AKS machine instances in a different Context/file.
+	// If ProvisionMode = AKSScriptless is no longer supported, their code/tests will be replaced with ProvisionMode = AKSMachineAPI.
+	Context("ProvisionMode = AKSScriptless", func() {
+		BeforeEach(func() {
+			testOptions = test.Options(test.OptionsFields{
+				ProvisionMode: lo.ToPtr(consts.ProvisionModeAKSScriptless),
```
tallaxes, line 51: These were not scriptless specific ... Are they considered scriptless specific now because boostrapping client mode is going away?

comtalyst:
(discussed some offline)

> Would like to understand a little better what is it that makes certain kinds of tests not applicable to AKSMachineAPI mode, and whether the differences are significant enough to warrant full split. Some of the invariants captured in these tests are fundamental enough that, if they break in a different mode, it is a reason for concern. (Though maybe this is just about different expectations re internal API calls?) 

Some of the big concerns that prevent easy unification, apart from API interface/fake expectations, are:
- The expectation/responsibility of managing NIC-related configurations in general
- The expectation/responsibility of managing CustomData-level configurations (e.g., networking labels)
  - Without this in machine mode, most tests will lose ties to `instancetype` module
- Running the same set of test code with different options: need some organization here

A "generalized" alternative would be to have if statements in each test: some just to have different validation implementation, and some to skip inapplicable tests (e.g., networking labels). This, together with potential recategorization (especially instancetype module) will be reconsidered in the next iteration. It would still be beneficial, just don't think it is the most prioritized given the effort to resolve the above.

This will also be up to the project's direction of whether we want to maintain multiple provision modes.

Expected behavioral differences are also noted in the design docs. Some are also more difficult to fit into the generalized/unified model. E.g., the API call phase that quota error shall return.

>Also wondering about exactly which mechanism is proposed for running different sets of tests in different modes, though that should not be too complicated.

It would be to change the options in `BeforeEach` for a different mode.

pkg/controllers/nodeclaim/garbagecollection/suite_test.go
comtalyst: In addition to the design docs, that will be done in https://github.com/Azure/karpenter-provider-azure/pull/1197, if applicable. This PR should be independent from machine API integration's implementation.

pkg/cloudprovider/suite_drift_test.go
comtalyst: We do not have drift tests for bootstrappingclient mode. So, technically, this is considered mode-specific. 
Doesn't mean it couldn't be changed to be shared though, but that is out of scope.

pkg/cloudprovider/suite_test.go
comtalyst: > Also, this set of tests was presumably working for bootstrapping client as well?

Most E2E acceptance tests (note that not all are really "E2E") we have are only testing Scriptless mode, except a small, specific set in `instancetype`. We have never support shared tests across modes.
Whether we could is a different question, but one concern is going to be the fact we should not test CSE for bootstrappingclient mode. Machine mode has the similar concern and could be resolved at the same time.

> I don't see this applied in other tests

It should be applied in other tests...?

pkg/controllers/nodeclaim/inplaceupdate/suite_test.go
comtalyst: Yes, I added some new tests here I believe. Are there concerns?

pkg/controllers/nodeclaim/inplaceupdate/suite_test.go
comtalyst: In addition to the design docs, that will be done in https://github.com/Azure/karpenter-provider-azure/pull/1197, if applicable. This PR should be independent from machine API integration's implementation.

pkg/controllers/nodeclaim/inplaceupdate/suite_test.go
comtalyst: Not sure what caused this change. But not seeing a problem restoring DescribeTable now (done).

pkg/controllers/nodeclaim/garbagecollection/suite_test.go
comtalyst: This would be used in machine tests (different PR). Technically it could be removed here and added there too, though I see it as a more general piece.

pkg/cloudprovider/suite_drift_test.go
tallaxes: Ack, resolving

pkg/cloudprovider/suite_test.go
tallaxes: Ack, resolving

pkg/controllers/nodeclaim/inplaceupdate/suite_test.go
tallaxes: No concerns, just wanted to make sure I understand the changes

PR #1186, opened 2025-09-24
refactor: distinguish instance type in inplaceupdate

comtalyst:
Fixes # 

**Description**

`inplaceupdate` module will be used for the upcoming [Machine API integration](https://github.com/Azure/karpenter-provider-azure/pull/1102) as well. Although, the module's implementation currently does not expect to handle more than VM instances. This refactor will designate VM instance code path as a separate path from upcoming AKS machine instances due to different assumptions. See some relevant discussions in the PoC PR.

*

- [ ] Yes, issue opened: # 
- [X] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/controllers/nodeclaim/inplaceupdate/utils.go
+++ b/pkg/controllers/nodeclaim/inplaceupdate/utils.go
@@ -36,8 +36,9 @@ type vmInPlaceUpdateFields struct {
 	Tags       map[string]string `json:"tags,omitempty"`
 }
 
-func (i *vmInPlaceUpdateFields) CalculateHash() (string, error) {
-	encoded, err := json.Marshal(i)
+// CalculateHash computes a hash for any JSON-marshalable struct
+func CalculateHash(data interface{}) (string, error) {
+	encoded, err := json.Marshal(data)
 	if err != nil {
 		return "", err
 	}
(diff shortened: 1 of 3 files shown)


pkg/controllers/nodeclaim/inplaceupdate/utils.go
```diff
@@ -36,8 +36,9 @@ type vmInPlaceUpdateFields struct {
 	Tags       map[string]string `json:"tags,omitempty"`
 }
 
-func (i *vmInPlaceUpdateFields) CalculateHash() (string, error) {
-	encoded, err := json.Marshal(i)
+// CalculateHash computes a hash for any JSON-marshalable struct
+func CalculateHash(data interface{}) (string, error) {
```
matthchr: Is there going to be a different struct for machine (since a different set of fields are in-place updatable?)
That makes sense.

How is `HashFromNodeClaim` going to work then though - won't it need to know which type to use? Is it going to grow some additional checks internally, or a different parameter, to determine that?

matthchr: style
```suggestion
func CalculateHash(data any) (string, error) {
```

comtalyst: There will be [different struct](https://github.com/Azure/karpenter-provider-azure/pull/1197/commits) for machine, but I figured this function (formerly method) can be shared.

comtalyst: For `HashFromNodeClaim` , there will just be an if statement based on NodeClaim. See https://github.com/Azure/karpenter-provider-azure/blob/ec2a6aa76d16209788afc1547f8faf5fa0b03fb2/pkg/controllers/nodeclaim/inplaceupdate/utils.go

PR #1187, opened 2025-09-24
refactor: rename zone util functions to be more accurate

comtalyst:
Fixes # 

**Description**

Current naming of these functions could be confusing in many use cases. See changes for specifics.

*

- [ ] Yes, issue opened: # 
- [X] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/providers/launchtemplate/launchtemplate.go
+++ b/pkg/providers/launchtemplate/launchtemplate.go
@@ -192,14 +192,15 @@ func (p *Provider) createLaunchTemplate(ctx context.Context, params *parameters.
 		StorageProfileSizeGB:      params.StorageProfileSizeGB,
 	}
 
-	if p.provisionMode == consts.ProvisionModeBootstrappingClient {
+	switch p.provisionMode {
+	case consts.ProvisionModeBootstrappingClient:
 		customData, cse, err := params.CustomScriptsNodeBootstrapping.GetCustomDataAndCSE(ctx)
 		if err != nil {
 			return nil, err
 		}
 		template.CustomScriptsCustomData = customData
 		template.CustomScriptsCSE = cse
-	} else {
+	case consts.ProvisionModeAKSScriptless:
 		// render user data
 		userData, err := params.ScriptlessCustomData.Script()
 		if err != nil {


pkg/utils/zone.go
```diff
@@ -23,25 +23,25 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/compute/armcompute/v7"
 )
 
-// MakeZone returns the zone value in format of <region>-<zone-id>.
-func MakeZone(location string, zoneID string) string {
+// GetAKSLabelZoneFromARMZone returns the zone value in format of <region>-<zone-id>.
+func GetAKSLabelZoneFromARMZone(location string, zoneID string) string {
```
matthchr: I think you went from maybe too generic to too specific. Plus `GetXyz` IMO is wrong since we are making something here (the `MakeXyz` name before was better in that regard).

Maybe `MakeFullyQualifiedZone`? It's definitely for more than just "AKS Label", it could be used for other things in the future - we should say what it is not what it's for (if possible). I think it is possible in this case.

pkg/utils/zone.go
```diff
@@ -23,25 +23,25 @@ import (
...
 	if zoneID == "" {
 		return ""
 	}
 	return fmt.Sprintf("%s-%s", strings.ToLower(location), zoneID)
 }
 
-// VM Zones field expects just the zone number, without region
-func MakeVMZone(zone string) []*string {
+// GetARMZonesFromAKSLabelZone returns the zone ID from <region>-<zone-id>.
+func GetARMZonesFromAKSLabelZone(zone string) []*string {
```
matthchr: Same I think `MakeXyz` is better here because we are sorta still making something, though I could be convinced Get is better on this one.

`MakeShortZoneFromFullZone`? or if you really prefer, `GetShortZoneFromFullZone`

pkg/utils/zone.go
matthchr: Or maybe `MakeFullZone` if you wanna exactly match below suggested name for `GetARMZonesFromAKSLabelZone`

pkg/utils/zone.go
```diff
@@ -23,25 +23,25 @@ import (
...
 		return []*string{}
 	}
 	zoneNum := zone[len(zone)-1:]
 	return []*string{&zoneNum}
 }
 
-// GetZone returns the zone for the given virtual machine, or an empty string if there is no zone specified
-func GetZone(vm *armcompute.VirtualMachine) (string, error) {
+// GetAKSLabelZoneFromVM returns the zone for the given virtual machine, or an empty string if there is no zone specified
+func GetAKSLabelZoneFromVM(vm *armcompute.VirtualMachine) (string, error) {
```
matthchr:
```suggestion
func GetFullZoneFromVM(vm *armcompute.VirtualMachine) (string, error) {
```

pkg/utils/zone.go
comtalyst: > MakeXyz name before was better in that regard

On the dimension of "action", I was hoping to be more generic, in fact. E.g., the idea of "Get" and "Make" shouldn't be exposed to the name in the first place, and I would hope to pick one and stick to it. Thoughts?

> It's definitely for more than just "AKS Label", it could be used for other things in the future - we should say what it is not what it's for (if possible). I think it is possible in this case

Although, on the dimension of "format', I think we should be specific here to prevent confusion. 
Labelling it to cover future use cases is a good point. Though, for this case specifically, I don't see if it is likely to have more use cases. Do you have significant example? 
Otherwise, labeling/categorizing it now would be unnecessary abstraction, leading to more cognitive load. Not doing that accurately can also lead to the case (in the future) where we thought it is re-usable, but not really.
Unless this "full zone" format is more of a universally-known format. But, is it?

comtalyst: (discussed offline) Sticking to this plan, but use `Make` instead of `Get` to align with SDK.

pkg/utils/zone.go
comtalyst: (see discussion in another comment above)

pkg/utils/zone.go
comtalyst: (see discussion in another comment above)

PR #1196, opened 2025-09-30
feat: armcontainerservice.ErrorDetail handlers in offerings

comtalyst:
Fixes # 

**Description**

`azcore.ResponseError` is the error type that is returned in VM instance provisioning. In upcoming AKS machine instance provisioning(https://github.com/Azure/karpenter-provider-azure/pull/1102), that error type will be `armcontainerservice.ErrorDetail`, thus the handling logic should be added in parallel.

There are changes on azure-sdk-for-go-extensions in [the PR there](https://github.com/Azure/azure-sdk-for-go-extensions/pull/316).

*

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


go.mod
```diff
@@ -6,7 +6,7 @@ require (
 	github.com/Azure/aks-middleware v0.0.41
 	github.com/Azure/azure-kusto-go v0.16.1
 	github.com/Azure/azure-sdk-for-go v68.0.0+incompatible
-	github.com/Azure/azure-sdk-for-go-extensions v0.2.0
+	github.com/Azure/azure-sdk-for-go-extensions v0.2.1-0.20250924093303-adf3c22e7580
```
comtalyst: Pending merge of https://github.com/Azure/azure-sdk-for-go-extensions/pull/316 to formalize this

pkg/providers/instance/offerings/errordetailhandlers.go
```diff
@@ -0,0 +1,99 @@
...
+	sdkerrors "github.com/Azure/azure-sdk-for-go-extensions/pkg/errors"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/containerservice/armcontainerservice/v7"
+	"github.com/Azure/karpenter-provider-azure/pkg/cache"
+	"github.com/Azure/skewer"
+	corecloudprovider "sigs.k8s.io/karpenter/pkg/cloudprovider"
+)
+
+type cloudErrorHandlerEntry struct {
+	match  func(cloudError armcontainerservice.CloudErrorBody) bool
+	handle errorHandle
```
matthchr, line 31: minor, IMO either `errorHandler` or `errorHandlerFunc` is better (see my comment on your other PR I think this is the same code just moved?)

pkg/providers/instance/offerings/clouderrorhandlers_test.go
```diff
@@ -0,0 +1,272 @@
...
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/containerservice/armcontainerservice/v7"
+	"github.com/Azure/karpenter-provider-azure/pkg/cache"
+	"github.com/Azure/skewer"
+	"github.com/stretchr/testify/assert"
+	karpv1 "sigs.k8s.io/karpenter/pkg/apis/v1"
+	corecloudprovider "sigs.k8s.io/karpenter/pkg/cloudprovider"
+)
+
+type cloudErrorTestCaseBuilder struct {
```
matthchr: Are these tests net-new or were they moved from somewhere?  I think moved from somewhere, right?

not seeing the deletes though.

pkg/providers/instance/offerings/errordetailhandlers.go
comtalyst: This is in parallel to the previous PR, let's refer to the discussion there

pkg/providers/instance/offerings/clouderrorhandlers_test.go
comtalyst: Also based on the one in the previous PR: https://github.com/Azure/karpenter-provider-azure/pull/1196/files.

comtalyst: It is not "move", but rather a new creation, just from the similar pattern.
This one handle `cloudErrorBody`. The previous one handle `azcore.ResponseError`.

pkg/providers/instance/offerings/clouderrorhandlers.go
```diff
@@ -0,0 +1,99 @@
...
+	"context"
+
+	sdkerrors "github.com/Azure/azure-sdk-for-go-extensions/pkg/errors"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/containerservice/armcontainerservice/v7"
+	"github.com/Azure/karpenter-provider-azure/pkg/cache"
+	"github.com/Azure/skewer"
+	corecloudprovider "sigs.k8s.io/karpenter/pkg/cloudprovider"
+)
+
+type cloudErrorHandlerEntry struct {
```
matthchr: Can you leave a comment somewhere in here explaining the difference between this and pkg/providers/instance/offerings/resperrorhandlers.go? as well as possibly a note in both of them that they should be kept in-sync (or, some tests that ensure that they are kept in-sync?)

Ideally the comment would explain why we need special error handling for `cloudError armcontainerservice.CloudErrorBody` which is different than the error handling that we have for `azcore.ResponseError`

Once you do this, I'm OK to merge this to unblock things (even though I don't love the split in error handling or the way we're doing error management for machine API today -- I think the error should be coming from AKS via `azcore.ResponseError` the same as it does for CRP)

comtalyst: There is also https://github.com/Azure/azure-sdk-for-go-extensions/pull/316 that this PR relies on. 
Not sure if there is a better way without further redesign (e.g., expose generic error handler instead of type-specific, as there is no common interface between these two, not worth it?).

matthchr (approved):

comtalyst:
Update: renames: https://github.com/Azure/azure-sdk-for-go-extensions/pull/316#issuecomment-3609699833

tallaxes (approved):

PR #1208, opened 2025-10-08
[WIP] Idea for metrics?

MaximilianoUribe:
Fixes # 

**Description**
First draft for adding a process to track metrics, generated a readme with ai to summarize the process.
(Also generated buckets, comments and steps. Subject to change.)

*

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/metrics/README.md
+++ b/pkg/metrics/README.md
@@ -1,0 +1,228 @@
+# Karpenter Azure Metrics Infrastructure
+
+This package provides a reusable, extensible metrics infrastructure for the Karpenter Azure provider. The design follows patterns from [cloud-provider-azure](https://github.com/kubernetes-sigs/cloud-provider-azure) to ensure consistency and ease of use.
+
+## Overview
+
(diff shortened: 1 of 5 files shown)


MaximilianoUribe:

MaximilianoUribe:
maybe change the name constant to karpenter-azure or something, to differentiate core from provider metrics?

MaximilianoUribe:
there is also the option to use library: https://pkg.go.dev/k8s.io/component-base/metrics
adding levels of stability to the metrics, that as previously mentioned would be good

MaximilianoUribe:
> there is also the option to use library: https://pkg.go.dev/k8s.io/component-base/metrics adding levels of stability to the metrics, that as previously mentioned would be good

this would introduce a breaking change with current metric ImageSelectionErrorCount, if we want to change to this package

rakechill:
For metrics namespace, I think we'll want to have something like `karpenter_azure` to prepend any metric

PR #1200, opened 2025-10-03
test(e2e+machine): Base structure for machine mode testing within E2E for oss

charliedmcb:
Fixes # 

**Description**
Adding the base structure for E2E testing using machine mode for oss. 
- Framework changs from this PR:
https://github.com/Azure/karpenter-provider-azure/pull/1204

Creates a BYO nodepool, and adapts the mode setting from here:
- https://github.com/Azure/karpenter-provider-azure/pull/1199/files
    - Does still need to handle the case for old expect statements not working for machine mode

**How was this change tested?**

E2E runs

scriptless:
- https://github.com/Azure/karpenter-provider-azure/actions/runs/18389347689
- https://github.com/Azure/karpenter-provider-azure/actions/runs/18397924767

aksmachineapi:
- https://github.com/Azure/karpenter-provider-azure/actions/runs/18389364777
- https://github.com/Azure/karpenter-provider-azure/actions/runs/18397941620

**Does this change impact docs?**
- [ ] Yes, PR includes docs updates
- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
NONE
```

PR #1212, opened 2025-10-09
test(e2e+machine): add in machine tests

charliedmcb:
Fixes # 

**Description**
Built off this underlaying framework:
https://github.com/Azure/karpenter-provider-azure/pull/1200

These tests cover:
- Assignment of post-provisioning networking labels that were previously handled by Karpenter
- Karpenter handling of DriftAction + kubelet causing DriftAction
- PUT machine during PUT MC

Tested with this PR:
- https://github.com/Azure/karpenter-provider-azure/pull/1204
NOTE: didn't get fully successful runs, and do expect the kubelet test to fail

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
NONE
```

### DIFF
--- a/.github/workflows/e2e-matrix.yaml
+++ b/.github/workflows/e2e-matrix.yaml
@@ -51,7 +51,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        suite: [ACR, BYOK, Chaos, Consolidation, Drift, GPU, InPlaceUpdate, Integration, KubernetesUpgrade, NodeClaim, Scheduling, Spot, Subnet, Utilization]
+        suite: [ACR, BYOK, Chaos, Consolidation, Drift, GPU, InPlaceUpdate, Integration, KubernetesUpgrade, NodeClaim, Scheduling, Spot, Subnet, Utilization, Machines]
     permissions:
       contents: read
       id-token: write
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -149,6 +149,7 @@ jobs:
           git_ref: ${{ inputs.git_ref }}
           location: ${{ inputs.location }}
           provisionmode: ${{ inputs.provisionmode }}
+          identity_type: ${{ inputs.suite == 'Machines' && 'UserAssigned' || 'SystemAssigned' }}
       - name: build and publish karpenter
         shell: bash
         run: AZURE_ACR_NAME=${{ env.ACR_NAME }} make az-build
--- a/Makefile-az.mk
+++ b/Makefile-az.mk
@@ -93,6 +93,14 @@ az-mkaks-cilium: az-mkacr ## Create test AKS cluster (with --network-dataplane c
 	az aks get-credentials --name $(AZURE_CLUSTER_NAME) --resource-group $(AZURE_RESOURCE_GROUP) --overwrite-existing
 	skaffold config set default-repo $(AZURE_ACR_NAME).$(AZURE_ACR_SUFFIX)/karpenter
 
+az-mkaks-cilium-userassigned: az-mkacr az-create-workload-msi ## Create test AKS cluster with user-assigned identity (supports custom kubelet identity)
+	$(eval KARPENTER_USER_ASSIGNED_IDENTITY_ID=$(shell az identity show --resource-group "${AZURE_RESOURCE_GROUP}" --name "${AZURE_KARPENTER_USER_ASSIGNED_IDENTITY_NAME}" --query 'id' -otsv))
+	az aks create          --name $(AZURE_CLUSTER_NAME) --resource-group $(AZURE_RESOURCE_GROUP) --attach-acr $(AZURE_ACR_NAME) \
+		--assign-identity $(KARPENTER_USER_ASSIGNED_IDENTITY_ID) --node-count 3 --generate-ssh-keys -o none --network-dataplane cilium --network-plugin azure --network-plugin-mode overlay \
+		--enable-oidc-issuer --enable-workload-identity $(if $(AZURE_VM_SIZE),--node-vm-size $(AZURE_VM_SIZE),)
+	az aks get-credentials --name $(AZURE_CLUSTER_NAME) --resource-group $(AZURE_RESOURCE_GROUP) --overwrite-existing
+	skaffold config set default-repo $(AZURE_ACR_NAME).$(AZURE_ACR_SUFFIX)/karpenter
+
 az-mkaks-overlay: az-mkacr ## Create test AKS cluster (with --network-plugin-mode overlay)
 	az aks create          --name $(AZURE_CLUSTER_NAME) --resource-group $(AZURE_RESOURCE_GROUP) --attach-acr $(AZURE_ACR_NAME) \
 		--enable-managed-identity --node-count 3 --generate-ssh-keys -o none --network-plugin azure --network-plugin-mode overlay \
@@ -162,7 +170,9 @@ az-perm-aksmachine: ## Create role assignments for AKS machine API operations
 	$(eval KARPENTER_USER_ASSIGNED_CLIENT_ID=$(shell az identity show --resource-group "${AZURE_RESOURCE_GROUP}" --name "${AZURE_KARPENTER_USER_ASSIGNED_IDENTITY_NAME}" --query 'principalId' -otsv))
 	az role assignment create --assignee-object-id $(KARPENTER_USER_ASSIGNED_CLIENT_ID) --assignee-principal-type "ServicePrincipal" --scope /subscriptions/$(AZURE_SUBSCRIPTION_ID)/resourceGroups/$(AZURE_RESOURCE_GROUP) --role "Azure Kubernetes Service Contributor Role"
 	az role assignment create --assignee-object-id $(KARPENTER_USER_ASSIGNED_CLIENT_ID) --assignee-principal-type "ServicePrincipal" --scope /subscriptions/$(AZURE_SUBSCRIPTION_ID)/resourceGroups/$(AZURE_RESOURCE_GROUP_MC) --role "Network Contributor"
-	$(eval CLUSTER_IDENTITY=$(shell az aks show --resource-group "${AZURE_RESOURCE_GROUP}" --name "${AZURE_CLUSTER_NAME}" --query 'identity.principalId' -otsv))
+	## TODO Do we need to do similar handling for user assigned on other perms commands
+	$(eval CLUSTER_IDENTITY_TYPE=$(shell az aks show --resource-group "${AZURE_RESOURCE_GROUP}" --name "${AZURE_CLUSTER_NAME}" --query 'identity.type' -otsv))
+	$(eval CLUSTER_IDENTITY=$(shell if [ "$(CLUSTER_IDENTITY_TYPE)" = "UserAssigned" ]; then echo "$(KARPENTER_USER_ASSIGNED_CLIENT_ID)"; else az aks show --resource-group "${AZURE_RESOURCE_GROUP}" --name "${AZURE_CLUSTER_NAME}" --query 'identity.principalId' -otsv; fi))
 	az role assignment create --assignee-object-id $(CLUSTER_IDENTITY) --assignee-principal-type "ServicePrincipal" --scope /subscriptions/$(AZURE_SUBSCRIPTION_ID)/resourceGroups/$(AZURE_RESOURCE_GROUP_MC) --role "Virtual Machine Contributor"
 	az role assignment create --assignee-object-id $(CLUSTER_IDENTITY) --assignee-principal-type "ServicePrincipal" --scope /subscriptions/$(AZURE_SUBSCRIPTION_ID)/resourceGroups/$(AZURE_RESOURCE_GROUP_MC) --role "Network Contributor"
 	az role assignment create --assignee-object-id $(CLUSTER_IDENTITY) --assignee-principal-type "ServicePrincipal" --scope /subscriptions/$(AZURE_SUBSCRIPTION_ID)/resourceGroups/$(AZURE_RESOURCE_GROUP_MC) --role "Managed Identity Operator"
--- a/test/pkg/environment/azure/expectations.go
+++ b/test/pkg/environment/azure/expectations.go
@@ -18,6 +18,7 @@ package azure
 
 import (
 	"fmt"
+	"os"
 	"slices"
 	"strings"
 	"time"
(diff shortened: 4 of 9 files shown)


test/suites/machines/machines_test.go
```diff
@@ -0,0 +1,251 @@
...
+			env.EventuallyExpectHealthyPodCount(selector, int(numPods))
+
+			By("Preforming a K8s upgrade")
+			availableKubernetesUpgrades := env.ExpectSuccessfulGetOfAvailableKubernetesVersionUpgradesForManagedCluster()
+			kubernetesUpgradeVersion := *lo.MaxBy(availableKubernetesUpgrades, func(a, b *containerservice.ManagedClusterPoolUpgradeProfileUpgradesItem) bool {
+				aK8sVersion := lo.Must(semver.Parse(*a.KubernetesVersion))
+				bK8sVersion := lo.Must(semver.Parse(*b.KubernetesVersion))
+				return aK8sVersion.GT(bK8sVersion)
+			}).KubernetesVersion
+			upgradedMC := env.ExpectSuccessfulUpgradeOfManagedCluster(kubernetesUpgradeVersion)
```
comtalyst, line 235: According to the implementation, this might unintentionally wait for PUT to be done, while we want PUT to be ongoing?

.github/workflows/e2e.yaml
```diff
@@ -149,6 +149,7 @@ jobs:
           git_ref: ${{ inputs.git_ref }}
           location: ${{ inputs.location }}
           provisionmode: ${{ inputs.provisionmode }}
+          identity_type: ${{ inputs.suite == 'Machines' && 'UserAssigned' || 'SystemAssigned' }}
```
matthchr, line 152: Why do we need machine identity to be UserAssigned? Can we leave a comment here or in create-cluster/action.yaml explaining?

test/pkg/environment/azure/expectations.go
```diff
@@ -96,6 +97,40 @@ func (env *Environment) ExpectCreatedInterface(networkInterface armnetwork.Inter
 	})
 }
 
+func (env *Environment) ExpectGetManagedCluster() containerservice.ManagedCluster {
```
matthchr, line 100:
```suggestion
func (env *Environment) ExpectGetManagedCluster() *containerservice.ManagedCluster {
```

test/pkg/environment/azure/expectations.go
```diff
@@ -96,6 +97,40 @@ func (env *Environment) ExpectCreatedInterface(networkInterface armnetwork.Inter
...
+	Expect(err).ToNot(HaveOccurred())
+	return managedClusterResponse.ManagedCluster
+}
+
+// WarnIfClusterNotInExpectedProvisioningState checks if the clusters provisioning state is equal to the
+// given expected provisioning state.
+func (env *Environment) WarnIfClusterNotInExpectedProvisioningState(expectedProvisioningState string) containerservice.ManagedCluster {
+	GinkgoHelper()
+	managedCluster := env.ExpectGetManagedCluster()
+	Expect(managedCluster.Properties.ProvisioningState).ToNot(BeNil())
```
matthchr, line 112: minor: assumes `Properties` is not nil

test/pkg/environment/azure/expectations.go
```diff
@@ -96,6 +97,40 @@ func (env *Environment) ExpectCreatedInterface(networkInterface armnetwork.Inter
...
+func (env *Environment) ExpectGetManagedCluster() containerservice.ManagedCluster {
+	GinkgoHelper()
+	managedClusterResponse, err := env.managedClusterClient.Get(env.Context, env.ClusterResourceGroup, env.ClusterName, nil)
+	Expect(err).ToNot(HaveOccurred())
+	return managedClusterResponse.ManagedCluster
+}
+
+// WarnIfClusterNotInExpectedProvisioningState checks if the clusters provisioning state is equal to the
+// given expected provisioning state.
+func (env *Environment) WarnIfClusterNotInExpectedProvisioningState(expectedProvisioningState string) containerservice.ManagedCluster {
```
matthchr, line 109:
```suggestion
func (env *Environment) WarnIfClusterNotInExpectedProvisioningState(expectedProvisioningState string) *containerservice.ManagedCluster {
```

.github/workflows/e2e-matrix.yaml
```diff
@@ -51,7 +51,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        suite: [ACR, BYOK, Chaos, Consolidation, Drift, GPU, InPlaceUpdate, Integration, KubernetesUpgrade, NodeClaim, Scheduling, Spot, Subnet, Utilization]
+        suite: [ACR, BYOK, Chaos, Consolidation, Drift, GPU, InPlaceUpdate, Integration, KubernetesUpgrade, NodeClaim, Scheduling, Spot, Subnet, Utilization, Machines]
```
matthchr, line 54: minor: Alphabetical?

TBH might be a good idea at this point to just break this guy and put 1 suite per line.
This is YAML after all we don't need to mash everything into a single line AFAIK.

test/pkg/environment/azure/identity.go
```diff
@@ -65,3 +69,106 @@ func (env *Environment) GetCurrentUserPrincipalID(ctx context.Context, cred azco
...
+}
+
+// IsClusterUserAssignedIdentity checks if the cluster uses user-assigned managed identity
+func (env *Environment) IsClusterUserAssignedIdentity(ctx context.Context) bool {
+	identityType := env.CheckClusterIdentityType(ctx)
+	return identityType == "UserAssigned"
+}
+
+// GetKubeletIdentityResourceID returns the current kubelet identity resource ID
+func (env *Environment) GetKubeletIdentityResourceID(ctx context.Context) string {
```
matthchr, line 161: minor, style:

These two methods could just be one `GetKubeletIdentity` that returns the identity object and the caller could just go `.ClientID` or `.ResourceID`.

Slightly less code and also works better if you end up wanting another field of the identity later

test/pkg/environment/azure/identity.go
```diff
@@ -65,3 +69,106 @@ func (env *Environment) GetCurrentUserPrincipalID(ctx context.Context, cred azco
...
+	resp, err := msiClient.CreateOrUpdate(ctx, env.NodeResourceGroup, identityName, identity, nil)
+	Expect(err).ToNot(HaveOccurred())
+
+	// Note: we don't register for cleanup in the env.tracker, in case there are more tests to run. We don't want to break the cluster by deleting the kubelet identity
+	return &resp.Identity
+}
+
+// ExpectUpdatedManagedClusterKubeletIdentityAsync updates the kubelet identity of a managed cluster asynchronously
+// Returns the poller so the caller can control when to wait for completion
+func (env *Environment) ExpectUpdatedManagedClusterKubeletIdentityAsync(ctx context.Context, newIdentity *armmsi.Identity) *runtime.Poller[containerservice.ManagedClustersClientCreateOrUpdateResponse] {
```
matthchr, line 97: Still not following why we want/need to do this...

Makefile-az.mk
```diff
@@ -93,6 +93,14 @@ az-mkaks-cilium: az-mkacr ## Create test AKS cluster (with --network-dataplane c
 	az aks get-credentials --name $(AZURE_CLUSTER_NAME) --resource-group $(AZURE_RESOURCE_GROUP) --overwrite-existing
 	skaffold config set default-repo $(AZURE_ACR_NAME).$(AZURE_ACR_SUFFIX)/karpenter
 
+az-mkaks-cilium-userassigned: az-mkacr az-create-workload-msi ## Create test AKS cluster with user-assigned identity (supports custom kubelet identity)
+	$(eval KARPENTER_USER_ASSIGNED_IDENTITY_ID=$(shell az identity show --resource-group "${AZURE_RESOURCE_GROUP}" --name "${AZURE_KARPENTER_USER_ASSIGNED_IDENTITY_NAME}" --query 'id' -otsv))
```
matthchr, line 97: Needs to merge and use my new cluster create + check tag pattern

test/suites/machines/suite_test.go
```diff
@@ -0,0 +1,56 @@
...
+
+var env *azure.Environment
+var nodeClass *v1beta1.AKSNodeClass
+var nodePool *karpv1.NodePool
+
+func TestMachines(t *testing.T) {
+	RegisterFailHandler(Fail)
+	BeforeSuite(func() {
+		env = azure.NewEnvironment(t)
+		// > Note: we want to run this test case in Machine Mode regardless of what the config is,
```
matthchr, line 38: This comment reads a bit weird to me. Can we clarify?

Makefile-az.mk
```diff
@@ -162,7 +170,9 @@ az-perm-aksmachine: ## Create role assignments for AKS machine API operations
 	$(eval KARPENTER_USER_ASSIGNED_CLIENT_ID=$(shell az identity show --resource-group "${AZURE_RESOURCE_GROUP}" --name "${AZURE_KARPENTER_USER_ASSIGNED_IDENTITY_NAME}" --query 'principalId' -otsv))
 	az role assignment create --assignee-object-id $(KARPENTER_USER_ASSIGNED_CLIENT_ID) --assignee-principal-type "ServicePrincipal" --scope /subscriptions/$(AZURE_SUBSCRIPTION_ID)/resourceGroups/$(AZURE_RESOURCE_GROUP) --role "Azure Kubernetes Service Contributor Role"
 	az role assignment create --assignee-object-id $(KARPENTER_USER_ASSIGNED_CLIENT_ID) --assignee-principal-type "ServicePrincipal" --scope /subscriptions/$(AZURE_SUBSCRIPTION_ID)/resourceGroups/$(AZURE_RESOURCE_GROUP_MC) --role "Network Contributor"
-	$(eval CLUSTER_IDENTITY=$(shell az aks show --resource-group "${AZURE_RESOURCE_GROUP}" --name "${AZURE_CLUSTER_NAME}" --query 'identity.principalId' -otsv))
+	## TODO Do we need to do similar handling for user assigned on other perms commands
```
matthchr, line 173: Do we need to answer this TODO before merge?

test/suites/machines/machines_test.go
```diff
@@ -0,0 +1,251 @@
...
+		env.EventuallyExpectRegisteredNodeClaimCount("==", 1)
+		env.EventuallyExpectCreatedMachineCount("==", 1)
+		env.EventuallyExpectHealthyPodCount(selector, int(numPods))
+
+		// Check that the node has the expected networking labels
+		Expect(node.Labels).To(HaveKeyWithValue("kubernetes.azure.com/ebpf-dataplane", consts.NetworkDataplaneCilium))
+		Expect(node.Labels).To(HaveKeyWithValue("kubernetes.azure.com/azure-cni-overlay", "true"))
+		Expect(node.Labels).To(HaveKeyWithValue("kubernetes.azure.com/podnetwork-type", consts.NetworkPluginModeOverlay))
+
+		// Note: these labels we only check their keys since, the values are dynamic
```
matthchr, line 109:
```suggestion
		// Note: these labels we only check their keys since the values are dynamic
```

test/suites/machines/machines_test.go
```diff
@@ -0,0 +1,251 @@
...
+		// TODO: improve E2E test to be dynamic, reusing the same provisioning logic we have for labels creation.
+		Expect(lo.Keys(node.Labels)).To(ContainElements([]string{
+			"kubernetes.azure.com/network-subnet",
+			"kubernetes.azure.com/nodenetwork-vnetguid",
+			"kubernetes.azure.com/network-stateless-cni",
+		}))
+	})
+
+	// NOTE: ClusterTests modify the actual cluster itself, which means that preforming tests after a cluster test
+	// might not have a clean environment, and might produce unexpected results. Ordering of cluster tests is important
```
matthchr, line 119: minor: Should we add a note here that this is safe in CI because each E2E runs on its own cluster?

test/suites/machines/machines_test.go
```diff
@@ -0,0 +1,251 @@
...
+			env.ExpectUpdatedManagedClusterKubeletIdentityAsync(env.Context, newIdentity)
+
+			By("granting ACR access to the new kubelet identity")
+			env.ExpectGrantedACRAccess(env.Context, newIdentity)
+
+			By("verifying the kubelet identity was updated")
+			// updatedKubeletIdentityResourceID := env.GetKubeletIdentityResourceID(env.Context)
+
+			// TODO: check if we want to have this possibly logged
+			// Expect(updatedKubeletIdentityResourceID).To(Equal(lo.FromPtr(newIdentity.ID)), "Expected updatedKubeletIdentityResourceID to match new kubelet resource id")
```
matthchr, line 180: Do we not want this assert? Not sure I follow why not

test/suites/machines/machines_test.go
```diff
@@ -0,0 +1,251 @@
...
+			Eventually(func(g Gomega) {
+				machines := env.EventuallyExpectCreatedMachineCount("==", 3)
+				for _, machine := range machines {
+					g.Expect(machine.Properties.Status.DriftAction).ToNot(BeNil())
+					g.Expect(*machine.Properties.Status.DriftAction).To(Equal(containerservice.DriftActionRecreate))
+				}
+			}).WithTimeout(3 * time.Minute).Should(Succeed())
+
+			By("expecting nodes to drift")
+			env.EventuallyExpectDriftedWithTimeout(15*time.Minute, nodeClaims...)
```
matthchr, line 193: 15m seems a long time?

matthchr: especially since this is just detection of drift, not actual drift? Does this really take 15m? Shouldn't it take more like 30s?

test/suites/machines/machines_test.go
matthchr: agree - I don't see how cluster is still in upgrading at the end of this test given the above.

test/suites/machines/machines_test.go
charliedmcb: This was a light concern on leaking the indentity info. If there is no concern, than yes this asset should be there

test/suites/machines/machines_test.go
charliedmcb: So, its safe so long as the suite itself is setup correctly to handle the ordering. However, there is leakage between test cases within the suite

test/suites/machines/machines_test.go
charliedmcb: Yea, good catch.

I think I had a local fix for this changing the func to be async, and you could optionally wait for polling after, so here is could avoid that, and in the regular k8s test it could use that. Must have not gotten pushed, and lost in the codespace now. Sorry about that. 

Should be updated.

test/suites/machines/suite_test.go
charliedmcb: Basically, my idea here is to have baseline tests of certain machine mode functionality that get tested regardless of what matrix configuration we have for running the CI pipeline (since the CI pipeline is where we select if we are testing in machine mode or not).

For these tests, we cover a few important cases, specifically:
- Provisioning during an 'ing state
- Kubelet update

.github/workflows/e2e.yaml
charliedmcb: This is required for handling kubelet update/upgrade.

You can't do the user assigned kubelet update test without the cluster being in a user assigned identity state.

test/pkg/environment/azure/identity.go
charliedmcb: In what way? The fact that its async, or just the test in general. 

Idea of async is that we can watch the test, and preform any needed actions while the op is ongoing.

test/suites/machines/machines_test.go
charliedmcb: yea, I think this came from the other drift tests. Agreed can be shortened; though, I feel like it'd be worth looking at the drift logic, or some metrics to make sure its now being shortened too far.

Makefile-az.mk
charliedmcb: I beleive this was more a question of cleaniness, and functionality of the makefile commands itself. Like if folks used the commands in a self-hosted/standalone testing environment, ensuring that user assigned worked in the other instances too. 

For the tests itself, I don't think its needed beyond this.

matthchr:
Replaced by https://github.com/Azure/karpenter-provider-azure/pull/1427

PR #1214, opened 2025-10-10
Design for metrics implementation

MaximilianoUribe:
Fixes # 

**Description**
Adding design for the implementation of metrics in azure provider

image

*E2E test and grafana

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


designs/metrics.md
```diff
@@ -0,0 +1,296 @@
...
+```
+
+## Design Principles
+
+### Metric Namespace Identification
+
+To distinguish Azure provider metrics from core Karpenter metrics, we use a **different namespace**:
+
+- **Core Karpenter metrics**: `karpenter_*` (from upstream Karpenter)
+- **Azure provider metrics**: `karpenter_aks_*` (Azure-specific implementations)
```
rakechill: nit: I think using `karpenter_azure_` here instead would be a bit better.

pkg/providers/instance/vminstance.go
```diff
@@ -643,11 +642,11 @@ func (p *DefaultProvider) createVirtualMachine(ctx context.Context, opts *create
 	}
 	vm := newVMObject(opts)
 	log.FromContext(ctx).V(1).Info("creating virtual machine", "vmName", opts.VMName, logging.InstanceType, opts.InstanceType.Name)
-	metrics.VMCreateStartMetric.Inc(ctx, "creating virtual machine", metrics.ImageID(opts.LaunchTemplate.ImageID))
+	VMCreateStartMetric.WithLabelValues(opts.LaunchTemplate.ImageID).Inc()
```
rakechill: Have you tested this in standalone? I think if you use the full `make az-all ` + `make az-mon-deploy` + `make az-mon-access` that should allow you to access the grafana dash

designs/metrics.md
MaximilianoUribe: i agree

pkg/providers/instance/vminstance.go
MaximilianoUribe: Screenshot 2025-10-13 at 3 09 36 PM

MaximilianoUribe:
do we think we should remove logmetrics as a whole? i think that if we want to keep consistency this will be a good addition to the PR

designs/metrics.md
```diff
@@ -0,0 +1,296 @@
...
+
+This design document explains the approach for organizing metrics in the Karpenter Azure provider, specifically focusing on **component-specific metrics that are co-located with their implementation** rather than being centralized.
+
+## First Example: Instance Provider Metrics
+
+The instance provider serves as the **first example** of this co-location pattern, demonstrating how metrics should be organized across the codebase.
+
+### 1. Created Component-Specific Metrics File
+
+**File Created:** `pkg/providers/instance/metrics.go`
```
matthchr, line 13: Did you consider a package rather than just a file? `pkg/providers/instance/metrics/metrics.go` or similar?

It has (I think?) all of the same organization benefits of co-location, with the added value of allowing a clear separation between exported/non-exported metrics functions (if we need any), and keeps metrics code/definitions from cluttering the main business-logic package of controllers

matthchr: In general I think we (Go authors) over-bloat individual packages. The unit of encapsulation is a package.

matthchr: FWIW this also seems to be what the core provider does? pkg/cloudprovider/metrics/cloudprovider.go, for example.

designs/metrics.md
```diff
@@ -0,0 +1,296 @@
...
+**Metric Naming Examples:**
+- `karpenter_aks_instance_vm_create_start_total` - Azure provider metric
+- `karpenter_nodeclaims_created_total` - Core Karpenter metric
+
+### Co-location Pattern
+
+Following Kubernetes best practices, **production Prometheus metrics are co-located with their implementation components**:
+
+1. **Provider-specific metrics** → Live within provider packages (e.g., `pkg/providers/instance/`)
+2. **Controller-specific metrics** → Live within controller packages (e.g., `pkg/controllers/*/`)
```
matthchr, line 96: I assume you mean within a _specific_ controller package, right? Like `controllers/nodeclaim` - not at the top level controllers folder, yes?

designs/metrics.md
```diff
@@ -0,0 +1,296 @@
...
+```
+
+**Label Selection Guidelines:**
+- Choose labels that help diagnose component-specific issues
+- Keep cardinality reasonable (avoid high-cardinality labels like instance IDs)
+- Use labels that align with how the component operates
+
+### Label Constants
+
+Defined locally in each component's metrics file:
```
matthchr: Aren't most labels going to be shared across various metrics? i.e. even though metrics themselves are per-component, labels need consistency across components. I think labels (at least labels we think are going to be common/shared across more than just this packages metrics) should probably be defined in a single common place. If we want to re-declare them in the individual metrics file that's probably fine, like:
```
myLabel = metrics.MyLabel
```

but I don't think we should be duplicating the actual strings across N files. Too easy to mismatch. TBH may just be better to put all of the labels in one shared file rather than next to the metrics that use them.

designs/metrics.md
```diff
@@ -0,0 +1,296 @@
...
+
+```go
+const (
+    componentSubsystem = "component_name"
+    label1 = "label_one"
+    label2 = "label_two"
+)
+```
+
+These are **component-specific** and don't need to be shared unless multiple components use identical labels.
```
matthchr: As mentioned above I think shared labels is going to be the common case. nodeClaimName, nodePoolName, etc.

designs/metrics.md
```diff
@@ -0,0 +1,296 @@
...
+
+### Metric Call Pattern
+
+Using standard Prometheus API:
+
+```go
+// Simple, direct metric recording
+ComponentMetric.WithLabelValues(value1, value2).Inc()
```
matthchr, line 176: I think we should prefer (require?) the map variant. Relying on value ordering is a recipe for mis-ordering versus the metric definition. Again [see core](https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/cloudprovider/metrics/cloudprovider.go#L158)

designs/metrics.md
```diff
@@ -0,0 +1,296 @@
...
+
+1. **Constants**
+   ```go
+   const (
+       Namespace = "karpenter_aks"  // Shared across all metrics
+   )
+   ```
+
+2. **Builder Patterns**
+   ```go
```
matthchr: Not sure I follow this.

If the metric definitions are per package, what do we need to construct here?

designs/metrics.md
```diff
@@ -0,0 +1,296 @@
...
+When adding metrics to other components, follow this pattern:
+
+### 1. Create Component Metrics File
+
+```go
+// pkg/providers/<component>/metrics.go
+package component
+
+import (
+    coremetrics "github.com/Azure/karpenter-provider-azure/pkg/metrics"
```
matthchr: minor: This isn't usually what we mean by `core`. Core means upstream (karpenter-core). If it's from karpenter-provider-azure, I don't' think it's `core`.

designs/metrics.md
```diff
@@ -0,0 +1,296 @@
...
+- Use labels that align with how the component operates
+
+### Label Constants
+
+Defined locally in each component's metrics file:
+
+```go
+const (
+    componentSubsystem = "component_name"
+    label1 = "label_one"
```
matthchr: may be worth calling out explicit expectation of `snake_case` in labels, and reference to prom docs where they recommend that (I assume it exists somewhere?)

designs/metrics.md
```diff
@@ -0,0 +1,296 @@
...
+
+var (
+    MyComponentMetric = prometheus.NewCounterVec(
+        prometheus.CounterOpts{
+            Namespace: coremetrics.Namespace,
+            Subsystem: componentSubsystem,
+            Name:      "operation_total",
+            Help:      "Total number of operations.",
+        },
+        []string{"operation_type"},
```
matthchr: probably always want the labels defined in consts.

designs/metrics.md
matthchr: Though I guess you could make the argument that if you just have a `GetLabelMap(string, string)` we're again relying on ordering (though at least we know we have the right number of arguments, which is something).

We could require it inline...
```
map[string]string{
		metricLabelController: "foo",
		metricLabelMethod:     "bar",
		metricLabelProvider:   "baz",
	}
```

As the prom docs point out, this is safer but also less performant so may need to be careful in hot paths...

matthchr: The more I think about this the more I think we should take a stance on how we're dealing with this risk.

Either tests, linting (not 100% sure how this would work), prefer maps in most cases... are the ideas I see.

pkg/metrics/constants.go
```diff
@@ -18,7 +18,7 @@ package metrics
 
 const (
 	// Namespace(s).
-	Namespace = "karpenter"
```
matthchr, line 21: Isn't this a breaking change?

selfhosted users can use the metrics we're emitting already.

Also checked and aws emits metrics in the standard Karpenter namespace, thinking possibly we should just match them?

pkg/providers/instance/metrics.go
```diff
@@ -0,0 +1,93 @@
...
+	//
+	// STABILITY: ALPHA - This metric may change or be removed without notice.
+	VMCreateStartMetric = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Namespace: coremetrics.Namespace,
+			Subsystem: instanceSubsystem,
+			Name:      "vm_create_start_total",
+			Help:      "Total number of VM creation operations started.",
+		},
+		[]string{imageID},
```
matthchr: Does ImageID contain OS as well? Wondering if we want that split out for easy azlinux vs ubuntu comparisons?

I know when Charlie added this we were only using it internally but I think we need to think more carefully about what we want here and what we can have.

Things like:
- VM size
- Zone
- Capacity type (spot vs dedicated)
- Network details (is this always the same for a whole cluster? If so, maybe don't need it)

Seem to make a lot of sense here and are probably OK from a cardinality perspective? Can we go over all the fields that we _could_ have and pick out the ones we decide we want/need?

designs/metrics.md
matthchr: Ah I see core does this but the aws provider doesn't seem to (at least not consistently, it does do it some I think)

MaximilianoUribe: In the long term (imo), the advantage of not bundling these into a single package is that keeping them per component eliminates the need to import unrelated metrics and prevents the metrics package from becoming overloaded with metrics from multiple components. But of course is only my opinion and what I see from aws, lmkwyt

designs/metrics.md
MaximilianoUribe: yes

designs/metrics.md
MaximilianoUribe: i agree that this would be more clear in the big picture, we can place these labels in the constants file in metrics and rename appropriately

designs/metrics.md
MaximilianoUribe: I agree that this should be a topic of discussion, I agree that maybe taking the map approach and also implementing a test for a specific metrics should be standard?

pkg/metrics/constants.go
MaximilianoUribe: Yes, this is a breaking change. I agree that we should follow the standard Karpenter namespace. Perhaps this should be a team discussion to determine whether there’s value in differentiating our provider metrics from the core metrics

pkg/providers/instance/metrics.go
MaximilianoUribe: yes, as you mentioned this is done with cardinality in mind, we can have a discussion about these fields sure

designs/metrics.md
```diff
@@ -0,0 +1,296 @@
...
+**After:**
+```go
+// No separate metrics import needed - metrics are in same package
+
+ComponentMetric.WithLabelValues(value).Inc()
+```
+
+## Design Principles
+
+### Metric Namespace Identification
```
rakechill, line 59: Let's add a section here on tradeoffs between using a provider-specific metrics namespace `karpenter_azure` vs. reusing core `karpenter`

Examples of emitting/decorating pre-existing core metrics in our provider code:
* https://github.com/Azure/karpenter-provider-azure/blob/30165d784b3293653320bbd5373145471a6e3431/cmd/controller/main.go#L63
* https://github.com/Azure/karpenter-provider-azure/blob/main/pkg/cloudprovider/cloudprovider.go#L224

Pre-existing provider metric that uses `karpenter` namespace, but unlikely that it has consumers/shouldn't be considered stable:
* https://github.com/Azure/karpenter-provider-azure/blob/30165d784b3293653320bbd5373145471a6e3431/pkg/metrics/metrics.go#L24-L34

designs/metrics.md
matthchr: I'm not saying they should be in a single package I'm saying a package per spot (`pkg/providers/instance/metrics/metrics.go`)

edit: but if _all_ that's in there is just the metrics defs and there are no unexported functions or anything like that, I'm fine with just a file too. As far as I can tell, core does some of each.

designs/metrics.md
MaximilianoUribe: Thinking more about this, going with the map solution I think will add more complexity to the solution (taking into consideration that it also requires order), but I do agree that adding testing for each metric definition should be standard.

matthchr:
> do we think we should remove logmetrics as a whole? i think that if we want to keep consistency this will be a good addition to the PR

There may be alerts based on that data that @charliedmcb wrote. Can you check if there are? If so, we can't remove until we've updated the alerts to use the metrics instead.

pkg/providers/instance/metrics.go
```diff
@@ -0,0 +1,90 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package instance
+
+import (
+	metricslabels "github.com/Azure/karpenter-provider-azure/pkg/metrics"
```
matthchr: minor, style, feel free to ignore: I'd personally import this as just `metrics` since:
- The values you're reading from it already say label, so `metrics.ImageLabel` is clear enough and `metricslabels.ImageLabel` is a bit repetitive.
- Technically this package isn't _just_ labels (at least, it doesn't say it is by its path) and there may end up being other "shared" metrics helpers there - just maybe not metrics themselves, and it would be weird to access those via `metricslabels` if they weren't labels.

rakechill:
> > do we think we should remove logmetrics as a whole? i think that if we want to keep consistency this will be a good addition to the PR
> 
> There may be alerts based on that data that @charliedmcb wrote. Can you check if there are? If so, we can't remove until we've updated the alerts to use the metrics instead.

Just checked. We don't have any relying on this.

pkg/metrics/constants.go
```diff
@@ -18,8 +18,12 @@ package metrics
...
+	Namespace = "karpenter_azure"
```
matthchr: I thought there was some agreement we were going to keep this `karpenter`?

designs/metrics.md
matthchr: Well raw maps inline doesn't require order:
```
	map[string]string{
		metricLabelController: "foo",
		metricLabelMethod:     "bar",
		metricLabelProvider:   "baz",
	}
```

Im fine with tests too, but can you show an example either in your doc or maybe even better in the new samples you added, of how that test would be written to assert that the metric looks right?

The specific thing I think we need to assert is that if we have a call like this:
```
ComponentMetric.WithLabelValues(value1, value2, value3, value4).Inc()
```

the values are ordered correctly for the labels they are supposed to be for. i.e. we dont accidentally do: `ComponentMetric.WithLabelValues(value2, value3, value1, value4).Inc()` - which I know when written this way looks easy to avoid but when the names of the values arent numbered it's a lot harder =)

designs/metrics.md
```diff
@@ -0,0 +1,304 @@
...
+
+We are intentionally keeping provider metrics in the core `karpenter_*` namespace so that:
+
+- **Compatibility with upstream tooling** remains intact—existing dashboards and alerts maintained by the Karpenter community continue to function. For example, our controller still emits core metrics by wrapping the provider implementation with `metrics.Decorate` in `cmd/controller/main.go` (see [`cmd/controller/main.go#L63`](https://github.com/Azure/karpenter-provider-azure/blob/main/cmd/controller/main.go#L63)).
+- **Stability expectations** stay aligned—metrics such as `ProvisioningDurationSeconds` in `pkg/cloudprovider/cloudprovider.go` (see [`cloudprovider.go#L224`](https://github.com/Azure/karpenter-provider-azure/blob/main/pkg/cloudprovider/cloudprovider.go#L224)) already ship in this namespace and have consumers.
+
+In practice, the controller will continue to emit the decorated core metrics for compatibility, and new Azure-specific metrics (such as the VM create counters) will also use the `karpenter` namespace to maintain a single, consistent surface area. This keeps expectations aligned with upstream Karpenter and simplifies future maintenance and documentation.
```
matthchr, line 89: minor: Worth clarifying that we'll have our own "subsystems" so it should still be relatively clear which metrics are from core versus us (plus we can document that too)

pkg/metrics/constants.go
MaximilianoUribe: missed the correction :(

pkg/metrics/constants.go
```diff
@@ -22,4 +22,12 @@ const (
 
 	// Subsystem(s).
 	imageFamilySubsystem = "image"
+
+	// Label key(s).
+	ImageLabel        = "image"
+	ErrorCodeLabel    = "error"
```
matthchr:
```suggestion
	ErrorCodeLabel    = "error_code"
```

Maybe?

pkg/providers/instance/metrics.go
```diff
@@ -0,0 +1,91 @@
...
+	//
+	// STABILITY: ALPHA - This metric may change or be removed without notice.
+	VMCreateSyncFailureMetric = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Namespace: metrics.Namespace,
+			Subsystem: instanceSubsystem,
+			Name:      "vm_create_sync_failure_total",
+			Help:      "Total number of synchronous VM creation failures.",
+		},
+		[]string{metrics.ImageLabel, metrics.SizeLabel, metrics.ZoneLabel, metrics.CapacityTypeLabel, metrics.NodePoolLabel},
```
matthchr: missing error code label?

pkg/providers/instance/metrics.go
```diff
@@ -0,0 +1,91 @@
...
+			Name:      "vm_create_async_failure_total",
+			Help:      "Failed to create virtual machine during LRO",
+		},
+		[]string{metrics.ImageLabel, metrics.SizeLabel, metrics.ZoneLabel, metrics.CapacityTypeLabel, metrics.NodePoolLabel},
+	)
+
+	// VMCreateResponseErrorMetric tracks VM creation response errors.
+	//
+	// STABILITY: ALPHA - This metric may change or be removed without notice.
+	VMCreateResponseErrorMetric = prometheus.NewCounterVec(
```
matthchr: This isn't actually used anywhere and I am not sure it makes sense as a standalone metric?

Metric should be failures - failures can have `error_code`?

pkg/providers/instance/metrics.go
```diff
@@ -0,0 +1,91 @@
...
+			Name:      "vm_create_sync_failure_total",
+			Help:      "Total number of synchronous VM creation failures.",
+		},
+		[]string{metrics.ImageLabel, metrics.SizeLabel, metrics.ZoneLabel, metrics.CapacityTypeLabel, metrics.NodePoolLabel},
+	)
+
+	// VMCreateAsyncFailureMetric tracks asynchronous VM creation failures.
+	//
+	// STABILITY: ALPHA - This metric may change or be removed without notice.
+	VMCreateAsyncFailureMetric = prometheus.NewCounterVec(
```
matthchr: @rakechill, @tallaxes: thoughts on distinguishing async_failure from sync_failure like this?

Part of me thinks if we're going to differentiate, having a single metric with label for `failure_time` (or `phase`?) or something (sync|async) makes more sense. OTOH, much (all?) of the time, we just care that it failed and we don't really care "where" (or we could determine at what phase from logs if we really needed to).

tl;dr: IMO we should combine `vm_create_async_failure_total` and `vm_create_sync_failure_total` into `vm_create_failure_total` and _maybe_ add a `phase` or similar label.

pkg/providers/instance/suite_test.go
```diff
@@ -94,6 +99,96 @@ var _ = AfterSuite(func() {
...
+			if metricLabelsMatch(metric, labels) {
+				return metric, true
+			}
+		}
+	}
+
+	return nil, false
+}
+
+func metricLabelsMatch(metric *dto.Metric, expected map[string]string) bool {
```
matthchr: I think there's an existing assertion you can do with `gomega` that should cover this?

```
			mapA := map[string]string{}
			mapB := map[string]string{}
			Expect(mapA).To(Equal(mapB))
```

Does that not work?

So you do:
```
Expect(metric.GetLabel()).To(Equal(expected))
```

it should give you a nice diff-based error message too so it's both shorter and better, I think?

pkg/providers/instance/suite_test.go
```diff
@@ -147,6 +242,102 @@ var _ = Describe("VMInstanceProvider", func() {
...
+		BeforeEach(func() {
+			instancemetrics.VMCreateStartMetric.Reset()
+			instancemetrics.VMCreateSyncFailureMetric.Reset()
+			instancemetrics.VMCreateAsyncFailureMetric.Reset()
+			instancemetrics.VMCreateResponseErrorMetric.Reset()
+		})
+
+		It("records VM create start metric during successful launch", func() {
+			ExpectApplied(ctx, env.Client, nodeClaim, nodePool, nodeClass)
+			instanceTypes, err := cloudProvider.GetInstanceTypes(ctx, nodePool)
```
matthchr: I don't think you should have to call `BeginCreate` for `VMInstanceProvider` yourself here. I think if you just do:
```
				ExpectApplied(ctx, env.Client, nodePool, nodeClass)
				pod := coretest.UnschedulablePod(coretest.PodOptions{})
```

and then do your assertions on the metrics, it'll work without you having to do the calls yourself?

pkg/providers/instance/suite_test.go
```diff
@@ -94,6 +99,96 @@ var _ = AfterSuite(func() {
...
+		return lo.FromPtr(ref.ID)
+	case ref.CommunityGalleryImageID != nil:
+		return lo.FromPtr(ref.CommunityGalleryImageID)
+	case ref.SharedGalleryImageID != nil:
+		return lo.FromPtr(ref.SharedGalleryImageID)
+	case ref.ExactVersion != nil:
+		return lo.FromPtr(ref.ExactVersion)
+	default:
+		return ""
+	}
```
matthchr:
```suggestion
	return lo.CoalesceOrEmpty(
		lo.FromPtr(ref.ID),
		lo.FromPtr(ref.CommunityGalleryImageID),
		lo.FromPtr(ref.SharedGalleryImageID),
		lo.FromPtr(ref.ExactVersion))
```

pkg/providers/instance/suite_test.go
```diff
@@ -94,6 +99,96 @@ var _ = AfterSuite(func() {
...
+}
+
+func vmSizeFromVM(vm *armcompute.VirtualMachine) string {
+	if vm == nil || vm.Properties == nil || vm.Properties.HardwareProfile == nil || vm.Properties.HardwareProfile.VMSize == nil {
+		return ""
+	}
+	return string(*vm.Properties.HardwareProfile.VMSize)
+}
+
+func zoneLabelFromVM(vm *armcompute.VirtualMachine) string {
```
matthchr:
```suggestion
func zoneFromVM(vm *armcompute.VirtualMachine) string {
```

Match your naming convention above?

pkg/providers/instance/suite_test.go
matthchr: If the issue here is that at different times Karpenter can choose different VM sizes, so you have a hard time writing a static assertion because the assertions keep changing you can do one of a few things:

1. Use a non-default nodePool that restricts things down to where there should only be 1 choice.
2. Use the default nodePool like you're doing now, _but_ just extract the VM create input like you're doing here: `createInput := azureEnv.VirtualMachinesAPI.VirtualMachineCreateOrUpdateBehavior.CalledWithInput.Pop()` and use that for building the label values you're asserting against.

pkg/providers/instance/suite_test.go
matthchr: Ah I see you're using this as part of a larger workflow where you want a boolean return because you need to find the metric in question.

Still shouldn't need to implement map-equality ourselves there are a bunch of helpers that exist already for it. You could use `reflect.DeepEqual` for example on maps. Maybe better than that is `maps.Equal`.

pkg/providers/instance/vminstance.go
```diff
@@ -642,11 +643,25 @@ func (p *DefaultVMProvider) createVirtualMachine(ctx context.Context, opts *crea
...
 	if err != nil {
-		metrics.VMCreateSyncFailureMetric.Inc(ctx, "failed to create virtual machine on initial put", metrics.ImageID(opts.LaunchTemplate.ImageID))
+		VMCreateFailureMetric.With(map[string]string{
+			metrics.ImageLabel:        opts.LaunchTemplate.ImageID,
+			metrics.SizeLabel:         opts.InstanceType.Name,
+			metrics.ZoneLabel:         opts.Zone,
+			metrics.CapacityTypeLabel: opts.CapacityType,
+			metrics.NodePoolLabel:     opts.NodePoolName,
+			metrics.PhaseLabel:        phaseSyncFailure,
+			metrics.ErrorCodeLabel:    err.Error(),
```
matthchr: I don't think this is actually the error_code we want, it's just going to be a  big string with a lot of possible values and blow out cardinality, isn't it?

I think we need to extract the error_code from the ARM HTTP response. See https://github.com/Azure/azure-sdk-for-go-extensions/blob/69ee2f9f2b8a9610e0d1f31738f364079b6cc262/pkg/errors/armerrors.go#L50C11-L50C26

I think we need to find the AzureError code and put that here. If we can't find the code we can fall back to Unknown.

matthchr: Same applies on the async side (the error parsing should be the same)

pkg/providers/instance/metrics.go
matthchr: Unresolving this just to give @rakechill and @tallaxes a chance to weigh in if they want.

pkg/metrics/test_helpers.go
```diff
@@ -0,0 +1,72 @@
...
+		}
+	}
+
+	return nil, false, nil
+}
+
+// FailureMetricLabels produces a failure metric label set by merging the provided base labels with
+// the phase label and any additional label maps. Later maps override earlier values when conflicts occur.
+func FailureMetricLabels(base map[string]string, phase string, extra ...map[string]string) map[string]string {
+	labels := make(map[string]string, len(base)+len(extra)+1)
```
matthchr: Probably easier to do with 
```
phaseMap := map[string]string{
	PhaseLabel: phase
}
return lo.Assign(base, phaseMap, extra...)
```

I think?

designs/metrics.md
```diff
@@ -0,0 +1,325 @@
...
+### 1. Created Component-Specific Metrics File
+
+**File Created:** `pkg/providers/instance/metrics.go`
+
+This file contains metrics specific to the instance provider component. As the first implementation of this pattern, it includes VM creation tracking metrics:
+
+```go
+package instance
+
+var (
```
matthchr, line 20: minor: correct this based on what you've actually got now.

pkg/providers/instance/suite_test.go
```diff
@@ -147,6 +200,95 @@ var _ = Describe("VMInstanceProvider", func() {
...
+			ExpectApplied(ctx, env.Client, nodeClaim, nodePool, nodeClass)
+			pod := coretest.UnschedulablePod(coretest.PodOptions{})
+			ExpectProvisioned(ctx, env.Client, cluster, cloudProvider, coreProvisioner, pod)
+			ExpectScheduled(ctx, env.Client, pod)
+
+			Expect(azureEnv.VirtualMachinesAPI.VirtualMachineCreateOrUpdateBehavior.CalledWithInput.Len()).To(BeNumerically(">=", 1))
+			createInput := azureEnv.VirtualMachinesAPI.VirtualMachineCreateOrUpdateBehavior.CalledWithInput.Pop()
+			labels := vmMetricLabelsFromCreateInput(createInput, nodePool.Name)
+
+			metric, ok, err := metrics.FindMetricWithLabelValues("karpenter_instance_vm_create_start_total", labels)
```
matthchr: Just confirming that this asserts the labels + values match right? (99% sure it does)

pkg/providers/instance/suite_test.go
```diff
@@ -147,6 +200,95 @@ var _ = Describe("VMInstanceProvider", func() {
...
+			Expect(err).NotTo(HaveOccurred())
+			Expect(ok).To(BeTrue())
+			Expect(metric.GetCounter().GetValue()).To(BeNumerically("==", 1))
+
+			syncFailureLabels := metrics.FailureMetricLabels(labels, "sync", map[string]string{metrics.ErrorCodeLabel: beginErr.Error()})
```
matthchr: Will need to fix this to expect the code is `OperationNotAllowed` not the whole err

pkg/providers/instance/vminstance.go
```diff
@@ -66,6 +66,19 @@ const (
...
+func errorCodeForMetrics(err error) string {
+	if err == nil {
+		return ""
+	}
+	if azErr := sdkerrors.IsResponseError(err); azErr != nil {
+		if azErr.ErrorCode != "" {
+			return azErr.ErrorCode
+		}
+	}
+	return err.Error()
```
matthchr: I don't think you want this. Return a sentinel like `Unknown` if we don't know.

pkg/providers/instance/vminstance.go
```diff
@@ -66,6 +66,19 @@ const (
 	cseNameLinux   = "cse-agent-karpenter"
 )
 
+// errorCodeForMetrics extracts a stable Azure error code for metric labeling when possible, preventing overcardinality in metrics with failures.
+func errorCodeForMetrics(err error) string {
```
matthchr: minor: Future item but I think we could upstream this to azure sdk for go extensions as:
`ErrorCode(err error) string` and just document that it returns a static string `Unknown` if the error is not of type ResponseError.

matthchr (approved):

pkg/providers/instance/vminstance.go
matthchr: Missing some unit tests.

Prefer to write not in Ginkgo (plain go test for UTs IMO)

pkg/metrics/constants.go
```diff
@@ -22,4 +22,13 @@ const (
 
 	// Subsystem(s).
 	imageFamilySubsystem = "image"
```
rakechill, line 24: do we need this subsystem anymore?

pkg/metrics/test_helpers.go
```diff
@@ -0,0 +1,67 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package metrics
+
+import (
+	"maps"
+
+	dto "github.com/prometheus/client_model/go"
```
rakechill, line 20: very minor nit: why is this called dto?

pkg/metrics/test_helpers.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/samber/lo"
+	crmetrics "sigs.k8s.io/controller-runtime/pkg/metrics"
+)
+
+// FindMetricWithLabelValues locates a metric by name and label values within the controller-runtime registry.
+// It returns the metric, whether it was found, and any error produced while gathering metrics.
```
rakechill: Do we need the bool here? Wouldn't a nil pointer to a metric be sufficient for saying the metric wasn't found?

pkg/providers/instance/suite_test.go
```diff
@@ -147,6 +200,95 @@ var _ = Describe("VMInstanceProvider", func() {
...
+			ExpectProvisioned(ctx, env.Client, cluster, cloudProvider, coreProvisioner, pod)
+			ExpectScheduled(ctx, env.Client, pod)
+
+			Expect(azureEnv.VirtualMachinesAPI.VirtualMachineCreateOrUpdateBehavior.CalledWithInput.Len()).To(BeNumerically(">=", 1))
+			createInput := azureEnv.VirtualMachinesAPI.VirtualMachineCreateOrUpdateBehavior.CalledWithInput.Pop()
+			labels := vmMetricLabelsFromCreateInput(createInput, nodePool.Name)
+
+			metric, ok, err := metrics.FindMetricWithLabelValues("karpenter_instance_vm_create_start_total", labels)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(ok).To(BeTrue())
```
rakechill: Left a comment in the utils function itself, but seems like we don't need to use this second rv "ok" + can instead just use `nil` ptr as our not found val

pkg/providers/instance/metrics.go
```diff
@@ -0,0 +1,65 @@
...
+		prometheus.CounterOpts{
+			Namespace: metrics.Namespace,
+			Subsystem: instanceSubsystem,
+			Name:      "vm_create_start_total",
+			Help:      "Total number of VM creation operations started.",
+		},
+		[]string{metrics.ImageLabel, metrics.SizeLabel, metrics.ZoneLabel, metrics.CapacityTypeLabel, metrics.NodePoolLabel},
+	)
+
+	// VMCreateFailureMetric tracks VM creation failures, regardless of phase.
```
rakechill, line 46: Maybe clarify that the two possible phases are "sync" and "async"

pkg/providers/instance/suite_test.go
```diff
@@ -147,6 +200,95 @@ var _ = Describe("VMInstanceProvider", func() {
...
+			Expect(metric.GetCounter().GetValue()).To(BeNumerically("==", 1))
+
+			_, ok, err = metrics.FindMetricWithLabelValues("karpenter_instance_vm_create_failure_total", metrics.FailureMetricLabels(labels, "sync"))
```
rakechill: Is this just saying, we expect create failure total metric **to not be emitted** during both sync and async portion of provisioning?

bc this code path shouldn't be entered if the vm was successfully created

pkg/providers/instance/vminstance.go
```diff
@@ -38,7 +38,7 @@ import (
 	"github.com/Azure/karpenter-provider-azure/pkg/cache"
 	"github.com/Azure/karpenter-provider-azure/pkg/consts"
 	"github.com/Azure/karpenter-provider-azure/pkg/logging"
-	"github.com/Azure/karpenter-provider-azure/pkg/metrics"
+	metrics "github.com/Azure/karpenter-provider-azure/pkg/metrics"
```
rakechill, line 41: It will always use the last word of the import statement, so I think in this case it will already use "metrics". Why do we need this extra aliasing?

pkg/providers/instance/vminstance.go
```diff
@@ -66,6 +66,19 @@ const (
 	cseNameLinux   = "cse-agent-karpenter"
 )
 
+// errorCodeForMetrics extracts a stable Azure error code for metric labeling when possible, preventing overcardinality in metrics with failures.
+func errorCodeForMetrics(err error) string {
+	if err == nil {
+		return ""
```
rakechill: Should we also return `Unknown` here? Wondering why we would end up with a failed VM w/o error code....

pkg/providers/instance/vminstance.go
rakechill: I think I agree with this wrt potential to have bad cardinality, but it would potentially erase important information/signals from these failures. 

Do we have any data on:
1. How many different error codes are we receiving now? I'm guessing we are logging them?
2. How many of these would fail the IsResponseError check?

pkg/providers/instance/vminstance.go
rakechill: https://github.com/Azure/karpenter-provider-azure/blob/main/pkg/providers/instance/azureresourcegraphlist_test.go

^^ smth like this just to verify this function works properly

pkg/metrics/constants.go
MaximilianoUribe: yes, this is another metric that bryce created, will need to update in the future

pkg/providers/instance/vminstance.go
matthchr: Reactivating: absent Kusto data showing this risk is low, I still think this should be unknown not the full error text... it's _probably_ the case that these error texts are static and this is relatively safe, but:

1. I want data showing that assuming we can get it today.
2. We should maybe leave a comment here explaining how we know it's OK.

PR #1229, opened 2025-10-21
chore: pin E2E to specific VM size that we have most quota for

matthchr:

* E2E test: https://github.com/Azure/karpenter-provider-azure/actions/runs/18668452187

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
NONE
```

### DIFF
--- a/.github/actions/e2e/create-cluster/action.yaml
+++ b/.github/actions/e2e/create-cluster/action.yaml
@@ -6,13 +6,13 @@ inputs:
   #   required: false
   #   default: "1.27"
   client-id:
-    description: 
+    description:
     required: true
   tenant-id:
-    description: 
+    description:
     required: true
   subscription-id:
-    description: 
+    description:
     required: true
   resource_group:
     description: "Name of the resource group to create the cluster within"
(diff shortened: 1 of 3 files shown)


.github/actions/e2e/create-cluster/action.yaml
```diff
@@ -55,8 +55,9 @@ runs:
       env:
         AZURE_CLUSTER_NAME: ${{ inputs.cluster_name }}
         AZURE_RESOURCE_GROUP: ${{ inputs.resource_group }}
-        AZURE_ACR_NAME: ${{ inputs.acr_name }} 
+        AZURE_ACR_NAME: ${{ inputs.acr_name }}
         AZURE_LOCATION: ${{ inputs.location }}
+        AZURE_VM_SIZE: Standard_D4ds_v5 # We have more quota for this VM size so force it to avoid quota errors
```
matthchr, line 60: Note that this means the tests that run in E2E will be a bit different than what you get with vanilla `make az-mkaks-cilium`. That's because in the vanilla make we don't pick a VM size we let AKS do it for us.

@tallaxes if we're worried about that I can just make the default in the makefile this size instead. LMK which you think is better.

matthchr:
will wait for E2E to pass before merging.

.github/actions/e2e/create-cluster/action.yaml
tallaxes: No, I think what you propose is fine - as long as we remember the difference. Maybe the makefile can print the value (if it is not already)

PR #1232, opened 2025-10-24
feat(NodeClass): add option to configure patch settings

squat:
This commit introduces a configuration option to the NodeClass CRD to
control the Linux VM patch settings.
xref: https://learn.microsoft.com/en-us/azure/virtual-machines/automatic-vm-guest-patching

This setting is useful for enabling automatic patching on the
provisioned node, which is required by many compliance frameworks to
improve the security posture of VMs.

Note that the field on the NodeClass spec is `Patching` rather than
`PatchSettings`, which would more closely match the upstream library.
My thinking is that this is less confusing than having fields suffixed
with various forms of `settings`, `options`, `config`, etc. Also, this
naming aligns with how the `Security` field drops the `Profile` suffix.

Signed-off-by: squat <lserven@gmail.com>


squat:

PR #1244, opened 2025-10-30
chore: add design template

matthchr:
Refactor designs folder a little to make it slightly easier to read over time

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note
NONE
```


tallaxes:
Anything useful from KEP template? https://github.com/kubernetes/enhancements/blob/master/keps/NNNN-kep-template/README.md

matthchr:
I borrowed goals/non-goals from KEP.

I don't mind Alternatives from KEP but I think it helps to put it front and center which is where Decisions came from. I could definitely be convinced to rename that to Proposal though rather than decisions - though I do like the "decision 1: what do we ...?" thing I think.

Dont think we need any of the test plan stuff atm, trying to keep it light/fast.

Maybe the biggest gap is a lack of "design details" section. I kinda think that'll come out as part of decisions/proposal, but if we want to change that to have maybe decisions and then design details where you get more into the meat of the options we chose and what they imply, I could be convinced of that too.

TBH mostly just wanted something and we can further refine as we go forward as well.

tallaxes (dismissed):
LGTM. Might also want to add a bit on testing & production readiness (enablement, observability, etc.) later, to make sure those are at least considered

matthchr:
Ill add a section at the bottom on testing and production readiness just as a catch-all for now

PR #1245, opened 2025-10-30
docs: add labels/requirements improvements design

matthchr:
- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note
NONE
```

### DIFF
--- a/designs/0006-requirements-and-labels.md
+++ b/designs/0006-requirements-and-labels.md
@@ -109,9 +109,9 @@ something can be added as a requirement we should keep that in mind.
 Here's a (probably not exhaustive) list of labels that AKS writes.
 The ones that I think Karpenter should allow scheduling on (but doesn't currently) are **in bold**.
 
-**Note**: AKS (observed or code) means we see the label written on nodes im practice by AKS. This doesn't necessarily mean that the AKS service writes every single one of these labels. Some of them are written directly
-by Kubelet with with integration from Azure CloudProvider, such as `topology.kubernetes.io/zone`, see [pkg/provider/azure_zones.go](https://github.com/kubernetes-sigs/cloud-provider-azure/blob/master/pkg/provider/azure_zones.go)
-for more details.
+**Note**: AKS (observed or code) means we see the label written on nodes im practice by AKS. This doesn't necessarily mean that the AKS service writes every single one of these labels. Some of them are written by other components such
+as directly CloudProvider [node controller](https://github.com/kubernetes/kubernetes/blob/d777de7741d36d1cc465162d94f39200e299070b/staging/src/k8s.io/cloud-provider/controllers/node/node_controller.go#L490). There may also be other
+label writers as well.
 
 | Label                                                   | AKS (documented) | AKS (observed or code) | Karpenter (schedulable) | Karpenter (written to node) | Notes                                                                                   |
 | ------------------------------------------------------- | ---------------- | ---------------------- | ----------------------- | --------------------------- | --------------------------------------------------------------------------------------- |
@@ -257,8 +257,8 @@ Machine API is not that far away, and we already write most critical labels. Som
 
 ## FAQ
 
-### How do the labels get onto the Node?
-There are currently two paths for labels to get onto the node:
+### How do Karpenter labels get onto the Node?
+There are currently two paths for Karpenter labels to get onto the node:
 * Written by Karpenter core onto the node object.
 * Written by Kubelet onto the node object (configured by the Azure Karpenter provider)
 


designs/0006-labels.md
```diff
@@ -0,0 +1,289 @@
...
+that it matches the static value.
+
+#### Conclusion: Neither option is appropriate for all labels. We need a combination.
+I propose the following guidance for choosing which labels fall into each category:
+* Choose **Option A (Block)** if it is impossible for the label to ever change or vary and/or it doesn't make any sense to attempt to schedule on it.
+  For example `kubernetes.azure.com/cluster` seems unlikely to ever change over time. The cluster resource group will be the resource group always.
+* Choose **Option B (Requirements)** if the label is static now but may theoretically be expanded/relaxed to be non-static/multi-valued in the future.
+  For example `topology.kubernetes.io/region` - today we do not support cross-region node provisioning but we theoretically could.
+
+**TODO**: Not sure the above is that well-defined
```
matthchr: Do folks agree with above... rules reasonable enough?

designs/0006-labels.md
```diff
@@ -0,0 +1,289 @@
...
+
+### Decision 2: What to do with system-determined cluster-wide labels?
+Labels for which there is one correct value for the whole cluster, but that value may change over time as the cluster topology changes (usually driven by changes to the AKS ManagedCluster data model).
+Examples: `kubernetes.azure.com/ebpf-dataplane`, `kubernetes.azure.com/network-name`
+
+**TODO**: Fill this in... answer for ebpf-dataplane may need to be special...?
```
matthchr: Calling out TODO to myself, need to fix

designs/0006-labels.md
```diff
@@ -0,0 +1,289 @@
...
+| node.kubernetes.io/instance-type                        | ✅               | ✅                     | ✅                      | ✅                          |                                                                                         |
+| topology.kubernetes.io/region                           | ✅               | ✅                     | ✅                      | ✅                          |                                                                                         |
+| topology.kubernetes.io/zone                             | ✅               | ✅                     | ✅                      | ✅                          |                                                                                         |
+| kubernetes.azure.com/cluster                            | ✅               | ✅                     | ❌                      | ✅                          | TODO: Today user can overwrite what AKS sets, we need to block                          |
```
matthchr: Im tracking this now I can remove this TODO

designs/0006-labels.md
```diff
@@ -0,0 +1,289 @@
...
+This is because:
+* It's marginally easier to reason about possible configurations when you can look at a well-typed object describing the options, which well-known labels does not support well.
+  This is especially the case with requirements which may come associated with complex configuration - it's not just `thing: on|off`, it's also `thingConfig: 5 options`.
+* Historical reasons (including following what upstream/AWS and other providers do).
+* Avoiding having too many default requirements which can cause cpu/memory burden.
+
+**Note**: Well-known requirements that aren't directly about the instance SKU selection and are instead about provisioning details such as OS or zone require special handling in the code to ensure
+that provisioning is done considering them. This is often accomplished via [createOfferings](https://github.com/Azure/karpenter-provider-azure/blob/main/pkg/providers/instancetype/instancetypes.go#L215).
+
+**TODO: Do we agree on this? Is this a principled enough stance?**
```
matthchr: Thoughts?

designs/0006-labels.md
```diff
@@ -0,0 +1,289 @@
...
+
+* kubernetes.azure.com/ebpf-dataplane
+* kubernetes.azure.com/os-sku-effective
+* kubernetes.azure.com/os-sku-requested
+* kubernetes.azure.com/security-type
+* kubernetes.azure.com/accelerator
+
+We will allow these labels to be set on the NodePool/NodeClaim (in addition to the labels we already support as schedulable outlined in the tables above):
+* kubernetes.azure.com/ebpf-dataplane
+* kubernetes.azure.com/cluster-health-monitor-checker-synthetic (due to its high usage...) **TODO:** Learn what this is
```
matthchr: Calling out TODO to myself

designs/0006-requirements-and-labels.md
```diff
@@ -0,0 +1,289 @@
...
+| karpenter.azure.com/sku-hyperv-generation           |                  |                        | ✅                      | ✅                          |       |
+
+#### Conclusion
+
+We will add these labels to WellKnownLabels + requirements:
+* kubernetes.azure.com/scalesetpriority
+* kubernetes.azure.com/node-image-version
+* kubernetes.azure.com/fips_enabled
+* kubernetes.azure.com/os-sku
+* kubernetes.azure.com/mode
```
matthchr, line 207: Actually we may not need to add this if we just require users always set it on their pool then we can schedule against it anyway.

matthchr: Though if we don't make this a requirement, and user pods request `mode: user`, we will set `mode: user` but won't be able to correctly simulate their pods will land on the pool, so it may be worth making it schedulable anyway...

designs/0006-requirements-and-labels.md
```diff
@@ -0,0 +1,289 @@
...
+* kubernetes.azure.com/storageprofile
+* kubernetes.azure.com/storagetier
+
+We already block usage of the `karpenter.azure.com` and `karpenter.sh` namespaces except for certain labels we understand. This will not change.
+
+We will also block all `kubernetes.azure.com` namespaced labels except for those called out above as either well known labels, or allowed.
+
+### Decision 5: How will we get labels that Karpenter doesn't currently write set on nodes?
+
+#### Option A: Wait for Machine API migration
```
comtalyst, line 247: Just FYI that public Machine API doc should be published soon. Feel free to note something like "see more details in Machine API doc".

designs/0006-labels.md
```diff
@@ -0,0 +1,289 @@
...
+**Author:** @matthchr
+
+**Last updated:** Oct 30, 2025
+
+**Status:** Proposed
+
+## Overview
+
+This document examines Karpenter and AKS managed labels and discusses what is needed to bring what we currently do more
+in line with AKS does.
```
tallaxes:
```suggestion
in line with what AKS does.
```
(and maybe single line?)

designs/0006-labels.md
```diff
@@ -0,0 +1,289 @@
...
+
+There are 2 main topics:
+* Label parity with AKS, setting labels that we don't currently set.
+* Support for scheduling simulation on labels we don't currently support, but should to enable easier migrations from cluster autoscaler to Karpenter.
```
tallaxes:
```suggestion
* Support for scheduling simulation on labels we don't currently support, but should; to enable easier migrations from cluster autoscaler to Karpenter.
```

designs/0006-requirements-and-labels.md
```diff
@@ -0,0 +1,289 @@
...
+* Label parity with AKS, setting labels that we don't currently set.
+* Support for scheduling simulation on labels we don't currently support, but should to enable easier migrations from cluster autoscaler to Karpenter.
+
+Today, Karpenter has the following categories of labels:
+
+1. Labels Karpenter understands and can schedule (Requirements)
+    * Can be set on Pod when not set on NodePool/NodeClaim
+    * Can be set on NodePool/NodeClaim
+    * May trigger behavior other than just setting a label (inform VM size, zone, etc)
+2. Labels written to the nodes but that Karpenter doesn't understand (can't schedule)
```
tallaxes, line 22: Is it worth distinguishing between those passed to kubelet by the provider vs. those set via other mechanisms?

designs/0006-labels.md
```diff
@@ -0,0 +1,289 @@
...
+
+Today, Karpenter has the following categories of labels:
+
+1. Labels Karpenter understands and can schedule (Requirements)
```
tallaxes: Maybe rather "target for scheduling"?

designs/0006-labels.md
```diff
@@ -0,0 +1,289 @@
...
+### Well-known requirements
+
+Well-known requirements (those which come automatically from [computeRequirements](https://github.com/Azure/karpenter-provider-azure/blob/e322adfbad3567eb5ff34d776a2c365e060a5009/pkg/providers/instancetype/instancetype.go#L135))
+help us schedule workloads to nodes by ensuring that the intersection of node requirements and workload requirements align.
+
+Even though the structure in Karpenter that requirements are transported on is `cloudprovider.InstanceType`, they can and should contain more information
+than _just_ information about the instance type itself. For example information about the OS and the instances deployment circumstances (zone, capacity reservation, etc)
+also make sense as well-known requirements.
+
+We currently shy away from including things that aren't directly related to the instance type, OS, or how it is provisioned in the set of well-known requirements.
```
tallaxes: An example of such thing would be helpful, maybe imageFamily?

designs/0006-labels.md
```diff
@@ -0,0 +1,289 @@
...
+This is because:
+* It's marginally easier to reason about possible configurations when you can look at a well-typed object describing the options, which well-known labels does not support well.
+  This is especially the case with requirements which may come associated with complex configuration - it's not just `thing: on|off`, it's also `thingConfig: 5 options`.
+* Historical reasons (including following what upstream/AWS and other providers do).
+* Avoiding having too many default requirements which can cause cpu/memory burden.
+
+**Note**: Well-known requirements that aren't directly about the instance SKU selection and are instead about provisioning details such as OS or zone require special handling in the code to ensure
```
tallaxes: "instance SKU" -> "instance type" or "VM SKU"

designs/0006-labels.md
```diff
@@ -0,0 +1,289 @@
...
+
+Well-known requirements (those which come automatically from [computeRequirements](https://github.com/Azure/karpenter-provider-azure/blob/e322adfbad3567eb5ff34d776a2c365e060a5009/pkg/providers/instancetype/instancetype.go#L135))
+help us schedule workloads to nodes by ensuring that the intersection of node requirements and workload requirements align.
+
+Even though the structure in Karpenter that requirements are transported on is `cloudprovider.InstanceType`, they can and should contain more information
+than _just_ information about the instance type itself. For example information about the OS and the instances deployment circumstances (zone, capacity reservation, etc)
+also make sense as well-known requirements.
+
+We currently shy away from including things that aren't directly related to the instance type, OS, or how it is provisioned in the set of well-known requirements.
+This is because:
```
tallaxes: It seems to me the primary reason is still conceptual: requirements (in NodePool or workload) are primarily a filter / selector (a collection of constraints on the existing set), where other things (e.g. settings in NodeClass) contribute to the desired state for the chosen alternative. Zone, Capacity Type, and OS can be thought of as part of this filter / selector. (Possibly to slightly different extent in different providers, e.g. I think on AWS capacity type and maybe even platform/OS are part of the internal capacity pool definition. It looks like both Azure and AWS have instance types that only support one OS ....)

This separation is also useful / easier to reason about from UX perspective and explains why requirements (=filter/selector) generally target simple-valued intrinsic instance type properties (though could match multiple values, and support negation), where other things (=settings) can have complex/structured values.

designs/0006-labels.md
```diff
@@ -0,0 +1,289 @@
...
+Machine API is not that far away, and we already write most critical labels. Some additional labels such as `kubernetes.azure.com/os-sku` and `kubernetes.azure.com/fips_enabled` will be enabled for scheduling
+(and thus be written to the nodes) based on decision 4, but other labels and full label parity will come with the machine API migration.
+
+## FAQ
+
+### How do the labels get onto the Node?
+
+* Karpenter core writes all NodeClaim labels to the node, see: [syncNode](https://github.com/kubernetes-sigs/karpenter/blob/38e728c99b530f660f685d53d01f2e9ec9696668/pkg/controllers/nodeclaim/lifecycle/registration.go#L128).
+* Karpenter core translates most requirements from the NodeClaim spec into labels, see [PopulateNodeClaimDetails](https://github.com/kubernetes-sigs/karpenter/blob/38e728c99b530f660f685d53d01f2e9ec9696668/pkg/controllers/nodeclaim/lifecycle/launch.go#L129)
+    * Note that `.Labels()` does exclude multi-valued requirements and labels defined in `WellKnownLabels` (registered by the cloud provider). See [Labels()](https://github.com/kubernetes-sigs/karpenter/blob/38e728c99b530f660f685d53d01f2e9ec9696668/pkg/scheduling/requirements.go#L270).
```
tallaxes: "`.Labels()` does exclude multi-valued requirements" - is that true? It excludes those that have [`.Any()`](https://github.com/kubernetes-sigs/karpenter/blob/565c33c673d28f5f5a442182dcc14db6cb6b44ea/pkg/scheduling/requirement.go#L233) produce an empty value, which would only exclude `DoesNotExist`? It looks like requirement with "In" would use the first value (even if there are multiple), and - IMO the most surprisingly from user perspective - generate random integer label values for "Exists/Gt/Lt" ... So when it does appear to skip multi-valued it is probably due to `WellKnownLabels` etc.?

"`.Labels()` does exclude ... labels defined in `WellKnownLabels` (registered by the cloud provider)" - probably should say "including those registered by cloud provider", some come from core. It also excludes labels under `RestrictedLabelDomains` ("prohibited by kubelet or reserved by karpenter", e.g. "kubernetes.io", "karpenter.sh", "karpenter.azure.com", ...) - unless under `LabelDomainExceptions` ("not used in a context where they may be passed as argument to kubelet", e.g. "node.kubernetes.io").

designs/0006-labels.md
```diff
@@ -0,0 +1,289 @@
...
+
+### What happens if user sets a non-schedulable (== non-requirements) label affinity on their pod?
+
+WellKnownLabels (like: `kubernetes.azure.com/cluster`): Infinite nodes if the label doesn't match the expected value - otherwise one node.
+Non-WellKnownLabels that we don't write (like: `kubernetes.azure.com/network-name`): Results in `"incompatible requirements, label \"kubernetes.azure.com/network-name\" does not have known values`
+Non-WellKnownLabels that we do write (like `kubernetes.azure.com/ebpf-dataplane`): Results in `incompatible requirements, label \"kubernetes.azure.com/kubelet-identity-clientid\" does not have known values"` - because the simulation engine in Karpenter doesn't know we're going to write it.
+
+### Where can I see more about how AKS manages labels?
+
+https://github.com/Azure/karpenter-poc/issues/1517#issuecomment-3470039930 discusses this further.
```
tallaxes: Should not link to a private repo from a public one?

designs/0006-labels.md
```diff
@@ -0,0 +1,289 @@
...
+
+### What's the difference between NodePool/NodeClaim spec.template.metadata.labels and spec.template.requirements?
+
+There isn't really a difference. NodePool labels _become_ requirements, see [pkg/controllers/provisioning/scheduling/nodeclaimtemplate.go](https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/controllers/provisioning/scheduling/nodeclaimtemplate.go#L75).
+
+### What happens if user sets a non-schedulable (== non-requirements) label affinity on their pod?
+
+WellKnownLabels (like: `kubernetes.azure.com/cluster`): Infinite nodes if the label doesn't match the expected value - otherwise one node.
+Non-WellKnownLabels that we don't write (like: `kubernetes.azure.com/network-name`): Results in `"incompatible requirements, label \"kubernetes.azure.com/network-name\" does not have known values`
+Non-WellKnownLabels that we do write (like `kubernetes.azure.com/ebpf-dataplane`): Results in `incompatible requirements, label \"kubernetes.azure.com/kubelet-identity-clientid\" does not have known values"` - because the simulation engine in Karpenter doesn't know we're going to write it.
```
tallaxes: Bulleted list would probably work better here?

```suggestion
* WellKnownLabels (like: `kubernetes.azure.com/cluster`): Infinite nodes if the label doesn't match the expected value - otherwise one node.
* Non-WellKnownLabels that we don't write (like: `kubernetes.azure.com/network-name`): Results in `"incompatible requirements, label \"kubernetes.azure.com/network-name\" does not have known values`
* Non-WellKnownLabels that we do write (like `kubernetes.azure.com/ebpf-dataplane`): Results in `incompatible requirements, label \"kubernetes.azure.com/kubelet-identity-clientid\" does not have known values"` - because the simulation engine in Karpenter doesn't know we're going to write it.
```

designs/0006-labels.md
```diff
@@ -0,0 +1,289 @@
...
+* Labels that get skipped by core (either due to `WellKnownLabels` or having multiple values) are added to the NodeClaim by AKS Karpenter provider in [vmInstanceToNodeClaim](https://github.com/Azure/karpenter-provider-azure/blob/b9c8c82edb289ac5b281c85b0851b5a0c45bc4bb/pkg/cloudprovider/cloudprovider.go#L476).
+  * Note that `WellKnownLabels` seems to serve two purposes:
+      1. It prevents Karpenter core from automatically including those labels on the node object (though note that AKS Karpenter provider ends up including most of them on the node object anyway).
+      2. It allows pods to ask for that label when the NodeClaim/requirements don't actually have it, and still have scheduling take place (I don't fully understand where/why this is needed but I can see the code for it [here](https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/scheduling/requirements.go#L175))
+* Requirements come from InstanceType via [computeRequirements](https://github.com/Azure/karpenter-provider-azure/blob/1f327c6d7ac62b3a3a1ad83c0f14f95001ab4ae8/pkg/providers/instancetype/instancetype.go#L135)
```
tallaxes: Narrative-wise, one would expect this to be about NodeClaim spec requirements - which [come](https://github.com/kubernetes-sigs/karpenter/blob/565c33c673d28f5f5a442182dcc14db6cb6b44ea/pkg/controllers/provisioning/scheduling/nodeclaimtemplate.go#L74-L85) from multiple sources (including NodePool requirements and labels) and don't include requirements derived from instance type? (Those get translated into labels by cloud provider ...)

designs/0006-labels.md
```diff
@@ -0,0 +1,289 @@
...
+
+## FAQ
+
+### How do the labels get onto the Node?
+
+* Karpenter core writes all NodeClaim labels to the node, see: [syncNode](https://github.com/kubernetes-sigs/karpenter/blob/38e728c99b530f660f685d53d01f2e9ec9696668/pkg/controllers/nodeclaim/lifecycle/registration.go#L128).
+* Karpenter core translates most requirements from the NodeClaim spec into labels, see [PopulateNodeClaimDetails](https://github.com/kubernetes-sigs/karpenter/blob/38e728c99b530f660f685d53d01f2e9ec9696668/pkg/controllers/nodeclaim/lifecycle/launch.go#L129)
+    * Note that `.Labels()` does exclude multi-valued requirements and labels defined in `WellKnownLabels` (registered by the cloud provider). See [Labels()](https://github.com/kubernetes-sigs/karpenter/blob/38e728c99b530f660f685d53d01f2e9ec9696668/pkg/scheduling/requirements.go#L270).
+* Labels that get skipped by core (either due to `WellKnownLabels` or having multiple values) are added to the NodeClaim by AKS Karpenter provider in [vmInstanceToNodeClaim](https://github.com/Azure/karpenter-provider-azure/blob/b9c8c82edb289ac5b281c85b0851b5a0c45bc4bb/pkg/cloudprovider/cloudprovider.go#L476).
```
tallaxes: This happens before core NodeClaim controllers update labels, right? (The wording may be confusing regarding order ...) These labels are populated on Create by the provider from single-valued instance type requirements, and by special logic for multi-valued ones. The associated requirements don't get added to NodeClaim as requirements, and thus don't get observed on NodeClaim by NodeClaim controllers, right?

designs/0006-requirements-and-labels.md
```diff
@@ -0,0 +1,289 @@
...
+This has already been accounted for in the tables above. We could add more labels there.
+
+#### Conclusion: Wait for machine API for complete parity
+
+Machine API is not that far away, and we already write most critical labels. Some additional labels such as `kubernetes.azure.com/os-sku` and `kubernetes.azure.com/fips_enabled` will be enabled for scheduling
+(and thus be written to the nodes) based on decision 4, but other labels and full label parity will come with the machine API migration.
+
+## FAQ
+
+### How do the labels get onto the Node?
```
tallaxes: The following describes the sequence in the reverse order, which makes sense, but probably worth noting explicitely

tallaxes (commented):
Capturing more comments/questions

designs/0006-labels.md
matthchr: Done

designs/0006-requirements-and-labels.md
matthchr: When you say "set by other mechanisms" you mean written by components other than Karpenter?

I do kindof already draw that distinction somewhat in the tables below
- Karpenter (schedulable) means that Karpenter supports it as a requirement _and_ writes it to Kubelet labels.
- Karpenter (written to node) means that Karpenter writes it to the node but doesn't support if for scheduling
- AKS (observed or code) means that it gets written somehow by AKS, either by AKS RP or some other mechanism

So labels that are "AKS (observed or code)" but not "Karpenter (written to node)" means that they are written by some other mechanism to the node.

Is what you're trying to get it "should we be writing it"?

tallaxes: > When you say "set by other mechanisms" you mean written by components other than Karpenter?

Yes. And Yes the distinction is already covered in the table, I was just wondering if is worth breaking down in the categorization here - but maybe not. As you said:

> Note that labels written by other parts of the system (daemonsets, operators, etc) generally fall into this category

designs/0006-labels.md
matthchr: imageFamily is actually `kubernetes.azure.com/os-sku`, I think? and I propose lower down that we do support scheduling on that.

I've added an example of `kubernetes.azure.com/podnetwork-subscription` here... though it's sorta flimsy because I don't know that we would _want_ to support scheduling on that... another possible consideration is `kubernetes.azure.com/fips_enabled` or `kubernetes.azure.com/mode`, both of which we do not allow today and both of which aren't _really_ about the instance... but I propose we add them both to requirements below.

designs/0006-labels.md
matthchr: Not 100% sure I am following this:
> requirements (in NodePool or workload) are primarily a filter / selector (a collection of constraints on the existing set), where other things (e.g. settings in NodeClass) contribute to the desired state for the chosen alternative

OS (for example) also is about the desired state. I see the argument about "complex vs simple" and "filters vs complex configuration" somewhat, but I don't know that it necessarily excludes many fields that (today at least) are excluded.

For example, FIPS is a feature of the OS. Insofar as we support scheduling on the OS (and we do), FIPS on/off impacts that and so is "a filter" too, isn't it? It _also_ does flow somewhere and impact the desired state, but then again so do all of the existing requirements like zone, etc.

matthchr: Another (possibly) interesting angle here that I haven't explored is "system determined vs user determined" w.r.t. the labels themselves.

We support scheduling on any labels the user wants as long as they put them into the NodePool. So the problem arises when the system _reserves_ certain labels for its own use (often as proxies for if feature XYZ is enabled), but those labels are also interesting for scheduling purposes or users are already using them for scheduling purposes in CAS.

This may be different between AKS/EKS because I think EKS just doesn't add as many labels as we do, see: https://repost.aws/questions/QUsppjW2m9Sy-5lZsFikMYHg/eks-node-selector-possible-labels and also https://docs.aws.amazon.com/eks/latest/userguide/create-node-pool.html.

designs/0006-requirements-and-labels.md
matthchr: Added. Also clarified there are two paths here. In the actual issue tracking this there is an item to remove one of the paths (IMO the non-kubelet one), at least for the "core" labels we own. User labels will still be set by Karpenter core via the path documented here.

designs/0006-labels.md
matthchr: oh good catch

designs/0006-labels.md
matthchr: I've refactored the wording here a bit which might help.

For this:
> The associated requirements don't get added to NodeClaim as requirements, and thus don't get observed on NodeClaim by NodeClaim controllers.

Correct (I tested locally too just to confirm). Instance requirements don't end up on the claim unless they are explicitly listed in the NodePool or Pod. Otherwise they are in the set of "options" returned via InstanceTypes but not actually present on the NodeClaim object.

designs/0006-labels.md
matthchr: You're right, the way I wrote this is confusing because I'm mixing requirements on InstanceType and "actual requirements".

I've cleaned this up some, I think. The focus here is on labels so we're interested in the requirements on instance type and less so on the actual requirements on the NodePool.

designs/0006-requirements-and-labels.md
matthchr: Yeah I think trying to keep this simple and just talk about the categories at a high-level that really matter for the purposes of "schedulable or not" is ideal. We can get into the nitty-gritty in the FAQ for the different sub-categories here if we want.

designs/0006-labels.md
matthchr: Fixed

designs/0006-labels.md
tallaxes: > OS (for example) also is about the desired state.

I was thinking the same thing - and then discovered that there are instance types / VM SKUs that only support one of the two. So, at least in that sense, it could still be thought of as filter ...

designs/0006-labels.md
matthchr: You're right this isn't actually accurate.

Have updated to:
```
    * Note that `.Labels()` excludes labels defined in `WellKnownLabels`, including both those registered by the cloud provider and core. It also excludes labels defined in `RestrictedLabelDomains`, which are
      labels that are prohibited by Kubelet or reserved by Karpenter, except for certain sub-domains within the restricted domains defined by `LabelDomainExceptions`.
```

designs/0006-labels.md
```diff
@@ -0,0 +1,289 @@
...
+Labels for which there is one value and that value is (at least for now) static. Examples: `topology.kubernetes.io/region`, `kubernetes.azure.com/cluster`
+
+#### Option A: Block
+Block setting these labels on the NodePool/NodeClaim.
+
+#### Option B: Requirements
+Allow setting these labels on the NodePool/NodeClaim, but specify them in requirements, which will effectively require that the user either doesn't request the label on their workload or NodePool/NodeClaim, or if they do
+that it matches the static value.
+
+#### Conclusion: Neither option is appropriate for all labels. We need a combination.
```
matthchr: I'm not sure if this conclusion is actually well-defined enough. What do people think?

designs/0006-requirements-and-labels.md
```diff
@@ -0,0 +1,318 @@
...
+We will add these labels to WellKnownLabels + requirements:
+* kubernetes.azure.com/scalesetpriority
+* kubernetes.azure.com/node-image-version
+* kubernetes.azure.com/fips_enabled
+* kubernetes.azure.com/os-sku
+* kubernetes.azure.com/mode
+
+We will consider adding these labels to WellKnownLabels + requirements in the future:
+
+* kubernetes.azure.com/ebpf-dataplane - we may need to add this to requirements in the future to support network-dataplane updates. In the migration from azure -> cilium,
```
matthchr, line 211: Calling this out here we do not currently allow network transitions so this isn't an issue today, but if we wanted to allow network transitions it ends up leading into this conversation and I _think_ at least one solution (IMO probably the best solution) ends up putting `kubernetes.azure.com/ebpf-dataplane` in well-known labels, which would remove the need for the user to correct set this value on all of their NodePools based on the network topology of the cluster (we effectively set it for them).

designs/0006-requirements-and-labels.md
```diff
@@ -0,0 +1,318 @@
...
+  which are basically specific VM sizes) and so would probably make sense as well-known requirements. Note that even though `kubernetes.azure.com/os-sku` is about `Ubuntu` vs `AzureLinux` which is specified in the NodeClass, it still
+  fits into requirements as well because it's about the OS which is very commonly required for scheduling.
+
+### Decision 4: Which labels will we block, and which labels will we allow scheduling on?
+
+Here's a (probably not exhaustive) list of labels that AKS writes.
+The ones that I think Karpenter should allow scheduling on (but doesn't currently) are **in bold**.
+
+| Label                                                   | AKS (documented) | AKS (observed or code) | Karpenter (schedulable) | Karpenter (written to node) | Notes                                                                                   |
+| ------------------------------------------------------- | ---------------- | ---------------------- | ----------------------- | --------------------------- | --------------------------------------------------------------------------------------- |
```
matthchr, line 117: For zone -- who writes it? Kubelet? CloudProvider?

matthchr: Figure out which ones are written by CloudProvider

designs/0006-labels.md
comtalyst: Format-wise, we could consider adding the dimension of decisions/options and categories to the table below in addition to this

designs/0006-labels.md
```diff
@@ -0,0 +1,318 @@
...
+| karpenter.azure.com/sku-family                      |                  |                        | ✅                      | ✅                          |       |
+| karpenter.azure.com/sku-version                     |                  |                        | ✅                      | ✅                          |       |
+| karpenter.azure.com/sku-storage-ephemeralos-maxsize |                  |                        | ✅                      | ✅                          |       |
+| karpenter.azure.com/sku-storage-premium-capable     |                  |                        | ✅                      | ✅                          |       |
+| karpenter.azure.com/sku-networking-accelerated      |                  |                        | ✅                      | ✅                          |       |
+| karpenter.azure.com/sku-hyperv-generation           |                  |                        | ✅                      | ✅                          |       |
+
+#### Conclusion
+We will add these labels to WellKnownLabels + requirements:
```
comtalyst: Would be great to include example implementations of adding one of these labels to well-known requirements too, and its example impact. E.g., mode could be easiest(?)

designs/0006-labels.md
matthchr: I reformatted this starting section a bunch just now to stop repeating myself so much so the information is more condensed. Not 100% sure that this resolves the ask you have but maybe it does?

designs/0006-labels.md
matthchr: I'll do this in a PR (I can even come back and update the design with a link when I do, good idea).

matthchr: Added that to the item tracking this.

designs/0006-requirements-and-labels.md
matthchr: Actually none of the labels are really written by CloudProvider from what I can tell (except maybe the beta scheduling labels). CloudProvider just configures kubelet to write the labels automatically. I've updated this to mention that.

designs/0006-labels.md
matthchr: After speaking with upstream maintainers I think the answer/consensus is "everything you want to schedule should go here even if it's not about the instance type". See https://kubernetes.slack.com/archives/C04JW2J5J5P/p1762367462502089

PR #1246, opened 2025-10-31
chore: fix claude settings

matthchr:
This fixes some claude settings so that the claude container works again. At least for me it did not work without these fixes.

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
NONE
```

### DIFF
--- a/designs/metrics.md
+++ b/designs/metrics.md
@@ -1,0 +1,332 @@
+# Metrics Design: Co-located Component-Specific Metrics
+
+## Overview
+
+This design document explains the approach for organizing metrics in the Karpenter Azure provider, specifically focusing on **component-specific metrics that are co-located with their implementation** rather than being centralized.
+
+## First Example: Instance Provider Metrics
+
+The instance provider serves as the **first example** of this co-location pattern, demonstrating how metrics should be organized across the codebase.
+
+### 1. Created Component-Specific Metrics File
+
+**File Created:** `pkg/providers/instance/metrics.go`
+
+This file contains metrics specific to the instance provider component. As the first implementation of this pattern, it currently ships two VM creation counters:
+
+```go
+package instance
+
+var (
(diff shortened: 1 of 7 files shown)


.claude/settings.json
```diff
@@ -1,9 +1,14 @@
...
+    "ANTHROPIC_AUTH_TOKEN": "sk-",
     "ANTHROPIC_BASE_URL": "http://localhost:4141",
-    "ANTHROPIC_AUTH_TOKEN": "dummy",
-    "ANTHROPIC_MODEL": "claude-sonnet-4",
-    "ANTHROPIC_SMALL_FAST_MODEL": "claude-3.7-sonnet",
+    "ANTHROPIC_DEFAULT_OPUS_MODEL": "claude-opus-41",
+    "ANTHROPIC_DEFAULT_SONNET_MODEL": "claude-sonnet-4.5",
+    "ANTHROPIC_DEFAULT_HAIKU_MODEL": "claude-haiku-4.5",
+    "ANTHROPIC_MODEL": "claude-sonnet-4.5",
+    "ANTHROPIC_SMALL_FAST_MODEL": "claude-haiku-4.5",
```
tallaxes: Based on docs, `ANTHROPIC_SMALL_FAST_MODEL` is deprecated, so maybe not needed? https://anthropic.mintlify.app/en/docs/claude-code/model-config#environment-variables

(Also not sure if `ANTRHOPIC_MODEL` is required, when all other defaults are available ...)

tallaxes (dismissed):

.claude/settings.json
matthchr: Ahh ok I can remove this one. I didn't actually check each individual one to see what fixed the issues I was having (and that the container seems to have now in codespaces). I _think_ it was the AUTH_TOKEN being moved to `sk-` but not 100% sure.

tallaxes (approved):

Issue #1247, opened 2025-11-02
Ephemeral OS disk placement config so NVMe isn’t always consumed


Pothulapati:
### Tell us about your request

On the v6 Azure VM sizes (e.g. D8ds_v6) Karpenter always places the Ephemeral OS disk on the local NVMe. That eats the entire device, so workloads that relied on the NVMe temp disk suddenly have nothing to mount. Azure already exposes a diffDiskPlacement option (CacheDisk, ResourceDisk, NVMe), but the AKSNodeClass spec doesn’t surface it. 

Could you add a field so we can pick Cache/Resource placement when the SKU supports it? 

Docs: https://learn.microsoft.com/en-us/azure/virtual-machines/ephemeral-os-disks

### Tell us about the problem you're trying to solve. What are you trying to do, and why is it hard?

to use local NVMe for application workloads and not as an ephermal os disk

### Are you currently working around this issue?

One option is to set `osDiskSize` to be bigger than the ephermal disk size, so that a managed disk is provisioned but it comes at a huge cost

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

alex-hunt-materialize:
I think this should maybe be split into two issues:
1. Support using managed disks instead of ephemeral disks, even when ephemeral disks are present and large enough.
2. Support specifying which ephemeral disk placement to use when an instance type supports multiple placements.

I suspect either of these options would likely solve @Pothulapati 's issue.

I have a draft PR that implements option 1.
https://github.com/Azure/karpenter-provider-azure/pull/1339

matthchr:
We discussed this in our standup and one challenge here is that we're somewhat subject to what AKS is doing with this field. AFAIK, AKS doesn't allow you to choose a preference, though there has been discussions of that I have heard so it may be coming. We don't want to support this in Karpenter until AKS itself agrees it's something that makes sense to do. That discussion is ongoing but not done yet.

alex-hunt-materialize:
[...]

PR #1256, opened 2025-11-19
Support for OS Disk Type in AKSNodeClass

### LINKED ISSUE #801: OS Disk Type add to AKSNodeClass

Please try to add OS Disk type to be configurable.
The default is Premium SSD (Premium_LRS), so we can configure the disk type on our own.

Please try to add OS Disk type to be configurable.
The default is Premium SSD (Premium_LRS), so we can configure the disk type on our own.

No

### Additional Context

_No response_

### Attachments

_No response_

### Community Note

nandorKollar:
Fixes #801

**Description**

This change adds a new option to AKSNodeClass to specify the required OS disk type.

- Ran `make az-all`, created an AKS with non-default OS disk type, which worked fine.
- Added new unit test to verify both validation of this new option as well as verify that the value is properly used when provisioning a VM.

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
New feature: support for OS Disk Type in AKSNodeClass
```

### DIFF
--- a/pkg/apis/crds/karpenter.azure.com_aksnodeclasses.yaml
+++ b/pkg/apis/crds/karpenter.azure.com_aksnodeclasses.yaml
@@ -539,6 +539,17 @@ spec:
                 maximum: 2048
                 minimum: 30
                 type: integer
+              osDiskType:
+                default: Premium_LRS
+                description: osDiskType is the type of the OS disk. If not specified,
+                  defaulted to Premium_LRS.
+                enum:
+                - Premium_LRS
+                - Premium_ZRS
+                - Standard_LRS
+                - StandardSSD_LRS
+                - StandardSSD_ZRS
+                type: string
               security:
                 description: Collection of security related karpenter fields
                 properties:
--- a/pkg/apis/v1beta1/aksnodeclass.go
+++ b/pkg/apis/v1beta1/aksnodeclass.go
@@ -45,6 +45,10 @@ type AKSNodeClassSpec struct {
 	// +kubebuilder:validation:Maximum=2048
 	// osDiskSizeGB is the size of the OS disk in GB.
 	OSDiskSizeGB *int32 `json:"osDiskSizeGB,omitempty"`
+	// osDiskType is the type of the OS disk. If not specified, defaulted to Premium_LRS.
+	// +kubebuilder:default=Premium_LRS
+	// +kubebuilder:validation:Enum:={Premium_LRS,Premium_ZRS,Standard_LRS,StandardSSD_LRS,StandardSSD_ZRS}
+	OSDiskType *string `json:"osDiskType,omitempty"`
 	// ImageID is the ID of the image that instances use.
 	// Not exposed in the API yet
 	ImageID *string `json:"-"`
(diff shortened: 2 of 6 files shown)


alex-hunt-materialize:
@nandorKollar I think there might be some conflicts between this PR and my PR https://github.com/Azure/karpenter-provider-azure/pull/1339, just because we both want to use the same field name in the AKSNodeClass. I think mine is less specific, and therefore would make more sense with the `osDiskType` field name.

Can I ask you to change your field name to something more specific? Maybe something like `osManagedDiskStorageAccountType`? That way, if someone ever implements the third related task of specifying which ephemeral disk type (NVMe, Cache, Resource), we could call that field `osEphemeralDiskPlacementType` and we'd have nice consistency in the naming. In that case, my `osDiskType` field would specify which of the other fields matters.

In my proposal:
1. `osDiskType` would be for specifying whether the OS disk should be `Managed` or `EphemeralWithFallbackToManaged`.
2. `osManagedDiskStorageAccountType` would be for the storage account type of the OS disk if using a managed disk (`Premium_LRS`, etc...).
3. `osEphemeralDiskPlacementType` would be for the placement type of the OS disk if ephemeral (`NVMe`, `Cache`, `Resource`).

No one has offered to implement the third thing, but there is [a ticket open for it](https://github.com/Azure/karpenter-provider-azure/issues/1247), which my PR only partially addresses.

nandorKollar:
> @nandorKollar I think there might be some conflicts between this PR and my PR #1339, just because we both want to use the same field name in the AKSNodeClass. I think mine is less specific, and therefore would make more sense with the `osDiskType` field name.
> 
> Can I ask you to change your field name to something more specific? Maybe something like `osManagedDiskStorageAccountType`? That way, if someone ever implements the third related task of specifying which ephemeral disk type (NVMe, Cache, Resource), we could call that field `osEphemeralDiskPlacementType` and we'd have nice consistency in the naming. In that case, my `osDiskType` field would specify which of the other fields matters.
> 
> In my proposal:
> 
> 1. `osDiskType` would be for specifying whether the OS disk should be `Managed` or `EphemeralWithFallbackToManaged`.
> 2. `osManagedDiskStorageAccountType` would be for the storage account type of the OS disk if using a managed disk (`Premium_LRS`, etc...).
> 3. `osEphemeralDiskPlacementType` would be for the placement type of the OS disk if ephemeral (`NVMe`, `Cache`, `Resource`).
> 
> No one has offered to implement the third thing, but there is [a ticket open for it](https://github.com/Azure/karpenter-provider-azure/issues/1247), which my PR only partially addresses.

Thanks @alex-hunt-materialize for noticing this name collision. Sure, I can rename the field. It seems that indeed we need two different configuration parameter: one is for selecting between Ephemeral (if available)/Managed, that's what your PR does, and maybe one to select a different manage instance type (this is `osManagedDiskStorageAccountType`), that I try to address here, and probably we need a followup PR for `osEphemeralDiskPlacementType`? Actually this approach diverges from AKS a bit, as far as I know there's no way to select a managed disk type, every node is created with PremiumSSD (at least that's what I recall), therefore I'm curious about the opinion of the maintainers of this project.

PR #1270, opened 2025-11-25
chore: additional labels refactoring

matthchr:

* Unit tests
* E2E tests: https://github.com/Azure/karpenter-provider-azure/actions/runs/19711035137

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note
NONE
```

### DIFF
--- a/pkg/providers/instancetype/suite_test.go
+++ b/pkg/providers/instancetype/suite_test.go
@@ -1607,7 +1607,7 @@ var _ = Describe("InstanceType Provider", func() {
 					for range 3 {
 						bindings = append(bindings, ExpectProvisionedNoBinding(ctx, env.Client, clusterBootstrap, cloudProviderBootstrap, coreProvisionerBootstrap, pod))
 					}
-					for i := 1; i < len(bindings); i++ {
+					for i := range len(bindings) {
 						Expect(lo.Values(bindings[i])[0].Node.Name).To(Equal(lo.Values(bindings[0])[0].Node.Name), "expected all bindings to have the same node name")
 					}
 				}
@@ -1645,7 +1645,7 @@ var _ = Describe("InstanceType Provider", func() {
 					for range 3 {
 						bindings = append(bindings, ExpectProvisionedNoBinding(ctx, env.Client, clusterBootstrap, cloudProviderBootstrap, coreProvisionerBootstrap, pod))
 					}
-					for i := 1; i < len(bindings); i++ {
+					for i := range len(bindings) {
 						Expect(lo.Values(bindings[i])[0].Node.Name).To(Equal(lo.Values(bindings[0])[0].Node.Name), "expected all bindings to have the same node name")
 					}
 				}


pkg/providers/labels/labels.go
```diff
@@ -137,6 +143,10 @@ func IsKubeletLabel(key string) bool {
 	}
 
 	namespace := getLabelNamespace(key)
+	if !isKubernetesLabel(namespace) {
+		return true
+	}
```
tallaxes: This addition warrants a comment, I think (and also invalidates the comment above, as it is not longer the same as IsKubeletLabel in kubelet)

pkg/providers/labels/labels.go
```diff
@@ -146,6 +156,48 @@ func IsKubeletLabel(key string) bool {
...
+	for key, req := range requirements {
+		if req.Len() == 1 && predicate(key, req) {
+			labels[key] = req.Values()[0]
+		}
+	}
+	return labels
+}
+
+func isKubernetesLabel(namespace string) bool {
```
tallaxes: Maybe something like `isKubernetesNamespace`?

pkg/providers/labels/labels_test.go
```diff
@@ -0,0 +1,235 @@
...
+	}{
+		{
+			name:           "Nil instance type",
+			requirements:   nil,
+			expectedLabels: map[string]string{},
+		},
+		{
+			name: "Single-valued requirements",
+			requirements: scheduling.NewRequirements(
+				scheduling.NewRequirement("node.kubernetes.io/instance-type", corev1.NodeSelectorOpIn, "Standard_D2s_v3"),
```
tallaxes:
```suggestion
				scheduling.NewRequirement(corev1.LabelInstanceTypeStable, corev1.NodeSelectorOpIn, "Standard_D2s_v3"),
```
(here and elsewhere)

pkg/providers/instancetype/suite_test.go
```diff
@@ -1597,6 +1600,17 @@ var _ = Describe("InstanceType Provider", func() {
 				value := item.ValueFunc()
 
 				pod := coretest.UnschedulablePod(coretest.PodOptions{NodeSelector: map[string]string{item.Label: value}})
+				// Simulate multiple scheduling passes before final binding, this ensures that when real scheduling happens we won't
+				// end up with a new node for each scheduling attempt
+				if item.Label != v1.LabelWindowsBuild { // TODO: special case right now as we don't support it
+					bindings := []Bindings{}
+					for i := 0; i < 3; i++ {
```
tallaxes: Could now use `range`? (here and elsewhere)

```suggestion
					for range 3 {
```

pkg/providers/instancetype/suite_test.go
```diff
@@ -1597,6 +1600,17 @@ var _ = Describe("InstanceType Provider", func() {
 				value := item.ValueFunc()
 
 				pod := coretest.UnschedulablePod(coretest.PodOptions{NodeSelector: map[string]string{item.Label: value}})
+				// Simulate multiple scheduling passes before final binding, this ensures that when real scheduling happens we won't
+				// end up with a new node for each scheduling attempt
```
tallaxes, line 1604: This looks like it would work ... I presume you can make this fail under "wrong" circumstances?

(FWIW, there are a number of tests for in-flight nodes in core that _might_ be helpful, like [this one](https://github.com/kubernetes-sigs/karpenter/blob/34398e54f24b281619bf39e728480b422938e8b3/pkg/controllers/provisioning/scheduling/suite_test.go#L2412) - and others in that file, search for "in-flight")

pkg/cloudprovider/cloudprovider.go
```diff
@@ -169,6 +169,11 @@ func (c *CloudProvider) createVMInstance(ctx context.Context, nodeClass *v1beta1
 	if err != nil {
 		return nil, err
 	}
+	// Propagate single-value wellKnownLabels from the nodeClaim requirements to the labels,
+	// as this is required for scheduling in core to work correctly. If this is not done, on subsequent scheduling passes before the Node is
+	// registered, the NodeClaim will not have the labels required to match the Pod and so a new NodeClaim will be created each time.
+	newNodeClaim.Labels = lo.Assign(newNodeClaim.Labels, labelspkg.GetWellKnownSingleValuedRequirementLabels(scheduling.NewNodeSelectorRequirementsWithMinValues(nodeClaim.Spec.Requirements...)))
```
tallaxes: Do we understand why AWS provider gets away with not doing this?

tallaxes (approved):

pkg/providers/labels/labels.go
matthchr: It is and isn't: If you look at how `IsKubeletLabel` is actually used, it has this check applied with it:
https://github.com/kubernetes/kubernetes/blob/master/cmd/kubelet/app/options/options.go#L152.

I've updated the comments to match

pkg/providers/labels/labels_test.go
matthchr: Done

pkg/providers/instancetype/suite_test.go
matthchr: hah good catch!

pkg/providers/labels/labels_test.go
matthchr: Also done in `pkg/providers/instancetype/suite_test.go`

pkg/providers/instancetype/suite_test.go
matthchr: Yes, I was able to reproduce the issue I was trying to reproduce with this pattern. Once I fixed, it fixed it here and also in the E2E test. This was really just a way to get the failure in a unit test rather than an E2E test since it was a bit easier to debug there.

It looks like they use a similar trick there (call `ExpectProvisionedNoBinding` a few times)

pkg/cloudprovider/cloudprovider.go
matthchr: It looks like they can derive all of their labels from the instance type + instance, so they get them from there rather than doing this. We could do the same but we might have to smuggle a few of the ones I'm planning to add that are AKS concepts but not CRP concepts through tags (we already do this with `NodePoolLabelKey`). Doing that has the advantage of ensuring that the NodeClaim we return from `List` and `Get` match labels with what's actually there in Kubernetes. OTOH, that's not actually required because Karpenter isn't using those NodeClaims to update the Kuberentes definition, they're just using it to do GC and so don't actually care about the labels on the resulting NodeClaims.

For an example of what I mean for AWS, take a look at [computeRequirements](https://github.com/aws/karpenter-provider-aws/blob/dee70a565f7da41dbef23642a53666c7fd982ba2/pkg/providers/instancetype/types.go#L158) and trace the labels to [their CloudProvider](https://github.com/aws/karpenter-provider-aws/blob/main/pkg/cloudprovider/cloudprovider.go#L456), for example:

https://github.com/aws/karpenter-provider-aws/blob/dee70a565f7da41dbef23642a53666c7fd982ba2/pkg/cloudprovider/cloudprovider.go#L456

Note that they also smuggle the `NodePoolLabelKey` through tags.

matthchr: I could also smuggle these other items through tags but I'm not sure it's needed, WDYT?

Either way, I'll update the comment here a bit to make this more clear.

tallaxes: No, I think the way you are doing it now is fine. Good to capture the rationale, this helps! (And makes it easier to revisit if we need to)

pkg/providers/instancetype/suite_test.go
```diff
@@ -1597,6 +1600,17 @@ var _ = Describe("InstanceType Provider", func() {
...
+				if item.Label != v1.LabelWindowsBuild { // TODO: special case right now as we don't support it
+					bindings := []Bindings{}
+					for range 3 {
+						bindings = append(bindings, ExpectProvisionedNoBinding(ctx, env.Client, clusterBootstrap, cloudProviderBootstrap, coreProvisionerBootstrap, pod))
+					}
+					for i := 1; i < len(bindings); i++ {
```
tallaxes: This can also use range? I expect this would work, right?
```suggestion
					for i := range len(bindings) {
```

tallaxes (approved):

PR #1271, opened 2025-11-26
feat: add support for kubernetes.azure.com/mode label

matthchr:
**Description**

Add scheduling support for the `kubernetes.azure.com/mode` label

* Unit test
* E2E test: https://github.com/Azure/karpenter-provider-azure/actions/runs/19914204524

- [x] Yes, existing issue tracking required doc updates already.
- [ ] No

**Release Note**

```release-note
Add scheduling support for the `kubernetes.azure.com/mode` label
```

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -21,6 +21,10 @@ updates:
         exclude-patterns:
           - "k8s.io/*"
           - "sigs.k8s.io/*"
+    ignore:
+      - dependency-name: "github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/containerservice/armcontainerservice"
+        versions:
+          - ">=v8.0.0" # AKS Machine API is externally in preview. Need to preserve beta versions of the module.
   - package-ecosystem: docker
     directories:
       - /.devcontainer/**/*


pkg/providers/imagefamily/customscriptsbootstrap/provisionclientbootstrap.go
```diff
@@ -180,7 +180,7 @@ func (p *ProvisionClientBootstrap) ConstructProvisionValues(ctx context.Context)
 		}
 	}
 
-	if modeString, ok := p.Labels["kubernetes.azure.com/mode"]; ok && modeString == "system" {
+	if modeString, ok := p.Labels[v1beta1.AKSLabelMode]; ok && modeString == "system" {
```
matthchr: Fix compare

tallaxes (dismissed):

tallaxes (approved):

matthchr:
Worth noting that it's a bit "weird" (AKS weird, not Karpenter weird) that if you create a NodePool without the `kubernetes.azure.com/mode` label, that NodePool can provision system or user nodes. I don't think we should necessarily recommend this to users as it can be a bit surprising. The issue is if we _dont_ put `mode` into WellKnownLabels and include it in `computeRequirements` then if the user doesn't request a specific mode on the NodePool we'll set the mode to `user` but we won't correctly scheduling simulate on it (so workloads that want user will not land there), which also is wrong.

So I do think just making it fully schedulable makes the most sense.

PR #1273, opened 2025-11-26
fix: fully support kuberentes.azure.com/cluster as WellKnownLabel

matthchr:
**Description**
Fixed bug where attempting to schedule against the `kuberentes.azure.com/cluster` label while it was not set on the NodePool would result in a large number of NodeClaims being created. Scheduling against `kuberentes.azure.com/cluster` is now fully supported.

* Unit tests
* E2E (https://github.com/Azure/karpenter-provider-azure/actions/runs/19841620656)

- [x] Yes, existing issue tracking required doc updates already.
- [ ] No

**Release Note**

```release-note
Fixed bug where attempting to schedule against the `kuberentes.azure.com/cluster` label while it was not set on the NodePool would result in a large number of NodeClaims being created. Scheduling against `kuberentes.azure.com/cluster` is now fully supported.
```

### DIFF
--- a/.github/workflows/ci-test.yml
+++ b/.github/workflows/ci-test.yml
@@ -13,7 +13,7 @@ jobs:
       statuses: write
     runs-on: ubuntu-latest
     strategy:
-      fail-fast: false # https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast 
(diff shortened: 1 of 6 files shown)


matthchr:
This isn't going to work because we need to support scheduling on this label (probably why it was in WellKnownLabels to begin with). That means we probably need to go the other direction and just fully support scheduling on this label. I will update this PR to do that.

PR #1296, opened 2025-12-05
pablotrivino/skaffold-fix

PabloTriv:
Fixes # 

**Description**
Make commands kept hitting problems about skaffold-overrides.yaml when trying to do skaffold run or skaffold dev, like in the picture below. This seems to fix it. 
image

Tested commands on my own cluster with make az-all and make az-run
*

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note

```

### DIFF
--- a/skaffold.yaml
+++ b/skaffold.yaml
@@ -24,21 +24,6 @@ manifests:
           controller.image.digest: "{{.IMAGE_DIGEST_controller}}"
         valuesFiles:
           - karpenter-values.yaml
-        overrides:
-          controller:
-            resources:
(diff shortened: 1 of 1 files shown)


tallaxes:
I have not observed this, so wondering what might be the difference ... Which version of skaffold do you have? Are you using codespace? (I don't have anything strong against the fix - though it is nice to keep this in a single file ...)

PabloTriv:
You're right, this only happens locally. Codespaces does not run into this. I'll close this PR.

PR #1233, opened 2025-10-24
[CRD change] Add LocalDNSProfile to AKSNodeClassSpec + NAP Integration Tests

saewoni:
Fixes # 

**Description**
We need to expose the localdns configuration in the CRD for Karpenter. This will allow NAP to leverage the same localdns currently used for node pools in AKS-RP.

**How was this change tested?**

Integration Test + Unit Tests

Integration Test Results
- [integration-test-12-9.txt](https://github.com/user-attachments/files/24065884/integration-test-12-9.txt)

- [example_aksnodeclass_integrationtest.txt](https://github.com/user-attachments/files/24042062/example_aksnodeclass_integrationtest.txt) - aksnodeclass config from integration test (the custom resource (CR) the cx needs to provide to enable localdns on NAP path) 

Need to learn how to do test like this PR: https://github.com/Azure/karpenter-provider-azure/pull/688
**Does this change impact docs?**
- [ ] Yes, PR includes docs updates
- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
New AKSNodeClass.Spec field (LocalDNSProfile)
```

PR #1301, opened 2025-12-10
chore: CEL validation for local DNS

tallaxes:
**Description**

Follow-up to #1233 - see https://github.com/Azure/karpenter-provider-azure/pull/1233#discussion_r2601652044 - re-introduces CEL validation that was [removed](https://github.com/Azure/karpenter-provider-azure/compare/21c7ab0..6beb899#diff-6ff166ad79fde51cb3cc3456223454a31df7515b758d0730d6f6a35e95ae3518), with some modifications.

> It looks like all of the current validation can be done in CEL, without limiting the list length (though IMO a reasonable limit should be there in any case) - except for uniqueness checking. Uniqueness in CEL (and elsewhere) is expensive, works in CEL only if we limit maxItems to 100-250 (each list). So we could leave it to runtime - though I am not even sure if listType=map will ever observe non-unique entries at runtime.

This PR includes the uniqueness validation and maxItems=100 in CEL (possibly making runtime validation of localDNS unnecessary); both could be removed if decided.

TODO:
* [ ] Review the added rules for correctness, as they were refactored from the original, and some more constraints were added or copied from runtime validation
* [x] Adding the above validation breaks some existing tests, because they don't always use a valid localDns ("." and "cluster.local" are not always there), these need to be fixed. Until this is fixed, expect all CI tests to fail - but note they are not failing early on CRD deployment due to CEL budget.
* [x] Check on whether listType=map enforces/validates uniqueness by itself, or just merges the entries with the same key (more likely), and whether non-unique key would ever be observed by the controller (pre-requisite for runtime validation) @tallaxes - turns out listType=map enforces uniqueness, which means no other enforcement is necessary (and also means non-unique key won't ever be observed by controller)
* [x] Decide on whether we will keep uniqueness check in CEL (with maxItems) or drop it (either altogether, or in favor of runtime validation) - from this [thread](https://github.com/Azure/karpenter-provider-azure/pull/1301#discussion_r2604984730), decided that the maxItems=100 for kubeDNSOverrides and vnetDNSOverrides is acceptable. It makes sense to keep a sensible limit even though the CEL uniqueness check that required it is no longer needed.
* [X] Cleanup (or remove) runtime validation based on the above. (Though even if it proves to be unnecessary for localDNS, probably good to keep the skeleton for reference, we will definitely need it for other things ...) - keeping validation.go as a skeleton for future validations that can't be done in CEL.

* `make presubmit`
* [integration-test-1212.txt](https://github.com/user-attachments/files/24132882/integration-test-1212.txt)

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/apis/v1alpha2/crd_validation_cel_test.go
+++ b/pkg/apis/v1alpha2/crd_validation_cel_test.go
@@ -32,19 +32,18 @@ import (
 	"sigs.k8s.io/karpenter/pkg/test"
 )
 
+var _ = AfterEach(func() {
+	// Clean up all AKSNodeClasses created during tests
+	nodeClassList := &v1alpha2.AKSNodeClassList{}
+	if err := env.Client.List(ctx, nodeClassList); err == nil {
+		for i := range nodeClassList.Items {
+			_ = env.Client.Delete(ctx, &nodeClassList.Items[i])
+		}
+	}
+})
+
 var _ = Describe("CEL/Validation", func() {
 	var nodePool *karpv1.NodePool
-	var createdNodeClasses []string
-
-	AfterEach(func() {
-		// Clean up only the AKSNodeClasses we created in this test
-		for _, name := range createdNodeClasses {
-			nodeClass := &v1alpha2.AKSNodeClass{}
-			nodeClass.Name = name
-			_ = env.Client.Delete(ctx, nodeClass)
-		}
-		createdNodeClasses = nil
-	})
 
 	// Helper function to create a complete LocalDNSZoneOverride with all required fields
 	// Use forwardToVnetDNS=true for root zone "." in vnetDNSOverrides
@@ -66,13 +65,7 @@ var _ = Describe("CEL/Validation", func() {
 		}
 	}
 
-	// Helper function to track nodeClass creation
-	trackNodeClass := func(nodeClass *v1alpha2.AKSNodeClass) {
-		createdNodeClasses = append(createdNodeClasses, nodeClass.Name)
-	}
-
 	BeforeEach(func() {
-		createdNodeClasses = nil
 		if env.Version.Minor() < 25 {
 			Skip("CEL Validation is for 1.25>")
 		}
@@ -109,7 +102,6 @@ var _ = Describe("CEL/Validation", func() {
 			}
 			if expected {
 				Expect(env.Client.Create(ctx, nodeClass)).To(Succeed())
-				trackNodeClass(nodeClass)
 			} else {
 				Expect(env.Client.Create(ctx, nodeClass)).ToNot(Succeed())
 			}
@@ -169,7 +161,6 @@ var _ = Describe("CEL/Validation", func() {
 				},
 			}
 			Expect(env.Client.Create(ctx, nodeClass)).To(Succeed())
-			trackNodeClass(nodeClass)
 		})
 
 		It("should accept complete LocalDNS configuration with all required fields", func() {
@@ -190,7 +181,6 @@ var _ = Describe("CEL/Validation", func() {
 				},
 			}
 			Expect(env.Client.Create(ctx, nodeClass)).To(Succeed())
-			trackNodeClass(nodeClass)
 		})
 
 		DescribeTable("should validate LocalDNSMode", func(mode v1alpha2.LocalDNSMode, expectedErr string) {
@@ -207,7 +197,6 @@ var _ = Describe("CEL/Validation", func() {
 			err := env.Client.Create(ctx, nodeClass)
 			if expectedErr == "" {
 				Expect(err).To(Succeed())
-				trackNodeClass(nodeClass)
 			} else {
 				Expect(err).To(HaveOccurred())
 				Expect(err.Error()).To(ContainSubstring(expectedErr))
@@ -243,7 +232,6 @@ var _ = Describe("CEL/Validation", func() {
 			err := env.Client.Create(ctx, nodeClass)
 			if expectedErr == "" {
 				Expect(err).To(Succeed())
-				trackNodeClass(nodeClass)
 			} else {
 				Expect(err).To(HaveOccurred())
 				Expect(err.Error()).To(ContainSubstring(expectedErr))
@@ -275,7 +263,6 @@ var _ = Describe("CEL/Validation", func() {
 			err := env.Client.Create(ctx, nodeClass)
 			if expectedErr == "" {
 				Expect(err).To(Succeed())
-				trackNodeClass(nodeClass)
 			} else {
 				Expect(err).To(HaveOccurred())
 				Expect(err.Error()).To(ContainSubstring(expectedErr))
@@ -303,7 +290,6 @@ var _ = Describe("CEL/Validation", func() {
 			err := env.Client.Create(ctx, nodeClass)
 			if expectedErr == "" {
 				Expect(err).To(Succeed())
-				trackNodeClass(nodeClass)
 			} else {
 				Expect(err).To(HaveOccurred())
 				Expect(err.Error()).To(ContainSubstring(expectedErr))
@@ -331,7 +317,6 @@ var _ = Describe("CEL/Validation", func() {
 			err := env.Client.Create(ctx, nodeClass)
 			if expectedErr == "" {
 				Expect(err).To(Succeed())
-				trackNodeClass(nodeClass)
 			} else {
 				Expect(err).To(HaveOccurred())
 				Expect(err.Error()).To(ContainSubstring(expectedErr))
@@ -360,7 +345,6 @@ var _ = Describe("CEL/Validation", func() {
 			err := env.Client.Create(ctx, nodeClass)
 			if expectedErr == "" {
 				Expect(err).To(Succeed())
-				trackNodeClass(nodeClass)
 			} else {
 				Expect(err).To(HaveOccurred())
 				Expect(err.Error()).To(ContainSubstring(expectedErr))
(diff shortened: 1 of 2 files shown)


pkg/apis/v1beta1/aksnodeclass.go
```diff
@@ -127,19 +127,30 @@ type LocalDNS struct {
...
 }
 
 // LocalDNSZoneOverride specifies DNS override configuration for a specific zone
+// +kubebuilder:validation:XValidation:message="'cluster.local' cannot be forwarded to VnetDNS",rule="!(self.zone.endsWith('cluster.local') && self.forwardDestination == 'VnetDNS')"
+// +kubebuilder:validation:XValidation:message="serveStale Verify cannot be used with protocol ForceTCP",rule="!(self.serveStale == 'Verify' && self.protocol == 'ForceTCP')"
 type LocalDNSZoneOverride struct {
 	// Zone is the DNS zone this override applies to (e.g., ".", "cluster.local").
 	// +required
 	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=254
```
tallaxes, line 150: I added MaxLength constraint, need to verify this is correct

pkg/apis/v1alpha2/aksnodeclass.go
```diff
@@ -127,19 +127,30 @@ type LocalDNS struct {
...
 	// +listMapKey=zone
+	// +kubebuilder:validation:XValidation:message="must contain required zones '.' and 'cluster.local'",rule="['.', 'cluster.local'].all(z, self.exists(x, x.zone == z))"
+	// +kubebuilder:validation:XValidation:message="must not contain duplicate zones",rule="self.all(x, self.exists_one(y, x.zone == y.zone))"
+	// +kubebuilder:validation:MaxItems=100
 	KubeDNSOverrides []LocalDNSZoneOverride `json:"kubeDNSOverrides,omitempty"`
 }
 
 // LocalDNSZoneOverride specifies DNS override configuration for a specific zone
+// +kubebuilder:validation:XValidation:message="'cluster.local' cannot be forwarded to VnetDNS",rule="!(self.zone.endsWith('cluster.local') && self.forwardDestination == 'VnetDNS')"
+// +kubebuilder:validation:XValidation:message="serveStale Verify cannot be used with protocol ForceTCP",rule="!(self.serveStale == 'Verify' && self.protocol == 'ForceTCP')"
```
tallaxes, line 145: Additional constraint migrated from runtime validation

pkg/apis/v1beta1/aksnodeclass.go
saewoni: i cross-checked with AI agent & this is the result. i think the `maxLength` can be decreased to 253.  
image

pkg/apis/v1alpha2/aksnodeclass.go
```diff
@@ -127,19 +127,30 @@ type LocalDNS struct {
...
+	// +kubebuilder:validation:XValidation:message="root zone '.' cannot be forwarded to ClusterCoreDNS from vnetDNSOverrides",rule="!self.exists(x, x.zone == '.' && x.forwardDestination == 'ClusterCoreDNS')"
+	// +kubebuilder:validation:MaxItems=100
+	VnetDNSOverrides []LocalDNSZoneOverride `json:"vnetDNSOverrides,omitempty"`
 	// KubeDNS overrides apply to DNS traffic from pods with dnsPolicy:ClusterFirst (referred to as KubeDNS traffic).
 	// +required
 	// +listType=map
 	// +listMapKey=zone
+	// +kubebuilder:validation:XValidation:message="must contain required zones '.' and 'cluster.local'",rule="['.', 'cluster.local'].all(z, self.exists(x, x.zone == z))"
+	// +kubebuilder:validation:XValidation:message="must not contain duplicate zones",rule="self.all(x, self.exists_one(y, x.zone == y.zone))"
+	// +kubebuilder:validation:MaxItems=100
```
saewoni, line 139: @SriHarsha001 / @yewmsft, if we move forward with doing this validation purely in CEL, we’ll need to cap the number of LocalDNSOverrides at 100 for both KubeDNSOverrides and VnetDNSOverrides.
In the earlier PR, we stated that customers would not have a limit on how many LocalDNSOverrides they can define. After reviewing Alex’s follow‑up comment (https://github.com/Azure/karpenter-provider-azure/pull/1233#discussion_r2603801677), should we continue allowing an unlimited number of overrides, or is it acceptable to introduce a limit of 100?

saewoni: alex's followup comment: 
image

saewoni: talked with @SriHarsha001  and @yewmsft  offline and they agreed that 100 limit is acceptable.

pkg/apis/v1beta1/aksnodeclass.go
tallaxes: The regex used currently allows trailing dot, which makes 254 correct, I think. If you want to disallow trailing dot, need to update the regex (and then can change the MaxLength to 253).

saewoni: let's just leave it as 254, in the internal repo, we allow for trailing dot. cc @SriHarsha001 / @yewmsft  for awareness

pkg/apis/v1beta1/crd_validation_cel_test.go
```diff
@@ -431,6 +444,172 @@ var _ = Describe("CEL/Validation", func() {
...
+							KubeDNSOverrides: []v1beta1.LocalDNSZoneOverride{
+								createCompleteLocalDNSZoneOverride(".", false),
+								createCompleteLocalDNSZoneOverride("cluster.local", false),
+							},
+						},
+					},
+				}
+				err := env.Client.Create(ctx, nodeClass)
+				if shouldSucceed {
+					Expect(err).To(Succeed())
```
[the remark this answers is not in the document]
saewoni: added

pkg/apis/v1alpha2/aksnodeclass.go
tallaxes: Note that, since we proved listType=map enforces uniqueness, CEL uniqueness check is no longer needed and consequently maxItems=100 is not even absolutely required - but still a good idea to have a reasonable limit.

pkg/apis/v1alpha2/crd_validation_cel_test.go
```diff
@@ -151,6 +151,7 @@ var _ = Describe("CEL/Validation", func() {
 				},
 			}
 			Expect(env.Client.Create(ctx, nodeClass)).To(Succeed())
+			Expect(env.Client.Delete(ctx, nodeClass)).To(Succeed())
```
tallaxes: nit: Deleting all AKSNodeClass instances in AfterEach instead would be better (and would fix leaks in other tests - however inconsequential they are, due to random names ...)

pkg/apis/v1beta1/crd_validation_cel_test.go
```diff
@@ -151,6 +151,7 @@ var _ = Describe("CEL/Validation", func() {
 				},
 			}
 			Expect(env.Client.Create(ctx, nodeClass)).To(Succeed())
+			Expect(env.Client.Delete(ctx, nodeClass)).To(Succeed())
```
tallaxes: nit: Deleting all AKSNodeClass instances in AfterEach instead would be better (and would fix leaks in other tests - however inconsequential they are, due to random names ...)

saewoni: ack

saewoni: https://github.com/Azure/karpenter-provider-azure/pull/1301/commits/141aba674e87baaef9807ae5e1ca68e123adc57a

saewoni:
> LGTM, thank you! Let's attach evidence of passing E2E tests, and it is good to go. Somebody else will have to approve since I started the PR :-)

the integration test result is here: https://github.com/user-attachments/files/24132882/integration-test-1212.txt

pkg/apis/v1beta1/crd_validation_cel_test.go
tallaxes: Sorry, I did not make myself clear, I was thinking rather a file-global AfterEach, that would list all AKSNodeClasses in cluster and delete them. This way you don't need to rely on specific global variable and do any kind of cleanup in each step. (We should really add provider-specific general AfterEach cleanup, but that can wait ...)

saewoni: this is my attempt of making a file-global AfterEach https://github.com/Azure/karpenter-provider-azure/pull/1301/commits/4d98c1f9df1bddde6e01285fca346e271224433e let me know if this is what you thought

pkg/apis/v1beta1/crd_validation_cel_test.go
```diff
@@ -675,3 +845,13 @@ var _ = Describe("CEL/Validation", func() {
 		})
 	})
 })
+
+var _ = AfterEach(func() {
+	// Clean up all AKSNodeClasses created during LocalDNS tests
```
tallaxes: nit: it is customary to place global After* handlers in front of the file. Also, this will cleanup _all_ AKSNodeCLasses (not just those created during LocalDNS tests) - which is fine and desired

saewoni: does this mean it'll delete the AKSNodeClasses from the customers as well?

saewoni: moved the AfterEach block to the front of the file, https://github.com/Azure/karpenter-provider-azure/pull/1301/commits/e4eb8d2cbc81360d917de79e97f52ec836db9edb

tallaxes: > does this mean it'll delete the AKSNodeClasses from the customers as well?

No, these unit tests use envtest clusters, spinning up temporary kube-apiserver and etcd just for the test. There is nothing in these except for what tests create. It is also not _critical_ to cleanup here (these resource use random names, so there is no conflict), but it is a good practice not to leak resources in tests.

comtalyst (approved):

Issue #1260, opened 2025-11-20
New NAP nodes don't have parallel image pull configured at the kubelet level


juanmaia:
### Version

__Karpenter Version:__ v0.0.0 (command doesn't work for me) but probably latest.

__Kubernetes Version:__ v1.32.3

### Expected Behavior

Images should be pulled in parallel in nodes created by Karpenter. Basically no waiting time should be seen for image pulling.

### Actual Behavior

Images are being pulled in series in nodes created by Karpenter, and in parallel for nodes created manually or by cluster autoscaler (VMSS).

When pods in a new node is pulling images, I'm seeing this message often:
```Successfully pulled image "foo:bar" in 2.786s (10m16.315s including waiting). Image size: 40538901 bytes.```

The `waiting` part made me realize that parallel image pulling is not set by default. When inspecting the kubelet config for nodes that Karpenter created, I don't see the config:
```"serializeImagePulls"```

And while inspecting nodes from regular node pools (VMSS) I do see that it's set to `false`:
```
 "serializeImagePulls": false
```

### Steps to Reproduce the Problem

In a new node created by Karpenter, check in the kubelet config if it has:
```
 "serializeImagePulls": false
```

### Resource Specs and Logs

I'm using NAP, so no access to logs.

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

comtalyst:
The code change for this is in a different (AKS) codebase. Expecting it to roll out in a week or two.
Update 03/14/2026: rollout haven't start yet.

comtalyst:
Rollout started. Could take up to 2 weeks to reach all regions.

PR #1311, opened 2025-12-19
feat: Add serializeImagePulls kubelet config parameter

nandorKollar:
Fixes # 

**Description**

Enable parallel image pulls by allowing to modify `serializeImagePulls` (or in our case `--serialize-image-pulls` flag) config parameter. The default value is unchanged, which means that by default image pulls are serial, but this PR allows to change it via config at least for an AKS node class. This partially addresses #1260 , although image pulls are still serial, but there's a way to change it for a given AKS node class.

Ran make test, and created a cluster and verified that I can change this config on the node class.

*

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note
Add serializeImagePulls option kubelet config parameter. If not explicitly set, then default value is true, hence image pulls are serial (this matches with current kubelet default value).
```


Homulvas:
Should this default to `true` instead? Since k8s 1.31 Parallel image pulls should be the default in AKS. This is currently not the case if NAP is used https://learn.microsoft.com/en-us/troubleshoot/azure/azure-kubernetes/availability-performance/container-image-pull-performance#serialized-and-parallel-image-pulls

nandorKollar:
> Should this default to `true` instead? Since k8s 1.31 Parallel image pulls should be the default in AKS. This is currently not the case if NAP is used https://learn.microsoft.com/en-us/troubleshoot/azure/azure-kubernetes/availability-performance/container-image-pull-performance#serialized-and-parallel-image-pulls

I personally don't mind setting the default to true, probably makes more sense (as AFAIK AKS is also puling images in parallel). @tallaxes what's your opinion? Would it make sense to expose this Kubelet parameter in AKSNodeClass? Should the default be aligned with AKS default: parallel image pulls?

tallaxes:
Ideally this would match defaulting logic in AKS - which means it won't have CRD-level default (unless we make it an enum with additional value of "auto"). 

On NAP side, lack of proper defaulting is a bug that we are looking into. Would fixing NAP help, or is there an important use case for making this work on self-hosted?

nandorKollar:
> Ideally this would match defaulting logic in AKS - which means it won't have CRD-level default (unless we make it an enum with additional value of "auto").
> 
> On NAP side, lack of proper defaulting is a bug that we are looking into. Would fixing NAP help, or is there an important use case for making this work on self-hosted?

I'm also fine with defaulting this kubelet config to parallel image pulls, and offering no CRD level option to change it. Is NAP different regarding configs and features like this? I expected that most of the configs and features are supposed to be the same in both NAP and self-hosted version, or at least there're just minimal difference between the supported features and config options.

tallaxes:
> Is NAP different regarding configs and features like this? I expected that most of the configs and features are supposed to be the same in both NAP and self-hosted version, or at least there're just minimal difference between the supported features and config options.

We are making an effort to keep differences to a minimum, but with NAP delegating some of the node bootstrapping configuration (currently CSE) to AKS, there are more ways to lean into existing defaulting or configuration logic, without having to bring it over into Karpenter provider codebase. In the future, self-hosted may be able to benefit from the same mechanisms, but for now we need to manage these differences explicitly.

alexkivars:
Any news on getting this released? Currently trying to get this config in self hosted Karpenter on AKS.

PR #1318, opened 2025-12-25
fix: add ZRS PV multi-zone scheduling support and test

tallaxes:
> **Note:** This temporarily references unreleased `sigs.k8s.io/karpenter` commit that includes includes [kubernetes-sigs/karpenter#2743](https://github.com/kubernetes-sigs/karpenter/pull/2743) fix - to be updated to the next release version when available

**Description**

Add an E2E test that validates Karpenter correctly schedules pods with ZRS (Zone Redundant Storage) persistent volumes when the pod targets a specific zone.

This test validates the fix in kubernetes-sigs/karpenter#2743, which ensures Karpenter considers all PV node affinity terms during scheduling, not just the first one. ZRS disks are accessible across zones in a region, so the PV node affinity includes multiple zone terms. Without the fix, a pod targeting a secondary zone could fail if only the first zone term was considered.

**Changes:**
- Add a ZRS PV multi-zone scheduling test using an immediate-binding `Premium_ZRS` Azure Disk `StorageClass`.
- Add `EventuallyExpectPVCBound()` to wait for dynamic provisioning and return the bound PV.
- Pin the test workload to the secondary ARM zone (`2`) so the test remains deterministic while still validating multi-zone PV scheduling.
- Assert the Deployment-backed workload becomes healthy and lands on a node in the requested zone.
- Temporarily replace `sigs.k8s.io/karpenter` with the upstream commit containing kubernetes-sigs/karpenter#2743 until the fix is available in a release.

- Integration test passed locally in westus3 region (Node created in zone 2 as expected, pod became healthy)
- E2E: 
  - https://github.com/Azure/karpenter-provider-azure/actions/runs/25981150131 ✅
  - [E2EMatrixTrigger · Azure/karpenter-provider-azure@cb05870](https://github.com/Azure/karpenter-provider-azure/actions/runs/26059918914) - everything except Consolidation passes

- No

**Release Note**

```release-note
```


go.mod
```diff
@@ -61,6 +61,8 @@ require (
 	sigs.k8s.io/karpenter v1.6.2
 )
 
+replace sigs.k8s.io/karpenter => github.com/tallaxes/karpenter-core v0.0.0-20251221063144-7863910ee631
```
matthchr: Dont forget to remove before merge.

test/suites/storage/suite_test.go
```diff
@@ -224,6 +224,47 @@ var _ = Describe("Persistent Volumes", func() {
...
+			})
+
+			env.ExpectCreated(nodeClass, nodePool, zrsStorageClass, pvc)
+			env.EventuallyExpectPVCBound(pvc)
+
+			// ZRS disks are available in all zones of the region. To test that Karpenter
+			// considers ALL PV node affinity terms (not just the first), we intentionally
+			// schedule the pod in zone 2. Without the sigs.k8s.io/karpenter fix, this would fail
+			// because the scheduler only looked at the first zone.
+			zone2 := utils.MakeAKSLabelZoneFromARMZone(env.Region, zones[1])
```
matthchr: Is it guaranteed that zones is always either len 0 or len 2? It can't be len 1?

If it's length 1, we'd get here and panic

tallaxes: I don't think we would ever observe  1, but strengthening the guard condition above just in case

go.mod
```diff
@@ -61,6 +61,8 @@ require (
 	sigs.k8s.io/karpenter v1.12.0
 )
 
+replace sigs.k8s.io/karpenter => github.com/kubernetes-sigs/karpenter v1.12.1-0.20260514173230-7e9d426919bd
```
tallaxes, line 64: To be removed prior to merging / after sig.k8s.io/karpenter update (unless we decide to release early with this pin)

matthchr (approved):
Approved: Assuming we aren't actually removing the replace as we may want to release

go.mod
tallaxes: Decided to release early with pin

PR #1321, opened 2025-12-26
test(e2e): extract and extend storage tests into dedicated test suite

tallaxes:
**Description**

Extracts storage-related E2E tests from the Integration test suite into a dedicated Storage test suite, and extends it with additional test cases for better coverage.

Changes:
- Created new `test/suites/storage/suite_test.go` containing the extracted storage E2E tests
- Removed `test/suites/integration/storage_test.go`
- Added `Storage` to the E2E test suite options/matrix
- Added `GetAvailableZones()` and `SupportsZones()` helpers to the Azure test environment
- Extended the zone provider to return available zone lists from the Azure locations API, and removed the hardcoded zonal-region fallback so zone support and zone-list behavior stay consistent on API failures
- Preserved `runner` coverage for the dynamic Azure Disk PVC test and added it to the dynamic allowed-topologies storage test

The new Storage test suite includes:
- **Persistent Volumes (Static)**: Pre-bound PV tests with storage class and topology constraint coverage
- **Persistent Volumes (Dynamic)**: Azure Disk CSI driver tests with dynamic provisioning and allowed topologies
- **Stateful workloads**: StatefulSet disruption and volume detachment tests
- **Ephemeral Storage**: `emptyDir` and memory-backed `emptyDir` tests

* Ran the new Storage test suite locally: `TEST_SUITE=Storage make az-e2etests` (all 11 tests passed in ~30 minutes)
* E2E CI run: 
  * https://github.com/Azure/karpenter-provider-azure/actions/runs/24979723195 (latest)
  * https://github.com/Azure/karpenter-provider-azure/actions/runs/20529632559/job/58982728922 ✅

- [ ] Yes, issue opened: #
- [x] No

```release-note
```


pkg/providers/zone/zone.go
```diff
@@ -56,9 +56,13 @@ type Provider struct {
 	subscriptionID   string
 	clock            Clock
 
-	// Cached zone support data - maps region name to zone support boolean
+	// Cached zone support data - maps region name to zone support boolean.
+	// Kept separate from zoneList because fallbackZonalRegions only has boolean data,
+	// allowing SupportsZones() to work even when the API fails and zoneList is empty.
```
comtalyst: Is there a possibility that this happen? The code below make it looks like both will come from the same API, though not sure if there are more behind the scenes.

test/suites/storage/suite_test.go
```diff
@@ -0,0 +1,409 @@
...
+
+		env.EventuallyExpectCreatedNodeCount(">=", 1)
+
+		// After the deletion timestamp is set and all pods are drained the node should be gone.
+		env.EventuallyExpectNotFound(nodeClaim, node)
+
+		// We expect the stateful workload to become healthy on new node before the 6-minute force detach timeout.
+		// We start timer after pod binds to node because volume attachment happens during ContainerCreating
+		env.EventuallyExpectCreatedNodeClaimCount("==", 1)
+		env.EventuallyExpectCreatedNodeCount(">=", 1)
```
comtalyst: Do we need to check for create node >= 1 this occasionally?
If we don't reset, or check whether there is no new node created, just once isenough?

comtalyst (dismissed):

pkg/providers/zone/zone.go
matthchr: I think the point is that we only hardcode the boolean collection (via `fallbackZonalRegions`) rather than the full list of zones which is what is being added here.

It is a bit weird though, I wonder if it would be better if we just hardcoded the full zonal list rather than a boolean collection and then we could avoid having a collection of booleans + a collection of list of zones. We could possibly update the fallback zonal regions to be read from a file we code-generate via calling the az cli + script (similar to other things) and then use `embed` or similar to embed that data into the application?

matthchr: This would also mean that `GetAvailableZones` had the same fallback behavior as `SupportsZones`, which would be nice.

pkg/providers/zone/zone.go
```diff
@@ -86,13 +91,29 @@ func (p *Provider) SupportsZones(ctx context.Context, region string) bool {
...
+	p.mu.Lock()
+	defer p.mu.Unlock()
+
+	p.ensureLoaded(ctx)
+	return p.zoneList[region]
+}
+
+// ensureLoaded attempts to load zone data from Azure API if not already loaded.
+// Must be called with p.mu held.
+func (p *Provider) ensureLoaded(ctx context.Context) {
```
matthchr: minor: You could call this `ensureLoadedLocked` (which is a slightly funky name) to make it more clear the expectation that this must be called w/ the lock held

matthchr (dismissed):

pkg/providers/zone/zone.go
tallaxes: Good points. Given that zone provider is currently only used in tests, I opted to remove the hardcoded fallback altogether. WDYT?

test/suites/storage/suite_test.go
tallaxes: Ack, removed

pkg/providers/zone/zone.go
tallaxes: Ack, renamed

pkg/providers/zone/zone.go
matthchr: Yeah this seems fair.

matthchr (approved):

PR #1327, opened 2026-01-04
chore: disable Azure SDK middleware logging by default in dev/E2E

tallaxes:
## Motivation

Follow-up to https://github.com/Azure/karpenter-provider-azure/pull/1031

Azure SDK middleware logging is verbose and makes logs hard to read in both E2E tests and local development. Each HTTP request/response to Azure APIs is logged, creating excessive noise that obscures actual test output and debugging information.

## Changes

- Change `ENABLE_AZURE_SDK_LOGGING` default from `true` to `false` in Makefile-az.mk

**Note:** The internal code flag default in `pkg/operator/options/options.go` remains `true` (enabled). This change only affects the Makefile environment variable override used for dev and E2E workflows. Production deployments are unaffected.

This is a configuration change. Middleware logging can still be enabled when needed by setting:
```bash
make az-configure-values ENABLE_AZURE_SDK_LOGGING=true
```

## Does this change impact docs?

- [x] No

## Release Note

```release-note
NONE
```

### DIFF
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -114,11 +114,11 @@ func NewOperator(ctx context.Context, operator *operator.Operator) (context.Cont
 
 	log.FromContext(ctx).V(0).Info("Initial AZConfig", "azConfig", azConfig.String())
 
-	cred, err := getCredential()
-	lo.Must0(err, "getting Azure credential")
-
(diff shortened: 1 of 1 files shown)


rakechill (approved):
lgtm

PR #1338, opened 2026-01-08
chore(e2e): deflake consolidation

tallaxes:
**Description**

Remove some more sources of flakiness in consolidation:
* Make allowance for metric-server migration
* Reduce reliance on utilization thresholds

* Repeated local E2E tests ✅
* https://github.com/Azure/karpenter-provider-azure/actions/runs/20807632466/job/59767665013 ✅
* https://github.com/Azure/karpenter-provider-azure/actions/runs/20824785702/job/59823134546 ✅

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/apis/v1beta1/labels.go
+++ b/pkg/apis/v1beta1/labels.go
@@ -65,6 +65,7 @@ var (
 
 		AKSLabelCluster,
 		AKSLabelMode,
+		AKSLabelScaleSetPriority,
 	)
 
 	RestrictedLabels = sets.New(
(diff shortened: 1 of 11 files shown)


tallaxes:
> Still wondering why this slowness/flakiness suddenly occurs though... Or has it always been like this?

Not sure, but metric server makes things non-deterministic (e.g. any change in its scaling config elsewhere can affect things), and relying on utilization thresholds is fragile too (changes in daemonset resource needs could shift things). Prefer to reduce (or account for) non-determinism rather than tracking down the exact source.

PR #1339, opened 2026-01-08
feat: Support specifying OS disk type

alex-hunt-materialize:
**Description**
Adds a new optional `osDiskType` field to the AKSNodeClass spec, aligned with the [AKS AgentPool API](https://learn.microsoft.com/en-us/rest/api/aks/agent-pools/create-or-update?view=rest-aks-2026-05-01&tabs=HTTP#osdisktype):

* unspecified (default)
    Matches the current behavior: an ephemeral OS disk is used when the VM size supports one of at least `osDiskSizeGB`, otherwise a managed disk.
* `Managed`
    Always uses a managed disk, even if the instance type supports an ephemeral disk large enough for the OS disk.

The field has no default so the spec hash of existing AKSNodeClasses is unchanged and no drift is triggered. An explicit `Ephemeral` value (as in the AKS API) is left for a follow-up since it needs additional validation.

Partially addresses https://github.com/Azure/karpenter-provider-azure/issues/1247.
While this does not implement the suggested solution in that issue, it does provide a better alternative than creating a massive OS disk.

This also relates to https://github.com/Azure/karpenter-provider-azure/issues/801, and conflicts with https://github.com/Azure/karpenter-provider-azure/pull/1256.
I think that PR should be updated to instead use a field name such as `osManagedDiskStorageAccountType` to better match the storage profile field names.

* New acceptance tests for both the VM and AKS machine provisioning paths, plus E2E tests.
* Manual testing in AKS was also performed.

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
feat: add osDiskType to AKSNodeClass to force a managed OS disk instead of ephemeral
```


alex-hunt-materialize:
@microsoft-github-policy-service agree company="Materialize"

uladzimirbabrou-tomtom:
@tallaxes Would be absolutely fantastic to see this change finally released.
It's quite a bit of a hassle to go rounds with keeping `karpenter-provider-azure` patched version over version to have access to such a fundamental setting.

Without this change it's literally impossible to efficiently use local NVMe backed VMs with 1 disk (via `local-csi`) and loosing 1 disk out of 2+ for OS, where managed disk would fit perfectly, resulting in quite a bit of a wasted effort and costs. It's such a shame that such a trivial and important change is being open for so long.

Is there a particular reason for not moving it furter?

alex-hunt-materialize:
@tallaxes Any chance I could get an update on this? Having this PR released is a blocker for us using Karpenter in Azure, and using Karpenter is the only reasonable way to handle node upgrades without downtime (Azure node groups force evict pods and there is no way around this).

wdarko1:
following up on @matthchr's comment here https://github.com/Azure/karpenter-provider-azure/issues/1247#issuecomment-4093591568 in #1247, Karpenter side behavior changes for disk type would have to align with the available options in AKS all up. Until AKS generally offers this option to select OS disk type, we wouldn't support this in Karpenter just yet. 

Once AKS agrees on making this path available we can continue with this change, though the discussion is still ongoing.

Alphadelta14:
> following up on Matthew Christopher (@matthchr)'s comment here [#1247 (comment)](https://github.com/Azure/karpenter-provider-azure/issues/1247#issuecomment-4093591568) in #1247, Karpenter side behavior changes for disk type would have to align with the available options in AKS all up. Until AKS generally offers this option to select OS disk type, we wouldn't support this in Karpenter just yet.
> 
> Once AKS agrees on making this path available we can continue with this change, though the discussion is still ongoing.

AKS does allow setting a storage profile and this PR is exposing them. The [imageFamilyResolver.Resolve](https://github.com/Azure/karpenter-provider-azure/blob/35c094f3555e3c3f1791f2e2cc7592d257e3f07a/pkg/providers/imagefamily/resolver.go#L116) is dynamically passing a storage profile to the [launch template API](https://github.com/Azure/karpenter-provider-azure/blob/35c094f3555e3c3f1791f2e2cc7592d257e3f07a/pkg/providers/launchtemplate/parameters/types.go#L59) (before this change, the default is to choose ephemeral solely based on disk size).
The [AKSMachineProvider](https://github.com/Azure/karpenter-provider-azure/blob/35c094f3555e3c3f1791f2e2cc7592d257e3f07a/pkg/providers/instance/aksmachineinstancehelpers.go#L53) is doing similarly via the [MachineOSProfile](https://github.com/Azure/karpenter-provider-azure/blob/35c094f3555e3c3f1791f2e2cc7592d257e3f07a/pkg/providers/instance/aksmachineinstancehelpers.go#L134).

pkg/apis/v1beta1/aksnodeclass.go
```diff
@@ -70,6 +70,12 @@ type AKSNodeClassSpec struct {
 	// +kubebuilder:validation:Pattern=`(?i)^\/subscriptions\/[^\/]+\/resourceGroups\/[a-zA-Z0-9_\-().]{0,89}[a-zA-Z0-9_\-()]\/providers\/Microsoft\.Network\/virtualNetworks\/[^\/]+\/subnets\/[^\/]+$`
 	// +optional
 	VNETSubnetID *string `json:"vnetSubnetID,omitempty"`
+	// OSDiskType is the type of disk to use for the OS.
+	// If EphemeralWithFallbackToManaged, but the VM type does not support ephemeral disks
+	// of at least OSDiskSizeGB size, we will fall back to Managed.
+	// +default="EphemeralWithFallbackToManaged"
+	// +kubebuilder:validation:Enum:={"EphemeralWithFallbackToManaged","Managed"}
```
comtalyst: Note that [AKS AgentPool API](https://learn.microsoft.com/en-us/rest/api/aks/agent-pools/create-or-update?view=rest-aks-2026-05-01&tabs=HTTP#osdisktype) have `Managed`, `Ephemeral`, and unspecified, in which unspecified should have the same behavior as our `EphemeralWithFallbackToManaged` here.

I can see some benefits from clear `EphemeralWithFallbackToManaged` comparing to unspecified, though it will introduce another difference with AKS API. @matthchr any thoughts?

pkg/providers/instance/vminstance.go
```diff
@@ -634,6 +634,8 @@ func setVMPropertiesOSDiskType(vmProperties *armcompute.VirtualMachineProperties
 			Placement: lo.ToPtr(placement),
 		}
 		vmProperties.StorageProfile.OSDisk.Caching = lo.ToPtr(armcompute.CachingTypesReadOnly)
+	} else {
+		vmProperties.StorageProfile.OSDisk.ManagedDisk = &armcompute.ManagedDiskParameters{}
 	}
 }
```
comtalyst, line 641: If not already, in addition to this "VM instance" path, you probably want to evaluate "AKS machine instance" path too, mainly in `aksmachineinstancehelpers.go`. Although I don't think any change is needed there.

For context, AKS machine path is a new node provisioning mode which will becomes the default soon for both self-hosted Karpenter and NAP. AKS API handles bootstrapping logic.

pkg/providers/instance/suite_test.go
```diff
@@ -676,6 +676,92 @@ var _ = Describe("VMInstanceProvider", func() {
...
+		Expect(azureEnv.VirtualMachinesAPI.VirtualMachineCreateOrUpdateBehavior.CalledWithInput.Len()).To(Equal(1))
+		vm := azureEnv.VirtualMachinesAPI.VirtualMachineCreateOrUpdateBehavior.CalledWithInput.Pop().VM
+
+		// Default behavior should prefer ephemeral disk when instance type supports it
+		Expect(vm.Properties.StorageProfile.OSDisk.DiffDiskSettings).ToNot(BeNil())
+		Expect(lo.FromPtr(vm.Properties.StorageProfile.OSDisk.DiffDiskSettings.Option)).To(Equal(armcompute.DiffDiskOptionsLocal))
+		Expect(vm.Properties.StorageProfile.OSDisk.ManagedDisk).To(BeNil())
+	})
+
 	Context("Update", func() {
```
comtalyst, line 765: Consider adding similar kind of tests in AKS machine path. They should be in `cloudprovider/suite_*_test.go`.

pkg/apis/v1beta1/aksnodeclass.go
comtalyst, line 77: Note that [AKS AgentPool API](https://learn.microsoft.com/en-us/rest/api/aks/agent-pools/create-or-update?view=rest-aks-2026-05-01&tabs=HTTP#osdisktype) have `Managed`, `Ephemeral`, and unspecified, in which unspecified should have the same behavior as our `EphemeralWithFallbackToManaged` here.

I think we could align ours here with AKS API by having `Managed` or unspecified (`Ephemeral` can be a follow-up, and there will more validation concerns there). 
Another side benefit is it will also spare the currently deployed NodeClasses from being outdated (in which hash version needs to be updated to prevent unnecessary drift trigger).

cc @matthchr any thoughts?


```suggestion
	// +optional
	// +kubebuilder:validation:Enum:={"Managed"}
```

test/suites/nodeclaim/eph_osdisk_test.go
```diff
@@ -87,4 +89,90 @@ var _ = Describe("Ephemeral OS Disk", func() {
...
+			},
+		)
+
+		nodeClass.Spec.OSDiskType = lo.ToPtr("Managed")
+		nodeClass.Spec.OSDiskSizeGB = lo.ToPtr[int32](50)
+
+		pod := test.Pod()
+		env.ExpectCreated(nodeClass, nodePool, pod)
+		env.EventuallyExpectHealthy(pod)
+		env.ExpectCreatedNodeCount("==", 1)
```
comtalyst, line 114: Maybe use a Deployment here and `EventuallyExpectHealthyDeployment`, like the existing tests in this file? [PR 1686](https://github.com/Azure/karpenter-provider-azure/pull/1686) switched them over because of the [Cilium startup-taint race](https://github.com/Azure/karpenter-provider-azure/issues/1625).

comtalyst (commented):
Appreciate the effort on E2E and testing, and sorry for late review.

alex-hunt-materialize:
@comtalyst I think I've addressed all your review comments and those of Copilot.

comtalyst:
E2Es are passing:
[aksscriptless](https://github.com/Azure/karpenter-provider-azure/actions/runs/34306714008), [aksmachineapi](https://github.com/Azure/karpenter-provider-azure/actions/runs/34306730592) (one scenario fail from flakiness, other runs are passing)

comtalyst (approved):
LGTM, thanks again for contributing

pkg/providers/instancetype/instancetypes.go
```diff
@@ -631,6 +631,9 @@ func supportsNVMeEphemeralOSDisk(sku *skewer.SKU) bool {
 }
 
 func UseEphemeralDisk(sku *skewer.SKU, nodeClass *v1beta1.AKSNodeClass) bool {
+	if lo.FromPtr(nodeClass.Spec.OSDiskType) == v1beta1.OSDiskTypeManaged {
+		return false
+	}
 	sizeGB, _ := FindMaxEphemeralSizeGBAndPlacement(sku)
 	return int64(*nodeClass.Spec.OSDiskSizeGB) <= sizeGB // use ephemeral disk if it is large enough
 }
```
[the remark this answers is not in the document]
comtalyst: Defaulting in multiple places is not good either. This part is also used in real NodeClass path only. Not worth fixing now.

PR #1345, opened 2026-01-09
fix: improve error message for cases where we cannot read the subnet

matthchr:
This should make it easier for users to self-diagnose issues with subnet access especially.

* Unit tests

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
Improve error message in NodeClass condition when Karpenter is unable to read the subnet for some reason
```

### DIFF
--- a/pkg/controllers/nodeclass/status/subnet.go
+++ b/pkg/controllers/nodeclass/status/subnet.go
@@ -44,9 +44,9 @@ func NewSubnetReconciler(subnetClient instance.SubnetsAPI) *SubnetReconciler {
 }
 
 const (
-	SubnetUnreadyReasonNotFound = "SubnetNotFound"
-
-	SubnetUnreadyReasonIDInvalid = "SubnetIDInvalid"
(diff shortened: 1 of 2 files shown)


pkg/controllers/nodeclass/status/subnet.go
```diff
@@ -122,6 +122,11 @@ func (r *SubnetReconciler) validateVNETSubnetID(ctx context.Context, nodeClass *
 			)
 			return reconcile.Result{RequeueAfter: time.Minute}, err
 		}
+		nodeClass.StatusConditions().SetFalse(
```
matthchr, line 125: Note that AWS doesn't do this (see [here](https://github.com/aws/karpenter-provider-aws/blob/main/pkg/controllers/nodeclass/subnet.go#L44)), but I think that's wrong. If this fails we probably ought to set the condition always. I don't really see a reason not to?

tallaxes: Same

PR #1353, opened 2026-01-17
test: fix tests to work in non-public clouds

matthchr:
For example:
```
ARM_CLOUD=AzureChinaCloud AZURE_LOCATION=chinaeast2 \
AZURE_RESOURCE_GROUP=matthchr-test AZURE_CLUSTER_NAME=matthchr-test-2 \
make az-e2etests
```
would run the tests against a cluster named matthchr-test-2 in chinaeast2 (assuming you had created one with that name).

Note that many of the tests do not pass in non-public clouds due to other assumptions like that certain SKUs exist. We will need to tackle solving that as a separate item.

* Manually (by running utilization against Azure China Cloud)

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
NONE
```

### DIFF
--- a/test/pkg/environment/azure/environment.go
+++ b/test/pkg/environment/azure/environment.go
@@ -30,6 +30,7 @@ import (
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/cloud"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
 	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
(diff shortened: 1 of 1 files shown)


test/pkg/environment/azure/environment.go
```diff
@@ -96,31 +99,69 @@ func readEnv(name string) string {
 	return value
 }
 
+func getCloudEnvironment() *auth.Environment {
+	cfg := auth.Config{}
+	err := cfg.Build()
+	if err != nil {
+		panic(fmt.Sprintf("Failed to build cloud environment: %v", err))
+	}
```
tallaxes: Maybe lo.Must - here and elsewhere? (It can take message as well)

tallaxes (dismissed):

tallaxes (approved):

PR #1361, opened 2026-01-21
chore: update sigs.k8s.io/karpenter to v1.8.2

tallaxes:
**Description**

Update `sigs.k8s.io/karpenter` to v1.8.2, enable [static capacity](https://karpenter.sh/docs/concepts/nodepools/#specreplicas) support

* `make presubmit`
* Manual `FOCUS=StaticCapacity make az-upstream-e2etests`. Proper CI integration for testing of this is best done via https://github.com/Azure/karpenter-provider-azure/pull/1363

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -172,5 +172,3 @@ require (
 	sigs.k8s.io/structured-merge-diff/v6 v6.3.0 // indirect
 	sigs.k8s.io/yaml v1.6.0 // indirect
 )
-
-replace sigs.k8s.io/karpenter => github.com/tallaxes/karpenter-core v0.0.0-20260207002041-d13eabd5d543
--- a/go.sum
+++ b/go.sum
@@ -314,8 +314,6 @@ github.com/stretchr/testify v1.8.0/go.mod h1:yNjHg4UonilssWZ8iaSj1OCr/vHnekPRkoO
 github.com/stretchr/testify v1.8.2/go.mod h1:w2LPCIKwWwSfY2zedu0+kehJoqGctiVI29o6fzry7u4=
 github.com/stretchr/testify v1.11.1 h1:7s2iGBzp5EwR7/aIZr8ao5+dra3wiQyKjjFuvgVKu7U=
 github.com/stretchr/testify v1.11.1/go.mod h1:wZwfW3scLgRK+23gO65QZefKpKQRnfz6sD981Nm4B6U=
-github.com/tallaxes/karpenter-core v0.0.0-20260207002041-d13eabd5d543 h1:9FE9AMMS9pzMxnI1VmhLjqa2PNDFBEmW9tZaOXKCexk=
-github.com/tallaxes/karpenter-core v0.0.0-20260207002041-d13eabd5d543/go.mod h1:SlMh0Ixoi1uj+cQ9meuWwk/iqfoXtcnqDx4wYheaeso=
 github.com/tidwall/gjson v1.18.0 h1:FIDeeyB800efLX89e5a8Y0BNH+LOngJyGrIWxG2FKQY=
 github.com/tidwall/gjson v1.18.0/go.mod h1:/wbyibRr2FHMks5tjHJ5F8dMZh3AcwJEMf5vlfC0lxk=
 github.com/tidwall/match v1.1.1 h1:+Ho715JplO36QYgwN9PGYNhgZvoUSc9X2c80KVTi+GA=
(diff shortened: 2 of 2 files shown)


tallaxes:
Manual run of upstream StaticCapacity passed 10/12, the failures are likely due to metric-server etc. sticky overprovisioning, but have not investigated yet

```
$ FOCUS=StaticCapacity make az-upstream-e2etests
...
Summarizing 2 Failures:
  [FAIL] StaticCapacity Dynamic NodeClaim Interaction [It] should provision dynamic nodeclaims when static nodeclaims cant satisfy pod constraints
  /go/pkg/mod/sigs.k8s.io/karpenter@v1.8.2/test/suites/regression/staticcapacity_test.go:423
  [FAIL] StaticCapacity Edge Cases [It] should handle NodePool deletion gracefully
  /go/pkg/mod/sigs.k8s.io/karpenter@v1.8.2/test/suites/regression/staticcapacity_test.go:507

Ran 12 of 74 Specs in 3486.442 seconds
FAIL! -- 10 Passed | 2 Failed | 0 Pending | 62 Skipped
```

charts/karpenter-crd/templates/karpenter.sh_nodepools.yaml
```diff
@@ -166,8 +169,24 @@ spec:
                       - type: string
                     pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
                     x-kubernetes-int-or-string: true
-                  description: Limits define a set of bounds for provisioning capacity.
+                  description: |-
+                    Limits define a set of bounds for provisioning capacity.
+                    Limits other than limits.nodes is not supported when replicas is set.
```
matthchr, line 174: minor: I dont think we can fix it here but this isn't correct English, is it?


```suggestion
                    Limits other than limits.nodes are not supported when replicas is set.
```

charts/karpenter/templates/deployment.yaml
```diff
@@ -99,7 +99,7 @@ spec:
                   divisor: "0"
                   resource: limits.memory
             - name: FEATURE_GATES
-              value: "SpotToSpotConsolidation={{ .Values.settings.featureGates.spotToSpotConsolidation }},NodeRepair={{ .Values.settings.featureGates.nodeRepair }},NodeOverlay={{ .Values.settings.featureGates.nodeOverlay }}"
+              value: "SpotToSpotConsolidation={{ .Values.settings.featureGates.spotToSpotConsolidation }},NodeRepair={{ .Values.settings.featureGates.nodeRepair }},NodeOverlay={{ .Values.settings.featureGates.nodeOverlay }},StaticCapacity={{ .Values.settings.featureGates.staticCapacity }}"
```
matthchr, line 102: Just confirming does this mean that the user can set those values in the CRDs but they'll just be ignored if this feature gate isn't enabled?

karpenter-values-template.yaml
```diff
@@ -57,3 +57,6 @@ serviceAccount:
 podLabels:
   azure.workload.identity/use: "true"
 logLevel: ${LOG_LEVEL}
+settings:
+  featureGates:
+    staticCapacity: true
```
matthchr, line 68: Why default this to true? We don't default any other alpha features to true, do we? Is this purely for testing sake, since this is our test Helm values file?

Maybe leave a comment to that effect here explaining why we set this to true even though it's alpha.

matthchr: and presumably we _won't_ set this to true in the managed NAP experience is that right?

charts/karpenter-crd/templates/karpenter.sh_nodepools.yaml
tallaxes: No can't fix here and No is not correct English, can submit upstream PR

charts/karpenter/templates/deployment.yaml
tallaxes: Yes (the static provisioning/deprovisioning controllers are not included)

karpenter-values-template.yaml
tallaxes: Yes, this is for testing. Will have to figure out the right approach here, especially if using upstream tests; I think some tests expect features to be on (some others maybe flip them dynamically). FWIW, AWS provider flips _all_ features on for development/testing: https://github.com/aws/karpenter-provider-aws/blob/10c4ef30c58b54476f67deff8f62fe81b31c415b/Makefile#L20-L24

(Let's have a separate discussion re state of this in NAP)

matthchr (dismissed):
Approved - feel free to ping me if you push updates and need me to reapprove

tallaxes:
Update on testing: There are three (upstream) tests failing. Two are issues with test code (in one case it loses our NodePool labels/startupTaints causing overprovisioning, another needs a bit of timing adjustment / Eventually to work for Azure provider. The third one however appears to be a real problem: The static provisioner controller does not check if NodePool is being deleted, and can keep provisioning NodeClaims. For some reason this does not surface in AWS testing (I presume). 

All of the above should be fixed by https://github.com/kubernetes-sigs/karpenter/pull/2840; we should hold this until fixes land.

tallaxes:
Temp point to fork for validation. All three previously-failing StaticCapacity tests pass with the tallaxes/karpenter-core fork (branch tallaxes/staticcapacity-fix-v1.8.2). The fork's fixes (DeletionTimestamp check in controller.go, preserving labels/startupTaints from provider NodePool, and Eventually() wrapper in the test) are validated end-to-end against the AKS cluster.

tallaxes:
> All of the above should be fixed by https://github.com/kubernetes-sigs/karpenter/pull/2840; we should hold this until fixes land.

The fix is in 1.10 core, retested on #1585, everything passes
* #1585

riyasyash:
@tallaxes @matthchr Would there be a release of version 1.8.x with these changes or are we expecting a direct jump to 1.10.x from 1.7.2?

matthchr:
A direct jump to 1.10 is the plan.

PR #1363, opened 2026-01-21
chore: incorporate upstream E2E tests

tallaxes:
**Description**

This PR adds AWS-style `source` field support to e2e tests, allowing tests to run from the Azure provider, upstream karpenter-core, or both. It also removes the duplicate Chaos test suite in favor of using the upstream version.

#### E2E Test Matrix (`e2e-matrix.yaml`)
- Converted test suite matrix from simple string array to object array with `name` and `source` fields
- Categorized test suites by source:
  - **Azure-only**: ACR, BYOK, Consolidation, GPU, InPlaceUpdate, KubernetesUpgrade, Scheduling, Spot, Subnet, Utilization
  - **Both (azure + upstream)**: Drift, Integration, NodeClaim
  - **Upstream-only**: Chaos, Expiration, StaticCapacity, Termination

#### E2E Workflow (`e2e.yaml`)
- Added `source` input parameter (`azure`, `upstream`, or `all`)
- Modified test execution to conditionally run:
  - `make e2etests` for Azure provider tests
  - `make upstream-e2etests` for upstream karpenter-core tests
  - Both when `source: all`

#### Makefile
- Updated `upstream-e2etests` target
- Supports `FOCUS` variable to filter specific test suites

#### Test Cleanup
- **Deleted `test/suites/chaos/`** - The Azure Chaos tests were functionally identical to upstream (same test cases, same helper functions). The upstream version works with Azure via the `--default-nodeclass` and `--default-nodepool` flags.

* https://github.com/Azure/karpenter-provider-azure/actions/runs/21202886269 - single upstream (Chaos)
* https://github.com/Azure/karpenter-provider-azure/actions/runs/21203977074 - full set

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -63,7 +63,7 @@ e2etests: ## Run the e2e suite against your local cluster
 		--ginkgo.vv
 
 upstream-e2etests: tidy download
-	# Upstream tests are in test/suites/. Use FOCUS to filter by Describe block name.
+	# Upstream tests are in test/suites/. Use FOCUS to filter by Describe block (== test suite) name.
 	# Available upstream tests: Chaos, Drift, Expiration, Integration, NodeClaim, StaticCapacity, Termination
 	AZURE_CLUSTER_NAME=${AZURE_CLUSTER_NAME} AZURE_ACR_NAME=${AZURE_ACR_NAME} AZURE_RESOURCE_GROUP=${AZURE_RESOURCE_GROUP} AZURE_SUBSCRIPTION_ID=${AZURE_SUBSCRIPTION_ID} AZURE_LOCATION=${AZURE_LOCATION} VNET_RESOURCE_GROUP=${VNET_RESOURCE_GROUP} \
 	cd $(KARPENTER_CORE_DIR) && go test \


Makefile
```diff
@@ -62,13 +62,15 @@ e2etests: ## Run the e2e suite against your local cluster
 		--ginkgo.grace-period=3m \
 		--ginkgo.vv
 
-upstream-e2etests:
+upstream-e2etests: tidy download
+	# Upstream tests are in test/suites/. Use FOCUS to filter by Describe block name.
+	# Available upstream tests: Chaos, Drift, Expiration, Integration, NodeClaim, StaticCapacity, Termination
```
matthchr, line 67: Would a link to upstream test location be better than a static list in this file?

.github/workflows/e2e.yaml
```diff
@@ -226,9 +242,18 @@ jobs:
...
           PROVISION_MODE: ${{ inputs.provisionmode }}
+          SOURCE: ${{ inputs.source }}
         run: |
           make az-creds
-          make e2etests
+
+          if [[ "$SOURCE" == 'azure' ]]; then
+            make e2etests
+          elif [[ "$SOURCE" == 'upstream' ]]; then
+            FOCUS="$TEST_SUITE" make upstream-e2etests
```
matthchr, line 252: Not `az-upstream-e2etests`?

matthchr: I assume we want the same tainting... no?

.github/workflows/e2e-matrix.yaml
```diff
@@ -26,20 +26,66 @@ on:
...
+          - name: Integration
+            source: all       # Run both; Azure has provider-specific integration tests
+          - name: NodeClaim
+            source: all       # Run both; Azure has provider-specific NodeClaim tests
+          # Upstream-only suites
+          - name: Chaos
+            source: upstream
+          - name: Expiration
+            source: upstream
+          #- name: StaticCapacity
```
matthchr, line 76: Why comment this out?

Feels like either:
1. Remove this entirely.
2. Uncomment it (I assume it won't work due to fact we aren't on 1.8+ yet)
3. Leave a short comment here explaining why it is commented out

Makefile
```diff
@@ -62,13 +62,15 @@ e2etests: ## Run the e2e suite against your local cluster
 		--ginkgo.grace-period=3m \
 		--ginkgo.vv
 
-upstream-e2etests:
+upstream-e2etests: tidy download
+	# Upstream tests are in test/suites/. Use FOCUS to filter by Describe block name.
```
matthchr:
```suggestion
	# Upstream tests are in test/suites/. Use FOCUS to filter by Describe block (== test suite) name.
```

maybe? Took me a second to realize they have a different structure than we do

.github/workflows/e2e.yaml
tallaxes: In CI install-karpenter step does the tainting. Other steps that az-e2etests does (cleanup and untainting at the end) are only useful for development loops - though I think it won't break anything here. So Yes could make it the same, or just leave it as is - which I think is what I would prefer

PR #1365, opened 2026-01-21
feat: add support for kubernetes.azure.com/os-sku

matthchr:

* Unit tests
* End to end tests: https://github.com/Azure/karpenter-provider-azure/actions/runs/21226225837

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
Add support for scheduling against kubernetes.azure.com/os-sku, and support for writing (but not scheduling against) kubernetes.azure.com/os-sku-effective and kubernetes.azure.com/os-sku-requested
```

### DIFF
--- a/pkg/providers/labels/labels_test.go
+++ b/pkg/providers/labels/labels_test.go
@@ -404,7 +404,7 @@ func TestLabelsGet(t *testing.T) {
 			},
 		},
 		{
-			name:              "Explicit Ubuntu2204 with k8s < 1.34",
+			name:              "Explicit Ubuntu2204 with k8s < 1.34 still uses Ubuntu2204",
 			imageFamily:       v1beta1.Ubuntu2204ImageFamily,
 			kubernetesVersion: "1.31.0",
 			expectedLabels: map[string]string{
@@ -424,7 +424,7 @@ func TestLabelsGet(t *testing.T) {
 			},
 		},
 		{
-			name:              "Explicit Ubuntu2404 with k8s < 1.34",
+			name:              "Explicit Ubuntu2404 with k8s < 1.34 still uses Ubuntu2404",
 			imageFamily:       v1beta1.Ubuntu2404ImageFamily,
 			kubernetesVersion: "1.31.0",
 			expectedLabels: map[string]string{
@@ -434,7 +434,7 @@ func TestLabelsGet(t *testing.T) {
 			},
 		},
 		{
-			name:              "Explicit Ubuntu2404 with k8s >= 1.34",
+			name:              "Explicit Ubuntu2404 with k8s >= 1.34 still uses Ubuntu2404",
 			imageFamily:       v1beta1.Ubuntu2404ImageFamily,
 			kubernetesVersion: "1.35.0",
 			expectedLabels: map[string]string{


pkg/apis/v1beta1/labels.go
```diff
@@ -112,6 +113,10 @@ var (
 	AKSLabelKubeletIdentityClientID = AKSLabelDomain + "/kubelet-identity-client-id"
 	AKSLabelMode                    = AKSLabelDomain + "/mode"             // "system" or "user"
 	AKSLabelScaleSetPriority        = AKSLabelDomain + "/scalesetpriority" // "spot" or "regular". Note that "regular" is never written by AKS as a label but we write it to make scheduling easier
+	AKSLabelOSSKU                   = AKSLabelDomain + "/os-sku"           // "Ubuntu" or "AzureLinux"
+
+	AKSLabelOSSKUEffective = AKSLabelDomain + "/os-sku-effective" // "Ubuntu2204", "Ubuntu2404", "AzureLinux2", "AzureLinux3"
+	AKSLabelOSSKURequested = AKSLabelDomain + "/os-sku-requested" // "Ubuntu", "Ubuntu2204", or "AzureLinux" (We don't currently allow users to explicitly request AzureLinux3 but if we did that would show up here too)
```
tallaxes, line 119: Wondering about the philosophy behind not having constants for AzureLinux2/3 here. Is it because the possible _values_ are not technically part of the API?

tallaxes (approved):

pkg/apis/v1beta1/labels.go
matthchr: yeah I could define them as constants elsewhere and then if/when we put them into the API move them here, but it felt wrong to put them here because they are not valid values for input (unlike for Ubuntu where we do support both the generic and the specific versions as inputs)

pkg/apis/v1beta1/labels_test.go
```diff
@@ -0,0 +1,71 @@
...
+
+import (
+	"testing"
+
+	"github.com/Azure/karpenter-provider-azure/pkg/apis/v1beta1"
+	. "github.com/onsi/gomega"
+)
+
+func TestGetOSSKUFromImageFamily(t *testing.T) {
+	cases := []struct {
```
theunrepentantgeek, line 27: No need to change these, but I'm currently favouring `map[string]struct{...}` style - promoting the test name to make it more visible (plus, the compiler then enforces uniqueness).

pkg/providers/labels/labels_test.go
```diff
@@ -375,3 +375,129 @@ func TestDoNotSyncTaintsLabel(t *testing.T) {
...
+			imageFamily:       v1beta1.UbuntuImageFamily,
+			kubernetesVersion: "1.34.0",
+			expectedLabels: map[string]string{
+				v1beta1.AKSLabelOSSKU:          "Ubuntu",
+				v1beta1.AKSLabelOSSKURequested: v1beta1.UbuntuImageFamily,
+				v1beta1.AKSLabelOSSKUEffective: v1beta1.Ubuntu2404ImageFamily,
+			},
+		},
+		{
+			name:              "Explicit Ubuntu2204 with k8s < 1.34",
```
theunrepentantgeek:
```suggestion
			name:              "Explicit Ubuntu2204 with k8s < 1.34 still uses Ubuntu2204",
```

pkg/providers/labels/labels_test.go
```diff
@@ -375,3 +375,129 @@ func TestDoNotSyncTaintsLabel(t *testing.T) {
...
+			imageFamily:       v1beta1.Ubuntu2204ImageFamily,
+			kubernetesVersion: "1.35.0",
+			expectedLabels: map[string]string{
+				v1beta1.AKSLabelOSSKU:          "Ubuntu",
+				v1beta1.AKSLabelOSSKURequested: v1beta1.Ubuntu2204ImageFamily,
+				v1beta1.AKSLabelOSSKUEffective: v1beta1.Ubuntu2204ImageFamily,
+			},
+		},
+		{
+			name:              "Explicit Ubuntu2404 with k8s < 1.34",
```
theunrepentantgeek:
```suggestion
			name:              "Explicit Ubuntu2404 with k8s < 1.34 still uses Ubuntu2404",
```

pkg/providers/labels/labels_test.go
```diff
@@ -375,3 +375,129 @@ func TestDoNotSyncTaintsLabel(t *testing.T) {
...
+			imageFamily:       v1beta1.Ubuntu2404ImageFamily,
+			kubernetesVersion: "1.31.0",
+			expectedLabels: map[string]string{
+				v1beta1.AKSLabelOSSKU:          "Ubuntu",
+				v1beta1.AKSLabelOSSKURequested: v1beta1.Ubuntu2404ImageFamily,
+				v1beta1.AKSLabelOSSKUEffective: v1beta1.Ubuntu2404ImageFamily,
+			},
+		},
+		{
+			name:              "Explicit Ubuntu2404 with k8s >= 1.34",
```
theunrepentantgeek:
```suggestion
			name:              "Explicit Ubuntu2404 with k8s >= 1.34 still uses Ubuntu2404",
```

theunrepentantgeek: What's important about the v1.34 boundary?

matthchr: Under 1.34, the default is 2204, above 1.34, the default is 2404. For explicitly chosen 2204 or 2404 it doesn't _really_ matter but these are just ensuring that the defaulting doesn't happen if the user has chosen a specific version

pkg/apis/v1beta1/labels_test.go
matthchr: Will leave as-is to match other table-driven tests in this file.

PR #1368, opened 2026-01-23
fix(e2e): pin down arch to amd64 for CGO

tallaxes:
**Description**

Follow-up to #1218 which broke E2E since we request both architectures but don't currently install cross-compiler tools that would be needed for CGO

* E2E: 
  * https://github.com/Azure/karpenter-provider-azure/actions/runs/21270048553/job/61218081251
  * https://github.com/Azure/karpenter-provider-azure/actions/runs/21270781852

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/.github/actions/e2e/create-cluster/action.yaml
+++ b/.github/actions/e2e/create-cluster/action.yaml
@@ -59,7 +59,7 @@ runs:
         AZURE_ACR_NAME: ${{ inputs.acr_name }}
         AZURE_LOCATION: ${{ inputs.location }}
         # Must be amd64 SKU - E2E builds only amd64 images (see KO_DEFAULTPLATFORMS in e2e.yaml)
-        AZURE_VM_SIZE: Standard_D4ds_v5
+        AZURE_VM_SIZE: Standard_D4ds_v5 # We have more quota for this VM size so force it to avoid quota errors
         K8S_VERSION: ${{ inputs.k8s_version }}
       run: make az-mkaks-cilium
     - name: az login 2


.github/actions/e2e/create-cluster/action.yaml
```diff
@@ -58,6 +58,7 @@ runs:
         AZURE_RESOURCE_GROUP: ${{ inputs.resource_group }}
         AZURE_ACR_NAME: ${{ inputs.acr_name }}
         AZURE_LOCATION: ${{ inputs.location }}
+        # Must be amd64 SKU - E2E builds only amd64 images (see KO_DEFAULTPLATFORMS in e2e.yaml)
```
matthchr, line 61: Standard_D4ds_v5 is amd64 right - this is just a note for our future selves in case we try to change it?

.github/workflows/e2e.yaml
```diff
@@ -200,6 +200,8 @@ jobs:
         shell: bash
         env:
           AZURE_ACR_NAME: ${{ env.ACR_NAME }}
+          # amd64 only: CGO cross-compilers not installed, so must match runner arch (and system pool SKU)
```
matthchr, line 203: Is this just for _our_ system pool? Not in general, right? Since users can definitely have ARM64 system pools, can't they?

.github/actions/e2e/create-cluster/action.yaml
tallaxes: Yes, exactly

.github/workflows/e2e.yaml
tallaxes: Yes. We could install tooling and build both, but right now we are using amd64, so building just for it. When which switch to building once, may be a good time to revisit. Right now seems like building faster is better.

PR #1372, opened 2026-01-23
feat: Add status validation to AKSNodeClass to ensure disk encryption set perms are present

rakechill:
**Description**
Adds additional validation so the AKSNodeClass will be marked as `NotReady` if Karpenter doesn't have permission to read disk encryption sets.

I've gone with a minimal approach s.t. the controller will attempt to get the DiskEncryptionSet with the supplied ID:
* fails with an auth error --> validation failed, requeue after 1 min
* fails with any other error --> return error, requeue soon
* succeeds --> validation succeeded, requeue after 1 hr (to re-check RBAC is still present)

* Updated existing byok e2e test cases to ensure AKSNodeClass is Ready and added an additional OSS-only test to simulate a scenario where the controlling identity (Karpenter workload identity in the case of OSS) tries to create VMs w/o having read access to the disk encryption set.
* Added acceptance tests for the Validation reconciler exercising the different code branches possible

- [X] Yes, PR includes docs updates
- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note
Add AKSNodeClass status validation for BYOK RBAC 
```

### DIFF
--- a/pkg/providers/instancetype/suite_test.go
+++ b/pkg/providers/instancetype/suite_test.go
@@ -140,6 +140,11 @@ var _ = Describe("InstanceType Provider", func() {
 	var nodePool *karpv1.NodePool
 
 	BeforeEach(func() {
+		// Reset testOptions and ctx in case a test edited them
+		// TODO: It would be nice to find a cleaner way to edit ctx/options in these tests...
+		testOptions = test.Options()
+		ctx = options.ToContext(ctx, testOptions)
+
 		nodeClass = test.AKSNodeClass()
 		test.ApplyDefaultStatus(nodeClass, env, testOptions.UseSIG)
 
@@ -157,7 +162,6 @@ var _ = Describe("InstanceType Provider", func() {
 			},
 		})
 
-		ctx = options.ToContext(ctx, test.Options())
 		cluster.Reset()
 		clusterNonZonal.Reset()
 		clusterBootstrap.Reset()
@@ -2295,6 +2299,7 @@ var _ = Describe("InstanceType Provider", func() {
 				Name      string
 				Label     string
 				ValueFunc func() string
+				SetupFunc func()
 				// ExpectedInKubeletLabels indicates if we expect to see this in the KUBELET_NODE_LABELS section of the custom script extension.
 				// If this is false it means that Karpenter will not set it on the node via KUBELET_NODE_LABELS.
 				// It does NOT mean that it will not be on the resulting Node object in a real cluster, as it may be written by another process.
@@ -2306,11 +2311,13 @@ var _ = Describe("InstanceType Provider", func() {
 				// or by another process. We're asserting on this distinction currently because it helps clarify who is doing what
 				ExpectedOnNode bool
 			}
+
+			// TODO: Is this stuff really about Provider List? Feels like no, should we put it elsewhere?
 			entries := []WellKnownLabelEntry{
 				// Well known
 				{Name: v1.LabelTopologyRegion, Label: v1.LabelTopologyRegion, ValueFunc: func() string { return fake.Region }, ExpectedInKubeletLabels: true, ExpectedOnNode: true},
 				{Name: karpv1.NodePoolLabelKey, Label: karpv1.NodePoolLabelKey, ValueFunc: func() string { return nodePool.Name }, ExpectedInKubeletLabels: true, ExpectedOnNode: true},
-				{Name: v1.LabelTopologyZone, Label: v1.LabelTopologyZone, ValueFunc: func() string { return fakeZone1 }, ExpectedInKubeletLabels: false, ExpectedOnNode: true},
+				{Name: v1.LabelTopologyZone, Label: v1.LabelTopologyZone, ValueFunc: func() string { return fakeZone1 }, ExpectedInKubeletLabels: true, ExpectedOnNode: true},
 				{Name: v1.LabelInstanceTypeStable, Label: v1.LabelInstanceTypeStable, ValueFunc: func() string { return "Standard_NC24ads_A100_v4" }, ExpectedInKubeletLabels: true, ExpectedOnNode: true},
 				{Name: v1.LabelOSStable, Label: v1.LabelOSStable, ValueFunc: func() string { return "linux" }, ExpectedInKubeletLabels: true, ExpectedOnNode: true},
 				{Name: v1.LabelArchStable, Label: v1.LabelArchStable, ValueFunc: func() string { return "amd64" }, ExpectedInKubeletLabels: true, ExpectedOnNode: true},
@@ -2336,6 +2343,22 @@ var _ = Describe("InstanceType Provider", func() {
 				{Name: v1beta1.AKSLabelScaleSetPriority + "=regular", Label: v1beta1.AKSLabelScaleSetPriority, ValueFunc: func() string { return "regular" }, ExpectedInKubeletLabels: true, ExpectedOnNode: true},
 				{Name: v1beta1.AKSLabelScaleSetPriority + "=spot", Label: v1beta1.AKSLabelScaleSetPriority, ValueFunc: func() string { return "spot" }, ExpectedInKubeletLabels: true, ExpectedOnNode: true},
 				{Name: v1beta1.AKSLabelOSSKU, Label: v1beta1.AKSLabelOSSKU, ValueFunc: func() string { return "Ubuntu" }, ExpectedInKubeletLabels: true, ExpectedOnNode: true},
+				{
+					Name:  v1beta1.AKSLabelFIPSEnabled,
+					Label: v1beta1.AKSLabelFIPSEnabled,
+					// Needs special setup because it only works on FIPS
+					SetupFunc: func() {
+						testOptions.UseSIG = true
+						ctx = options.ToContext(ctx, testOptions)
+
+						nodeClass.Spec.FIPSMode = &v1beta1.FIPSModeFIPS
+						nodeClass.Spec.ImageFamily = lo.ToPtr(v1beta1.AzureLinuxImageFamily)
+						test.ApplyDefaultStatus(nodeClass, env, testOptions.UseSIG)
+					},
+					ValueFunc:               func() string { return "true" },
+					ExpectedInKubeletLabels: true,
+					ExpectedOnNode:          true,
+				},
 				// Deprecated Labels -- note that these are not expected in kubelet labels or on the node.
 				// They are written by CloudProvider so don't need to be sent to kubelet, and they aren't required on the node object because Karpenter does a mapping from
 				// the new labels to the old labels for compatibility.
@@ -2346,7 +2369,7 @@ var _ = Describe("InstanceType Provider", func() {
 				{Name: v1.LabelInstanceType, Label: v1.LabelInstanceType, ValueFunc: func() string { return "Standard_NC24ads_A100_v4" }, ExpectedInKubeletLabels: false, ExpectedOnNode: false},
 				{Name: "topology.disk.csi.azure.com/zone", Label: "topology.disk.csi.azure.com/zone", ValueFunc: func() string { return fakeZone1 }, ExpectedInKubeletLabels: false, ExpectedOnNode: false},
 				// Unsupported labels
-				{Name: v1.LabelWindowsBuild, Label: v1.LabelWindowsBuild, ValueFunc: func() string { return "window" }, ExpectedInKubeletLabels: false, ExpectedOnNode: false},
+				{Name: v1.LabelWindowsBuild, Label: v1.LabelWindowsBuild, ValueFunc: func() string { return "window" }, ExpectedInKubeletLabels: true, ExpectedOnNode: false},
 				// Cluster Label
 				{Name: v1beta1.AKSLabelCluster, Label: v1beta1.AKSLabelCluster, ValueFunc: func() string { return "test-resourceGroup" }, ExpectedInKubeletLabels: true, ExpectedOnNode: true},
 			}
@@ -2359,6 +2382,9 @@ var _ = Describe("InstanceType Provider", func() {
 					entry WellKnownLabelEntry
 				}
 				for _, item := range entries {
+					if item.SetupFunc != nil {
+						continue // can't support nonstandard setup here as we're putting all labels on one pod
+					}
 					podDetails = append(podDetails, struct {
 						pod   *v1.Pod
 						entry WellKnownLabelEntry
@@ -2389,16 +2415,20 @@ var _ = Describe("InstanceType Provider", func() {
 					node := ExpectScheduled(ctx, env.Client, detail.pod)
 					if detail.entry.ExpectedOnNode {
 						Expect(node.Labels[key]).To(Equal(detail.pod.Spec.NodeSelector[key]))
+					} else {
+						Expect(node.Labels).ToNot(HaveKey(key))
 					}
 
 					// Get the VM creation input and decode custom data
-					if detail.entry.ExpectedInKubeletLabels {
-						// Extract the vm name from the provider ID
-						vmName, err := nodeclaimutils.GetVMName(node.Spec.ProviderID)
-						Expect(err).ToNot(HaveOccurred())
+					// Extract the vm name from the provider ID
+					vmName, err := nodeclaimutils.GetVMName(node.Spec.ProviderID)
+					Expect(err).ToNot(HaveOccurred())
 
-						vm := vmInputs[vmName].VM
+					vm := vmInputs[vmName].VM
+					if detail.entry.ExpectedInKubeletLabels {
 						ExpectKubeletNodeLabelsInCustomData(&vm, detail.entry.Label, detail.entry.ValueFunc())
+					} else {
+						ExpectKubeletNodeLabelsNotInCustomData(&vm, detail.entry.Label, detail.entry.ValueFunc())
 					}
 				}
 			})
@@ -2406,6 +2436,10 @@ var _ = Describe("InstanceType Provider", func() {
 			DescribeTable(
 				"should support individual instance type labels (when all pods scheduled individually)",
 				func(item WellKnownLabelEntry) {
+					if item.SetupFunc != nil {
+						item.SetupFunc()
+					}
+
 					ExpectApplied(ctx, env.Client, nodePool, nodeClass)
 					value := item.ValueFunc()
 
@@ -2418,6 +2452,7 @@ var _ = Describe("InstanceType Provider", func() {
 							bindings = append(bindings, ExpectProvisionedNoBinding(ctx, env.Client, clusterBootstrap, cloudProviderBootstrap, coreProvisionerBootstrap, pod))
 						}
 						for i := range len(bindings) {
+							Expect(lo.Values(bindings[i])).ToNot(BeEmpty())
 							Expect(lo.Values(bindings[i])[0].Node.Name).To(Equal(lo.Values(bindings[0])[0].Node.Name), "expected all bindings to have the same node name")
 						}
 					}
@@ -2426,14 +2461,18 @@ var _ = Describe("InstanceType Provider", func() {
 
 					if item.ExpectedOnNode {
 						Expect(node.Labels[item.Label]).To(Equal(value))
+					} else {
+						Expect(node.Labels).ToNot(HaveKey(item.Label))
 					}
 
 					// Get the VM creation input and decode custom data
+					Expect(azureEnv.VirtualMachinesAPI.VirtualMachineCreateOrUpdateBehavior.CalledWithInput.Len()).To(Equal(1))
+					vmInput := azureEnv.VirtualMachinesAPI.VirtualMachineCreateOrUpdateBehavior.CalledWithInput.Pop()
+					vm := vmInput.VM
 					if item.ExpectedInKubeletLabels {
-						Expect(azureEnv.VirtualMachinesAPI.VirtualMachineCreateOrUpdateBehavior.CalledWithInput.Len()).To(Equal(1))
-						vmInput := azureEnv.VirtualMachinesAPI.VirtualMachineCreateOrUpdateBehavior.CalledWithInput.Pop()
-						vm := vmInput.VM
 						ExpectKubeletNodeLabelsInCustomData(&vm, item.Label, value)
+					} else {
+						ExpectKubeletNodeLabelsNotInCustomData(&vm, item.Label, value)
 					}
 				},
 				lo.Map(entries, func(item WellKnownLabelEntry, _ int) TableEntry {
@@ -2444,6 +2483,10 @@ var _ = Describe("InstanceType Provider", func() {
 			DescribeTable(
 				"should support individual instance type labels (when all pods scheduled individually) on bootstrap API",
 				func(item WellKnownLabelEntry) {
+					if item.SetupFunc != nil {
+						item.SetupFunc()
+					}
+
 					ExpectApplied(ctx, env.Client, nodePool, nodeClass)
 					value := item.ValueFunc()
 
@@ -2456,6 +2499,7 @@ var _ = Describe("InstanceType Provider", func() {
 							bindings = append(bindings, ExpectProvisionedNoBinding(ctx, env.Client, clusterBootstrap, cloudProviderBootstrap, coreProvisionerBootstrap, pod))
 						}
 						for i := range len(bindings) {
+							Expect(lo.Values(bindings[i])).ToNot(BeEmpty())
 							Expect(lo.Values(bindings[i])[0].Node.Name).To(Equal(lo.Values(bindings[0])[0].Node.Name), "expected all bindings to have the same node name")
 						}
 					}
@@ -2465,14 +2509,17 @@ var _ = Describe("InstanceType Provider", func() {
 
 					if item.ExpectedOnNode {
 						Expect(node.Labels[item.Label]).To(Equal(value))
+					} else {
+						Expect(node.Labels).ToNot(HaveKey(item.Label))
 					}
 
 					// Get the bootstrap API input
+					Expect(azureEnvBootstrap.NodeBootstrappingAPI.NodeBootstrappingGetBehavior.CalledWithInput.Len()).To(Equal(1))
+					bootstrapInput := azureEnvBootstrap.NodeBootstrappingAPI.NodeBootstrappingGetBehavior.CalledWithInput.Pop()
 					if item.ExpectedInKubeletLabels {
-						Expect(azureEnvBootstrap.NodeBootstrappingAPI.NodeBootstrappingGetBehavior.CalledWithInput.Len()).To(Equal(1))
-						bootstrapInput := azureEnvBootstrap.NodeBootstrappingAPI.NodeBootstrappingGetBehavior.CalledWithInput.Pop()
-
 						Expect(bootstrapInput.Params.ProvisionProfile.CustomNodeLabels).To(HaveKeyWithValue(item.Label, value))
+					} else {
+						Expect(bootstrapInput.Params.ProvisionProfile.CustomNodeLabels).ToNot(HaveKeyWithValue(item.Label, value))
 					}
 				},
 				lo.Map(entries, func(item WellKnownLabelEntry, _ int) TableEntry {
@@ -2481,7 +2528,8 @@ var _ = Describe("InstanceType Provider", func() {
 			)
 
 			It("entries should cover every WellKnownLabel", func() {
-				Expect(lo.Map(entries, func(item WellKnownLabelEntry, _ int) string { return item.Label })).To(ContainElements(append(karpv1.WellKnownLabels.UnsortedList(), lo.Keys(karpv1.NormalizedLabels)...)))
+				expectedLabels := append(karpv1.WellKnownLabels.UnsortedList(), lo.Keys(karpv1.NormalizedLabels)...)
+				Expect(lo.Map(entries, func(item WellKnownLabelEntry, _ int) string { return item.Label })).To(ContainElements(expectedLabels))
 			})
 
 			nonSchedulableLabels := map[string]string{
--- a/pkg/apis/v1beta1/labels.go
+++ b/pkg/apis/v1beta1/labels.go
@@ -67,6 +67,7 @@ var (
 		AKSLabelMode,
 		AKSLabelScaleSetPriority,
 		AKSLabelOSSKU,
+		AKSLabelFIPSEnabled,
 	)
 
 	RestrictedLabels = sets.New(
@@ -114,6 +115,7 @@ var (
 	AKSLabelMode                    = AKSLabelDomain + "/mode"             // "system" or "user"
 	AKSLabelScaleSetPriority        = AKSLabelDomain + "/scalesetpriority" // "spot" or "regular". Note that "regular" is never written by AKS as a label but we write it to make scheduling easier
 	AKSLabelOSSKU                   = AKSLabelDomain + "/os-sku"           // "Ubuntu" or "AzureLinux"
+	AKSLabelFIPSEnabled             = AKSLabelDomain + "/fips_enabled"     // "true" or not specified
 
 	AKSLabelOSSKUEffective = AKSLabelDomain + "/os-sku-effective" // "Ubuntu2204", "Ubuntu2404", "AzureLinux2", "AzureLinux3"
 	AKSLabelOSSKURequested = AKSLabelDomain + "/os-sku-requested" // "Ubuntu", "Ubuntu2204", or "AzureLinux" (We don't currently allow users to explicitly request AzureLinux3 but if we did that would show up here too)
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -43,6 +43,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 	karpv1 "sigs.k8s.io/karpenter/pkg/apis/v1"
+	karpv1alpha1 "sigs.k8s.io/karpenter/pkg/apis/v1alpha1"
+
 	"sigs.k8s.io/karpenter/pkg/operator"
 	coreoptions "sigs.k8s.io/karpenter/pkg/operator/options"
 
@@ -313,6 +315,7 @@ func WaitForCRDs(ctx context.Context, timeout time.Duration, config *rest.Config
 	var requiredGVKs = []schema.GroupVersionKind{
 		gvk(&karpv1.NodePool{}),
 		gvk(&karpv1.NodeClaim{}),
+		gvk(&karpv1alpha1.NodeOverlay{}),
 		gvk(&v1beta1.AKSNodeClass{}),
 	}
 
--- a/pkg/providers/instancetype/instancetype.go
+++ b/pkg/providers/instancetype/instancetype.go
@@ -177,6 +177,7 @@ func computeRequirements(
 		scheduling.NewRequirement(v1beta1.AKSLabelMode, corev1.NodeSelectorOpIn, v1beta1.ModeSystem, v1beta1.ModeUser),
 		scheduling.NewRequirement(v1beta1.AKSLabelScaleSetPriority, corev1.NodeSelectorOpIn, v1beta1.ScaleSetPriorityRegular, v1beta1.ScaleSetPrioritySpot),
 		scheduling.NewRequirement(v1beta1.AKSLabelOSSKU, corev1.NodeSelectorOpIn, v1beta1.GetOSSKUFromImageFamily(lo.FromPtr(nodeClass.Spec.ImageFamily))),
+		scheduling.NewRequirement(v1beta1.AKSLabelFIPSEnabled, corev1.NodeSelectorOpDoesNotExist), // AKS only sets this label if FIPS is enabled, otherwise it's expected to be empty
 
 		// composites
 		scheduling.NewRequirement(v1beta1.LabelSKUName, corev1.NodeSelectorOpDoesNotExist),
@@ -205,6 +206,9 @@ func computeRequirements(
 	setRequirementsHyperVGeneration(requirements, sku)
 	setRequirementsGPU(requirements, sku, vmsize)
 	setRequirementsVersion(requirements, vmsize)
+	if lo.FromPtr(nodeClass.Spec.FIPSMode) == v1beta1.FIPSModeFIPS {
+		requirements[v1beta1.AKSLabelFIPSEnabled].Insert("true")
+	}
 
 	return requirements
 }
--- a/pkg/providers/labels/labels.go
+++ b/pkg/providers/labels/labels.go
@@ -111,6 +111,10 @@ func Get(
 	labels[v1beta1.AKSLabelScaleSetPriority] = v1beta1.ScaleSetPriorityRegular
 	// Add os-sku label based on imageFamily
 	labels[v1beta1.AKSLabelOSSKU] = v1beta1.GetOSSKUFromImageFamily(lo.FromPtr(nodeClass.Spec.ImageFamily))
+	if lo.FromPtr(nodeClass.Spec.FIPSMode) == v1beta1.FIPSModeFIPS {
+		labels[v1beta1.AKSLabelFIPSEnabled] = "true"
+	}
+
 	// Add os-sku-requested label that exactly matches the imageFamily specified on the NodeClass
 	labels[v1beta1.AKSLabelOSSKURequested] = lo.FromPtr(nodeClass.Spec.ImageFamily)
 	// nil static parameters here is safe only because we're not using the resulting imageFamily for anything except to get its name
--- a/test/suites/scheduling/suite_test.go
+++ b/test/suites/scheduling/suite_test.go
@@ -86,6 +86,11 @@ var _ = Describe("Scheduling", Ordered, ContinueOnFailure, func() {
 			v1beta1.LabelSKUGPUName,
 		)
 
+		if env.InClusterController {
+			// Can't test FIPS on self-hosted Karpenter
+			selectors.Insert(v1beta1.AKSLabelFIPSEnabled)
+		}
+
 		// If no spec with Label("GPU") ran (e.g., `-label-filter='!GPU'`),
 		// ignore GPU labels in the coverage assertion.
 		if !Label("GPU").MatchesLabelFilter(GinkgoLabelFilter()) {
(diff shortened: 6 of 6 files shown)


pkg/controllers/nodeclass/status/validation.go
```diff
@@ -18,23 +18,145 @@ package status
...
 
-type ValidationReconciler struct{}
+type ValidationReconciler struct {
+	kubeClient            client.Client
+	diskEncryptionSetsAPI instance.DiskEncryptionSetsAPI
+	options               *options.Options
+
+	// Cache to avoid redundant DES RBAC validation checks
+	// Stores the DES ID and the timestamp when it was last successfully validated
+	validatedDES   string
```
rakechill: nit: probably not necessary to further store the ID here since it's available from opts...

pkg/controllers/nodeclass/status/validation.go
```diff
@@ -18,23 +18,145 @@ package status
...
+)
 
-	"github.com/Azure/karpenter-provider-azure/pkg/apis/v1beta1"
+const (
+	ValidationFailedReasonDESRBACMissing     = "DESRBACMissing"
+	ValidationFailedReasonDESRBACCheckFailed = "DESRBACCheckFailed"
+	// DESValidationCacheTTL defines how long to cache successful DES RBAC validations
+	// After this duration, validation will be re-run to detect permission revocations
+	// 1 minute balances performance with timely detection of permission changes
+	DESValidationCacheTTL = 1 * time.Minute
```
rakechill: TTL should probably be higher since we suspect this RBAC will not change very often if ever.

If it's successful on first try, we maybe _don't_ want to recheck at all or very rarely.

If it's failure, maybe we should check each time (or at least frequently) to see if the role assignments have been added by the user yet

pkg/controllers/nodeclass/status/validation.go
```diff
@@ -18,23 +18,145 @@ package status
...
-	return &ValidationReconciler{}
+func NewValidationReconciler(
+	kubeClient client.Client,
+	diskEncryptionSetsAPI instance.DiskEncryptionSetsAPI,
+	opts *options.Options,
+) *ValidationReconciler {
+	return &ValidationReconciler{
+		kubeClient:            kubeClient,
+		diskEncryptionSetsAPI: diskEncryptionSetsAPI,
+		options:               opts,
```
rakechill: maybe instead of passing in opts, we just set the DES id as one of the fields of the reconciler

pkg/controllers/nodeclass/status/validation.go
```diff
@@ -18,23 +18,145 @@ package status
...
+	_, err = r.diskEncryptionSetsAPI.Get(ctx, parsedID.ResourceGroupName, parsedID.Name, nil)
+	if err != nil {
+		// Check if it's an authorization error
+		if isAuthorizationError(err) {
+			return fmt.Errorf(
+				"controlling identity does not have Reader role on Disk Encryption Set '%s'. "+
+					"Grant the Reader role on the DiskEncryptionSet to the controlling identity. "+
+					"For self-hosted installations, this is the Karpenter workload identity. "+
+					"For managed/AKS-hosted installations, this is the AKS cluster control plane identity. "+
+					"See https://learn.microsoft.com/en-us/azure/aks/use-karpenter for details",
```
rakechill: should instead have this link to the DES doc

pkg/controllers/nodeclass/status/validation_test.go
```diff
@@ -57,7 +60,29 @@ var _ = Describe("Validation Reconciler", func() {
 
 	BeforeEach(func() {
 		ctx = context.Background()
-		reconciler = status.NewValidationReconciler()
```
rakechill, line 60: TODO: add unit tests for failure vs. success scenarios and how they impact caching here. maybe instead of "success time".

If we decide to do away with the cache for successful validation, then we could instead have a field on the reconciler:
 `validatedDES bool`

pkg/providers/instancetype/suite_test.go
```diff
@@ -1293,7 +1293,7 @@ var _ = Describe("InstanceType Provider", func() {
 					UseSIG: lo.ToPtr(true),
 				})
 				ctx = options.ToContext(ctx)
-				statusController := status.NewController(env.Client, azureEnv.KubernetesVersionProvider, azureEnv.ImageProvider, env.KubernetesInterface, azureEnv.SubnetsAPI)
+				statusController := status.NewController(env.Client, azureEnv.KubernetesVersionProvider, azureEnv.ImageProvider, env.KubernetesInterface, azureEnv.SubnetsAPI, azureEnv.DiskEncryptionSetsAPI, options)
```
rakechill: `options` will be removed if DES id is part of the reconciler constructor

test/pkg/environment/azure/rbac_manager.go
```diff
@@ -85,3 +88,52 @@ func (r *RBACManager) EnsureRoleWithPrincipalType(ctx context.Context, scope, ro
 	}, nil)
 	return err
 }
+
+// HasRole checks if principalID has roleDefinitionID at scope.
+func (r *RBACManager) HasRole(ctx context.Context, scope, principalID, roleDefinitionID string) bool {
```
rakechill: remove ==> no longer used

test/pkg/environment/azure/rbac_manager.go
```diff
@@ -85,3 +88,52 @@ func (r *RBACManager) EnsureRoleWithPrincipalType(ctx context.Context, scope, ro
...
+				*ra.Properties.RoleDefinitionID == roleDefinitionID {
+				return true
+			}
+		}
+	}
+	return false
+}
+
+// RemoveRole removes all role assignments of principalID at scope.
+func (r *RBACManager) RemoveRole(ctx context.Context, scope, principalID string) error {
```
rakechill: remove ==> no longer used

test/suites/byok/suite_test.go
```diff
@@ -160,6 +172,56 @@ var _ = Describe("BYOK", func() {
...
 	})
+
+	It("should mark AKSNodeClass as NotReady when DES ID is set but Reader RBAC is missing", func() {
+		if !env.InClusterController {
+			Skip("This test requires InClusterController mode to configure DES environment variable")
+		}
+
+		ctx := context.Background()
+
+		// TODO(rbac): Use a shared constant for the RBAC error message here and in the implementation
```
rakechill: take care of this ^^

docs/e2e-testing-guide.md
```diff
@@ -0,0 +1,151 @@
...
+
+---
+
+## 3. Running E2E Tests via GitHub Actions (Manually)
+
+
+
+There are multiple workflows available:
+- **E2EMatrix**: Runs all e2e suites across multiple configurations. Trigger this with `/test e2ematrix` or by running the **E2EMatrixTrigger** workflow manually from the Actions tab.
+- **E2E**: Runs a single selected test suite. Trigger this with `/test <suite>` or by running the **E2E** workflow manually from the Actions tab and selecting a suite.
```
tallaxes: Slash commands are not in place. Manual run requires additional permissions

pkg/controllers/nodeclass/status/validation.go
```diff
@@ -18,23 +18,156 @@ package status
...
+			if isAuthorizationError(err) {
+				reason = ValidationFailedReasonDESRBACMissing
+			}
+			nodeClass.StatusConditions().SetFalse(
+				v1beta1.ConditionTypeValidationSucceeded,
+				reason,
+				err.Error(),
+			)
+			// Return success (not error) to allow status update to persist
+			return reconcile.Result{}, nil
```
tallaxes: If the idea is to check periodically, this should include ReueueAfter

pkg/controllers/nodeclass/status/validation.go
```diff
@@ -18,23 +18,156 @@ package status
...
+type ValidationReconciler struct {
+	kubeClient            client.Client
+	diskEncryptionSetsAPI instance.DiskEncryptionSetsAPI
+	diskEncryptionSetID   string
+
+	// validatedSuccessfully tracks whether DES validation has passed
+	// Once set to true, we use the cache TTL to avoid redundant checks
+	validatedSuccessfully bool
+	lastValidationTime    time.Time
+	validationMu          sync.RWMutex
```
tallaxes: It is likely possible to simplify this custom logic with cache.Cache pattern we use elsewhere

docs/e2e-testing-guide.md
tallaxes, line 1: Let's maybe have E2E guide in a separate PR? (And we can discuss the content there; e.g. neither automatic E2E nor slash commands exist right now ...)

docs/identity-model.md
tallaxes, line 1: Let's have this as a separate PR as well? (Maybe can sharpen the terminology a bit ...)

pkg/controllers/nodeclass/status/validation.go
tallaxes: To check periodically / detect changes - which I think it the intent? - Result here (and elsewhere) should include `RequeueAfter` ... (Unit tests that force-drive reconciliation runs won't catch this, but E2E tests should, is it tested?)

pkg/controllers/nodeclass/status/validation.go
tallaxes: It is likely possible to simplify the related custom logic with cache.Cache pattern we use elsewhere? (Could also cache the failure ...) The less one has to reason about cache eviction etc. - the better

pkg/controllers/nodeclass/status/validation.go
tallaxes: A quick summary of the intended logic - maybe in the PR description - would be helpful; things like periodicity / TTL (and maybe how fast change to allow or deny or is detected), which condition is set etc. Among other things, this also helps validate the implementation against the intent. 

A quick note on validating global setting via per-NodeClass validation would be good. (This is a bit unusual, but I think not unreasonable - especially given we might switch to per-NodeClass DES ID in the future.)

pkg/controllers/nodeclass/status/validation.go
```diff
@@ -18,23 +18,156 @@ package status
...
 }
 
 func (r *ValidationReconciler) Reconcile(ctx context.Context, nodeClass *v1beta1.AKSNodeClass) (reconcile.Result, error) {
-	// This is currently a skeleton for future runtime validations that cannot be expressed
-	// declaratively in the CRD schema (e.g., validations requiring external API calls,
-	// cross-resource checks, or complex contextual business logic).
+	logger := log.FromContext(ctx)
+
+	// Check BYOK RBAC if DES ID is configured
+	if r.diskEncryptionSetID != "" {
```
tallaxes: Wondering if the DES auth validation would be needed in Machine API provisioning mode - I suspect not; @comtalyst @matthchr for an opinion. If not - let's make it conditional on the provisioning mode.

pkg/controllers/nodeclass/status/validation.go
```diff
@@ -18,23 +18,156 @@ package status
...
+		if isAuthorizationError(err) {
+			// Mark validation as failed - will retry on every reconcile
+			r.validationMu.Lock()
+			r.validatedSuccessfully = false
+			r.validationMu.Unlock()
+			return fmt.Errorf(
+				"%s '%s'. "+
+					"Grant the Reader role on the DiskEncryptionSet to the controlling identity. "+
+					"For self-hosted installations, this is the Karpenter workload identity. "+
+					"For managed/AKS-hosted installations, this is the AKS cluster control plane identity. "+
```
tallaxes:
```suggestion
					"For NAP, this is the AKS cluster identity. "+
```

pkg/controllers/nodeclass/status/validation.go
```diff
@@ -18,23 +18,156 @@ package status
...
+		r.validationMu.RUnlock()
+		return nil
+	}
+	r.validationMu.RUnlock()
+
+	// Parse the DES resource ID
+	parsedID, err := arm.ParseResourceID(r.diskEncryptionSetID)
+	if err != nil {
+		return fmt.Errorf("invalid DiskEncryptionSet ID: %w", err)
+	}
```
tallaxes: nit: the ID is (currently) global and does not change; so pre-validated and does not really have to be parsed every reconcile

pkg/controllers/nodeclass/status/validation.go
tallaxes: Hmm The parent controller persists status updates irrespective of success or failure ... And non-auth API failures should likely return the error (to rely on controller-runtime retry) and leave the condition unchanged?

Otherwise, to re-check periodically and detect auth changes - which I think it the intent? - `Result` here (and elsewhere) should include `RequeueAfter` ... (Unit tests that force-drive reconciliation runs won't catch this, but E2E tests should; is it tested?)

So maybe something like this:
* Expected/recoverable failure (auth) -> Condition=False, RequeueAfter (ensure retry), no error
* Successful validation -> Condition=True, RequeueAfter (ensure retry), no error
* Unexpected errors -> don't change Condition, return error, rely on controller-runtime retry (could also use smarter retry depending on error, like respect RetryAfter etc.)

tallaxes (commented):
Nice! Left some comments re errors/retries handling, caching and such.

A short summary of the intended logic - maybe in the PR description - would be helpful; things like periodicity, caching/TTL (and maybe how fast change to allow or deny or is detected), which condition is set etc. Among other things, this also helps validate the implementation against the intent. (A quick note on validating global setting via per-NodeClass status validation reconciler would be good too. This is a bit unusual, but I think not unreasonable - especially given we might switch to per-NodeClass DES ID in the future ...)

test/pkg/environment/azure/rbac_manager.go
```diff
@@ -26,6 +26,7 @@ import (
 	"github.com/google/uuid"
 )
 
+
```
tallaxes: CI is failing because of this extra line

pkg/controllers/nodeclass/status/validation.go
rakechill: I will look further at the different patterns around this already in our codebase. I did not previouslu see the RequeueAfter option + will understand when + how conditions should be set.

rakechill:
Queued the e2e again: https://github.com/Azure/karpenter-provider-azure/actions/runs/21692919976

pkg/controllers/nodeclass/status/validation.go
```diff
@@ -18,23 +18,135 @@ package status
...
+}
 
-func NewValidationReconciler() *ValidationReconciler {
-	return &ValidationReconciler{}
+func NewValidationReconciler(
+	kubeClient client.Client,
+	diskEncryptionSetsAPI instance.DiskEncryptionSetsAPI,
+	opts *options.Options,
+) *ValidationReconciler {
+	var parsedID *arm.ResourceID
```
matthchr: Would it make more sense to do this in options.go?

Advantages I see:
1. Avoids the error case here in the constructor, which is nice.
2. We are already parsing it (sortof) in `parseDiskEncryptionSetID` -- we could cache it in options itself after parsing it there. Advantage being that if anybody else ends up needing it they don't need to re-do this bit here because we already stash it on the options struct?

Disadvantages I see (not really much of one):
1. Not a pattern we have today in options

pkg/controllers/nodeclass/status/validation.go
matthchr: I think we still want it because the experience with Machine API given a bad DES would be that individual machines fail to provision (errors/events on the machine/node/etc), but _probably_ up-front checks like this on the NodeClass status are better/clearer for users?

pkg/controllers/nodeclass/status/validation.go
```diff
@@ -18,23 +18,135 @@ package status
...
+	"github.com/Azure/karpenter-provider-azure/pkg/operator/options"
+	"github.com/Azure/karpenter-provider-azure/pkg/providers/instance"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
 
-	"github.com/Azure/karpenter-provider-azure/pkg/apis/v1beta1"
+const (
+	ValidationFailedReasonDESRBACMissing = "DESRBACMissing"
```
matthchr: IMO Spell out DiskEncryptionSet. DESRBAC is _quite_ the abbreviation =P

pkg/controllers/nodeclass/status/validation.go
```diff
@@ -18,23 +18,135 @@ package status
...
+	DESValidationFailureRequeueInterval = 1 * time.Minute
+	// DESRBACErrorMessage is the error message shown when the controlling identity lacks Reader permissions
+	DESRBACErrorMessage = "controlling identity does not have Reader role on Disk Encryption Set"
 )
 
-type ValidationReconciler struct{}
+type ValidationReconciler struct {
+	kubeClient            client.Client
+	diskEncryptionSetsAPI instance.DiskEncryptionSetsAPI
+	diskEncryptionSetID   string
```
matthchr: minor, style: We do we have/want both this and the parsed variant?

Feels like just the parsed variant is enough? I think you can get the raw string by doing `parsed.String()` on it

pkg/controllers/nodeclass/status/validation.go
```diff
@@ -18,23 +18,135 @@ package status
...
+	}
 
+	// All validations passed - requeue to detect permission revocations
 	nodeClass.StatusConditions().SetTrue(v1beta1.ConditionTypeValidationSucceeded)
-	return reconcile.Result{}, nil
+	return reconcile.Result{RequeueAfter: DESValidationSuccessRequeueInterval}, nil
+}
+
+func (r *ValidationReconciler) validateDiskEncryptionSetRBAC(ctx context.Context) error {
+	// Use pre-parsed ID if available, otherwise parse now
```
matthchr: We already tried this in the constructor and it feels like it either worked or it didn't, right?

Why do it again now?

pkg/controllers/nodeclass/status/validation.go
```diff
@@ -18,23 +18,135 @@ package status
...
+	_, err := r.diskEncryptionSetsAPI.Get(ctx, parsedID.ResourceGroupName, parsedID.Name, nil)
+	if err != nil {
+		if isAuthorizationError(err) {
+			// Wrap the original error to preserve the error chain for isAuthorizationError checks
+			return fmt.Errorf(
+				"%s '%s'. "+
+					"Grant the Reader role on the DiskEncryptionSet to the controlling identity. "+
+					"For self-hosted installations, this is the Karpenter workload identity. "+
+					"For NAP, this is the AKS cluster identity. "+
+					"See https://learn.microsoft.com/en-us/azure/aks/azure-disk-customer-managed-keys for details: %w",
```
matthchr: minor: trick with these URLs - drop the `en-us` bit so it'll redirect to the users language. (It _may_ do that automatically even if en-us is in the URL I am not sure, but it definitely will if you cut it out)

pkg/controllers/nodeclass/status/validation.go
```diff
@@ -18,23 +18,135 @@ package status
...
+		}
+		return fmt.Errorf("failed to validate DiskEncryptionSet '%s': %w", r.diskEncryptionSetID, err)
+	}
+
+	log.FromContext(ctx).V(1).Info("DES RBAC validation passed", "desID", r.diskEncryptionSetID)
+	return nil
+}
+
+// isAuthorizationError checks if an error is a 401 or 403 authorization error
+func isAuthorizationError(err error) bool {
```
matthchr: Could put it into https://github.com/Azure/azure-sdk-for-go-extensions/blob/main/pkg/errors/armerrors.go -- not a blocker though

pkg/controllers/nodeclass/status/validation_test.go
```diff
@@ -96,10 +107,188 @@ var _ = Describe("Validation Reconciler", func() {
...
+			Expect(err).To(HaveOccurred())
+			Expect(err.Error()).To(ContainSubstring("failed to validate DiskEncryptionSet"))
+			Expect(result).To(Equal(reconcile.Result{})) // No RequeueAfter, error triggers retry
+
+			// Condition should still be True (unchanged)
+			condition := nodeClass.StatusConditions().Get(v1beta1.ConditionTypeValidationSucceeded)
+			Expect(condition.IsTrue()).To(BeTrue())
+		})
+
+		It("should call API on every reconcile (no caching)", func() {
```
matthchr: This seems like a kinda weird test?

We didn't implement any caching... so why are we worried about caching?

test/suites/byok/suite_test.go
```diff
@@ -85,13 +87,24 @@ var _ = Describe("BYOK", func() {
...
+			retrieved := &v1beta1.AKSNodeClass{}
+			err := env.Client.Get(ctx, client.ObjectKeyFromObject(nodeClass), retrieved)
+			g.Expect(err).ToNot(HaveOccurred())
+			condition := retrieved.StatusConditions().Get(v1beta1.ConditionTypeValidationSucceeded)
+			g.Expect(condition.IsTrue()).To(BeTrue())
+			readyCondition := retrieved.StatusConditions().Get(status.ConditionReady)
+			g.Expect(readyCondition.IsTrue()).To(BeTrue())
+		}).WithTimeout(2 * time.Minute).WithPolling(10 * time.Second)
+
+		By("Phase 5: Waiting for VM to be created and node to be registered")
```
matthchr, line 99: Not that I object to logging in these tests, but worth noting that `EventuallyExpectCreatedNodeCount` _already_ logs and we could make others like  `EventuallyExpectHealthy` do so too.

test/suites/byok/suite_test.go
```diff
@@ -160,6 +173,53 @@ var _ = Describe("BYOK", func() {
 			Expect(lo.FromPtr(vm.Properties.StorageProfile.OSDisk.ManagedDisk.DiskEncryptionSet.ID)).To(Equal(diskEncryptionSetID))
 		}
 	})
+
+	It("should mark AKSNodeClass as NotReady when DES ID is set but Reader RBAC is missing", func() {
+		if !env.InClusterController {
+			Skip("This test requires InClusterController mode to configure DES environment variable")
+		}
+
+		ctx := context.Background()
```
matthchr: I _think_ Ginkgo supports a `ctx SpecContext` parameter on the `It` functions which you could pass in from above which would honor overall test suite cancellation/interruption better than using `context.Background()`

test/suites/byok/suite_test.go
```diff
@@ -160,6 +173,53 @@ var _ = Describe("BYOK", func() {
 			Expect(lo.FromPtr(vm.Properties.StorageProfile.OSDisk.ManagedDisk.DiskEncryptionSet.ID)).To(Equal(diskEncryptionSetID))
 		}
 	})
+
+	It("should mark AKSNodeClass as NotReady when DES ID is set but Reader RBAC is missing", func() {
+		if !env.InClusterController {
```
matthchr, line 175: Is this actually true?

Other tests seem to work in non-InClusterController mode, and the NODE_OSDISK_DISKENCRYPTIONSET_ID env variable here is only checked to be on the deployment (which is not accessible in out of cluster controller anyway). The test doesnt actually use it?

Is the issue that this test requires a misconfigured DES, which we cant really pass to the suite as a whole because then obviously it wouldn't work for the others? Maybe just need to improve the comment/log?

pkg/controllers/nodeclass/status/validation.go
rakechill: That is very fair haha

pkg/controllers/nodeclass/status/validation.go
rakechill: AI artifact... was about to push the fix before this round of reviews

pkg/controllers/nodeclass/status/validation.go
rakechill: I like that especially with the fact that we already do options validation. Makes sense to me.

pkg/controllers/nodeclass/status/validation_test.go
rakechill: Older impl had caching, so still have to remove

test/suites/byok/suite_test.go
rakechill: Yes, the issue requires us to be able to create an additional disk encryption set that we don't have RBAC for, associate them with managed Karpenter (NAP), and pass both of those in to this suite. We only have our deployment variables as options to pass that in. 

I tried two approaches to testing this validation:
1. Try to add + revoke RBAC, which we wouldn't be able to do for the control plane identity AND propagation delays took too long/are flaky for the purposes of these tests.
2. Create two different disk encryption sets and override the NODE_OSDISK_DISKENCRYPTIONSET_ID on the deployment to switch between the two. We can't do that for the tests running !InClusterController.

I'll update the comment to better reflect this.

rakechill: For now, this is true, but we may be able to somehow bypass this in the future with adjustments to how we run managed Karpenter e2e tests.

test/suites/byok/suite_test.go
rakechill: I think I still lean towards "I enjoy seeing these By statements in the output", but maybe being more intentional on ensuring the Eventually* functions log could be good in the future. I also really liked your SpecContext comment. I think that's probably something we should use any time we need a ctx specifically for text execution.

rakechill:
CI is failing b/c had to merge in larger Machine API PR. Updating those tests to use the correct status.NewController(...). After that, things should be fine.

pkg/controllers/nodeclass/status/validation.go
rakechill: https://github.com/Azure/azure-sdk-for-go-extensions/pull/364

rakechill:
Running e2e again here: https://github.com/Azure/karpenter-provider-azure/actions/runs/21882910683

edit: failed because the key vault isn't consistently using unique names per test case :"(

rakechill:
and...one final e2e run: https://github.com/Azure/karpenter-provider-azure/actions/runs/21914717467

pkg/controllers/nodeclass/status/validation.go
rakechill: @tallaxes mentioned that bumping the extensions version might be a bit more complicated. prefer to do it in a separate PR

pkg/operator/options/options.go
```diff
@@ -92,6 +93,7 @@ type Options struct {
 	AdditionalTags             map[string]string `json:"additionalTags,omitempty"`
 	EnableAzureSDKLogging      bool              `json:"enableAzureSDKLogging,omitempty"` // Controls whether Azure SDK middleware logging is enabled
 	DiskEncryptionSetID        string            `json:"diskEncryptionSetId,omitempty"`
+	ParsedDiskEncryptionSetID  *arm.ResourceID   `json:"-"`
```
matthchr: minor: may be worth sequestering this field at the bottom and having it in a `// computed properties` section, just to make it clear it's not an input it's calculated/computed

pkg/operator/options/options_validation.go
```diff
@@ -224,54 +225,41 @@ func (o *Options) validateDiskEncryptionSetID() error {
...
+	if !strings.HasPrefix(o.DiskEncryptionSetID, "/subscriptions/") {
+		return fmt.Errorf("disk-encryption-set-id is invalid: must start with /subscriptions/")
 	}
 
+	// Split and validate segment count
+	// Expected format: /subscriptions/{sub}/resourceGroups/{rg}/providers/Microsoft.Compute/diskEncryptionSets/{name}
 	parts := strings.Split(o.DiskEncryptionSetID, "/")
 	if len(parts) != 9 {
-		return nil, fmt.Errorf("disk-encryption-set-id is invalid: expected format /subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Compute/diskEncryptionSets/{diskEncryptionSetName}, got %s", o.DiskEncryptionSetID)
+		return fmt.Errorf("disk-encryption-set-id is invalid: expected format /subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Compute/diskEncryptionSets/{diskEncryptionSetName}")
```
matthchr: Why drop the `got %s`?

pkg/operator/options/options_validation.go
```diff
@@ -224,54 +225,41 @@ func (o *Options) validateDiskEncryptionSetID() error {
...
-	}
-
-	return o.validateDiskEncryptionSetValues(parts)
-}
-
-func (o *Options) parseDiskEncryptionSetID() ([]string, error) {
-	if !strings.HasPrefix(o.DiskEncryptionSetID, "/") {
-		return nil, fmt.Errorf("disk-encryption-set-id is invalid: must start with /subscriptions/, got %s", o.DiskEncryptionSetID)
+	// Must start with /subscriptions/
+	if !strings.HasPrefix(o.DiskEncryptionSetID, "/subscriptions/") {
```
matthchr: minor: Curious why do this parsing here when you could do:
```
	// Parse with Azure SDK for additional validation
	parsedID, err := arm.ParseResourceID(o.DiskEncryptionSetID)
	if err != nil {
		return fmt.Errorf("invalid DiskEncryptionSet ID: %w", err)
	}
	o.ParsedDiskEncryptionSetID = parsedID
```

and then use the fields there?

In particular IIRC that already verifies the `/subscriptions` prefix, and you can easily extract the provider, resource type, etc from it to validate them

test/suites/byok/suite_test.go
matthchr: Yeah, more logging is not really going to hurt us either - probably better to err on the side of "too much" than "too little" (not that I think this is too much, basically just saying my comment is maybe not correct/I am optimizing too much which can result in too few logs == hard to debug test)

matthchr (dismissed):
LGTM, a few minor suggestions

pkg/controllers/nodeclass/status/validation.go
tallaxes: Dependabot was insisting, so I found a way to bump it ([extracted](https://github.com/Azure/karpenter-provider-azure/pull/1415/changes/70e39942fdc9c7cbc04aa6b48045765e37a05cbf) one helper function from cloud-provider-azure), already merged: https://github.com/Azure/karpenter-provider-azure/pull/1415

pkg/operator/options/options.go
rakechill: I like that, updating

pkg/operator/options/options_validation.go
rakechill: ack, let me adjust this to utilize parse resource id better

pkg/controllers/nodeclass/status/validation.go
```diff
@@ -18,23 +18,110 @@ package status
...
+	// All validations passed - requeue to detect permission revocations
 	nodeClass.StatusConditions().SetTrue(v1beta1.ConditionTypeValidationSucceeded)
-	return reconcile.Result{}, nil
+	return reconcile.Result{RequeueAfter: ValidationSuccessRequeueInterval}, nil
+}
+
+func (r *ValidationReconciler) validateDiskEncryptionSetRBAC(ctx context.Context) error {
+	// Attempt to read the DiskEncryptionSet
+	// This uses the controller's current credentials (DefaultAzureCredential)
+	_, err := r.diskEncryptionSetsAPI.Get(ctx, r.parsedDiskEncryptionSetID.ResourceGroupName, r.parsedDiskEncryptionSetID.Name, nil)
```
tallaxes, line 94: Is getting rid of caching altogether intentional? A bit concerned about number of API calls (ValidationSuccessRequeueInterval x number of NodeClasses - even though having many NodeClasses might be unusual ...)

The original had caching, the last feeback was around potentially using a simpler caching mechanism, not removing caching overall (though open to arguments that we don't need it)
> It is likely possible to simplify the related custom logic with cache.Cache pattern we use elsewhere? (Could also cache the failure ...) The less one has to reason about cache eviction etc. - the better

pkg/controllers/nodeclass/status/controller.go
```diff
@@ -58,14 +59,17 @@ func NewController(
 	nodeImageProvider imagefamily.NodeImageProvider,
 	inClusterKubernetesInterface kubernetes.Interface,
 	subnetClient instance.SubnetsAPI,
+	diskEncryptionSetsClient instance.DiskEncryptionSetsAPI,
+	parsedDiskEncryptionSetID *arm.ResourceID,
```
theunrepentantgeek, line 65: The number of parameters to `NewController()` is getting high - 7 parameters is a lot. The only saving grace is that they are different types, making it difficult to get wrong when calling it (e.g. a transposition error is impossible).

That said, we should still consider converting this into some kind of builder struct in a later PR to make the calling code easier to read.

pkg/controllers/nodeclass/status/validation.go
rakechill: It was intentional, because I was thinking RequeueAfter() would ensure the controller acts appropriately and limits too many calls being made:
* fails with an auth error --> validation failed, requeue after 1 min
* fails with any other error --> return error, requeue soon
* succeeds --> validation succeeded, requeue after 1 hr (to re-check RBAC is still present)

Happy to discuss adjusting these timeouts, though. I was thinking the auth error being higher could result in too long of a wait time for the user after RBAC propagates. 

was based on the ideas from your original comment here: https://github.com/Azure/karpenter-provider-azure/pull/1372#discussion_r2752740570

tallaxes: These are kind of orthogonal to caching, especially given this reconciler runs per NodeClass. And we can decide not to cache failure - the original code did not, IIRC. But maybe the total/expected number of calls is not too bad ... @matthchr, WDYT?

rakechill: Yeah I understand what you mean from the perspective of "we can use requeue, but also go ahead and keep track of the success across calls to this status validation controller"

For instance, maybe in the future there are multiple reasons (aside from DES) that this controller could get requeued.

In those cases, we wouldn't necessarily want this API call to happen again

rakechill: So maybe... we don't requeue in the case of success + instead we have a cached field w/ TTL defined on the nodeclass itself. And this validation only happens during controller reconcile if the TTL is expired.

matthchr: Gets have pretty high throttling/throughput limits, so I'm not too worried about it, but it does bring up a good question about what the UX is if you remove your RBAC and how much do we care to optimize for that experience.

I'd expect that RBAC removal is not common, and when it happens the UX would be that we keep trying to provision nodes for a while (would get an error when attempting, due to no RBAC?) before eventually moving to realizing that the NodeClass status is not healthy once the timer hits and we re-check RBAC.

So I suppose the question is, the user is already going to have to deal with an hours worth of RBAC failures at the node provisioning level when they remove their RBAC - do we actually want to keep re-checking here, once this has passed once? Maybe we have an even simpler cache which is just "once the RBAC check passes a single time we assume it's good for the lifetime of the process?"

I think @tallaxes is right to call this out though, from a future-proofing of this function perspective it does seem like some form of "we may be making X checks, but we may not want to repeat this check every time we repeat the other checks, and we're sharing the same controller"

pkg/controllers/nodeclass/status/validation.go
```diff
@@ -18,23 +18,110 @@ package status
...
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
+	"github.com/Azure/karpenter-provider-azure/pkg/apis/v1beta1"
+	"github.com/Azure/karpenter-provider-azure/pkg/providers/instance"
+	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
 
-	"github.com/Azure/karpenter-provider-azure/pkg/apis/v1beta1"
+const (
+	RBACMissing = "DiskEncryptionSetRBACMissing"
```
matthchr: minor: I think you still want `DiskEncryption` or `DES` in this variable name, since it's in the error and there might be different RBAC errors in the future?

pkg/controllers/nodeclass/status/validation.go
tallaxes: > So I suppose the question is, the user is already going to have to deal with an hours worth of RBAC failures at the node provisioning level when they remove their RBAC

... only until we do check here and notice, because then NodeClass will be marked as unready and we will stop even trying to provision, right? (Hmm, maybe we could set this condition on NodeClass right away when we observe auth failure on the main codepath?)

> "once the RBAC check passes a single time we assume it's good for the lifetime of the process?"

I'd be Ok with this, since the scenario of withdrawing RBAC would presumably be very rare. The caveat is that (in NAP) the user does not have much control over the process (though probably not zero control)

Given the throttling/throughput is not a major concern, I'd be Ok with getting this in as is (maybe with a couple of comments / potential follow-ups)

pkg/controllers/nodeclass/status/validation.go
```diff
@@ -18,23 +18,110 @@ package status
...
+	CheckFailed = "DiskEncryptionSetRBACCheckFailed"
+	// ValidationSuccessRequeueInterval defines how often to re-validate DES RBAC after success
+	// Set to 1 hour since RBAC changes are infrequent in production
+	ValidationSuccessRequeueInterval = 1 * time.Hour
+	// ValidationFailureRequeueInterval defines how often to retry DES RBAC validation after auth failure
+	// Set to 1 minute to detect when permissions are granted
```
theunrepentantgeek: Might be good to indicate why 1m is a good balance between _system load_ and _quick response_.

```suggestion
	// Set to 1 minute to detect when permissions are granted without creating a high system load
```

pkg/controllers/nodeclass/status/validation.go
```diff
@@ -18,23 +18,110 @@ package status
...
+	ValidationFailureRequeueInterval = 1 * time.Minute
+	// RBACErrorMessage is the error message shown when the controlling identity lacks Reader permissions
+	RBACErrorMessage = "controlling identity does not have Reader role on Disk Encryption Set"
 )
 
-type ValidationReconciler struct{}
+type ValidationReconciler struct {
+	diskEncryptionSetsAPI     instance.DiskEncryptionSetsAPI
+	parsedDiskEncryptionSetID *arm.ResourceID // parsed by options.Validate(), will be nil if DES ID is not set
```
theunrepentantgeek: DES = Digital Encryption Standard, or something else?

theunrepentantgeek: Maybe `Disk Encryption Set` - seen later in the PR?

pkg/controllers/nodeclass/status/validation.go
```diff
@@ -18,23 +18,110 @@ package status
...
+func isAuthorizationError(err error) bool {
+	if err == nil {
+		return false
+	}
+	var respErr *azcore.ResponseError
+	if errors.As(err, &respErr) {
+		return respErr.StatusCode == http.StatusForbidden ||
+			respErr.StatusCode == http.StatusUnauthorized
+	}
+	return false
```
theunrepentantgeek: I haven't (yet) spotted which REST endpoint this is being used with ...

Any chance of this being used with an endpoint that returns 200 + error payload containing an ErrorCode that needs to be checked? ARM doesn't have very many of these, but it does have a few ...

pkg/fake/diskencryptionsetsapi.go
```diff
@@ -0,0 +1,48 @@
...
+import (
+	"context"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/compute/armcompute/v7"
+	"github.com/Azure/karpenter-provider-azure/pkg/providers/instance"
+	"github.com/samber/lo"
+)
+
+type DiskEncryptionSetsAPI struct {
+	GetFunc func(ctx context.Context, resourceGroupName string, diskEncryptionSetName string, options *armcompute.DiskEncryptionSetsClientGetOptions) (armcompute.DiskEncryptionSetsClientGetResponse, error)
```
theunrepentantgeek: For methods with more than a couple parameters, I favour formatting as one parameter per line to make them easier to read/scan:

```suggestion
	GetFunc func(
	    ctx context.Context, 
	    resourceGroupName string, 
	    diskEncryptionSetName string, 
	    options *armcompute.DiskEncryptionSetsClientGetOptions,
) (armcompute.DiskEncryptionSetsClientGetResponse, error)
```

TBH, the trailing comma on the last parameter irritates me, but I've learned to live with it.

pkg/fake/diskencryptionsetsapi.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+var _ instance.DiskEncryptionSetsAPI = &DiskEncryptionSetsAPI{}
+
+func (d *DiskEncryptionSetsAPI) Get(ctx context.Context, resourceGroupName string, diskEncryptionSetName string, options *armcompute.DiskEncryptionSetsClientGetOptions) (armcompute.DiskEncryptionSetsClientGetResponse, error) {
```
theunrepentantgeek: Ditto here.

```suggestion
func (d *DiskEncryptionSetsAPI) Get(
    ctx context.Context, 
    resourceGroupName string, 
    diskEncryptionSetName string, 
    options *armcompute.DiskEncryptionSetsClientGetOptions,
) (armcompute.DiskEncryptionSetsClientGetResponse, error) {
```

theunrepentantgeek (commented):
A few minor comments.

pkg/controllers/nodeclass/status/validation.go
rakechill: > ... (Hmm, maybe we could set this condition on NodeClass right away when we observe auth failure on the main codepath?)

Ah, that is a good thought.

> Given the throttling/throughput is not a major concern, I'd be Ok with getting this in as is (maybe with a couple of comments / potential follow-ups)
```
// TODO: May want to rethink how we handle successful validation + potential for RBAC removal. 
// See this PR comment for considerations:
// https://github.com/Azure/karpenter-provider-azure/pull/1372#discussion_r2795367386
```

rakechill: A few scenarios to consider:
* Do we even want to requeue Successful validations? Should re-verification be tied to pod restart?
* Should we update the Nodeclass's status to Failed if create VM fails due to role assignment/permissions issues?
* Should we add a cache to the controller so that if it gets requeued for other reasons (not DES), then we can only re-verify once the `DiskEncryptionSetRBACValidated` TTL expires?
* How should we handle potential future scenarios where multiple DES can be associated with a cluster? (Not sure if this is even allowed at AKS level...)

pkg/controllers/nodeclass/status/validation.go
rakechill: `op.AZClient.DiskEncryptionSetsClient()` -- let me double check on this.

rakechill: https://github.com/Azure/azure-sdk-for-go/blob/sdk/resourcemanager/compute/armcompute/v7.3.0/sdk/resourcemanager/compute/armcompute/diskencryptionsets_client.go#L205-L234

pkg/controllers/nodeclass/status/controller.go
rakechill: added a comment

pkg/controllers/nodeclass/status/validation.go
rakechill: updated

pkg/controllers/nodeclass/status/validation.go
rakechill: updated

pkg/fake/diskencryptionsetsapi.go
rakechill: accepted

pkg/fake/diskencryptionsetsapi.go
rakechill: accepted

pkg/controllers/nodeclass/status/validation.go
rakechill: updated

pkg/controllers/nodeclass/status/validation.go
rakechill: I added a comment w/ a TODO + link to this thread

rakechill:
Ok, one more review needed. Commits since last review:
* https://github.com/Azure/karpenter-provider-azure/pull/1372/changes/a15516ab76bd1f7b3bfee0ba374e30b6a9c80860..fb3d9c5e69178019042767e862e6675a629b4733 

Running e2e again:
* https://github.com/Azure/karpenter-provider-azure/actions/runs/21929638693

rakechill:
Passing e2e on latest: https://github.com/Azure/karpenter-provider-azure/actions/runs/21965973636

theunrepentantgeek (approved):

rakechill:
running full e2e matrix as well: https://github.com/Azure/karpenter-provider-azure/actions/runs/21998351570

pkg/controllers/nodeclass/status/validation.go
```diff
@@ -32,21 +30,23 @@ import (
...
+	// https://github.com/Azure/karpenter-provider-azure/pull/1372#discussion_r2795367386
 	// ValidationSuccessRequeueInterval defines how often to re-validate DES RBAC after success
 	// Set to 1 hour since RBAC changes are infrequent in production
 	ValidationSuccessRequeueInterval = 1 * time.Hour
 	// ValidationFailureRequeueInterval defines how often to retry DES RBAC validation after auth failure
-	// Set to 1 minute to detect when permissions are granted
+	// Set to 1 minute to detect when permissions are granted without creating a high system load
 	ValidationFailureRequeueInterval = 1 * time.Minute
 	// RBACErrorMessage is the error message shown when the controlling identity lacks Reader permissions
 	RBACErrorMessage = "controlling identity does not have Reader role on Disk Encryption Set"
```
matthchr: This is also a `DiskEncryptionSetRBACErrorMessage`, rather than a general `RBACErrorMessage`, right?

rakechill:
✅ And one more byok e2e test run: 
https://github.com/Azure/karpenter-provider-azure/actions/runs/22321621526

PR #1397, opened 2026-02-02
feat: Adding artifact streaming

kevinkrp93:
Fixes # 

**Description**
Adding artifact streaming to managed Karpenter (NAP). Using the node classes, customers can enable [Artifact Streaming for their AKS clusters](https://learn.microsoft.com/azure/aks/artifact-streaming). For Artifact Streaming to work, customers will also need to [enable Artifact Streaming on ACR](https://learn.microsoft.com/azure/container-registry/container-registry-artifact-streaming). Artifact Streaming is disabled by default with Karpenter (NAP).

Added new test integration for artifact streaming and tested on NAP, deploying new crd definition:
Screenshot 2026-03-02 at 10 52 48 AM

*

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


pkg/providers/imagefamily/customscriptsbootstrap/provisionclientbootstrap_test.go
```diff
@@ -984,3 +984,211 @@ func TestFIPSEnablement(t *testing.T) {
...
+		{
+			name:              "Azure Linux v3 with artifact streaming enabled",
+			ossku:             customscriptsbootstrap.ImageFamilyOSSKUAzureLinux3,
+			arch:              karpv1.ArchitectureAmd64,
+			artifactStreaming: lo.ToPtr(v1beta1.ArtifactStreamingModeEnabled),
+			expectedValue:     true,
+			description:       "Azure Linux v3 should allow artifact streaming to be enabled if configured",
+		},
+		{
+			name:              "ARM64 with artifact streaming enabled",
```
allyford: Arm64 not supported with artifact streaming

charts/karpenter-crd/templates/karpenter.azure.com_aksnodeclasses.yaml
```diff
@@ -619,6 +619,16 @@ spec:
               spec is the top level specification for the AKS Karpenter Provider.
               This will contain configuration necessary to launch instances in AKS.
             properties:
+              artifactStreaming:
+                description: |-
+                  artifactStreaming controls whether artifact streaming is enabled for provisioned nodes.
+                  Artifact streaming allows container images to be streamed to nodes rather than fully downloaded before starting.
```
juliusl: To avoid context confusion, should add that this is only supported for images that have streaming enabled in ACR

pkg/apis/v1beta1/aksnodeclass.go
```diff
@@ -34,6 +34,18 @@ var (
 	FIPSModeDisabled = FIPSMode("Disabled")
 )
 
+// +kubebuilder:validation:Enum:={Unspecified,Disabled,Enabled}
+type ArtifactStreamingMode string
```
juliusl: Eventually, I'd like to add more options to configure for streaming. 

Should this be an object instead? i.e

```
struct ArtifactStreamingSettings {
    mode ArtifactStreamingMode
}
```

juliusl (approved):

pkg/apis/v1beta1/aksnodeclass.go
```diff
@@ -424,6 +441,28 @@ func (in *AKSNodeClass) GetEncryptionAtHost() bool {
...
+// IsArtifactStreamingEnabled returns whether artifact streaming should be enabled for this node class.
+// Returns true for Enabled mode, false for Disabled mode, and for Unspecified mode,
+// returns false (the current default, which may change in the future).
+func (in *AKSNodeClass) IsArtifactStreamingEnabled() bool {
+	if in.Spec.ArtifactStreaming == nil || *in.Spec.ArtifactStreaming == "" {
+		// Unspecified/nil defaults to disabled (for now)
+		return false
+	}
+
+	switch *in.Spec.ArtifactStreaming {
```
MaximilianoUribe: nit: do not need all cases, can go with checking if enable, else return false

pkg/providers/imagefamily/bootstrap/aksbootstrap.go
```diff
@@ -355,6 +358,19 @@ func (a AKS) applyOptions(nbv *NodeBootstrapVariables) {
 	nbv.KubeletFlags = strings.Join(lo.MapToSlice(kubeletFlags, func(k, v string) string {
 		return fmt.Sprintf("%s=%s", k, v)
 	}), " ")
+
+	// Set EnableArtifactStreaming based on ArtifactStreamingMode
+	if a.ArtifactStreaming != nil {
```
MaximilianoUribe: nit: same comment as before, not needed until we have this by default

pkg/providers/imagefamily/customscriptsbootstrap/provisionclientbootstrap.go
```diff
@@ -104,12 +105,20 @@ func (p *ProvisionClientBootstrap) ConstructProvisionValues(ctx context.Context)
...
 
-	// artifact streaming is not yet supported for Arm64, for Ubuntu 20.04, Ubuntu 24.04, and for Azure Linux v3
-	// enableArtifactStreaming := p.Arch == karpv1.ArchitectureAmd64 &&
-	//		(p.OSSKU == ImageFamilyOSSKUUbuntu2204 || p.OSSKU == ImageFamilyOSSKUAzureLinux2)
-	// Temporarily disable artifact streaming altogether, until node provisioning performance is fixed
-	// (or until we make artifact streaming configurable)
+	// Artifact streaming is configurable through the AKSNodeClass spec
+	// Compute effective value based on ArtifactStreamingMode
 	enableArtifactStreaming := false
+	if p.ArtifactStreaming != nil {
```
MaximilianoUribe: nit: same as before

charts/karpenter-crd/templates/karpenter.azure.com_aksnodeclasses.yaml
ganeshkumarashok: nit: "...streamed on demand.."

pkg/apis/v1beta1/aksnodeclass.go
ganeshkumarashok: +1

pkg/apis/v1alpha2/aksnodeclass.go
```diff
@@ -34,6 +34,26 @@ var (
 	FIPSModeDisabled = FIPSMode("Disabled")
 )
 
+// +kubebuilder:validation:Enum:={Unspecified,Disabled,Enabled}
+type ArtifactStreamingMode string
+
+const (
+	// If not specified, use the default behavior (currently disabled, but may change in the future)
+	ArtifactStreamingModeUnspecified ArtifactStreamingMode = "Unspecified"
```
MaximilianoUribe: Was discussing with Alex that using Unspecified here maybe its not the best description, I would go with something like "Default", lmk what you think

pkg/apis/v1alpha2/aksnodeclass.go
```diff
@@ -95,6 +115,10 @@ type AKSNodeClassSpec struct {
 	// For more details see aka.ms/aks/localdns.
 	// +optional
 	LocalDNS *LocalDNS `json:"localDNS,omitempty"`
+	// artifactStreaming configures artifact streaming for provisioned nodes.
+	// Artifact streaming allows container images to be streamed on demand to nodes rather than fully downloaded before starting.
+	// +optional
+	ArtifactStreaming *ArtifactStreamingSettings `json:"artifactStreaming,omitempty"`
```
matthchr:
```suggestion
	ArtifactStreaming *ArtifactStreaming `json:"artifactStreaming,omitempty"`
```

Setting suffix is meaningless, IMO.

pkg/apis/v1alpha2/aksnodeclass.go
matthchr: It's unclear to me what this even means. Does it mean "delegate it to the system"? What is the system? Karpenter? AKS doesnt have this behavior for this field, right?

Why do we even want this value at all? So we can reserve the right to change it in the future? I don't know that we actually want/need an _explicit_ value for this - we could just document that unset (== nil) means we select the value.

With that said... is there risk that even changing the internal value selected by this can be breaking?

I'm sorta wondering if we just want to have unspecified (==nil) mean enabled, and the user can turn it off if they want? I think that's what @ganeshkumarashok wants the AKS API behavior to be too (though it's not there yet).

pkg/apis/v1alpha2/aksnodeclass.go
```diff
@@ -34,6 +34,26 @@ var (
...
+	// Disable artifact streaming
+	ArtifactStreamingModeDisabled ArtifactStreamingMode = "Disabled"
+	// Enable artifact streaming
+	ArtifactStreamingModeEnabled ArtifactStreamingMode = "Enabled"
+)
+
+// ArtifactStreamingSettings configures artifact streaming for provisioned nodes.
+// Artifact streaming allows container images to be streamed on demand to nodes rather than fully downloaded before starting.
+type ArtifactStreamingSettings struct {
+	// mode controls whether artifact streaming is enabled for provisioned nodes.
```
matthchr: Docs here should be improved - maybe from the REST API docs:


```suggestion
	// mode controls the artifact streaming mode. Artifact streaming speeds up the cold-start of containers on a node through on-demand image loading. To use this feature, container images must also enable artifact streaming on ACR.
```

pkg/apis/v1alpha2/aksnodeclass.go
matthchr: The value in the AKS REST API is going GA (soon) as a boolean, so unless we _really_ want this behavior in some ways it seems better (as a matter of policy) to just match the upstream AKS API here?

pkg/apis/v1beta1/aksnodeclass.go
```diff
@@ -34,6 +34,26 @@ var (
 	FIPSModeDisabled = FIPSMode("Disabled")
 )
 
+// +kubebuilder:validation:Enum:={Unspecified,Disabled,Enabled}
```
matthchr: My same comments from alpha apply here too.

pkg/providers/labels/labels.go
```diff
@@ -156,6 +157,12 @@ func Get(
 		labels[AKSLocalDNSStateLabelKey] = "disabled"
 	}
 
+	if nodeClass.IsArtifactStreamingEnabled() {
```
matthchr: As mentioned, this is wrong it doesn't match what AKS RP does. (We should only set the label when it's enabled and we should set the label to the correct/matching value that AKS RP sets, not `enabled|disabled`.)

pkg/providers/labels/labels.go
```diff
@@ -37,13 +37,14 @@ import (
...
-	AKSLabelNetworkStatelessCNI = v1beta1.AKSLabelDomain + "/network-stateless-cni"
-	AKSLocalDNSStateLabelKey    = v1beta1.AKSLabelDomain + "/localdns-state"
+	AKSLabelEBPFDataplane             = v1beta1.AKSLabelDomain + "/ebpf-dataplane"
+	AKSLabelAzureCNIOverlay           = v1beta1.AKSLabelDomain + "/azure-cni-overlay"
+	AKSLabelSubnetName                = v1beta1.AKSLabelDomain + "/network-subnet"
+	AKSLabelVNetGUID                  = v1beta1.AKSLabelDomain + "/nodenetwork-vnetguid"
+	AKSLabelPodNetworkType            = v1beta1.AKSLabelDomain + "/podnetwork-type"
+	AKSLabelNetworkStatelessCNI       = v1beta1.AKSLabelDomain + "/network-stateless-cni"
+	AKSLocalDNSStateLabelKey          = v1beta1.AKSLabelDomain + "/localdns-state"
+	AKSArtifactStreamingStateLabelKey = v1beta1.AKSLabelDomain + "/artifact-streaming-state"
```
matthchr: This is not the correct label to use.

These `AKSLabelDomain` labels should match the labels that AKS RP sets. It doesn't set this label, it sets `kubernetes.azure.com/artifactstreaming-enabled`. Note that it only sets that label if artifact streaming is enabled, which behavior we should probably match.

We also need to decide if that label is schedulable or not. According to @ganeshkumarashok it's restricted on Windows and Ubuntu18.04, which we don't currently support. It also doesn't currently work on ARM64 VHDs although that's getting fixed "soon".

I don't know if we need to make the label schedulable but I think we probably want/need something similar to what we have for LocalDNS where pkg/providers/instancetype/instancetypes.go `List` filters out excluded instance types (ARM64?). Though I suppose if @ganeshkumarashok is OK with not blocking that here since we aren't in AKS RP we can just rely on the fix for ARM64 landing and then maybe we don't need any filtering.

pkg/apis/v1beta1/aksnodeclass.go
```diff
@@ -397,7 +421,7 @@ type AKSNodeClass struct {
 // 1. A field changes its default value for an existing field that is already hashed
 // 2. A field is added to the hash calculation with an already-set value
 // 3. A field is removed from the hash calculations
-const AKSNodeClassHashVersion = "v3"
+const AKSNodeClassHashVersion = "v4"
```
matthchr: I don't think this is required because none of the above 3 requirements are true?

test/suites/integration/artifactstreaming_test.go
```diff
@@ -0,0 +1,307 @@
...
+}
+
+var _ = Describe("ArtifactStreaming", func() {
+	// Note: These tests verify Karpenter's configuration flow (labels, drift detection)
+	// AND check for artifact streaming infrastructure on nodes.
+	// Infrastructure verification (overlaybd daemon/config) requires NPS on NAP/managed clusters.
+
+	BeforeEach(func() {
+		if env.InClusterController {
+			Skip("ArtifactStreaming tests require NPS (Node Provisioning Service) - only supported in NAP/managed Karpenter mode")
```
matthchr, line 47: How were these tests tested, given this is set.

Can you walk me through the testing done?

test/suites/integration/artifactstreaming_test.go
```diff
@@ -0,0 +1,307 @@
...
+const (
+	// Label key for artifact streaming state
+	artifactStreamingStateLabelKey = "kubernetes.azure.com/artifact-streaming-state"
+
+	// Test timeouts
+	artifactStreamingTestTimeout = 5 * time.Minute
+)
+
+// ArtifactStreamingTestResult holds the results of artifact streaming verification
+type ArtifactStreamingTestResult struct {
```
matthchr: This whole test seems overengineered to me? I don't see why we need this type at all -- just have the test helper extract the logs, parse them, determine if artifact streaming is enabled and then return an error if expectation fails? The error can contain the log text?

pkg/providers/imagefamily/customscriptsbootstrap/provisionclientbootstrap.go
```diff
@@ -65,6 +65,7 @@ type ProvisionClientBootstrap struct {
 	NodeBootstrappingProvider      types.NodeBootstrappingAPI
 	FIPSMode                       *v1beta1.FIPSMode
 	LocalDNSProfile                *v1beta1.LocalDNS
+	ArtifactStreaming              *v1beta1.ArtifactStreamingSettings
```
matthchr: Should we support non-NPS modes too, given that direction seems to be interesting for various customers (and we've discussed we're keeping that mode going forward too?) cc @comtalyst

test/suites/integration/artifactstreaming_test.go
```diff
@@ -0,0 +1,307 @@
...
+func expectArtifactStreamingResult(result ArtifactStreamingTestResult, expectedEnabled bool) {
+	By(fmt.Sprintf("Artifact streaming verification - Expected enabled: %v, Actual enabled: %v", expectedEnabled, result.Enabled))
+	By(fmt.Sprintf("Verification logs:\n%s", result.Logs))
+	Expect(result.Success).To(BeTrue(), "Artifact streaming verification should succeed")
+	Expect(result.Enabled).To(Equal(expectedEnabled), fmt.Sprintf("Artifact streaming should be %v", map[bool]string{true: "enabled", false: "disabled"}[expectedEnabled]))
+}
+
+// expectNodeArtifactStreamingLabel verifies that a node has the expected artifact-streaming-state label value.
+// This function waits for the label to appear on the node with the correct value.
+func expectNodeArtifactStreamingLabel(node *corev1.Node, expectedValue string) {
```
matthchr: I don't think we need this at all.

`EventuallyExpectCreatedNodeCount` should return a node. That node should have the label. Just get the label off `enabledNode` and check that?

test/suites/integration/artifactstreaming_test.go
MaximilianoUribe: updated crds and created snapshot in rp standalone,  created a suite in rp for this artifact streaming e2e and used upstream steps for e2ev3

pkg/apis/v1alpha2/aksnodeclass.go
MaximilianoUribe: I see that the current use is this:
az aks nodepool add \
    --resource-group myResourceGroup \
    --cluster-name myAKSCluster \
    --name myNodePool \
    --enable-artifact-streaming
to keep in policy I think the best approach is to got with enable when the user specifies it, and if its (==nil) we enable and also keep the possibility of disabling

pkg/providers/imagefamily/customscriptsbootstrap/provisionclientbootstrap.go
MaximilianoUribe: I discussed this with Alex and its not in the priority list at the moment, can be done in a separate pr in the future

pkg/apis/v1alpha2/aksnodeclass.go
```diff
@@ -34,6 +34,15 @@ var (
 	FIPSModeDisabled = FIPSMode("Disabled")
 )
 
+// ArtifactStreaming configures artifact streaming for provisioned nodes.
+// Artifact streaming allows container images to be streamed on demand to nodes rather than fully downloaded before starting.
+type ArtifactStreaming struct {
+	// enabled controls whether artifact streaming is enabled for provisioned nodes.
```
matthchr: Still should improve docs? See this comment:
https://github.com/Azure/karpenter-provider-azure/pull/1397#discussion_r2897994798

pkg/apis/v1alpha2/aksnodeclass.go
```diff
@@ -421,6 +434,21 @@ func (in *AKSNodeClass) GetEncryptionAtHost() bool {
 	return false
 }
 
+// IsArtifactStreamingEnabled returns whether artifact streaming should be enabled for this node class
+// based on the architecture. ARM64 nodes do not support artifact streaming.
+// If not explicitly specified (nil), defaults to true for AMD64 and false for ARM64.
+func (in *AKSNodeClass) IsArtifactStreamingEnabled(arch string) bool {
```
matthchr: Don't we use the beta type for all actual controllers?

So I think you don't need this method here technically since nobody will be calling it?

pkg/apis/v1beta1/aksnodeclass.go
matthchr: This is still true I think

pkg/providers/imagefamily/customscriptsbootstrap/provisionclientbootstrap.go
```diff
@@ -103,12 +104,12 @@ func (p *ProvisionClientBootstrap) ConstructProvisionValues(ctx context.Context)
...
 
-	// artifact streaming is not yet supported for Arm64, for Ubuntu 20.04, Ubuntu 24.04, and for Azure Linux v3
-	// enableArtifactStreaming := p.Arch == karpv1.ArchitectureAmd64 &&
-	//		(p.OSSKU == ImageFamilyOSSKUUbuntu2204 || p.OSSKU == ImageFamilyOSSKUAzureLinux2)
-	// Temporarily disable artifact streaming altogether, until node provisioning performance is fixed
-	// (or until we make artifact streaming configurable)
-	enableArtifactStreaming := false
+	// Artifact streaming is configurable through the AKSNodeClass spec
+	// If not specified, defaults to enabled for AMD64 and disabled for ARM64
+	enableArtifactStreaming := p.Arch != karpv1.ArchitectureArm64
```
matthchr: Duplicating the defaulting logic you already defined in the AKSNodeClass?

Can we use that here? Looks like maybe no because we don't have access to the NodeClass here? Can we put the method onto `b1beta1.ArtifactStreaming` instead, maybe?

Would like to avoid duplicate codepaths for defaulting logic. Too easy for them to get out of sync.

pkg/providers/imagefamily/customscriptsbootstrap/provisionclientbootstrap_test.go
```diff
@@ -760,13 +760,13 @@ func TestArtifactStreamingEnablement(t *testing.T) {
 			description:                      "Artifact streaming should be disabled for ARM64 architecture even with supported OS",
 		},
 		{
-			name:                             "AMD64 AzureLinux3 - Artifact streaming disabled",
+			name:                             "AMD64 AzureLinux3 - Artifact streaming enabled (default, duplicate)",
```
matthchr: I'm confused this test looks the same as `"AMD64 AzureLinux3 - Artifact streaming enabled (default)",` above?

pkg/apis/v1beta1/aksnodeclass.go
```diff
@@ -421,6 +434,21 @@ func (in *AKSNodeClass) GetEncryptionAtHost() bool {
...
+func (in *AKSNodeClass) IsArtifactStreamingEnabled(arch string) bool {
+	// If explicitly set, use that value regardless of architecture
+	if in.Spec.ArtifactStreaming != nil && in.Spec.ArtifactStreaming.Enabled != nil {
+		return *in.Spec.ArtifactStreaming.Enabled
+	}
+
+	// If not specified (nil), default based on architecture
+	// ARM64 does not support artifact streaming, so default to false
+	// AMD64 supports artifact streaming, so default to true
+	return arch != "arm64"
```
matthchr: So what does this mean if the user requests artifact streaming enabled explicitly on their NodeClass? I _think_ it means we should exclude arm64 in that case then, right?

Behavior is:
- Artifact streaming not set one way or the other, default behavior which gets it enabled for amd64 but not arm64.
- If set true, only allow amd64 nodes.
- If set false, all any nodes.

If you agree, can we:
1. Add a test to ensure that this is the behavior?
2. Update pkg/providers/instancetype/instancetypes.go `(p *DefaultProvider) List` to exclude instance types that we can't enable Artifact streaming on if they explicitly requested it?

pkg/providers/imagefamily/customscriptsbootstrap/provisionclientbootstrap_test.go
```diff
@@ -984,3 +972,196 @@ func TestFIPSEnablement(t *testing.T) {
...
+			artifactStreaming:              nil,
+			expectedArtifactStreamingValue: true,
+			description:                    "When artifactStreaming is nil, artifact streaming should default to enabled",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			g := NewWithT(t)
+			// Create a copy of the base bootstrapper and modify for this test
```
matthchr: minor: Curious why do this rather than just construct the baseBootstrapper inline here?

It actually seems like less code to do that because you avoid the copy?

pkg/providers/imagefamily/customscriptsbootstrap/provisionclientbootstrap_test.go
```diff
@@ -984,3 +972,196 @@ func TestFIPSEnablement(t *testing.T) {
...
+			// Setup context with options
+			ctx := options.ToContext(context.Background(), &options.Options{
+				VMMemoryOverheadPercent: 0.075,
+				KubeletIdentityClientID: "test-kubelet-client-id",
+			})
+
+			values, err := bootstrapper.ConstructProvisionValues(ctx)
+
+			// Expect success for all cases
+			g.Expect(err).ToNot(HaveOccurred(), tt.description)
```
matthchr: minor: A bit odd to include `tt.description` as the additional help message here (and below on ln 1048). It's already basically in the test name which will show up in the test failure -- maybe we can just do away with description entirely?

pkg/providers/imagefamily/customscriptsbootstrap/provisionclientbootstrap_test.go
```diff
@@ -984,3 +972,196 @@ func TestFIPSEnablement(t *testing.T) {
...
+
+func TestArtifactStreamingWithDifferentOSSKUs(t *testing.T) {
+	// Test artifact streaming configuration with different OS SKUs
+	tests := []struct {
+		name              string
+		ossku             string
+		arch              string
+		artifactStreaming *v1beta1.ArtifactStreaming
+		expectedValue     bool
+		description       string
```
matthchr: Same description comment here.

test/suites/integration/artifactstreaming_test.go
```diff
@@ -0,0 +1,224 @@
...
+	It("should set artifact streaming label and enable infrastructure when not specified (defaults to enabled)", func() {
+		// nodeClass.Spec.ArtifactStreaming is nil by default
+
+		pod := coretest.Pod()
+		env.ExpectCreated(nodeClass, nodePool, pod)
+
+		node := env.EventuallyExpectCreatedNodeCount("==", 1)[0]
+		env.EventuallyExpectHealthy(pod)
+
+		expectNodeHasArtifactStreamingEnabledLabel(node)
```
matthchr: I think you can just use `EventuallyExpectNodeCountWithSelector` and do away with this helper entirely?

Selector will just be selecting on this label.

matthchr: Or maybe better yet do a selector for some other label that will always be there ( `v1beta1.AKSLabelMode` exists or something)? and then assert the resutling nodes also have the expected label of yours. Then you can use that for both the `expectNodeDoesNotHaveArtifactStreamingLabel` and `expectNodeHasArtifactStreamingEnabledLabel` cases and it'll make the tests run faster because we don't need 30s of `consistently` in the `expectNodeDoesNotHaveArtifactStreamingLabel` case

matthchr: Ahh wait even better:
`EventuallyExpectInitializedNodeCount` -- I think if you use that we expect it'll have all the labels by then.

pkg/providers/imagefamily/customscriptsbootstrap/provisionclientbootstrap_test.go
MaximilianoUribe: I agree, taking off

pkg/providers/imagefamily/customscriptsbootstrap/provisionclientbootstrap_test.go
MaximilianoUribe: it made more sense to me to do a copy, creation seems repetitive

test/suites/integration/artifactstreaming_test.go
MaximilianoUribe: yes, this makes more sense, will change

pkg/providers/imagefamily/customscriptsbootstrap/provisionclientbootstrap_test.go
```diff
@@ -984,3 +963,184 @@ func TestFIPSEnablement(t *testing.T) {
...
+
+			// Verify artifact streaming configuration
+			g.Expect(values.ProvisionProfile.ArtifactStreamingProfile.Enabled).To(Equal(lo.ToPtr(tt.expectedArtifactStreamingValue)),
+				"Artifact streaming enablement mismatch. Expected: %t, Actual: %t",
+				tt.expectedArtifactStreamingValue, *values.ProvisionProfile.ArtifactStreamingProfile.Enabled)
+		})
+	}
+}
+
+func TestArtifactStreamingWithDifferentOSSKUs(t *testing.T) {
```
matthchr: This test seems very similar to `TestArtifactStreamingEnablement` -- would it make sense to combine them? I think you could do so by adding an `artifactStreaming` parameter to the above test and then having it be `nil` (== default) for the existing cases and add these new cases in where it is not nil and instead explicitly specified.

pkg/providers/instancetype/suite_test.go
```diff
@@ -836,6 +836,68 @@ var _ = Describe("InstanceType Provider", func() {
...
+			},
+			Entry("when artifact streaming is not set (default) - includes ARM64",
+				nil, true),
+			Entry("when artifact streaming is explicitly enabled - excludes ARM64",
+				&v1beta1.ArtifactStreaming{Enabled: lo.ToPtr(true)}, false),
+			Entry("when artifact streaming is explicitly disabled - includes ARM64",
+				&v1beta1.ArtifactStreaming{Enabled: lo.ToPtr(false)}, true),
+		)
+
+		Context("Cache invalidation with ArtifactStreaming", func() {
```
matthchr: Test seems duplicate to above?

pkg/providers/labels/labels_test.go
```diff
@@ -494,11 +494,100 @@ func TestLabelsGet(t *testing.T) {
...
+			labelMap, err := labels.Get(ctx, nodeClass, "amd64")
 			g.Expect(err).ToNot(HaveOccurred())
 			for key, expectedValue := range tc.expectedLabels {
 				g.Expect(labelMap).To(HaveKeyWithValue(key, expectedValue), "label %s mismatch", key)
 			}
 		})
 	}
 }
+
+func TestArtifactStreamingLabel(t *testing.T) {
```
matthchr: Can you instead combine this with the above test and just have it take more parameters (variable arch?) It already has variable artifact streaming (via variable NodeClass).

matthchr (approved):

PR #1414, opened 2026-02-07
chore: update sigs.k8s.io/karpenter to v1.9.x

tallaxes:
**Description**

Update `sigs.k8s.io/karpenter` to v1.9.0

<details><summary>Initial set of changes</summary>

## Karpenter v1.9.0 Dependency Migration Notes

Summary of changes required when updating `sigs.k8s.io/karpenter` from v1.8.x to v1.9.0.

## 1. `NodeSelectorRequirementWithMinValues` struct change

The struct no longer embeds `v1.NodeSelectorRequirement` as a named field. Instead, `Key`, `Operator`, and `Values` are now direct fields on the struct.

**Old (v1.8.x):**
```go
type NodeSelectorRequirementWithMinValues struct {
    v1.NodeSelectorRequirement `json:",inline"`
    MinValues *int              `json:"minValues,omitempty"`
}
```

**New (v1.9.0):**
```go
type NodeSelectorRequirementWithMinValues struct {
    Key       string                      `json:"key"`
    Operator  v1.NodeSelectorOperator     `json:"operator,omitempty"`
    Values    []string                    `json:"values,omitempty"`
    MinValues *int                        `json:"minValues,omitempty"`
}
```

### Struct literal fix

```go
// Before
karpv1.NodeSelectorRequirementWithMinValues{
    NodeSelectorRequirement: corev1.NodeSelectorRequirement{
        Key:      corev1.LabelOSStable,
        Operator: corev1.NodeSelectorOpIn,
        Values:   []string{string(corev1.Linux)},
    },
}

// After
karpv1.NodeSelectorRequirementWithMinValues{
    Key:      corev1.LabelOSStable,
    Operator: corev1.NodeSelectorOpIn,
    Values:   []string{string(corev1.Linux)},
}
```

### Field access fix

```go
// Before
func convert(item karpv1.NodeSelectorRequirementWithMinValues) corev1.NodeSelectorRequirement {
    return item.NodeSelectorRequirement
}

// After
func convert(item karpv1.NodeSelectorRequirementWithMinValues) corev1.NodeSelectorRequirement {
    return corev1.NodeSelectorRequirement{Key: item.Key, Operator: item.Operator, Values: item.Values}
}
```

### Files affected
- `pkg/controllers/nodeclass/status/images.go`
- `pkg/test/nodeclass.go`
- `test/pkg/environment/common/environment.go`
- `pkg/cloudprovider/suite_test.go`
- `pkg/cloudprovider/suite_aksmachineapi_features_test.go`
- `pkg/cloudprovider/suite_aksmachineapi_offerings_test.go`
- `pkg/cloudprovider/suite_aksmachineapi_integration_test.go`
- `pkg/providers/instance/offerings/offerings_test.go`
- `pkg/providers/instancetype/suite_test.go`
- `pkg/apis/v1alpha2/crd_validation_cel_test.go`
- `pkg/apis/v1beta1/crd_validation_cel_test.go`
- `test/suites/integration/zone_constraints_test.go`
- `test/suites/drift/suite_test.go`
- `test/suites/consolidation/suite_test.go`
- `test/suites/scheduling/suite_test.go`
- `test/suites/nodeclaim/nodeclaim_test.go`
- `test/suites/nodeclaim/eph_osdisk_test.go`
- `test/suites/spot/suite_test.go`
- `test/suites/gpu/suite_test.go`
- `test/suites/byok/suite_test.go`
- `test/suites/chaos/suite_test.go`
- `test/suites/utilization/suite_test.go`

## 2. `LabelDomainExceptions` removed

`karpv1.LabelDomainExceptions` was removed in v1.9.0. There is no replacement — the concept of "exceptions to restricted label domains" no longer exists. Tests that iterated over this set were removed.

### Files affected
- `pkg/apis/v1alpha2/crd_validation_cel_test.go` — removed 2 test cases ("should allow restricted domains exceptions" in Requirements and Labels contexts)
- `pkg/apis/v1beta1/crd_validation_cel_test.go` — removed 2 test cases (same)

## 3. `v1alpha1.NodeSelectorRequirement` type for NodeOverlay

`NodeOverlaySpec.Requirements` changed from `[]corev1.NodeSelectorRequirement` to `[]karpv1alpha1.NodeSelectorRequirement`. The new type has the same fields (`Key`, `Operator`, `Values`) but is a distinct Go type.

```go
// Before
Requirements: []corev1.NodeSelectorRequirement{...}
coretest.ReplaceOverlayRequirements(overlay, corev1.NodeSelectorRequirement{...})

// After
Requirements: []karpv1alpha1.NodeSelectorRequirement{...}
coretest.ReplaceOverlayRequirements(overlay, karpv1alpha1.NodeSelectorRequirement{...})
```

### Files affected
- `test/suites/scheduling/suite_test.go`
- `test/suites/consolidation/suite_test.go`

## 4. Misc: format string fix

Fixed `%f` → `%q` for a string argument (`cost.MeterID`) in `hack/code/instancetype_testdata_gen/main.go`. (This may have been a pre-existing bug exposed by updated linting.)

</details>

* unit tests

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/charts/karpenter-crd/templates/karpenter.sh_nodeclaims.yaml
+++ b/charts/karpenter-crd/templates/karpenter.sh_nodeclaims.yaml
@@ -3,7 +3,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.20.0
+    controller-gen.kubebuilder.io/version: v0.19.0
   name: nodeclaims.karpenter.sh
 spec:
   group: karpenter.sh
--- a/charts/karpenter-crd/templates/karpenter.sh_nodeoverlays.yaml
+++ b/charts/karpenter-crd/templates/karpenter.sh_nodeoverlays.yaml
@@ -3,7 +3,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.20.0
+    controller-gen.kubebuilder.io/version: v0.19.0
   name: nodeoverlays.karpenter.sh
 spec:
   group: karpenter.sh
--- a/charts/karpenter-crd/templates/karpenter.sh_nodepools.yaml
+++ b/charts/karpenter-crd/templates/karpenter.sh_nodepools.yaml
@@ -3,7 +3,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.20.0
+    controller-gen.kubebuilder.io/version: v0.19.0
   name: nodepools.karpenter.sh
 spec:
   group: karpenter.sh
--- a/pkg/apis/crds/karpenter.sh_nodeclaims.yaml
+++ b/pkg/apis/crds/karpenter.sh_nodeclaims.yaml
@@ -3,7 +3,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.20.0
+    controller-gen.kubebuilder.io/version: v0.19.0
   name: nodeclaims.karpenter.sh
 spec:
   group: karpenter.sh
--- a/pkg/apis/crds/karpenter.sh_nodeoverlays.yaml
+++ b/pkg/apis/crds/karpenter.sh_nodeoverlays.yaml
@@ -3,7 +3,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.20.0
+    controller-gen.kubebuilder.io/version: v0.19.0
   name: nodeoverlays.karpenter.sh
 spec:
   group: karpenter.sh
--- a/pkg/apis/crds/karpenter.sh_nodepools.yaml
+++ b/pkg/apis/crds/karpenter.sh_nodepools.yaml
@@ -3,7 +3,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.20.0
+    controller-gen.kubebuilder.io/version: v0.19.0
   name: nodepools.karpenter.sh
 spec:
   group: karpenter.sh


tallaxes:
Capturing agent-generated background, completed and remaining work on restricted labels changes in 1.9.x, to be reviewed

<details><summary>Details</summary>

## Background

Karpenter core v1.9 simplified how restricted labels work:
- Removed `kubernetes.io` and `k8s.io` from `RestrictedLabelDomains` (now only `karpenter.sh` + cloud-provider-injected domains)
- Deleted `LabelDomainExceptions` concept entirely
- Deleted `IsRestrictedNodeLabel` function
- Shifted non-well-known label resolution from NodeClaim launch to NodeClaim creation time
- Removed `GetLabelsWithRestricted` — cloud providers must handle their own kubelet label filtering

Upstream thinking: [kubernetes-sigs/karpenter#2586 comment](https://github.com/kubernetes-sigs/karpenter/pull/2586#issuecomment-3747509240).
AWS reference implementation: [aws/karpenter-provider-aws#8888](https://github.com/aws/karpenter-provider-aws/pull/8888).

## Already Completed

- [x] **Karpenter core version updated** — `sigs.k8s.io/karpenter v1.9.0` in go.mod
- [x] **`NodeSelectorRequirementWithMinValues` struct changes** — flattened embedded struct across ~22 files
- [x] **`LabelDomainExceptions` references removed** — deleted from CRD CEL validation tests
- [x] **CRD CEL rules updated** — `kubernetes.io` and `k8s.io` domain restrictions removed from all CRD templates; only `karpenter.sh`, `kubernetes.io/hostname`, and `karpenter.azure.com` restrictions remain
- [x] **`v1alpha1.NodeSelectorRequirement` type for NodeOverlay** — updated
- [x] **Kubelet label filtering** — `IsKubeletLabel` filtering added in both the VM/launch-template path (`launchtemplate.go`) and the AKS Machine API path (`aksmachineinstancehelpers.go`)
- [x] **Unit tests re-enabled and passing** — CEL validation tests correct for v1.9 semantics; no stale `LabelDomainExceptions`/`IsRestrictedNodeLabel` references remain
- [x] **No custom CRD-level re-blocking needed** for `kubernetes.io`/`k8s.io` — v1.9 intentionally opens these for user scheduling; Azure's `IsKubeletLabel` correctly filters them before kubelet; Karpenter core syncs labels to Node directly

## Remaining Steps

### 1. Audit `AllowUndefinedWellKnownAndRestrictedLabels` correctness

**File:** `pkg/apis/v1beta1/labels.go` (line 76)

This compatibility option unions `WellKnownLabels` with `RestrictedLabels` (currently just `LabelSKUHyperVGeneration`). Since core v1.9 narrowed `RestrictedLabelDomains` to only `karpenter.sh` (+ `karpenter.azure.com` via Azure init), verify the resulting set still covers the right labels for compatibility checks in:
- `pkg/providers/imagefamily/resolver.go`
- `pkg/providers/instancetype/instancetypes.go`

### 2. Rework TEMPORARY label stripping in `aksmachineinstancehelpers.go`

**File:** `pkg/providers/instance/aksmachineinstancehelpers.go` (line ~265)

The hardcoded removal block marked `TODO(mattchr)` strips beta labels, GA well-known labels, and all `kubernetes.azure.com/` prefixed labels. With `IsKubeletLabel` filtering now applied upstream of this code, audit which removals are still necessary (many may be redundant) and replace with a principled approach.

### 3. Update stale comments and add v1.9 migration test coverage

- Update comments in `pkg/providers/launchtemplate/launchtemplate.go` and `pkg/providers/instance/aksmachineinstancehelpers.go` that reference removed core functions
- Add test cases verifying:
  - `node-restriction.kubernetes.io/*` labels in NodePool requirements are accepted by CRD validation but filtered before kubelet
  - Well-known Azure labels flow correctly through scheduling → NodeClaim → node

## Further Considerations (separate from v1.9 migration)

### `kubernetes.azure.com` domain restriction (design doc feature work)
The [design doc Decision 2](designs/0006-requirements-and-labels.md) calls for restricting the `kubernetes.azure.com` domain — adding it to `RestrictedLabelDomains` with an explicit allowlist of `kubernetes.azure.com/*` labels users can set on NodePool. This is feature work from the design doc, not a v1.9 core migration requirement. Track separately.

### AWS comparison: `RejectForbiddenLabels` vs `IsKubeletLabel`
AWS added a blocklist approach (`kubernetes.io`, `k8s.io`, `node-restriction.kubernetes.io`). Azure uses an allowlist via `IsKubeletLabel` which is more precise — it allows `node.kubernetes.io/*` and `kubelet.kubernetes.io/*` through, matching kubelet's actual admission rules. Both approaches are correct; Azure's is more accurate to kubelet behavior. Worth a comment explaining the design choice.

### Taint sanitization
The `TODO(mattchr)` at `aksmachineinstancehelpers.go` mentions doing similar sanitization for taints. Separate follow-up.

</details>

matthchr:
Here are my notes on the labels stuff:

# Matt's notes

- They dropped the kuberentes.io/k8s.io restriction entirely in CEL
- Above is _I think_ because there are two things: "What do we allow on NodeClaims" and "What do we set ourselves on nodes" and they're optimizing for the first (it's possible that certain labels WILL be on the node, so ideally they could be on the NodeClaim too for scheduling reasons)
- Also about label sync (what does Karpenter put on the Node object) - they're writing everything on the NodeClaim to the node, which is "everything" now, since they've removed the kuberentes.io/k8s.io restrictions

They said:

> Additionally, the behavior today is inconsistent since only certain labels are synced (e.g. well-known labels are ignored).

They changed things to set all labels on NodeClaim, and then the registration controller will sync all labels on the NodeClaim to the Node.

As mentioned above there are two places where we process/deal with labels:
- NodeClaim creation time (== "What labels do we put on the NodeClaim")
- VM/Machine creation time (== "What labels do we pass to Kubelet to set on start")

## We should (General):

- Decide if we still want to set labels via Kubelet? **I think yes**
- Rename our `IsKubeletLabel` - I don't love the name.

## For NodeClaim creation time (cloudprovider.go `createVMInstance`):

- Assign the custom labels from requirements. Custom means it's not well known (karpenter.sh, karpenter.azure.com, some kubernetes.azure.com) - Done by core at NodeClaim creation time as of 1.9.
- Set Well Known labels (We have this split between `vmInstanceToNodeClaim` and `createVMInstance`, with some coming from the instanceType and others coming from the nodeClaim requirements) -- I _think_ the reason we have to do from both here is because some requirements on the NodeClaim aren't single-valued and are only selected by the InstanceType, so we get those from the InstanceType and the rest from the NodeClaim requirements - we already do this and I think it stays as-is.

So I think we're good here. I don't think there's any optimizations here with the core change because they're not setting the well known labels and what we're doing is focused on that, so we still need everything we're doing.

## For VM/Machine creation (`getLaunchTemplate` and friends):

We're already filtering Kubelet labels out from the LaunchTemplate parameters, so I think we're already doing the right thing here and when the NodeClaim ends up with labels Kubelet can't set they'll get filtered here so we don't try to set them via Kubelet. Note that these labels WILL now get set on the Node by Karpenter after the Node is provisioned.

So I think we're good here too.

## Existing tests

I think a lot of this should be covered by the test at `should support individual instance type labels (when all pods scheduled at once)`. This goes over every WellKnown label and asserts that it's set on the node and Kubelet configuration. 

**What we should probably do though is add some tests here that cover the previously reserved labels, and ensure that ExpectedInKubeletLabels: false, ExpectedOnNode: true**

tallaxes:
> What we should probably do though is add some tests here that cover the previously reserved labels, and ensure that ExpectedInKubeletLabels: false, ExpectedOnNode: true

Added in https://github.com/Azure/karpenter-provider-azure/pull/1414/commits/1e01a09c86b07f79a486d67b015da93c6ff0a250

charts/karpenter-crd/templates/karpenter.sh_nodeclaims.yaml
```diff
@@ -3,7 +3,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.20.0
+    controller-gen.kubebuilder.io/version: v0.19.0
```
matthchr: Shouldn't this not change? If we're running generation from a devcontainer, shouldn't it has a consistent version of controller-gen?

matthchr: Or do we copy the yaml as-is from upstream and _they_ keep changing it? maybe that's the nuance I am missing

pkg/apis/v1alpha2/crd_validation_cel_test.go
```diff
@@ -81,10 +81,8 @@ var _ = Describe("CEL/Validation", func() {
 						},
 						Requirements: []karpv1.NodeSelectorRequirementWithMinValues{
 							{
-								NodeSelectorRequirement: corev1.NodeSelectorRequirement{
```
matthchr, line 84: They got rid of `json inline` here, I guess? So this serializes the same as before but structurally a bit different in Go code?

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -108,7 +108,13 @@ func (p *Provider) GetTemplate(
 	instanceType *cloudprovider.InstanceType,
 	additionalLabels map[string]string,
 ) (*Template, error) {
-	staticParameters, err := p.getStaticParameters(ctx, instanceType, nodeClass, lo.Assign(nodeClaim.Labels, additionalLabels))
+	// Filter nodeClaim.Labels to only include kubelet-compatible labels.
+	// In karpenter v1.9+, nodeClaim.Labels may contain custom labels resolved from requirements
+	// (e.g. node-restriction.kubernetes.io/*) that should not be passed to kubelet.
```
matthchr: Are you missing a similar change in pkg/providers/instance/aksmachineinstancehelpers.go?

It also does:
```
	// We need to get all single-valued requirement labels from the instance type and the nodeClaim to pass down to kubelet.
	// We don't just include single-value labels from the instance type because in the case where the label is NOT single-value on the instance
	// (i.e. there are options), the nodeClaim may have selected one of those options via its requirements which we want to include.
	// These may contain restricted labels from the pod that we need to filter out. We don't bother filtering the instance type requirements below because
	// we know those can't be restricted since they're controlled by the provider and none use the kubernetes.io domain.
	claimLabels := labels.GetFilteredSingleValuedRequirementLabels(
		scheduling.NewNodeSelectorRequirementsWithMinValues(nodeClaim.Spec.Requirements...),
		func(k string, req *scheduling.Requirement) bool {
			return labels.CanKubeletSetLabel(k)
		},
	)
	nodeLabels := lo.Assign(nodeClaim.Labels, claimLabels, labels.GetAllSingleValuedRequirementLabels(instanceType.Requirements), map[string]string{karpv1.CapacityTypeLabelKey: capacityType})
```

Which I think you want to change?

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -108,7 +108,13 @@ func (p *Provider) GetTemplate(
...
+	filteredClaimLabels := lo.PickBy(nodeClaim.Labels, func(k string, _ string) bool {
```
matthchr: Rather than doing this here, which is a second layer of filtering, we could pull this up into the caller which is in `vminstance.go` `getLaunchTemplate`, which already does this:
```
	// We need to get all single-valued requirement labels from the instance type and the nodeClaim to pass down to kubelet.
	// We don't just include single-value labels from the instance type because in the case where the label is NOT single-value on the instance
	// (i.e. there are options), the nodeClaim may have selected one of those options via its requirements which we want to include.

	// These may contain restricted labels from the pod that we need to filter out. We don't bother filtering the instance type requirements below because
	// we know those can't be restricted since they're controlled by the provider and none use the kubernetes.io domain.
	claimLabels := labels.GetFilteredSingleValuedRequirementLabels(
		scheduling.NewNodeSelectorRequirementsWithMinValues(nodeClaim.Spec.Requirements...),
		func(k string, req *scheduling.Requirement) bool {
			return labels.CanKubeletSetLabel(k)
		},
	)
	additionalLabels := lo.Assign(
		claimLabels,
		labels.GetAllSingleValuedRequirementLabels(instanceType.Requirements),
		map[string]string{karpv1.CapacityTypeLabelKey: capacityType},
	)
```

Though it would be a bit awkward because nobody is stopping the user from just using nodeClaim.Labels in the `getStaticParameters` function... so maybe we should pull this filtering down there instead?

May be worth noting that it's probably possibly to simplify this now that core is doing something slightly different than they were doing before.

What core sets: See https://github.com/kubernetes-sigs/karpenter/commit/37f5de7ac39f63457d341e0db995dc30e40efe0a#diff-cb076a97445792a289e4aaf2ccac938d9b051d5eb294a32c5bf7d174db9ac56dR81-R88

Since core already sets all the non-wellknown/non-restricted labels, we could limit our requirements -> labels propagation to _just_ WellKnownLabels (I think?). So that would remove the `CanKubeletSetLabel` call at the `getLaunchTemplate` level and we could just have a single layer of filtering on the nodeClaim.labels themselves here, which _is_ required because core would have set labels on `nodeClaim.Labels` that kubelet cannot set.

A bit rambly but... wdyt?





`PopulateNodeClaimDetails`

pkg/apis/v1alpha2/crd_validation_cel_test.go
tallaxes: Yes

charts/karpenter-crd/templates/karpenter.sh_nodeclaims.yaml
tallaxes: We copy them as-is. 1.8 went to 0.20, 1.9 reverted to 0.19. I presume they had a reason but have not checked. But this is not an indication of a problem.

pkg/providers/launchtemplate/launchtemplate.go
tallaxes: You may be right. I thought it already does it - it filters _everything_ by IsLabelKubeletManaged later - but this quite what's needed. Fixing (and adding a test)

tallaxes: Done in https://github.com/Azure/karpenter-provider-azure/pull/1414/commits/a20467f735c3ea778cfa36f582117284f21fe172, PTAL (CC @comtalyst). One thing to watch out for: The "contract" for labels passed into launch template (eventually passed into kubelet) and labels passed into Machine API might be different; would also expect Machine API to do additional filtering internally.

pkg/providers/launchtemplate/launchtemplate.go
tallaxes: Pushing filtering down seems cleaner / simpler (even if a bit of an overkill), PTAL https://github.com/Azure/karpenter-provider-azure/pull/1414/commits/1639f5bae8f103e0297499f962a83ef6e05eb0c8

(Can look into further simplification, but this should work)

pkg/providers/instance/aksmachineinstancehelpers.go
```diff
@@ -255,10 +251,12 @@ func configureLabelsAndMode(nodeClaim *karpv1.NodeClaim, instanceType *corecloud
 	}
 
 	// TODO: also do the same for taints (which don't have sanitization logic like this yet)
-	// Remove all labels with kubernetes.azure.com prefix, as well as those managed by kubelet.
-	// Also remove legacy AKS managed labels
+	// Remove labels that shouldn't be sent to the API:
+	// - AKS-managed labels (kubernetes.azure.com/*) and legacy AKS labels
+	// - Kubelet-managed labels (set automatically by kubelet, e.g. hostname, zone)
+	// - Labels kubelet can't set (e.g. kubernetes.io/*, k8s.io/* outside allowed namespaces)
```
matthchr, line 257: AKS only blocks:
```
			"disallowed-label-list": [
				"beta.kubernetes.io/arch",
				"beta.kubernetes.io/instance-type",
				"beta.kubernetes.io/os",
				"failure-domain.beta.kubernetes.io/region",
				"failure-domain.beta.kubernetes.io/zone",
				"failure-domain.kubernetes.io/zone",
				"failure-domain.kubernetes.io/region",
				"kubernetes.io/arch",
				"kubernetes.io/hostname",
				"kubernetes.io/os",
				"kubernetes.io/role",
				"kubernetes.io/instance-type",
				"node.kubernetes.io/instance-type",
				"topology.kubernetes.io/region",
				"topology.kubernetes.io/zone"
			]
```

So you might be right that we could technically allow these non-kubelet labels here. I still think it's right to block them because then Machine + VM are the same and just rely on Karpenter-core to set them for this case.

I also don't actually know _how_ AKS allows you to set these labels as I could've sworn they just passed them all to kubelet...

comtalyst: Note that this omission here will be silent removal, and provisioned nodes won't see the labels. Well, assuming that they get through CEL validations and it is possible to hit this.
If we are blocking beyond what we have today, that could be a behavior change?

(so far, before this PR, AKS Machine API's blocks and our blocks/omission are generally aligned)

matthchr: Provisioned nodes will still get the labels because Karpenter core assigns them to the node before it removes the Karpenter initialization taint.

tallaxes: > If we are blocking beyond what we have today, that could be a behavior change?

I don't think we are blocking beyond what we have today, it is just that before these were blocked by CEL, and now they are not.

PR #1427, opened 2026-02-13
chore: base updates to support Machine tests

matthchr:
**Description**
* Add provision_mode parameter everywhere
* Refactor makefile to avoid a bit of duplication
* Add new `az-mkaks-cilium-userassigned` make target
* Add some Machine-specific assertions/cleanup in the base test Environment struct

This is an update to the old PR: https://github.com/Azure/karpenter-provider-azure/pull/1200

* E2E: https://github.com/Azure/karpenter-provider-azure/actions/runs/21995351107

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note
NONE
```

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -50,7 +50,8 @@ on:
         type: choice
         description: "Azure location to run the e2e test in"
         default: westus3
-        # TODO: Why only these 3 regions?
+        # Limiting to a subset of regions to give a nice dropdown and also
+        # ensure the regions are configured to pass with all of our tests (capacity, etc)
         options:
           - westus3
           - southcentralus
--- a/test/pkg/environment/azure/environment.go
+++ b/test/pkg/environment/azure/environment.go
@@ -91,18 +91,21 @@ type Environment struct {
 	RBACManager *RBACManager
 }
 
-func readEnv(name string, required bool) string {
+func readEnvRequired(name string) string {
 	value, exists := os.LookupEnv(name)
 	if !exists {
-		if required {
-			panic(fmt.Sprintf("Environment variable %s is not set", name))
-		}
-		return ""
+		panic(fmt.Sprintf("Environment variable %s is not set", name))
 	}
 	if value == "" {
-		if required {
-			panic(fmt.Sprintf("Environment variable %s is set to an empty string", name))
-		}
+		panic(fmt.Sprintf("Environment variable %s is set to an empty string", name))
+	}
+	return value
+}
+
+func readEnvOptional(name string) string {
+	value, exists := os.LookupEnv(name)
+	if !exists {
+		return ""
 	}
 	return value
 }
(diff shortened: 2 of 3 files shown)


.github/workflows/e2e.yaml
```diff
@@ -50,14 +50,12 @@ on:
         type: choice
         description: "Azure location to run the e2e test in"
         default: westus3
+        # TODO: Why only these 3 regions?
```
matthchr: Can/should I expand this?

.github/actions/e2e/install-karpenter/action.yaml
```diff
@@ -28,7 +28,7 @@ inputs:
   location:
     description: "the azure location to run the e2e test in"
     default: "eastus"
-  provisionmode:
+  provision_mode:
```
comtalyst, line 31: I noticed that even earlier ones are inconsistent on this. E.g. `subscription-id` vs `cluster_name`. Is `_` the correct one?

matthchr: I mean I am not sure there is a "correct" one, but it's the one that:
- Matches the other parameter Charlie added (identity_type).
- Has the most parameters following that pattern (3 `-`, 4 `_` before this one which would make it 5).

We should probably make it consistent but I think either is more correct than `provisionmode` because we don't have any examples at least in this file of a "alloneword" pattern with no separators.

.github/workflows/e2e.yaml
theunrepentantgeek: Definitely - at least because you have four regions listed ... ;-)

.github/actions/e2e/install-karpenter/action.yaml
theunrepentantgeek: Working towards consistency is a good idea - but it can happen in a later PR. I don't have much of an opinion which way it goes.

test/pkg/environment/azure/environment.go
```diff
@@ -88,13 +91,18 @@ type Environment struct {
 	RBACManager *RBACManager
 }
 
-func readEnv(name string) string {
+func readEnv(name string, required bool) string {
```
theunrepentantgeek: Nitpick - I dislike bool parameters like this because while they are clear in the implementation, they're opaque at the call site:

``` go
readEnv("AZURE_RESOURCE_GROUP", true),
```

There's no way to know what the `true` means.

_Either_ split this into two methods (say, `readEnvOptional` and `readEnvMandatory` 
_or_ use an enum so the call sites are explicit (e.g. `readEnv("AZURE_RESOURCE_GROUP", mandatory)`)

.github/actions/e2e/install-karpenter/action.yaml
matthchr: Keeping this as `provision_mode` for now.

I'll send a separate PR fixing the pattern for the rest of them too, to be consistent.

test/pkg/environment/azure/environment.go
matthchr: Swapped to `readEnvOptional` and `readEnvRequired`.

Issue #1323, opened 2026-01-02
Improve handling on quota failures (or avoid them altogether)


matthchr:
We should:
- [ ] Investigate via metrics how often NAP/Karpenter is hitting quota issues today.
- [ ] Improve the existing quota handling, OR pre-check and just avoid attempting to allocate sizes that have very low/no quota.

matthchr:
Existing quota handling was improved by being more intelligent/reactive.

Leaving this open as I'd like to be proactive, which we have a number of APIs we could use to do, but even today the reactivity to quota errors is pretty good after the most recent updates.

PR #1433, opened 2026-02-17
fix: improve NodePool weight fallback and add pre-launch circuit breaker for quota/capacity errors

GuetaHen:
Fixes https://github.com/Azure/karpenter-provider-azure/issues/1323

## Summary

This PR addresses a critical issue where Karpenter fails to fall back to lower-weight NodePools when higher-weight pools encounter quota exhaustion or capacity outages during large-scale scheduling.

### Problem

When a high-weight NodePool's SKU family hits Azure quota limits or capacity outages:
1. The ICE cache TTL (3 min default) for non-zero quota errors was too short, causing offerings to recycle and be retried before the scheduler could exhaust all SKUs and fall through to lower-weight pools
2. Regional quota errors updated no cache at all, leading to infinite retry loops
3. During prompt-scale events (e.g., 5000 cores), all NodeClaims launched simultaneously would hit the same failed SKUs, wasting dozens of Azure API calls

### Changes

**Commit 1: Fix ICE cache gaps for quota errors** (`commonerrorhandlers.go`)
- Non-zero SKU family quota errors now use 15-minute TTL (was 3 min default), preventing the recycling loop
- Regional quota errors now update the ICE cache with 30-minute TTL before returning `InsufficientCapacityError`
- Added `SKUFamilyQuotaNonZeroTTL` and `RegionalQuotaExhaustedTTL` constants

**Commit 2: Pre-launch circuit breaker** (`offerings.go`, `vminstance.go`, `aksmachineinstance.go`)
- New `PreLaunchFilter()` re-checks instance types against the live ICE cache at launch time
- New `NewLiveCacheAvailabilityCheck()` helper avoids code duplication between VM and AKS Machine providers
- Acts as a circuit breaker: once the first failure updates the cache, subsequent NodeClaims skip the failed SKU immediately
- Fail-open design: if cache lookup fails, all instance types remain available

### Impact

| Scenario | Before | After |
|---|---|---|
| Quota exhausted, fallback | Infinite recycling loop | Immediate fallback after 15min cache |
| Regional quota hit | No cache update, repeated failures | 30min cache, clean fallback |
| 5000-core prompt scale | ~78 wasted API calls | ~1-3 wasted calls (race window only) |

### Tests

- 3 new tests for error handler changes (regional quota on-demand/spot, SKU family non-zero TTL)
- 11 new tests for PreLaunchFilter (nil fail-open, all/partial/no availability, multi-zone, spot/on-demand, large-scale circuit breaker, sequential progressive filtering scenario)

All existing tests pass. No new Azure API dependencies.

### DIFF
--- a/pkg/providers/instance/offerings/commonerrorhandlers.go
+++ b/pkg/providers/instance/offerings/commonerrorhandlers.go
@@ -41,7 +41,6 @@ var (
 	SKUNotAvailableSpotTTL           = 1 * time.Hour
 	SKUNotAvailableOnDemandTTL       = 23 * time.Hour
 	SKUFamilyQuotaNonZeroTTL         = 15 * time.Minute // Longer TTL for non-zero quota limits to prevent recycling before lower-weight NodePools are tried
-	RegionalQuotaExhaustedTTL        = 30 * time.Minute // Regional quota is unlikely to free up quickly
 )
 
 type errorHandle func(ctx context.Context, unavailableOfferings *cache.UnavailableOfferings, sku *skewer.SKU, instanceType *corecloudprovider.InstanceType, zone, capacityType, errorCode, errorMessage string) error
@@ -154,20 +153,6 @@ func handleOverconstrainedAllocationFailureError(ctx context.Context, unavailabl
 }
 
 func handleRegionalQuotaError(ctx context.Context, unavailableOfferings *cache.UnavailableOfferings, sku *skewer.SKU, instanceType *corecloudprovider.InstanceType, zone, capacityType, errorCode, errorMessage string) error {
-	// Regional quota is exhausted — no instance type of any size can be created for this capacity type.
-	// Mark offerings as unavailable with a long TTL so the scheduler doesn't keep retrying.
-	// Without this cache update, the scheduler would repeatedly select instance types from this capacity type,
-	// and each launch would fail with the same regional quota error, preventing fallback to lower-weight NodePools.
-	if capacityType == karpv1.CapacityTypeSpot {
-		unavailableOfferings.MarkSpotUnavailableWithTTL(ctx, RegionalQuotaExhaustedTTL)
-	} else {
-		// For on-demand regional quota, mark this specific instance type unavailable in all zones.
-		// This helps the scheduler skip it on subsequent loops. Since regional quota affects ALL instance types,
-		// each one that fails will also be marked, progressively draining the high-weight pool's available offerings
-		// until the scheduler falls through to lower-weight pools.
-		markAllZonesUnavailableForBothCapacityTypes(ctx, unavailableOfferings, instanceType, SubscriptionQuotaReachedReason, RegionalQuotaExhaustedTTL)
-	}
-
 	// InsufficientCapacityError is appropriate here because trying any other instance type will not help
 	return corecloudprovider.NewInsufficientCapacityError(
 		fmt.Errorf(
--- a/pkg/providers/instance/offerings/resperrorhandlers_test.go
+++ b/pkg/providers/instance/offerings/resperrorhandlers_test.go
@@ -247,28 +247,11 @@ func setupTestCases() []responseErrorTestCase {
 			expectAvailable(defaultTestOfferingInfo(testZone3, karpv1.CapacityTypeSpot)).
 			build(),
 
-		newTestCase("Regional quota exceeded for on-demand marks all zones unavailable").
+		newTestCase("Regional quota exceeded").
 			withInstanceType(zone2OnDemand, zone3Spot).
 			withZoneAndCapacity(testZone2, karpv1.CapacityTypeOnDemand).
 			withResponseError(sdkerrors.OperationNotAllowed, sdkerrors.RegionalQuotaExceededTerm).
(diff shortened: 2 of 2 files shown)


GuetaHen:
@microsoft-github-policy-service agree company="Microsoft"

GuetaHen:
This PR fixes #1323

pkg/providers/instance/offerings/commonerrorhandlers.go
```diff
@@ -148,6 +154,20 @@ func handleOverconstrainedAllocationFailureError(ctx context.Context, unavailabl
 }
 
 func handleRegionalQuotaError(ctx context.Context, unavailableOfferings *cache.UnavailableOfferings, sku *skewer.SKU, instanceType *corecloudprovider.InstanceType, zone, capacityType, errorCode, errorMessage string) error {
+	// Regional quota is exhausted — no instance type of any size can be created for this capacity type.
```
matthchr: Is this correct? This is for _regional_ quota errors. This means you have no remaining quota for _any_ instance type, AFAIK.

Basically AFAIK this error maps to the "Total regional vCPUs" error. If it were for a single family it would be the `isSKUFamilyQuotaExceeded` case.

I agree that we should probably do some update to the unavailableOfferings cache here, but I think it should be a short-lived one, not 30m like you've got here, and not only for a single instance type, because AFAIK this error is saying "nothing will work"

pkg/providers/instance/vminstance.go
```diff
@@ -217,6 +218,14 @@ func (p *DefaultVMProvider) BeginCreate(
 	instanceTypes []*corecloudprovider.InstanceType,
 ) (*VirtualMachinePromise, error) {
 	instanceTypes = offerings.OrderInstanceTypesByPrice(instanceTypes, scheduling.NewNodeSelectorRequirementsWithMinValues(nodeClaim.Spec.Requirements...))
+	// Pre-launch filter: re-check instance types against the live ICE cache.
+	// This acts as a circuit breaker during large-scale scheduling — if parallel NodeClaims
+	// have already failed and updated the cache, we skip those SKUs immediately.
+	instanceTypes = offerings.PreLaunchFilter(ctx, instanceTypes, offerings.NewLiveCacheAvailabilityCheck(
```
matthchr, line 224: Agree we need something that takes capacity into account at this layer, though I think it should probably be a "provider" of sorts (maybe `PriorityProvider` or something?) that integrates the following signals:
- Price
- Unavailable offerings (built from respones we get from the API like we're doing now)
- Quota (pre-check available quota and exclude or at least de-prioritize sizes that have low/no quota)
- Capacity - use capacity APIs to check capacity

These are all signals that play into "what size should we pick". Very similar to allocation strategy mentioned here: https://karpenter.sh/docs/faq/#how-does-karpenter-dynamically-select-instance-types

**What does this mean for this change?**
I'm working on something similar to the above, but don't necessarily want to block this on it as it's more involved... what I can do is do an initial refactoring take to give you a place to put this and then we can update this PR to use that, and then build/improve the provider to take additional signals in from there.

I'll update with a link here when I've sent that PR so you can see what I mean.

pkg/providers/instance/offerings/commonerrorhandlers.go
GuetaHen: You're right - regional quota (isRegionalQuotaExhausted) means "Total regional vCPUs exhausted", not a single family.
  Updated to:
   - TTL reduced from 30min → 15min (was too aggressive)
   - Now marks ALL instance types unavailable, not just the one that triggered the error - I added MarkOnDemandUnavailableWithTTL to the cache layer (mirroring the existing MarkSpotUnavailableWithTTL wildcard pattern), so a single cache entry blocks all on-demand offerings region-wide.

  Data backing the 15-min choice: I analyzed 7 days of real quota exhaustion data from our fleet using AIRS/AzureCM Kusto (EADSv5 as proxy since v6 families have no quota pressure yet):
  With a 15-min TTL, we catch ~89% of brief recovery windows in northeurope (our worst active region). 30min would miss 55% of them.

matthchr: I still am not sure that this makes sense. What we were doing before (directly returning insufficientCapacityError without modifying cache) seems correct/OK here?

For a regional quota error, at least _in theory_ no VM allocations will succeed. At this point, marking everything in the cache as blocked is just going to prevent all Karp VM allocation attempts for 15m.

Given that, it seems better to NOT block (what we were doing before) and let Karp keep trying.

What problem are you actually trying to solve here? you're talking about fallback and windows but again, this is regional quota for all families, if this gets hit nothing is going to allocate successfully so there's nothing to fall back to?

GuetaHen: You're right, fixed

matthchr:
I think this PR has been superceded by the one I just merged here:
https://github.com/Azure/karpenter-provider-azure/pull/1551

The filter bit we don't need (it already happens at provisioning time via InstanceTypes.List call).
The unavailable offerings cache, I did something similar in the above PR, but also fixed the "family" relationship when we hit quota errors so D8 renews D16's out of quota error (it didn't before).

If you disagree, please let me know, otherwise I will close this PR sometime next week. (after ~3/30)

matthchr:
Closing as I mentioned above.

PR #1434, opened 2026-02-17
chore: adding label runner to tests

MaximilianoUribe:
Fixes # 

**Description**
Adding label runners to tests that would be relevant to include in runners

I went with the label-based approach to avoid code duplication and to make test updates easier. The main trade-off is that we don’t get the full coverage a complete runner suite would provide, but I don’t see a real benefit in maintaining two nearly identical suites in different places

Opening discussion for the tests that are included

*

- [ ] Yes, issue opened: # 
- [X] No

**Release Note**

```release-note

```

### DIFF
--- a/.github/actions/e2e/create-cluster/action.yaml
+++ b/.github/actions/e2e/create-cluster/action.yaml
@@ -55,9 +55,9 @@ runs:
     - name: create cluster
       shell: bash
       env:
-        AZURE_SUBSCRIPTION_ID: ${{ inputs.subscription-id }}
-        AZURE_CLIENT_ID: ${{ inputs.client-id }}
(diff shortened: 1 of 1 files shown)


test/suites/gpu/suite_test.go
```diff
@@ -52,7 +52,7 @@ var _ = AfterEach(func() { env.AfterEach() })
 
 var _ = Describe("GPU", func() {
 	DescribeTable("should provision one GPU node and one GPU Pod",
-		Label("GPU"),
+		Label("GPU", "runner"),
```
matthchr: As discussed, may need to disable this

matthchr (approved):
LGTM but I think we may need to cut the GPU scenario

PR #1438, opened 2026-02-18
feat: add support for scheduling against kubernetes.azure.com/fips_enabled

matthchr:
* This adds the ability to simulate scheduling against the label but note that workloads cannot control the FIPS mode of their nodes by specifying kubernetes.azure.com/fips_enabled: true. The actual FIPS mode of provisioned nodes is still controlled by the field on the AKSNodeClass.
* As before, FIPS capability only works in NAP, not OSS Karpenter.

* Unit tests
* E2E tests
  * Run 1: https://github.com/Azure/karpenter-provider-azure/actions/runs/22164063622 - this discovered I missed adding a label to one E2E suite. Fixed for run 2.
  * Run 2: https://github.com/Azure/karpenter-provider-azure/actions/runs/22191927595
  * Run 3 (after behavior refactor): https://github.com/Azure/karpenter-provider-azure/actions/runs/22242212053

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
add support for scheduling against kubernetes.azure.com/fips_enabled
```

### DIFF
--- a/pkg/providers/instancetype/suite_test.go
+++ b/pkg/providers/instancetype/suite_test.go
@@ -140,6 +140,11 @@ var _ = Describe("InstanceType Provider", func() {
 	var nodePool *karpv1.NodePool
 
 	BeforeEach(func() {
+		// Reset testOptions and ctx in case a test edited them
+		// TODO: It would be nice to find a cleaner way to edit ctx/options in these tests...
+		testOptions = test.Options()
+		ctx = options.ToContext(ctx, testOptions)
+
 		nodeClass = test.AKSNodeClass()
 		test.ApplyDefaultStatus(nodeClass, env, testOptions.UseSIG)
 
@@ -157,7 +162,6 @@ var _ = Describe("InstanceType Provider", func() {
 			},
 		})
 
-		ctx = options.ToContext(ctx, test.Options())
 		cluster.Reset()
 		clusterNonZonal.Reset()
 		clusterBootstrap.Reset()


tallaxes (approved):

matthchr:
E2E uncovered a (test) issue, where I had failed to add the label to the `should support well-known labels for instance type selection` test.

I've fixed that and requeued a fresh E2E

pkg/providers/instancetype/suite_test.go
```diff
@@ -2336,6 +2339,22 @@ var _ = Describe("InstanceType Provider", func() {
 				{Name: v1beta1.AKSLabelScaleSetPriority + "=regular", Label: v1beta1.AKSLabelScaleSetPriority, ValueFunc: func() string { return "regular" }, ExpectedInKubeletLabels: true, ExpectedOnNode: true},
 				{Name: v1beta1.AKSLabelScaleSetPriority + "=spot", Label: v1beta1.AKSLabelScaleSetPriority, ValueFunc: func() string { return "spot" }, ExpectedInKubeletLabels: true, ExpectedOnNode: true},
 				{Name: v1beta1.AKSLabelOSSKU, Label: v1beta1.AKSLabelOSSKU, ValueFunc: func() string { return "Ubuntu" }, ExpectedInKubeletLabels: true, ExpectedOnNode: true},
+				{
+					Name:  v1beta1.AKSLabelFIPSEnabled,
+					Label: v1beta1.AKSLabelFIPSEnabled,
+					// Needs special setup because it only works on FIPS
+					SetupFunc: func() {
+						testOptions.UseSIG = true
+						ctx = options.ToContext(ctx, testOptions)
```
tallaxes, line 2352: Wouldn't this (once executed) persist for subsequent tests here? (Even if yes, maybe it does not matter since tests pass, but feels weird ...)

test/suites/scheduling/suite_test.go
```diff
@@ -84,6 +84,8 @@ var _ = Describe("Scheduling", Ordered, ContinueOnFailure, func() {
 			corev1.LabelWindowsBuild,
 			// VM SKU with GPU we are using does not populate this; won't be tested
 			v1beta1.LabelSKUGPUName,
+			// Can't test FIPS on self-hosted Karpenter
+			v1beta1.AKSLabelFIPSEnabled,
```
tallaxes: Could maybe make this (and test) conditional on `env.InClusterController`?

pkg/providers/instancetype/suite_test.go
matthchr: `ctx = options.ToContext(ctx, test.Options())` is in BeforeEach, so I think no?

test/suites/scheduling/suite_test.go
matthchr: yeah, fair enough - added a test and made this conditional

pkg/providers/instancetype/suite_test.go
tallaxes: Hmm, yes, so ctx should be fine. But top-level BeforeEach also does `test.ApplyDefaultStatus(nodeClass, env, testOptions.UseSIG)`, do we care about images in status? Probably not ...

tallaxes (dismissed):

pkg/providers/instancetype/suite_test.go
matthchr: Ok I fixed this by resetting test options too, and moving the ctx reset to the top of the `BeforeEach`. Good catch.

I _think_ this fixes it but it's still a bit icky. I left a TODO there as well as I didn't see an obvious better way without more sweeping test initialization changes.

PR #1452, opened 2026-02-25
feat: instance promise WaitGroup for async provisioning tracking

comtalyst:
Fixes # 

**Description**

With async provisioning, the async routine may still run in the background while unit test assertions (e.g., [ExpectProvisioned](https://github.com/Azure/karpenter-provider-azure/blob/e79c1b876053a05ba4bb6b1465a95562e411b45d/pkg/cloudprovider/suite_aksmachineapi_integration_test.go#L756-L757) after [cloudProvider.Create()](https://github.com/Azure/karpenter-provider-azure/blob/e79c1b876053a05ba4bb6b1465a95562e411b45d/pkg/cloudprovider/suite_aksmachineapi_integration_test.go#L397)) are being done. 
This PR adds a mechanism to wait for those async routine to be done. Currently, only unit tests will be using it. But it is open to more production uses.

Although, so far, we do not have any assertion that relies on the state after async provisioning. The first of that will be the [GET-based polling](https://github.com/Azure/karpenter-provider-azure/pull/1453) with [batched AKS machine creates](https://github.com/Azure/karpenter-provider-azure/pull/1455).

* Same test w/ Machine API

- [ ] Yes, issue opened: # 
- [X] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/cloudprovider/suite_aksmachineapi_drift_test.go
+++ b/pkg/cloudprovider/suite_aksmachineapi_drift_test.go
@@ -71,6 +71,7 @@ var _ = Describe("CloudProvider", func() {
 		})
 
 		AfterEach(func() {
+			// Wait for any async polling goroutines to complete before resetting
 			cloudProvider.WaitForInstancePromises()
 			cluster.Reset()
 			azureEnv.Reset()
--- a/pkg/cloudprovider/suite_aksmachineapi_features_test.go
+++ b/pkg/cloudprovider/suite_aksmachineapi_features_test.go
@@ -73,6 +73,7 @@ var _ = Describe("CloudProvider", func() {
 		})
 
 		AfterEach(func() {
+			// Wait for any async polling goroutines to complete before resetting
 			cloudProvider.WaitForInstancePromises()
 			cluster.Reset()
 			azureEnv.Reset()
--- a/pkg/cloudprovider/suite_aksmachineapi_integration_test.go
+++ b/pkg/cloudprovider/suite_aksmachineapi_integration_test.go
@@ -639,6 +639,7 @@ var _ = Describe("CloudProvider", func() {
 		})
 
 		AfterEach(func() {
+			// Wait for any async polling goroutines to complete before resetting
 			cloudProvider.WaitForInstancePromises()
 			cluster.Reset()
 			azureEnv.Reset()
@@ -677,6 +678,7 @@ var _ = Describe("CloudProvider", func() {
 		})
 
 		AfterEach(func() {
+			// Wait for any async polling goroutines to complete before resetting
 			cloudProvider.WaitForInstancePromises()
 			cluster.Reset()
 			azureEnv.Reset()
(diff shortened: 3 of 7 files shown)


comtalyst (commented):
(AI-generated)

Minimal, surgical change. Three production lines, zero risk. Test helpers are well-designed. No issues.

comtalyst (commented):
(AI-generated)

Reviewed. This is clean and well-structured. The `sync.WaitGroup` on `CloudProvider` is the right primitive — simple, no goroutine leaks, and the test helpers are well-abstracted behind the `instancePromiseWaiter` interface to avoid import cycles.

A few observations, but nothing blocking:

1. **`WaitForInstancePromises` is public** — good forward-thinking since you noted it's "open to more production uses." If it turns out we never need it outside tests, we could make it private later, but exposing it now is the right call given the GET-based poller in #1453 will need it.

2. The `CreateAndWaitForPromises` helper does the right thing by simulating the Launched condition set (which the core lifecycle controller does in production). The fresh GET + status-only update is correct for avoiding spec immutability conflicts.

3. The bulk of the diff is mechanical `ExpectProvisioned` → `ExpectProvisionedAndWaitForPromises` replacements across test files, which is expected. The pattern is consistent.

LGTM — no substantive concerns.

comtalyst:
(AI-generated)

## Design note: why `WaitForInstancePromises()` stays in test `AfterEach` rather than moving into `azureEnv.Reset()` or `cluster.Reset()`

Considered moving the `cloudProvider.WaitForInstancePromises()` call into `azureEnv.Reset()` or `cluster.Reset()` to reduce one-line boilerplate across test suites. Decided against it for the following reasons:

**`azureEnv.Reset()` — coupling problem.** `test.Environment` doesn't hold a reference to `CloudProvider`. The `CloudProvider` (which owns the `sync.WaitGroup`) is constructed *outside* `azureEnv` in each test suite's `BeforeSuite`/`BeforeEach`. To make this work, we'd need to either pass `cloudProvider` into `azureEnv` after construction (creating a circular-feeling dependency, since `azureEnv` is used to *build* the `CloudProvider`) or have `azureEnv` create the `CloudProvider` itself (a bigger refactor that changes the test setup pattern).

**`cluster.Reset()` — not our code.** `cluster` is `*state.Cluster` from core karpenter (`sigs.k8s.io/karpenter/pkg/controllers/state`). Its `Reset()` has no concept of cloud-provider-specific cleanup.

**Ordering matters.** Promises must complete *before* resetting fake APIs, otherwise goroutines still running from `handleInstancePromise` would hit reset/nil fakes and panic. Embedding the wait inside `azureEnv.Reset()` would require `azureEnv` to know about `cloudProvider` just to sequence this correctly.

**Current cost is low.** The boilerplate is one line per `AfterEach`, and the `ExpectProvisionedAndWaitForPromises` / `CreateAndWaitForPromises` helpers in `expectations.go` already absorb the majority of wait calls in the happy path. The `AfterEach` calls are the safety net for test failures where the helper wasn't reached.

PR #1451, opened 2026-02-24
refactor: separate azclient into its own package and rename opts to clientopts

comtalyst:
Fixes # 

**Description**

- Move AZClient, API interfaces, and client constructors from pkg/providers/instance/ to pkg/providers/azclient/
- Rename pkg/utils/opts/ to pkg/utils/clientopts/
- Add accessor methods on AZClient for private fields
- Update all import references across the codebase

The first beneficiary of this is batch machine API.

**How was this change tested?**

* Same test w/ Machine API

**Does this change impact docs?**
- [ ] Yes, PR includes docs updates
- [ ] Yes, issue opened: # 
- [X] No

**Release Note**

```release-note

```

PR #1454, opened 2026-02-25
feat: allow spot nodes on system mode for AKS machine API

comtalyst:
Fixes

**Description**

Add per-call Azure SDK policy that sets `AllowSpotVMSystemNode: true` header on PUT requests through the AKS Machines client, enabling spot instances on system-mode node pools.

The RP-side support for this was recently updated: `AllowSpotVMSystemNode` is now a standalone boolean HTTP header, not embedded in `AKSHTTPCustomFeatures`. `UseCustomizedOSImage` is handled server-side by the RP and does not need to be sent by Karpenter.

Depends on #1451 (azclient-refactor) — already merged.

* Verified spot + system mode machine creation via NAP (Machine API) on standalone with RP fix. Node labels: `kubernetes.azure.com/mode=system`, `karpenter.sh/capacity-type=spot`, `kubernetes.azure.com/agentpool=aksmanagedap`

- [X] No

**Release Note**

```release-note

```


comtalyst (commented):
(AI generated)

Clean and correct implementation. One thing to be aware of:

### `Header.Set` vs `Header.Add`

The per-call policy uses `Header.Set` which replaces the entire `AKSHTTPCustomFeatures` header value. If any other SDK policy or middleware also sets this header in the future, this policy will silently overwrite it. The comment in the code already acknowledges this. Consider using `Header.Add` or reading the existing value and appending. Not blocking since no other policies currently set this header.

comtalyst (commented):
Reviewed. The approach is correct — per-call policy on the Machines client is the right injection point. The copy of `opts` to avoid mutating the shared options is good.

**One substantive concern:**

The `Header.Set` call replaces the entire `AKSHTTPCustomFeatures` header value. The comment acknowledges this by re-adding `UseCustomizedOSImage`. But this is fragile: if any other code path (e.g., another policy, or the SDK itself) sets this header before our policy runs, we'd silently drop those values.

A more defensive approach would be to use `Header.Add` or read the existing header value and append to it:

\`\`\`go
func (p *spotSystemNodePolicy) Do(req *policy.Request) (*http.Response, error) {
    if req.Raw().Method == http.MethodPut {
        existing := req.Raw().Header.Get("AKSHTTPCustomFeatures")
        features := "Microsoft.ContainerService/UseCustomizedOSImage,Microsoft.ContainerService/AllowSpotVMSystemNode"
        if existing != "" {
            features = existing + "," + features
        }
        req.Raw().Header.Set("AKSHTTPCustomFeatures", features)
    }
    return req.Next()
}
\`\`\`

This way we don't need to manually track which features were previously set, and we're safe if the SDK or another policy ever adds to this header.

Not blocking merge, but worth considering.

LGTM otherwise — small, focused, correct intent.

PR #1460, opened 2026-02-27
chore: remove rapid gc polling at start

matthchr:
There's no need for this and it increases requests sent from a throttling perspective

* Unit Tests
* E2E: https://github.com/Azure/karpenter-provider-azure/actions/runs/22498890021

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note

```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -44,6 +44,7 @@ require (
 	github.com/prometheus/client_golang v1.23.2
 	github.com/prometheus/client_model v0.6.2
 	github.com/samber/lo v1.52.0
+	github.com/stretchr/testify v1.11.1
 	go.uber.org/multierr v1.11.0
 	go.yaml.in/yaml/v2 v2.4.3
 	golang.org/x/sync v0.19.0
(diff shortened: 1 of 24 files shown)


matthchr:
@tallaxes if there's a reason we're wanting to keep this behavior LMK and I'm happy to do so. I was just going through optimizations and saw this and I don't see a lot of need in fast requeue after the first time through because we'll clean everything up on boot so checking again in a short amount of time doesn't seem super valuable? If you remember why we had this though I can leave a comment.

matthchr:
For fail-forward/fail-back I assume we're keeping it for now, but @comtalyst should weigh in.

tallaxes:
> @tallaxes if there's a reason we're wanting to keep this behavior LMK and I'm happy to do so. I was just going through optimizations and saw this and I don't see a lot of need in fast requeue after the first time through because we'll clean everything up on boot so checking again in a short amount of time doesn't seem super valuable? If you remember why we had this though I can leave a comment.

The assumption was that VM orphaning is likely to be somewhat clustered - that is, new orphans are more likely to be discovered shortly after existing orphans were discovered - and that there are benefits to cleaning up quickly. But this optimization may well have been premature, or incorrect (if the observations don't support it), or costly. I am fine with removing it.

comtalyst:
@tallaxes @matthchr we still need instance garbage collection for Machine API, which will treat a machine as if it is a VM (e.g., handle not having associated NodeClaim).  But we don't need NIC garbage collection.

The current state of code already handle this. NIC garbage collection only sees NIC from non-Machine, and "instance" garbage collection includes machines.

PR #1461, opened 2026-02-27
fix: reduce IsDrifted throttling by removing subnet drift and improving imageID drift

matthchr:
Reduce API call usage in IsDrifted by:
* Using nodeClaim.Status.ImageID instead of a call to Azure to get image ID, for both machine and VM codepaths.
* Using static drift (via nodeclass hash) instead of going to the Azure to check the NIC for subnet drift.

* Unit tests
* E2E: https://github.com/Azure/karpenter-provider-azure/actions/runs/22498890021

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/controllers/nodeclaim/garbagecollection/instance_garbagecollection.go
+++ b/pkg/controllers/nodeclaim/garbagecollection/instance_garbagecollection.go
@@ -41,16 +41,14 @@ import (
 )
 
 type Instance struct {
-	kubeClient      client.Client
-	cloudProvider   corecloudprovider.CloudProvider
-	successfulCount uint64 // keeps track of successful reconciles for more aggressive requeuing near the start of the controller
+	kubeClient    client.Client
+	cloudProvider corecloudprovider.CloudProvider
 }
 
 func NewInstance(kubeClient client.Client, cloudProvider corecloudprovider.CloudProvider) *Instance {
 	return &Instance{
-		kubeClient:      kubeClient,
-		cloudProvider:   cloudProvider,
-		successfulCount: 0,
+		kubeClient:    kubeClient,
+		cloudProvider: cloudProvider,
 	}
 }
 
@@ -76,7 +74,7 @@ func (c *Instance) Reconcile(ctx context.Context) (reconciler.Result, error) {
 	if err := c.kubeClient.List(ctx, nodeList); err != nil {
 		return reconciler.Result{}, err
 	}
-	clusterProviderIDs := sets.New[string](lo.FilterMap(clusterNodeClaims.Items, func(n karpv1.NodeClaim, _ int) (string, bool) {
+	clusterProviderIDs := sets.New(lo.FilterMap(clusterNodeClaims.Items, func(n karpv1.NodeClaim, _ int) (string, bool) {
 		return n.Status.ProviderID, n.Status.ProviderID != ""
 	})...)
 	errs := make([]error, len(cloudNodeClaims))
@@ -95,8 +93,7 @@ func (c *Instance) Reconcile(ctx context.Context) (reconciler.Result, error) {
 	if err = multierr.Combine(errs...); err != nil {
 		return reconciler.Result{}, err
 	}
-	c.successfulCount++
-	return reconciler.Result{RequeueAfter: lo.Ternary(c.successfulCount <= 20, time.Second*10, time.Minute*2)}, nil
+	return reconciler.Result{RequeueAfter: time.Minute * 2}, nil
 }
 
 func (c *Instance) garbageCollect(ctx context.Context, nodeClaim *karpv1.NodeClaim, nodeList *v1.NodeList) error {


pkg/cloudprovider/drift.go
```diff
@@ -201,98 +179,24 @@ func (c *CloudProvider) isImageVersionDrifted(
...
-	logger.V(1).Info("drift triggered as actual image id was not found in the set of currently available node images",
+	logger.V(1).Info("drift triggered as actual image version was not found in the set of currently available node images",
 		"driftType", ImageDrift,
-		"actualImageID", vmImageID)
+		"actualImageVersion", nodeClaim.Status.ImageID)
 	return ImageDrift, nil
 }
 
-// isSubnetDrifted returns drift if the nic for this nodeclaim does not match the expected subnet
-func (c *CloudProvider) isSubnetDrifted(ctx context.Context, nodeClaim *karpv1.NodeClaim, nodeClass *v1beta1.AKSNodeClass) (cloudprovider.DriftReason, error) {
```
matthchr, line 269: This is already covered by NodeClass spec drift, because (AFAIK) it's not possible in AKS to have an opts -> opts Subnet drift... the subnet field is write once on the cluster.

So the only way the subnet can change is it being set (or unset) on the NodeClass.Spec.

matthchr: I can leave this in a comment in the static drift function if folks think that is useful.

matthchr: I toyed with the idea of adding another annotation/label with another hash, like `GlobalOptionsHash` or something, just to guard against the chance of global drift, but it seemed overkill for this one field... so didn't do it here.

happy to discuss if folks think that's something we should consider... only case I can think of it is something like OSS Karpenter that's misconfigured to use wrong subnet, then updated to use the right one, it won't drift the old nodes automatically now whereas it would've before.

pkg/cloudprovider/drift.go
```diff
@@ -75,7 +72,6 @@ func (c *CloudProvider) isNodeClassDrifted(ctx context.Context, nodeClaim *karpv
 			c.isK8sVersionDrifted,
 			c.isKubeletIdentityDrifted,
 			c.isImageVersionDrifted,
-			c.isSubnetDrifted,
```
tallaxes, line 78: nit: `SubnetDrift` const is now also unused

tallaxes (approved):
LGTM. Great to have this cleanup! (Some earlier discussion in  https://github.com/Azure/karpenter-provider-azure/pull/486#discussion_r1791275488)

pkg/cloudprovider/drift.go
tallaxes: Yeah, thinking the same, drift on change in global options is asking to be born for a while now - but maybe not yet :-)

pkg/cloudprovider/drift.go
comtalyst, line 78: Could you state this intention in the PR title? Would make historical investigations easier.

pkg/cloudprovider/drift.go
comtalyst: > I toyed with the idea of adding another annotation/label with another hash, like `GlobalOptionsHash`

Do you mean for cluster-wide values?
For Machine API (which would soon apply for OSS), that would be covered by `DriftAction` (as they are RP-controlled). But might still be fair as a framework for non-AKS use cases though(?)

pkg/cloudprovider/drift.go
```diff
@@ -169,30 +165,11 @@ func (c *CloudProvider) isImageVersionDrifted(
...
-		if aksMachine.Properties == nil || aksMachine.Properties.NodeImageVersion == nil {
-			// TODO (charliedmcb): this seems like an error case to me, but maybe not one to hard fail on? Is it even possible?
-			//     We may want to collect this with the other errors up a level as to not block other drift conditions.
-			return "", nil
-		}
-
-		aksMachineNodeImageVersion := lo.FromPtr(aksMachine.Properties.NodeImageVersion) // E.g., AKSUbuntu-2204gen2containerd-2022.10.03
+	if nodeClaim.Status.ImageID == "" {
+		return "", fmt.Errorf("no image ID found in nodeClaim status")
+	}
```
comtalyst, line 170: Can we technically reach this point (e.g., core call this while cloudprovider instance creation isn't done)? A comment/justification on that could be useful.

pkg/cloudprovider/drift.go
matthchr: yeah I mean for cluster wide values

comtalyst: But yes, something similar for NAP (e.g., passing this hash through the deployment) is still a useful workaround for "disruption controller is purely event-driven" issue we discussed earlier. But I guess the difference is the need for actual visibility into those values.

pkg/cloudprovider/drift.go
matthchr: Which intention? Not 100% sure I am following. Removing isSubnetDrifted?

PR title because you want to search it on GitHub or you want it in Git commit history? for commit history, I'll make sure when I merge to make sure my description bit here:
```
Reduce API call usage in IsDrifted by:

Using nodeClaim.Status.ImageID instead of a call to Azure to get image ID, for both machine and VM codepaths.
Using static drift (via nodeclass hash) instead of going to the Azure to check the NIC for subnet drift.
```

is in the commit message, so you should be able to find it via `git log --grep` searching for subnet drift.

pkg/cloudprovider/drift.go
matthchr: Yes, I think core can call this when creation isn't done. AWS has this same check in their image version drift.

matthchr: added a small comment to that effect.

pkg/cloudprovider/drift.go
matthchr: Good catch, removed

comtalyst: In PR title, as the commits are squashed. Example uses are in:
- Main branch's commit history, for timing of changes
- Individual file history
- `git blame` for individual line history

> Which intention?

Yes, removing `isSubnetDrifted`. Maybe "remove ineffective subnet drift" etc.

pkg/cloudprovider/drift.go
matthchr: and yes one of the reasons I didn't add it was because it seemed sorta pointless in a machine API world.

pkg/cloudprovider/drift.go
matthchr: Commits are squashed but you can control the commit message yourself manually (which I sometimes do), so it will be there in the commit message.

but I also changed the title to `fix: reduce IsDrifted throttling by removing subnet drift and improving imageID drift` which is a bit long for my taste but I think covers the 2 cases.

PR #1462, opened 2026-02-27
fix: reduce unneeded Get calls during delete

matthchr:

* Unit tests
* E2E: https://github.com/Azure/karpenter-provider-azure/actions/runs/22602466369

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/controllers/nodeclaim/inplaceupdate/controller.go
+++ b/pkg/controllers/nodeclaim/inplaceupdate/controller.go
@@ -252,6 +252,7 @@ func (c *Controller) Register(_ context.Context, m manager.Manager) error {
 		For(
 			&karpv1.NodeClaim{},
 			builder.WithPredicates(
+				nodeclaimutils.UsingAKSNodeClassPredicate(),
 				predicate.Or(
 					predicate.GenerationChangedPredicate{}, // Note that this will trigger on pod restart for all Machines.
 				),
--- a/pkg/utils/nodeclaim/nodeclaim.go
+++ b/pkg/utils/nodeclaim/nodeclaim.go
@@ -24,6 +24,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	karpv1 "sigs.k8s.io/karpenter/pkg/apis/v1"
 
 	armcompute "github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/compute/armcompute/v7"
@@ -32,6 +33,20 @@ import (
 	"github.com/Azure/karpenter-provider-azure/pkg/providers/instance"
 	"github.com/Azure/karpenter-provider-azure/pkg/utils"
 )
+
+// UsingAKSNodeClassPredicate creates a predicate to filter node claim using AKS node class.
+func UsingAKSNodeClassPredicate() predicate.Funcs {
+	return predicate.NewPredicateFuncs(func(object client.Object) bool {
+		nodeClaim, ok := object.(*karpv1.NodeClaim)
+		if !ok {
+			return false
+		}
+		if nodeClaim.Spec.NodeClassRef == nil {
+			return false
+		}
+		return nodeClaim.Spec.NodeClassRef.Kind == "AKSNodeClass"
+	})
+}
 
 // GetAKSNodeClass resolves the AKSNodeClass from the NodeClaim's NodeClassRef.
 // If the NodeClass for the nodeClaim has DeletionTimestamp set, an error is returned.


pkg/providers/instance/aksmachineinstance.go
```diff
@@ -132,6 +134,8 @@ type DefaultAKSMachineProvider struct {
 	aksMachinesPoolName     string // Only support one AKS machine pool at a time, for now.
 	aksMachinesPoolLocation string
 	errorHandling           *offerings.ErrorDetailHandler
+	deletingMachines        sets.Set[string] // tracks in-flight delete operations by machine name
+	deletingMachinesMu      sync.RWMutex
```
tallaxes, line 139: Maybe `sync.Map`?

tallaxes (approved):

pkg/providers/instance/aksmachineinstance.go
```diff
@@ -369,7 +382,13 @@ func (p *DefaultAKSMachineProvider) deleteMachine(ctx context.Context, aksMachin
 		return fmt.Errorf("failed to begin delete AKS machine %q: %w", aksMachineName, err)
 	}
 
+	p.deletingMachinesMu.Lock()
+	p.deletingMachines.Insert(aksMachineName)
+	p.deletingMachinesMu.Unlock()
 	_, err = poller.PollUntilDone(ctx, nil)
+	p.deletingMachinesMu.Lock()
+	p.deletingMachines.Delete(aksMachineName)
+	p.deletingMachinesMu.Unlock()
```
comtalyst: Could be more defensive (e.g., against later-caught panic)/clean with `defer`?

pkg/providers/instance/aksmachineinstance.go
```diff
@@ -369,7 +382,13 @@ func (p *DefaultAKSMachineProvider) deleteMachine(ctx context.Context, aksMachin
 		return fmt.Errorf("failed to begin delete AKS machine %q: %w", aksMachineName, err)
 	}
 
+	p.deletingMachinesMu.Lock()
+	p.deletingMachines.Insert(aksMachineName)
+	p.deletingMachinesMu.Unlock()
```
comtalyst: Why don't we cover `BeginDeleteMachines()` too? Thought that would also prevent duplicated `BeginDeleteMachines()` and unusually long `BeginDeleteMachines()`.

pkg/providers/instance/aksmachineinstance.go
```diff
@@ -278,6 +283,14 @@ func (p *DefaultAKSMachineProvider) Delete(ctx context.Context, aksMachineName s
...
 	}
 
+	// If there's already an in-flight delete for this machine, return immediately.
+	p.deletingMachinesMu.RLock()
+	deleting := p.deletingMachines.Has(aksMachineName)
+	p.deletingMachinesMu.RUnlock()
+	if deleting {
+		return nil
+	}
+
```
comtalyst, line 294: This will not completely prevent multiple GETs, given the "acquisition" (i.e., `deletingMachines.Insert()`) occurs after `Get()`. But should cover most cases as it is `deleteMachine()` that runs long.
The alternatives seems not easy, given the two-layered (one with `Get()`, one without) approach we have.

Though, this fact is better noted as a comment here?

pkg/providers/instance/aksmachineinstance.go
matthchr: I actually avoided it because (AFAIK) there are some mem usage risks with sync.Map in certain access patterns and in high write percentage cases it's better to just use a RWMutex anyway

See https://oneuptime.com/blog/post/2026-01-25-sync-map-vs-mutex-maps-go/view for one discussion of this (there are others):

> Quick Decision Framework
> Ask these questions about your use case:
>
> Is the map written once and read many times? Use sync.Map.
> Do goroutines access mostly disjoint keys? Use sync.Map.
> Do you need len(), iteration, or bulk operations? Use mutex + map.
> Is your write ratio above 10-20%? Use mutex + map.
> Is the access pattern unpredictable or changing? Start with mutex + map. It's easier to reason about and optimize later.

Our write ratio is (probably) above 10-20% here. I think delete usually takes 30s so we'll have ~6 reads and 1 write to add + 1 to delete for a ratio of ~25%. ofc I didn't actually benchmark but even the stdlib says:
```
// The Map type is specialized. Most code should use a plain Go map instead,
// with separate locking or coordination, for better type safety and to make it
// easier to maintain other invariants along with the map content.
//
// The Map type is optimized for two common use cases: (1) when the entry for a given
// key is only ever written once but read many times, as in caches that only grow,
// or (2) when multiple goroutines read, write, and overwrite entries for disjoint
// sets of keys. In these two cases, use of a Map may significantly reduce lock
// contention compared to a Go map paired with a separate [Mutex] or [RWMutex].
//
```

I _think_ delete is a second write so the first case is out.
Second case, we dont have disjoint sets of keys because different goroutines (== workers in controller-runtime) may read/write the same keys (though the same goroutine should always do both writes, others may read).

pkg/providers/instance/vminstance.go
```diff
@@ -941,6 +953,46 @@ func (p *DefaultVMProvider) cleanupAzureResources(ctx context.Context, resourceN
...
+		vmName,
+		&armcompute.VirtualMachinesClientBeginDeleteOptions{ForceDeletion: lo.ToPtr(true)},
+	)
+	if err != nil {
+		return err
+	}
+
+	p.deletingVMsMu.Lock()
+	p.deletingVMs.Insert(vmName)
+	p.deletingVMsMu.Unlock()
```
comtalyst: (same comment with https://github.com/Azure/karpenter-provider-azure/pull/1462/changes#r2866237646)

pkg/providers/instance/vminstance.go
```diff
@@ -941,6 +953,46 @@ func (p *DefaultVMProvider) cleanupAzureResources(ctx context.Context, resourceN
...
+
+	_, err = poller.PollUntilDone(ctx, nil)
+	p.deletingVMsMu.Lock()
+	p.deletingVMs.Delete(vmName)
+	p.deletingVMsMu.Unlock()
```
comtalyst: (same comment with https://github.com/Azure/karpenter-provider-azure/pull/1462/changes#r2866232841)

pkg/providers/instance/aksmachineinstance.go
tallaxes: Ack, TIL. Not totally sure the wins beat the simplicity here, but Yes it looks like a better match for the pattern. (It is a pity one has to think about it ...)

Also wondering if we should review other uses of `sync.Map` in the codebase, we have a few ...

pkg/providers/instance/aksmachineinstance.go
matthchr: Good call, I'll move it up to above `BeginDeleteMachines`

pkg/providers/instance/aksmachineinstance.go
matthchr: In general the pattern I have seen with these "lock multiple times in a single method" is that they don't use defer on the locks and just do them inline, and anyway if we have a panic in here we do have a bigger problem because if we recover it somewhere above we'll possibly fail to remove the item from `deletingMachines` which wouldn't be great (depends on where we panic, I guess, but the likely places would be `PollUntilDone` or `BeginDeleteMachines` and both would leak a `deletingMachines` entry, so we're already somewhat reliant on them not panicing)

matthchr: I guess what I _could_ do (curious what @tallaxes thinks) is put the whole `lock + delete + unlock` in a defer right after the insert, which solves both of those problems...

pkg/providers/instance/aksmachineinstance.go
matthchr: Did this

pkg/providers/instance/aksmachineinstance.go
matthchr: The comment I would think to put here is something like:
```
// This get exists to deal with the case where the operator restarted during the course of a deletion.
// With it, we may do an extra unneeded get before delete, but without it we may erroneously issue 
// 2 deletes if the instance was being deleted and the operator restarted.
// Since get quota is generally higher, we prefer to check w/ get rather than issue 2 deletes
```

WDYT? If you like, I can add in both machine and VM spots.

pkg/providers/instance/vminstance.go
matthchr: Moved it up

pkg/providers/instance/aksmachineinstance.go
comtalyst: > if we have a panic in here we do have a bigger problem because if we recover it somewhere above we'll possibly fail to remove the item from deletingMachines

`defer` would help ensure the removal though? Also curious why those pattern you have seen don't prefer `defer`.
(yes, I meant "whole `lock + delete + unlock` in a defer")

pkg/providers/instance/aksmachineinstance.go
comtalyst: In fact, I wasn't sure if we really need to have this extra `Get()` (even in restart case, 404 on delete isn't that bad, and did we really care about DELETE throttling more than GET throttling?). It is a candidate to be checked later on both modes. But, if you are confident that we don't need it, I'm good with you doing that here.

Otherwise, my suggestion was more about "how this 2-layer delete will work with locks" rather than "why do we need 2-layer delete". Though having both is better.

matthchr: It may not be 404 on delete it may be that the operation is already ongoing on delete (409?), but 2x delete is more quota-expensive than 1x Get 1x Delete so I still think this way makes sense? I'll add a comment though

pkg/providers/instance/aksmachineinstance.go
matthchr: Done

PR #1465, opened 2026-03-03
feat: adding new gpu support

MaximilianoUribe:
Fixes # 

**Description**

*

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/providers/instancetype/instancetype.go
+++ b/pkg/providers/instancetype/instancetype.go
@@ -269,10 +269,15 @@ func computeCapacity(ctx context.Context, sku *skewer.SKU, nodeClass *v1beta1.AK
 	}
 
 	// Add GPU capacity based on manufacturer
+	// Always include nvidia.com/gpu in capacity (set to 0 for non-GPU nodes)
 	skuName := sku.GetName()
 	if utils.IsNvidiaEnabledSKU(skuName) {
 		capacity[corev1.ResourceName("nvidia.com/gpu")] = *gpuNvidiaCount(sku)
-	} else if utils.IsAMDGPUEnabledSKU(skuName) {
+	} else {
+		capacity[corev1.ResourceName("nvidia.com/gpu")] = *resources.Quantity("0")
+	}
+
+	if utils.IsAMDGPUEnabledSKU(skuName) {
 		capacity[corev1.ResourceName("amd.com/gpu")] = *gpuAMDCount(sku)
 	}
 


pkg/providers/instancetype/instancetype.go
```diff
@@ -230,11 +230,17 @@ func setRequirementsHyperVGeneration(requirements scheduling.Requirements, sku *
...
-	if utils.IsNvidiaEnabledSKU(sku.GetName()) {
+	skuName := sku.GetName()
+	if utils.IsNvidiaEnabledSKU(skuName) {
 		requirements[v1beta1.LabelSKUGPUManufacturer].Insert(v1beta1.ManufacturerNvidia)
 		if vmsize.AcceleratorType != nil {
 			requirements[v1beta1.LabelSKUGPUName].Insert(*vmsize.AcceleratorType)
 		}
+	} else if utils.IsAMDGPUEnabledSKU(skuName) {
+		requirements[v1beta1.LabelSKUGPUManufacturer].Insert(v1beta1.ManufacturerAMD)
+		if vmsize.AcceleratorType != nil {
```
matthchr, line 241: minor, style:

```suggestion
		if vmSize.AcceleratorType != nil {
```

pkg/providers/instancetype/instancetype.go
```diff
@@ -255,16 +261,31 @@ func getArchitecture(architecture string) string {
...
+		corev1.ResourceCPU:              *cpu(sku),
+		corev1.ResourceMemory:           *memoryWithoutOverhead(ctx, sku),
+		corev1.ResourceEphemeralStorage: *ephemeralStorage(nodeClass),
+		corev1.ResourcePods:             *pods(ctx, nodeClass),
+	}
+
+	// Add GPU capacity based on manufacturer
+	// Always include nvidia.com/gpu in capacity (set to 0 for non-GPU nodes)
+	skuName := sku.GetName()
+	if utils.IsNvidiaEnabledSKU(skuName) {
```
matthchr, line 274: We were already doing this above:
```
corev1.ResourceName("nvidia.com/gpu"): *gpuNvidiaCount(sku),
```

so I think the `gpuNvidiaCount` method already returns 0 if the SKU isn't GPU enabled, so I think you can just drop this if/else and continue doing what we were doing above?

pkg/providers/instancetype/instancetype.go
```diff
@@ -255,16 +261,31 @@ func getArchitecture(architecture string) string {
...
+		capacity[corev1.ResourceName("nvidia.com/gpu")] = *gpuNvidiaCount(sku)
+	} else {
+		capacity[corev1.ResourceName("nvidia.com/gpu")] = *resources.Quantity("0")
+	}
+
+	if utils.IsAMDGPUEnabledSKU(skuName) {
```
matthchr, line 280: Are we supposed to have conditional capacity considerations (where the label sometimes is returned, but sometimes isn't?)

That's not what we're doing for NVidia GPUs which makes me think it's also probably not what we should be doing for AMD?

pkg/utils/gpu.go
```diff
@@ -33,40 +33,52 @@ const (
...
-	}
-	for _, skus := range nvidiaSKUConfig.MarinerNvidiaEnabledSKUFamilies {
-		for _, sku := range skus {
-			marinerNvidiaEnabledSKUs[sku] = true
+		if len(config.OS) == 0 {
+			panic("os field must not be empty for SKU: " + sku)
+		}
+		for _, os := range config.OS {
+			// NOTE: Windows is validated here for future support, but Windows nodes are not currently supported
+			if os != OSUbuntu && os != OSAzureLinux && os != OSAzureLinux3 && os != OSWindows {
```
matthchr, line 79: minor: probably cleaner to define these as an array and then use something like `slices.Contains()` or `lo.Contains()`.

Array can be a static var (or at least defined outside this loop)

pkg/utils/gpu.go
```diff
@@ -33,40 +33,52 @@ const (
 
 	NvidiaGridDriverVersion = "550.144.06"
 	AKSGPUGridVersionSuffix = "20250512225043"
+
+	// GPU types
+	GPUNvidia = "nvidia"
+	GPUAMD    = "amd"
+
+	// OS types
+	OSUbuntu      = "ubuntu"
```
matthchr, line 42: What exactly does this OS bit mean?
The device works regardless of OS... doesn't it? I guess the thing that is maybe confusing me is that I _think_ this OS bit is about "which OSes will we automatically install the device drivers for?"

Which sorta made sense pre-AMD skus, but now that we support AMD skus (which we don't ever install drivers for) it makes me wonder why we don't also expose the nvidia SKUs which we don't automatically install the drivers for, because the users could still manually install the drivers?

Or am I missing something?

pkg/utils/gpu.go
```diff
@@ -83,15 +95,55 @@ func IsNvidiaEnabledSKU(vmSize string) bool {
...
 	// Trim the optional _Promo suffix.
 	vmSize = strings.ToLower(vmSize)
 	vmSize = strings.TrimSuffix(vmSize, "_promo")
-	return marinerNvidiaEnabledSKUs[vmSize]
+	return IsSKUSupportedOnOS(vmSize, OSAzureLinux)
+}
+
+// IsAMDGPUEnabledSKU determines if a VM SKU has AMD GPU driver support
+func IsAMDGPUEnabledSKU(vmSize string) bool {
+	vmSize = strings.ToLower(vmSize)
```
matthchr, line 112: This bit is duplicated 4x.

Helper method?

```
func lookupSKU(vmSize string) (string, bool) {
	vmSize = strings.ToLower(vmSize)
	vmSize = strings.TrimSuffix(vmSize, "_promo")
	return gpuSKUConfigs[vmSize]
}
```

go.mod
```diff
@@ -144,6 +144,7 @@ require (
 	go.opentelemetry.io/otel/trace v1.40.0 // indirect
 	go.uber.org/mock v0.6.0 // indirect
 	go.uber.org/zap v1.27.0 // indirect
+	go.yaml.in/yaml/v2 v2.4.3 // indirect
 	go.yaml.in/yaml/v3 v3.0.4 // indirect
```
matthchr, line 148: Yours: gopkg.in/yaml.v3 v3.0.1

Others: 
go.yaml.in/yaml/v2 v2.4.3
go.yaml.in/yaml/v3 v3.0.4


https://github.com/go-task/task/issues/2171 (and others) suggest that gopkg.in/yaml.v3 is unmaintained. I don't think we want to take a dependency on it? The GitHub repo for that package is marked archived too, which suggests that is correct.

pkg/utils/gpu.go
```diff
@@ -33,40 +33,52 @@ const (
...
 	if err != nil {
 		panic(err)
 	}
-	for _, skus := range nvidiaSKUConfig.NvidiaEnabledSKUFamilies {
-		for _, sku := range skus {
-			nvidiaEnabledSKUs[sku] = true
+	// Validate the configuration
+	for sku, config := range gpuSKUConfigs {
+		if config.GPU != GPUNvidia && config.GPU != GPUAMD {
+			panic("gpu field must be either nvidia or amd for SKU: " + sku)
```
theunrepentantgeek, line 72: Not really specific to your changes (as this behaviour was happening before), but is it really appropriate to **panic** for a validation failure? This kills the entire process/pod, including interrupting anything/everything else that's going on in other goroutines.

pkg/utils/gpu.go
```diff
@@ -33,40 +33,52 @@ const (
...
 	NvidiaGridDriverVersion = "550.144.06"
 	AKSGPUGridVersionSuffix = "20250512225043"
+
+	// GPU types
+	GPUNvidia = "nvidia"
+	GPUAMD    = "amd"
+
+	// OS types
+	OSUbuntu      = "ubuntu"
+	OSAzureLinux  = "azurelinux"
```
theunrepentantgeek, line 43: This constant seems to be specifically used for Azure Linux _**2**_ (aka Mariner) only - but the name implies _all_ versions. Consider renaming it to `OSAzureLinux2`.

pkg/utils/gpu.go
theunrepentantgeek: We could even turn all four methods into wrappers that call a single implementation with different parameters.

pkg/utils/gpu_test.go
```diff
@@ -117,3 +131,77 @@ func TestIsMarinerEnabledGPUSKU(t *testing.T) {
...
+			g.Expect(result).To(Equal(test.output), "Failed for input: %s", test.input)
+		})
+	}
+}
+
+func TestIsGPUEnabledSKU(t *testing.T) {
+	tests := []struct {
+		name   string
+		input  string
+		output bool
```
theunrepentantgeek, line 163: Minor nit, but I found `output` opaque until I read the rest of the test. What about renaming it to `expected` or even `expectedGPUEnabled`?

pkg/providers/instancetype/instancetype.go
MaximilianoUribe: file has this type of format for vmsize, would like to keep to ensure consistency

pkg/utils/gpu.go
MaximilianoUribe: This is a valid point that I overlooked, I guess this is a question of design now of what do we want to do and what are the concerns of exposing all skus and only install drivers for os+sku valid configurations, that in my opinion should be what we should do here

test/suites/gpu/suite_test.go
```diff
@@ -71,8 +72,8 @@ var _ = Describe("GPU", func() {
 			})
 
 			nodePool.Spec.Limits = karpv1.Limits{
-				corev1.ResourceCPU:                    resource.MustParse("25"),
-				corev1.ResourceName("nvidia.com/gpu"): resource.MustParse("1"),
+				corev1.ResourceCPU:                   resource.MustParse("25"),
```
theunrepentantgeek, line 75: Do these values need units specified for clarity?

PR #1467, opened 2026-03-04
perf: add LIST-based machine cache to reduce GET throttling

comtalyst:
**Description**

Adds a LIST-based machine cache (`machineListCache`) that converts O(N) individual GET Machine API calls into O(1) cached lookups. At 1K-node scale, individual GET calls from drift checks and reconciliation cause severe throttling (6,594 GET 429s at 17-29s latency each).

- Cache is populated by `List()`, consulted by `Get()`, and invalidated by mutating operations (Create, Update, Delete)
- TTL-based expiry (30s default) ensures bounded staleness acceptable for drift and reconciliation
- Thread-safe with `sync.RWMutex`

This PR adds the cache infrastructure only. Wiring into `DefaultAKSMachineProvider` is in #1473.

* 9 unit tests covering: empty cache, populated cache, TTL expiry, invalidation (single/all), replacement updates, nil name handling, zero TTL, freshness before update
* `make presubmit`

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note

```


comtalyst (commented):
## Cache returns shared pointers — callers can corrupt cached state

`get()` returns the same `*armcontainerservice.Machine` pointer that's stored in the map. Any caller that mutates the returned object (e.g., the in-place update controller's `CalculateAKSMachinePatch`, which mutates the machine in-place to build the PUT body) will corrupt the cached entry.

This isn't a bug in this PR in isolation (the cache is well-implemented on its own), but it's a design constraint that PR #1473 (the integration PR) needs to handle. The in-place update controller does `Get()` → mutate in-place → `Update()` (which invalidates cache). Between the mutation and the invalidation, a concurrent reader could see partially-mutated data, and with `MaxConcurrentReconciles=10` this is a real window.

Two options:
1. Have `get()` return a deep copy (JSON round-trip like the in-place update controller already does for its own comparison copy), or
2. Document that callers must not mutate returned objects, and fix the in-place update caller to copy first.

Option 1 is safer since it makes the cache defensively correct regardless of caller behavior. The JSON round-trip cost is negligible compared to the GET API call it replaces.

PR #1468, opened 2026-03-04
fix: read AKS machine creation timestamp from MachineStatus instead of tag

comtalyst:
**Description**

Migrates creation timestamp reading from client-side tag (`karpenter.azure.com_aksmachine_creationtimestamp`) to server-side `MachineStatus.CreationTimestamp`.

* Existing unit tests updated; all pass
* `make presubmit`
* Same E2E w/ Machine API

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/providers/instance/aksmachineinstanceutils_test.go
+++ b/pkg/providers/instance/aksmachineinstanceutils_test.go
@@ -37,6 +37,7 @@ import (
 	"sigs.k8s.io/karpenter/pkg/scheduling"
 
 	"github.com/Azure/karpenter-provider-azure/pkg/apis/v1beta1"
+	"github.com/Azure/karpenter-provider-azure/pkg/consts"
 	"github.com/Azure/karpenter-provider-azure/pkg/providers/launchtemplate"
 
 	. "github.com/onsi/ginkgo/v2"
@@ -495,7 +496,7 @@ var _ = Describe("AKSMachineInstanceUtils Helper Functions", func() {
 		It("should return true when provisioning state is Deleting", func() {
 			machine := &armcontainerservice.Machine{
 				Properties: &armcontainerservice.MachineProperties{
-					ProvisioningState: lo.ToPtr("Deleting"),
+					ProvisioningState: lo.ToPtr(consts.ProvisioningStateDeleting),
 				},
 			}
 
--- a/pkg/consts/consts.go
+++ b/pkg/consts/consts.go
@@ -40,4 +40,13 @@ const (
 	ProvisionModeAKSScriptless       = "aksscriptless"
 	ProvisionModeBootstrappingClient = "bootstrappingclient"
 	ProvisionModeAKSMachineAPI       = "aksmachineapi"
+
+	// Provisioning states for AKS Machine objects.
+	// The SDK's Machine.Properties.ProvisioningState is typed as *string (no typed constants).
+	// Suggestion: find a constant from azure-sdk-for-go if one becomes available.
+	ProvisioningStateCreating  = "Creating"
+	ProvisioningStateUpdating  = "Updating"
+	ProvisioningStateSucceeded = "Succeeded"
+	ProvisioningStateFailed    = "Failed"
+	ProvisioningStateDeleting  = "Deleting"
 )
--- a/pkg/fake/aksmachinesapi.go
+++ b/pkg/fake/aksmachinesapi.go
@@ -27,6 +27,7 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/containerservice/armcontainerservice/v8"
+	"github.com/Azure/karpenter-provider-azure/pkg/consts"
 	"github.com/Azure/karpenter-provider-azure/pkg/providers/azclient"
 	"github.com/samber/lo"
 )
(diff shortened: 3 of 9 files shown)


pkg/providers/launchtemplate/tags.go
```diff
@@ -29,7 +29,7 @@ import (
 const (
 	KarpenterManagedTagKey                     = "karpenter.azure.com_cluster"
 	KarpenterAKSMachineNodeClaimTagKey         = "karpenter.azure.com_aksmachine_nodeclaim"
-	KarpenterAKSMachineCreationTimestampTagKey = "karpenter.azure.com_aksmachine_creationtimestamp" // TODO: stop using tag and use MachineStatus.CreationTimestamp when the change is live. See inplaceupdate/patch.go for details.
+	KarpenterAKSMachineCreationTimestampTagKey = "karpenter.azure.com_aksmachine_creationtimestamp" // Deprecated: no longer written. Kept for backward compatibility reading. Use MachineStatus.CreationTimestamp instead.
```
comtalyst: No need. Just get rid of it entirely. It is not in production yet. Also get rid of the related code. Pretend that this tag never exist and we will just use real creation timestamp from machine.

comtalyst: (AI-generated) Done. Removed the `KarpenterAKSMachineCreationTimestampTagKey` constant, the `AKSMachineTimestampToTag()` / `AKSMachineTimestampFromTag()` functions, the tag fallback reading in `BuildNodeClaimFromAKSMachineTemplate()`, all related tests, and migration comments throughout the codebase. The tag is now treated as if it never existed.

pkg/cloudprovider/suite_aksmachineapi_integration_test.go
```diff
@@ -239,7 +239,7 @@ func runSharedAKSMachineAPITests() {
 			}
 			Expect(ourNodeClaim).ToNot(BeNil())
 
-			// CreationTimestamp should be zero due to parsing failure
+			// CreationTimestamp should be zero because Status.CreationTimestamp is nil
 			Expect(ourNodeClaim.CreationTimestamp.IsZero()).To(BeTrue())
 		})
```
comtalyst: No need for this test. Malformation is not expected in AKS machine API fields.
Also check for other unnecessary tests given the above assumption.

pkg/controllers/nodeclaim/garbagecollection/suite_aksmachine_test.go
```diff
@@ -203,35 +192,32 @@ var _ = Describe("Instance Garbage Collection", func() {
 				ExpectCleanedUp(ctx, env.Client)
 			})
 
-			It("Instance created -> Tag deleted -> In-place update -> Garbage collection false positive -> Create() completed -> Registered -> In-place update", func() {
+			It("Instance created -> Tag deleted -> In-place update -> GC does NOT false-positive (Status.CreationTimestamp protects) -> Create() completed -> Registered -> In-place update", func() {
```
comtalyst: Given AKS machine's creation timestamp will always be genuine, for each of the tests, verify whether we really need them. E.g., in this one, "tag deleted" and "in-place update" holds no meaning.

pkg/providers/instance/aksmachineinstanceutils_test.go
```diff
@@ -161,8 +160,8 @@ var _ = Describe("AKSMachineInstanceUtils Helper Functions", func() {
 		})
 
 		It("should handle missing creation time gracefully", func() {
```
comtalyst: Like said in other comments, this case is not possible.

pkg/cloudprovider/suite_aksmachineapi_integration_test.go
comtalyst: Done. Removed the "malformed timestamp during List" integration test since AKS machine API fields are always well-formed. Also audited: no other unnecessary malformation tests found in the integration test file.

pkg/controllers/nodeclaim/garbagecollection/suite_aksmachine_test.go
comtalyst: Done. Removed the "without timestamp" GC test (impossible case). Simplified the first complex test by removing the timestamp-protection framing ("GC does NOT false-positive" / "Status.CreationTimestamp protects") and the in-place-update-before-ProviderID steps that only made sense in the old tag model. The remaining tests now focus purely on nodeclaim tag repair scenarios. Renamed the context and test names to say "NodeClaim tag deleted" instead of the ambiguous "Tag deleted".

pkg/providers/instance/aksmachineinstanceutils_test.go
comtalyst: Done. Removed the "should handle missing creation time gracefully" test since this case is not possible with AKS machine API.

pkg/controllers/nodeclaim/garbagecollection/suite_aksmachine_test.go
comtalyst: Updated: removed the entire "Complex tags manipulation scenarios with in-place updates" context (all 5 tests + BeforeEach/AfterEach). The whole premise was the since-removed creation timestamp tag — tag deletion, in-place update repair, and GC false-positive protection are all moot now. Cleaned up the unused imports (v1beta1, object, metav1) that were only needed by those tests.

comtalyst:
This cleanly implements **Option E** from the [Machine API Additional Design Discussions](https://dev.azure.com/msazure/CloudNativeCompute/_wiki/wikis/CloudNativeCompute.wiki/516726/) (Section 18, "Creation timestamp existence") — the TODO in `patch.go` and `tags.go` that said "stop using tag and use `MachineStatus.CreationTimestamp` when the change is live" is now fulfilled.

**What this solves:** The tag-based approach (Option B1) was a workaround for `Status.CreationTimestamp` originally reflecting VM creation time, not machine creation time. Now that the Machine API has repurposed that field to reflect machine creation time, the tag is unnecessary — and with it, all the complexity around tag tampering, in-place update repair, and malformed-timestamp fallbacks. Good riddance to ~580 lines.

**Assumptions this relies on (for the record):**

| Assumption | Impact if violated |
|---|---|
| `Status.CreationTimestamp` reflects **machine** creation time, not inner VM creation time | GC grace period would start late (shorter effective window). The test change from `creationTime + 10min` → `creationTime` confirms this expectation. |
| `Status.CreationTimestamp` is present immediately after PUT completes (not delayed to async phase) | If delayed, machines in the async window would have zero timestamp → immediate GC. We believe this is always present in the PUT response. |
| `Status.CreationTimestamp` is immutable once set | If it changed (e.g., on VM recreate), GC grace period would reset unexpectedly. |

All three are strictly better than the old tag-based assumptions (client-side timing, user-tamperability, repair logic). No compatibility concerns since the tag was never deployed to production.

PR #1469, opened 2026-03-04
fix: suppress machine drift GET during node creation

comtalyst:
**Description**

Skips the `isMachineDrifted()` GET Machine API call when the NodeClaim has not yet reached `Registered=True`. Core's drift controller calls `IsDrifted()` once the NodeClaim is "Launched" (PUT accepted), but the machine may still be provisioning. During burst scale-up with batch creation, this generates unnecessary API load and 404 errors.

Follows the same pattern as `getNodeForDrift()` which returns `(nil, nil)` for nodes that haven't joined the cluster yet.

* New unit test: `should not be drifted if the nodeclass does not exist and NodeClaim is not registered`
* Existing tests pass
* `make presubmit`

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/cloudprovider/drift.go
+++ b/pkg/cloudprovider/drift.go
@@ -22,6 +22,7 @@ import (
 	"strings"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/containerservice/armcontainerservice/v8"
+	"github.com/go-logr/logr"
 	"github.com/samber/lo"
 	v1 "k8s.io/api/core/v1"
 	"sigs.k8s.io/controller-runtime/pkg/log"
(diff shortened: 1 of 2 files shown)


comtalyst:
## Code Review: PR #1469 — Suppress Machine Drift GET During Node Creation

### What It Does
- Skip `isMachineDrifted` (GET Machine for DriftAction) when `ConditionTypeRegistered` is not true
- Extract `parseDriftAction` as standalone function (reduces cyclomatic complexity)

### Why It's Correct
- Follows established pattern (same guard used in in-place update controller, line 134)
- Nil-safe: `Get()` → `IsTrue()` on nil returns false correctly
- Other drift checks (static fields, K8s version, image version) still run
- DriftAction during creation is meaningless — node hasn't joined cluster yet
- Even if drift were detected, core can't act on it (node isn't ready for disruption)

### Tests: Good
- Existing tests updated with `SetTrue(ConditionTypeRegistered)` precondition
- New test proves suppression works: sets `DriftAction=Recreate` on machine, asserts no drift returned when not registered
- All 124 cloudprovider tests pass

### Risk: None
Staleness window is only during Launched→Registered phase — exactly when drift GETs are wasteful and throttle-harmful. At 1k scale, this eliminates O(N) unnecessary GETs per reconciliation cycle.

🚀

PR #1472, opened 2026-03-04
fix: add throttling-tolerant retry for GET-after-PUT in machine creation

comtalyst:
**Description**

Adds application-level exponential backoff retry for GET Machine calls during AKS machine creation that are vulnerable to 429 throttling.

- **GET-once-after-PUT**: The GET call immediately after `BeginCreateOrUpdate` to retrieve VMResourceID can fail with 429 during burst creation at scale (1000+ nodes)
- **Pre-create existence check**: The GET call to check if a machine already exists (for restart reuse) can also fail with 429, causing Karpenter to skip reuse and attempt duplicate creation
- Only 429 errors are retried; all other errors (404, 400, 500, etc.) are returned immediately
- Backoff parameters: 5s base, 2x factor, 30s cap, 5 attempts (~2min total worst-case)

The Azure SDK retries 429s 3 times with exponential backoff (800ms → 1.6s → 3.2s), totaling ~5.6s. Telescope data from 1k/5k node scale-up runs shows GET 429 responses lasting 17-29 seconds, exceeding SDK retry budget.

* Unit tests for `isThrottlingError` helper
* `make presubmit`

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/providers/instance/aksmachineinstance.go
+++ b/pkg/providers/instance/aksmachineinstance.go
@@ -363,7 +363,7 @@ func (p *DefaultAKSMachineProvider) getMachineWithRetry(ctx context.Context, aks
 		Duration: 5 * time.Second,
 		Factor:   2.0,
 		Jitter:   0.1,
-		Steps:    5,  // 5s, 10s, 20s, 40s, 80s (capped at 30s) → attempts at ~0s, 5s, 15s, 35s, 65s
+		Steps:    5, // 5s, 10s, 20s, 40s, 80s (capped at 30s) → attempts at ~0s, 5s, 15s, 35s, 65s
 		Cap:      30 * time.Second,
 	}
 


comtalyst:
## Code Review: PR #1472 — GET-after-PUT Throttling-Tolerant Retry

### What It Does
Adds `wait.ExponentialBackoffWithContext` retry (5 attempts, 5s×2.0 factor, 30s cap, ~93s worst case) for two `getMachine` calls in `beginCreateMachine`:
1. **GET-after-PUT** — where 429 → false failure → node deletion
2. **Pre-create existence check** — where 429 → skip reuse → duplicate creation

### What's Correct
- Uses idiomatic K8s retry; only retries 429 (correct error classification via `isThrottlingError`)
- Error wrapping safe for downstream `IsAKSMachineOrMachinesPoolNotFound`
- ~93s worst-case sufficient for observed 17-29s throttling windows
- 0.1 jitter prevents thundering herd; public `Get()` correctly unaffected

### Minor Suggestions
1. `wait.Interrupted()` conflates "retries exhausted" with "context cancelled" → check `ctx.Err()` separately for clearer error messages
2. Comment says "~2min" but actual worst-case is ~93s
3. Consider adding unit test for `getMachineWithRetry` itself (counting mock, N fails then success)

### Impact
At 1k scale, this prevents ~5-10s of wasted p90 latency from false failure → deletion → recreation cycles.

🚀

PR #1480, opened 2026-03-05
fix: handle silenced errors in deep copy and instance type cache key

comtalyst:
## Summary
- **In-place update controller**: `json.Marshal` error was silenced during deep copy of AKS Machine object for diff comparison. If marshal failed, nil bytes would flow into `json.Unmarshal`, producing a confusing downstream error instead of a clear root cause.
- **Instance type provider**: `hashstructure.Hash` error was silenced when computing the cache key for fully-initialized instance types. If hashing failed, `kcHash` would be 0, causing cache key collisions across different kubelet configs — potentially returning incorrect cached instance types for a NodeClass. Now the cache is skipped entirely when hashing fails, trading performance for correctness.

## Test plan
- [x] `go test ./pkg/controllers/nodeclaim/inplaceupdate/...` passes
- [x] `go test ./pkg/providers/instancetype/...` passes
- [x] `go build ./...` compiles clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pkg/controllers/nodeclaim/inplaceupdate/controller.go
+++ b/pkg/controllers/nodeclaim/inplaceupdate/controller.go
@@ -215,9 +215,12 @@ func (c *Controller) applyAKSMachinePatch(
 	aksMachine *armcontainerservice.Machine,
 ) error {
 	// Create a deep copy of the original for diff comparison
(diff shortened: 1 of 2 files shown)


comtalyst:
Closing — changes reverted by maintainer.

PR #1487, opened 2026-03-05
feat: add AzureNodeClass CRD at karpenter.azure.com/v1alpha1

comtalyst:
**Description**

Add `AzureNodeClass` CRD in karpenter.azure.com/v1alpha1 for non-AKS Azure VM provisioning. Defines a simpler, AKS-independent node class with `imageID`, `userData`, `managedIdentities`, `vnetSubnetID`, `osDiskSizeGB`, `tags`, and `security` fields.

Includes status controller interfaces (`StatusConditions`, `GetConditions`, `SetConditions`), scheme registration, and deep copy generation. Adds CRD YAML to charts and api packages.

*

- [ ] Yes, issue opened: # 
- [X] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/apis/v1alpha1/azurenodeclass.go
+++ b/pkg/apis/v1alpha1/azurenodeclass.go
@@ -53,12 +53,11 @@ type AzureNodeClassSpec struct {
 	// +kubebuilder:validation:MaxLength=1024
 	// +optional
 	ImageID *string `json:"imageID,omitempty"`
-	// userData is the base64-encoded custom data that will be passed to the VM at creation time.
-	// The caller must pre-encode their cloud-init or bootstrap script to base64, as the Azure API
-	// expects this field to contain a base64-encoded string.
-	// The user is fully responsible for providing valid bootstrap/cloud-init data.
+	// userData is the cloud-init or bootstrap script that will be passed to the VM at creation time.
+	// Provide the raw text (e.g. a cloud-init YAML or shell script); the provider will base64-encode
+	// it before passing to the Azure API. The user is responsible for providing valid bootstrap data.
 	// When this field is set, no Karpenter-managed bootstrapping is performed.
-	// +kubebuilder:validation:MaxLength=87380
+	// +kubebuilder:validation:MaxLength=65536
 	// +optional
 	UserData *string `json:"userData,omitempty"`
 	// managedIdentities is a list of user-assigned managed identity resource IDs
--- a/charts/karpenter-crd/templates/karpenter.azure.com_azurenodeclasses.yaml
+++ b/charts/karpenter-crd/templates/karpenter.azure.com_azurenodeclasses.yaml
@@ -61,6 +61,15 @@ spec:
               spec is the top level specification for the Azure Karpenter Provider.
               This will contain configuration necessary to launch instances in Azure.
             properties:
+              dataDiskSizeGB:
+                description: |-
+                  dataDiskSizeGB is the size of an additional data disk in GB to attach for
+                  container runtime storage (e.g., containerd root). A Premium_LRS managed disk
+                  is attached at LUN 0. The image's bootstrap must handle mounting this disk.
+                format: int32
+                maximum: 4096
+                minimum: 10
+                type: integer
               imageID:
                 description: |-
                   imageID is the ARM resource ID of the image that instances use.
@@ -73,6 +82,21 @@ spec:
                     /CommunityGalleries/{gallery}/images/{image}/versions/{version}
                 maxLength: 1024
                 pattern: (?i)^(\/subscriptions\/[^\/]+\/resourceGroups\/[^\/]+\/providers\/Microsoft\.Compute\/.*|\/CommunityGalleries\/[^\/]+\/images\/[^\/]+\/versions\/[^\/]+)$
+                type: string
+              instanceTypes:
+                description: |-
+                  instanceTypes overrides the instance types discovered by the pricing/SKU provider.
+                  When set, only these instance types are considered for this NodeClass.
+                  Use this for pinned capacity (e.g. GPU SKUs) where the SKU may not appear
+                  in standard pricing APIs or may be filtered by default compatibility checks.
+                items:
+                  type: string
+                maxItems: 100
+                type: array
+              location:
+                description: |-
+                  location overrides the controller-level LOCATION.
+                  When set, VMs for this NodeClass are created in this Azure region.
                 type: string
               managedIdentities:
                 description: |-
@@ -89,6 +113,11 @@ spec:
                 maximum: 4096
                 minimum: 30
                 type: integer
+              resourceGroup:
+                description: |-
+                  resourceGroup overrides the controller-level AZURE_NODE_RESOURCE_GROUP.
+                  When set, VMs for this NodeClass are created in this resource group.
+                type: string
               security:
                 description: security is a collection of security related karpenter
                   fields.
@@ -100,6 +129,12 @@ spec:
                       https://learn.microsoft.com/en-us/azure/virtual-machines/disk-encryption#encryption-at-host---end-to-end-encryption-for-your-vm-data
                     type: boolean
                 type: object
+              subscriptionID:
+                description: |-
+                  subscriptionID overrides the controller-level AZURE_SUBSCRIPTION_ID.
+                  When set, VMs for this NodeClass are created in this Azure subscription.
+                  The controller's Azure identity must have appropriate RBAC on the target subscription.
+                type: string
               tags:
                 additionalProperties:
                   type: string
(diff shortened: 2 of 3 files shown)


pkg/apis/v1alpha1/azurenodeclass.go
```diff
@@ -0,0 +1,146 @@
...
+	// +kubebuilder:validation:Pattern=`(?i)^(\/subscriptions\/[^\/]+\/resourceGroups\/[^\/]+\/providers\/Microsoft\.Compute\/.*|\/CommunityGalleries\/[^\/]+\/images\/[^\/]+\/versions\/[^\/]+)$`
+	// +kubebuilder:validation:MaxLength=1024
+	// +optional
+	ImageID *string `json:"imageID,omitempty"`
+	// userData is the base64-encoded custom data that will be passed to the VM at creation time.
+	// The caller must pre-encode their cloud-init or bootstrap script to base64, as the Azure API
+	// expects this field to contain a base64-encoded string.
+	// The user is fully responsible for providing valid bootstrap/cloud-init data.
+	// When this field is set, no Karpenter-managed bootstrapping is performed.
+	// +kubebuilder:validation:MaxLength=87380
```
comtalyst: **Suggestion:** The imageID regex uses `Microsoft\.Compute\/.*` which is overly permissive — it accepts any resource under Microsoft.Compute, including non-image resources like `/Microsoft.Compute/disks/someDisk` or even just `/Microsoft.Compute/` (trailing slash, empty path). Consider tightening to `.+` instead of `.*` to at least require something after the trailing slash, or better yet, enumerate the supported resource types:

```
(?i)^(\/subscriptions\/[^\/]+\/resourceGroups\/[^\/]+\/providers\/Microsoft\.Compute\/(galleries|images)\/.+|\/CommunityGalleries\/[^\/]+\/images\/[^\/]+\/versions\/[^\/]+)$
```

At minimum, change `.*` to `.+` to prevent accepting a bare `Microsoft.Compute/` path.

pkg/apis/v1alpha1/azurenodeclass.go
```diff
@@ -0,0 +1,146 @@
...
+	// +kubebuilder:validation:Maximum=4096
+	// +optional
+	OSDiskSizeGB *int32 `json:"osDiskSizeGB,omitempty"`
+	// tags to be applied on Azure resources like instances.
+	// +kubebuilder:validation:XValidation:message="tags keys must be less than 512 characters",rule="self.all(k, size(k) <= 512)"
+	// +kubebuilder:validation:XValidation:message="tags keys must not contain '<', '>', '%', '&', or '?'",rule="self.all(k, !k.matches('[<>%&?]'))"
+	// +kubebuilder:validation:XValidation:message="tags keys must not contain '\\'",rule="self.all(k, !k.contains('\\\\'))"
+	// +kubebuilder:validation:XValidation:message="tags values must be less than 256 characters",rule="self.all(k, size(self[k]) <= 256)"
+	// +optional
+	Tags map[string]string `json:"tags,omitempty" hash:"ignore"`
```
comtalyst, line 86: **Suggestion:** `managedIdentities` has a `MaxItems=10` constraint but no per-item validation (e.g., pattern or format). This means users can pass arbitrary strings like `"not-a-resource-id"` and get a cryptic Azure API error at VM creation time instead of a CRD validation error. Consider adding a `+kubebuilder:validation:items:Pattern` or `+kubebuilder:validation:items:MinLength=1` to catch obviously invalid entries at admission time.

pkg/apis/v1alpha1/crd_validation_cel_test.go
```diff
@@ -0,0 +1,256 @@
...
+
+	azv1alpha1 "github.com/Azure/karpenter-provider-azure/pkg/apis/v1alpha1"
+	"github.com/Pallinder/go-randomdata"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/samber/lo"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var _ = AfterEach(func() {
```
comtalyst, line 30: **Issue:** There are two `AfterEach` blocks registered — one in `crd_validation_cel_test.go` (manual deletion loop) and one in `suite_test.go` (`ExpectCleanedUp`). Both clean up resources after each test, which is redundant. The `ExpectCleanedUp` in `suite_test.go` should be sufficient — remove the manual `AfterEach` in this file to avoid double-cleanup confusion.

pkg/apis/v1alpha1/doc.go
```diff
@@ -0,0 +1,37 @@
...
+
+// +k8s:openapi-gen=true
+// +k8s:deepcopy-gen=package,register
+// +k8s:defaulter-gen=TypeMeta
+// +groupName=karpenter.azure.com
+package v1alpha1 // doc.go is discovered by codegen
+
+import (
+	"github.com/Azure/karpenter-provider-azure/pkg/apis"
+	corev1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
comtalyst, line 25: **Nit:** The alias `corev1` for `k8s.io/apimachinery/pkg/apis/meta/v1` is misleading — `corev1` conventionally refers to `k8s.io/api/core/v1`. This is consistent with the existing `v1beta1/doc.go` so not a blocker, but worth noting for a future cleanup. The standard alias for `k8s.io/apimachinery/pkg/apis/meta/v1` is `metav1`.

designs/0007-azurevm-provision-mode.md
```diff
@@ -0,0 +1,136 @@
...
+  dataDiskSizeGB: 256
+  subscriptionID: "aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee"
+  resourceGroup: "my-custom-rg"
+  location: "westus2"
+  managedIdentities:
+    - "/subscriptions/.../userAssignedIdentities/my-identity"
+  tags:
+    environment: production
+  security:
+    encryptionAtHost: true
```
comtalyst, line 58: **Issue:** The design doc YAML example includes fields that don't exist in this PR's CRD: `subscriptionID`, `resourceGroup`, `location`, and `dataDiskSizeGB`. These are added in PR #1497 later in the chain. This creates a confusing artifact where the design doc advertises an API surface that doesn't match the CRD in this commit. Consider either:
1. Stripping these fields from the design doc and adding them in PR #1497, or
2. Adding a note like "*Fields marked with `†` are added in a subsequent PR*" to set expectations.

pkg/apis/v1alpha1/suite_test.go
```diff
@@ -0,0 +1,60 @@
...
+
+	"github.com/Azure/karpenter-provider-azure/pkg/apis"
+	"github.com/Azure/karpenter-provider-azure/pkg/operator/options"
+	"github.com/Azure/karpenter-provider-azure/pkg/test"
+)
+
+var ctx context.Context
+var env *coretest.Environment
+var azureEnv *test.Environment
+
```
comtalyst, line 41: **Issue:** `azureEnv` is initialized in `BeforeSuite` but never referenced in any test. This is dead code — the CEL validation tests only use `env.Client` (the core test environment). Either remove the `azureEnv` declaration and initialization, or add a comment explaining it's needed for side-effects (if `test.NewEnvironment` registers something necessary).

pkg/apis/v1alpha1/register.go
```diff
@@ -0,0 +1,37 @@
...
+
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+const Group = "karpenter.azure.com"
```
comtalyst, line 25: **Issue:** `const Group = "karpenter.azure.com"` is defined in both `v1alpha1/register.go` (this PR) and `v1beta1/register.go` (existing). This creates two independent constants with the same value. If the group name ever changes, one could be updated while the other is missed. Consider using the existing `apis.Group` constant (already imported in `doc.go`) instead of redeclaring it here. Note: this is consistent with the existing `v1beta1` pattern, so not a blocker if you want to keep parity.

pkg/apis/apis.go
```diff
@@ -38,6 +40,7 @@ var (
 	NodeOverlayCRD []byte
 	CRDs           = []*apiextensionsv1.CustomResourceDefinition{
 		object.Unmarshal[apiextensionsv1.CustomResourceDefinition](AKSNodeClassCRD),
+		object.Unmarshal[apiextensionsv1.CustomResourceDefinition](AzureNodeClassCRD),
```
comtalyst, line 43: **Question:** Adding `AzureNodeClassCRD` to the global `CRDs` slice means both AKSNodeClass and AzureNodeClass CRDs will be installed in every cluster, regardless of provision mode. Is this intentional? In AKS mode, the AzureNodeClass CRD is unused; in AzureVM mode, the AKSNodeClass CRD is unused. This is harmless (extra CRD definition on the API server) but could confuse users who see both CRDs via `kubectl get crd`. Later PR #1489 makes controller registration mode-aware — consider whether CRD installation should also be conditional.

PR #1489, opened 2026-03-05
[OUTDATED] feat: support generic K8s cluster w/ Azure VMs through AzureNodeClasses and AzureVM provision mode

comtalyst:
**Description**

TODO

*

- [ ] Yes, issue opened: # 
- [X] No

**Release Note**

```release-note

```


pkg/cloudprovider/cloudprovider.go
```diff
@@ -129,6 +130,11 @@ func (c *CloudProvider) validateNodeClass(nodeClass *v1beta1.AKSNodeClass) error
 }
 
 func (c *CloudProvider) Create(ctx context.Context, nodeClaim *karpv1.NodeClaim) (*karpv1.NodeClaim, error) {
+	// In AzureVM mode, resolve AzureNodeClass instead of AKSNodeClass and adapt it
+	if options.FromContext(ctx).IsAzureVMMode() {
+		return c.createAzureVMInstance(ctx, nodeClaim)
+	}
```
comtalyst: If we want the same layering with other modes, maybe we should encase the rest in `createAKSInstance(...)` etc.?

pkg/cloudprovider/cloudprovider.go
```diff
@@ -177,6 +183,40 @@ func (c *CloudProvider) Create(ctx context.Context, nodeClaim *karpv1.NodeClaim)
...
+	azureNodeClass, err := nodeclaimutils.GetAzureNodeClass(ctx, c.kubeClient, nodeClaim)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			c.recorder.Publish(cloudproviderevents.NodeClaimFailedToResolveNodeClass(nodeClaim))
+		}
+		return nil, cloudprovider.NewInsufficientCapacityError(fmt.Errorf("resolving AzureNodeClass, %w", err))
+	}
+
+	// Adapt to AKSNodeClass for the VM provider path
+	nodeClass := nodeclaimutils.AKSNodeClassFromAzureNodeClass(azureNodeClass)
```
comtalyst: I think we should either use a common interface or just have two distinct (yet parallel) paths. This is to avoid treating AKSNodeClass as the main type.

pkg/cloudprovider/cloudprovider.go
comtalyst: Still, it looks like a lot of code in that could be shared. Maybe we can drop this `createAzureVMInstance` and rely on this one function, rely on if statements for type-specific (e.g., validateNodeClass), but having common nodeclass interface for sharable part like `resolveInstanceTypes`?

comtalyst: And even if 1556 may want (does it?) special handling, we may also allow some NodeClass type to not have to implement optional "features" in the interface?

comtalyst: Or complete separation might make the ownership more clear with less noise from the other (at the cost of less sharable code = duplicated maintenance in some places)...

pkg/cloudprovider/cloudprovider.go
comtalyst: (also see the comments above)

pkg/cloudprovider/cloudprovider.go
```diff
@@ -551,8 +591,23 @@ func (c *CloudProvider) RepairPolicies() []cloudprovider.RepairPolicy {
...
+			return nil, err
+		}
+		if !azureNodeClass.DeletionTimestamp.IsZero() {
+			return nil, utils.NewTerminatingResourceError(schema.GroupResource{Group: v1alpha1.Group, Resource: "azurenodeclasses"}, azureNodeClass.Name)
+		}
+		return nodeclaimutils.AKSNodeClassFromAzureNodeClass(azureNodeClass), nil
+	}
+
+	// Default: AKSNodeClass (existing behavior)
 	nodeClass := &v1beta1.AKSNodeClass{}
```
comtalyst: To ensure same layering, maybe we should still put the below in an else statement? Which one is the better pattern?

comtalyst: Or, again, same shareability/adapter pattern concern with `Create()`. Perhaps refactoring there will make it clearer of what to do here.

pkg/apis/v1beta1/aksnodeclass.go
```diff
@@ -74,6 +74,16 @@ type AKSNodeClassSpec struct {
...
+	// UserData is the raw cloud-init or bootstrap script from AzureNodeClass.
+	// The provider base64-encodes it before setting osProfile.CustomData.
+	// Only used in AzureVM provision mode via the AzureNodeClass adapter.
+	// Not exposed in the AKSNodeClass API.
+	UserData *string `json:"-"`
+	// ManagedIdentities is a list of user-assigned managed identity resource IDs
+	// from AzureNodeClass. Only used in AzureVM provision mode via the adapter.
+	// These are merged with the global --node-identities at VM creation time.
+	// Not exposed in the AKSNodeClass API.
+	ManagedIdentities []string `json:"-"`
```
comtalyst: Is there a way to not have to expose this/disturb AKSNodeClass CRD at all? Maybe try moving away from adapter pattern per the comment in `cloudprovider.Create()`?

pkg/cloudprovider/drift.go
```diff
@@ -57,7 +57,12 @@ func (c *CloudProvider) isNodeClassDrifted(ctx context.Context, nodeClaim *karpv
 	}
 
 	var checks []func(ctx context.Context, nodeClaim *karpv1.NodeClaim, nodeClass *v1beta1.AKSNodeClass) (cloudprovider.DriftReason, error)
-	if _, isAKSMachine := instance.GetAKSMachineNameFromNodeClaim(nodeClaim); isAKSMachine {
+	if options.FromContext(ctx).IsAzureVMMode() {
+		// AzureVM mode: only hash-based drift (no AKS-specific k8s version, image, kubelet identity checks)
+		checks = []func(ctx context.Context, nodeClaim *karpv1.NodeClaim, nodeClass *v1beta1.AKSNodeClass) (cloudprovider.DriftReason, error){
+			c.areStaticFieldsDrifted,
+		}
+	} else if _, isAKSMachine := instance.GetAKSMachineNameFromNodeClaim(nodeClaim); isAKSMachine {
```
comtalyst: This wouldn't support the case where we mix AzureVM mode-provisioned and AKS modes-provisioned instances in one cluster and want to maintain them both. 
Instead, we should look at how AKS machine instance manage to maintain that with (AKS) VM instance--by looking at NodeClaim. Though it would likely need changes in the function signature per NodeClass type. (see other comments in `cloudprovider.Create()` on types patterning)

pkg/controllers/nodeclass/termination/events.go
```diff
@@ -20,19 +20,19 @@ import (
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"sigs.k8s.io/karpenter/pkg/events"
 
-	"github.com/Azure/karpenter-provider-azure/pkg/apis/v1beta1"
 	"github.com/Azure/karpenter-provider-azure/pkg/utils"
 )
 
-func WaitingOnNodeClaimTerminationEvent(nodeClass *v1beta1.AKSNodeClass, names []string) events.Event {
+func WaitingOnNodeClaimTerminationEvent(nodeClass client.Object, names []string) events.Event {
```
comtalyst, line 30: Maybe a common interface rather than `client.Object()`?

pkg/controllers/controllers.go
```diff
@@ -62,18 +65,32 @@ func NewControllers(
...
+			nodeclasshash.NewAzureNodeClassController(kubeClient),
+			nodeclassstatus.NewAzureNodeClassController(kubeClient),
+			nodeclasstermination.NewAzureNodeClassController(kubeClient, recorder),
+			status.NewController[*v1alpha1.AzureNodeClass](kubeClient, mgr.GetEventRecorderFor("karpenter")),
+		)
+	} else {
+		controllers = append(controllers,
+			nodeclasshash.NewController(kubeClient),
+			nodeclassstatus.NewController(kubeClient, kubernetesVersionProvider, nodeImageProvider, inClusterKubernetesInterface, subnetsClient, diskEncryptionSetsClient, parsedDiskEncryptionSetID),
+			nodeclasstermination.NewController(kubeClient, recorder),
```
comtalyst: It should be fair to rename `NewController` to `NewAKSNodeClassesController` per "AKSNodeClass should be on the same layer as AzureNodeClass" philosophy discussed elsewhere in this PR.

pkg/operator/options/options.go
```diff
@@ -203,6 +210,14 @@ func FromContext(ctx context.Context) *Options {
...
 // with the additional assumption of DNS prefix being the first 33 chars of FQDN
 func getAKSClusterID(apiServerFQDN string) string {
+	if len(apiServerFQDN) < 33 {
+		// Non-AKS endpoints may have short hostnames. Use what we have.
+		// This path should not be reached in normal AKS operation.
+		h := fnv.New64a()
+		h.Write([]byte(apiServerFQDN))
+		r := rand.New(rand.NewSource(int64(h.Sum64()))) //nolint:gosec
+		return fmt.Sprintf("%08d", r.Uint32())[:8]
+	}
```
comtalyst, line 220: This function is literally called `getAKSClusterID()`. Non-AKS cases shouldn't even use it. Non-AKS cases needing something like this, plus other differences in validation/options requirements is another concern that needs to be addressed.

pkg/operator/options/options_extensions.go
```diff
@@ -31,3 +31,8 @@ func (o *Options) IsCiliumNodeSubnet() bool {
 func (o *Options) IsNetworkPluginNone() bool {
 	return o.NetworkPlugin == consts.NetworkPluginNone
 }
+
+// IsAzureVMMode returns true if the provision mode is AzureVM (non-AKS generic Azure VM provisioning).
+func (o *Options) IsAzureVMMode() bool {
+	return o.ProvisionMode == consts.ProvisionModeAzureVM
```
comtalyst: Do we really need this, and couldn't we just use `==` like other places?
Or, if we want something like this, I would rather have it like `IsAKSProvisioningMode` or something similar, where ProvisionModeAzureVM is not, but the rest is. Still, see if we need it.

pkg/providers/instance/vminstancehelpers.go
```diff
@@ -270,11 +310,44 @@ func configureNetworkProfile(nicReference string) *armcompute.NetworkProfile {
...
+		// In AzureVM mode, base64-encode the raw userData from AzureNodeClass and
+		// set as osProfile.CustomData. The Azure API expects base64-encoded data.
+		if nodeClass.Spec.UserData != nil && *nodeClass.Spec.UserData != "" {
+			encoded := base64.StdEncoding.EncodeToString([]byte(*nodeClass.Spec.UserData))
+			osProfile.CustomData = &encoded
+		}
+	} else {
+		// AKS modes: SSH key and admin username are required
+		osProfile.AdminUsername = lo.ToPtr(opts.LinuxAdminUsername)
+		osProfile.LinuxConfiguration = &armcompute.LinuxConfiguration{
```
comtalyst, line 331: Most of these seems to be an unnecessary duplication? I am seeing only CustomData logic is different (but might miss others, please double-check), which can just be handled with the if statement that we already have below.

comtalyst: I think we should look at every other changes we are making in this `instance` module (comparing to the base branch). Most of them seems to have the same category of problem, and that should be reworked.

pkg/utils/nodeclaim/nodeclaim.go
```diff
@@ -67,6 +68,58 @@ func GetAKSNodeClass(ctx context.Context, kubeClient client.Client, nodeClaim *k
...
+	}
+
+	return nodeClass, nil
+}
+
+// AKSNodeClassFromAzureNodeClass creates a compatible AKSNodeClass from an AzureNodeClass.
+// This adapter allows the VM provider (which expects AKSNodeClass) to work with AzureNodeClass
+// by mapping the shared fields. Fields that only exist on AKSNodeClass (e.g. ImageFamily, Kubelet,
+// MaxPods) are left at their zero values, and the VM provider's AzureVM-mode logic handles them.
+func AKSNodeClassFromAzureNodeClass(azureNC *v1alpha1.AzureNodeClass) *v1beta1.AKSNodeClass {
```
comtalyst: (as commented in cloudprovider module) Let's not use adapter pattern. It creates uneven layering and debt.

pkg/cloudprovider/cloudprovider.go
comtalyst: (AI-generated)

Done. Wrapped the AKS path in `createAKSInstance()`. `Create()` now routes by `NodeClassRef.Kind` instead of global provision mode flag — this supports mixed-mode clusters where both NodeClass types coexist. (commit `c2ab8f63`)

pkg/cloudprovider/cloudprovider.go
comtalyst: (AI-generated)

Created `VMNodeClass` interface in `pkg/apis/types/vmnodeclass.go`. Both `AKSNodeClass` and `AzureNodeClass` implement it. The interface covers shared VM-creation fields (SubnetID, OSDiskSizeGB, ImageID, UserData, Tags, EncryptionAtHost, ManagedIdentities). AKS-specific fields (Kubelet, ImageFamily, etc.) stay on AKSNodeClass only.

The adapter still exists as an implementation detail for the VM provider path (which currently takes `*v1beta1.AKSNodeClass`), but the interface is the primary abstraction going forward. Migrating the VM provider to accept `VMNodeClass` directly is the next step to fully remove the adapter.

pkg/cloudprovider/cloudprovider.go
comtalyst: (AI-generated)

Partially addressed. `Create()` now has two parallel entry points (`createAKSInstance` and `createAzureVMInstance`) routed by `NodeClassRef.Kind`. They share the same downstream VM provider (`createVMInstance`).

Full unification into one function would require changing `resolveInstanceTypes` to accept the `VMNodeClass` interface instead of `*v1beta1.AKSNodeClass` — that's a deeper refactor touching the instance type provider's `List()` signature. Keeping as a follow-up to avoid scope creep in this PR.

comtalyst: (AI-generated)

The `VMNodeClass` interface only has the shared fields needed for VM creation. AKS-specific features (Kubelet config, ImageFamily, LocalDNS, ArtifactStreaming) are NOT on the interface — they're accessed directly via `*v1beta1.AKSNodeClass` in the AKS code path. PR #1556 (instanceTypes) adds the field to `AzureNodeClassSpec` and maps it via adapter — no special handling needed on the interface since the instance type provider reads it from AKSNodeClass regardless of source.

comtalyst: (AI-generated)

Current approach: shared VM provider with mode-aware branching at key points (bootstrap resolution, NIC config, extensions). The `VMNodeClass` interface reduces coupling but doesn't fully separate.

Full separation (duplicate VM provider) would mean ~2000 lines of duplicated code with independent maintenance. Given both paths share 90%+ of the Azure VM lifecycle (NIC creation, VM creation, ARG queries, garbage collection, deletion), the shared approach with if-statements at the 3-4 branch points seems like the better trade-off for now. Happy to revisit if ownership boundaries become clearer.

pkg/cloudprovider/cloudprovider.go
comtalyst: (AI-generated)

Addressed above — `createAzureVMInstance` is now a parallel sibling to `createAKSInstance`, both routed by `NodeClassRef.Kind`. The `VMNodeClass` interface provides the shared type contract.

pkg/cloudprovider/cloudprovider.go
comtalyst: (AI-generated)

This is in the `resolveNodeClass` path (for Get/Delete). Updated drift checks to use `NodeClassRef.Kind` instead of mode flag. The resolveNodeClass function still uses mode check for now — migrating it to NodeClassRef-based routing is part of the same follow-up as unifying `Create()`.

comtalyst: (AI-generated)

Agreed. The adapter pattern is the root cause of the layering issues. The `VMNodeClass` interface is the path to removing it. Once the VM provider's function signatures accept `VMNodeClass` instead of `*AKSNodeClass`, the adapter becomes unnecessary and these resolve/Get paths become clean.

pkg/apis/v1beta1/aksnodeclass.go
comtalyst: (AI-generated)

The `json:"-"` fields on AKSNodeClass (UserData, ManagedIdentities) are the adapter's mechanism. With the `VMNodeClass` interface, these can eventually be removed — the VM provider would read from the interface methods instead of from AKSNodeClass struct fields. For this PR they remain as the adapter bridge, but the interface makes the removal path clear.

pkg/cloudprovider/drift.go
comtalyst: (AI-generated)

Fixed. Drift now checks `nodeClaim.Spec.NodeClassRef.Kind == "AzureNodeClass"` instead of `IsAzureVMMode()`. This correctly supports mixed-mode clusters where both NodeClass types exist, and each NodeClaim is checked against its own NodeClass type's drift rules.

pkg/controllers/nodeclass/termination/events.go
comtalyst: (AI-generated)

The `VMNodeClass` interface in `pkg/apis/types/` can serve this purpose. The termination event currently uses `client.Object` to be generic across both CRD types. Using `VMNodeClass` would be more type-safe. Will update if you prefer the interface here over `client.Object`.

pkg/controllers/controllers.go
comtalyst: (AI-generated)

Done. Renamed `NewController` → `NewAKSNodeClassController` in hash, status, and termination controller packages. The AzureNodeClass controllers already use `NewAzureNodeClassController` naming, so this makes them symmetric.

pkg/operator/options/options.go
comtalyst: (AI-generated)

Fixed. Changed from `ProvisionMode \!= AzureVM` (negative check) to explicit positive check for AKS modes: `ProvisionMode == AKSScriptless || BootstrappingClient || AKSMachineAPI`. Now only AKS modes call `getAKSClusterID()`.

pkg/operator/options/options_extensions.go
comtalyst: (AI-generated)

Kept the helper but reduced its usage. `Create()` and drift now use `NodeClassRef.Kind` checks instead of `IsAzureVMMode()`. The helper remains for code paths without a NodeClaim (controller registration, validation) with a comment recommending NodeClassRef-based checks where possible.

Didn't flip to `IsAKSProvisioningMode` because that would be true for 3 different modes (aksscriptless, bootstrappingclient, aksmachineapi) — those aren't always interchangeable so a blanket "is AKS" helper could mask important distinctions.

pkg/providers/instance/vminstancehelpers.go
comtalyst: (AI-generated)

Fixed. Merged the two branches — SSH configuration is now shared across all modes (the if-else for optional vs required SSH is really a validation concern, not an OS profile concern). Only CustomData handling differs, via a clean `switch provisionMode` at the end. Reduced from 55 lines to 40.

comtalyst: (AI-generated)

Reviewed all changes in the instance module. The remaining mode checks are:

1. `resolveBootstrapAndImageData` — early return for azurevm (calls `resolveAzureVMBootstrapData`). This is a clean separation, not duplication — the two bootstrap paths have no shared logic.

2. `buildAndCreateNIC` — skips LB/NSG in azurevm. Single if-block, clean.

3. `configureStorageProfile` — different image reference format (ARM ID vs community gallery). Single if-else, clean.

4. VM extensions skip — `skipAKSExtensions` check in the promise callback. Single if-block, clean.

These are all small, localized branching points. The configureOSProfile was the only one with actual duplication, now fixed.

pkg/utils/nodeclaim/nodeclaim.go
comtalyst: (AI-generated)

Agreed on the direction. Created `VMNodeClass` interface as the replacement abstraction. The adapter (`AKSNodeClassFromAzureNodeClass`) remains for now because the VM provider's function signatures still take `*v1beta1.AKSNodeClass` — changing those to accept the interface is a larger refactor that touches the instance type provider, launch template resolution, and test infrastructure.

Plan: migrate VM provider functions to `VMNodeClass` incrementally, then delete the adapter and the `json:"-"` fields on AKSNodeClass. The interface is in place to make this possible.

pkg/controllers/nodeclass/termination/events.go
comtalyst: (AI-generated)

Keeping `client.Object` here. The event function passes the object to the Kubernetes event recorder which requires `client.Object` (for metadata like name, namespace, GVK). `VMNodeClass` is a domain interface for VM-creation fields — embedding `client.Object` into it would conflate domain concerns with Kubernetes runtime concerns. Both `AKSNodeClass` and `AzureNodeClass` already satisfy `client.Object` via their embedded `metav1.ObjectMeta`, so the current signature works correctly for both types without coupling.

PR #1456, opened 2026-02-25
refactor: share the same acceptance tests between machine instance and VM instance 

comtalyst:
Fixes # 

**Description**

Consolidate VM-based and AKS Machine API test suites into unified parameterized suites, eliminating duplication.

- Merge `suite_drift_test.go` + `suite_aksmachineapi_drift_test.go` into unified suite
- Merge `suite_features_test.go` + `suite_aksmachineapi_features_test.go` into unified suite
- Merge `suite_offerings_test.go` + `suite_aksmachineapi_offerings_test.go` into unified suite
- Consolidate `garbagecollection/suite_test.go` and `inplaceupdate/suite_test.go` into their AKS machine counterparts
- Net -790 lines

**How was this change tested?**

* Same tests, reorganized

**Does this change impact docs?**
- [ ] Yes, PR includes docs updates
- [ ] Yes, issue opened: # 
- [X] No

**Release Note**

```release-note

```

PR #1455, opened 2026-02-25
feat: AKS machine creation batching mode

comtalyst:
Fixes # 

**Description**

Batch creation for AKS machines — groups multiple concurrent create requests with identical configurations into a single API call using the `BatchPutMachine` header.
See `designs/0009-aks-machines-batch-creation.md` for full details.

**How was this change tested?**

* Same tests w/ Machine API
* Acceptance tests in #1506

**Does this change impact docs?**
- [ ] Yes, PR includes docs updates
- [ ] Yes, issue opened: # 
- [X] No

**Release Note**

```release-note

```

PR #1491, opened 2026-03-06
test: add AKSMachineAPI + Batch as additional test configuration

comtalyst:
**Description**

Add batch creation as a new test configuration alongside the existing non-batch Machine API and AKSScriptless configs. Introduces `setupAKSMachineAPIModeWithBatch()` shared helper and new "AKSMachineAPI + Batch" contexts across all test suites. Exercises the batch grouper → coordinator → GET poller path through the same shared test suites.

Replaces #1482 (closed due to base branch rebuild). Depends on #1456 (test-reunification) and #1455 (batch-system).

* Unit tests here

- [ ] Yes, issue opened: # 
- [X] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/cloudprovider/suite_features_test.go
+++ b/pkg/cloudprovider/suite_features_test.go
@@ -908,6 +908,28 @@ var _ = Describe("CloudProvider - Features", func() {
 		})
 	})
 
+	Context("ProvisionMode = AKSMachineAPI + Batch", func() {
+		BeforeEach(func() { setupAKSMachineAPIModeWithBatch() })
+		AfterEach(func() { teardownTestEnvironment() })
+
+		runCommonGPUTests()
+		runCommonEphemeralDiskTests()
+		runCommonAdditionalTagsTests()
+		runCommonSubnetTests()
+		runCommonKubeletConfigTests()
+	})
+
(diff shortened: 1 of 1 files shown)


comtalyst:
(AI-generated)

### E2E results and RP-side bug found

E2E runs with this PR's batch image + `BATCH_CREATION_ENABLED=true` uncovered a bug in the RP's batch machine handling: the primary machine (from the PUT URL path) loses all per-machine tags and zones.

**Root cause:** The RP's `cleanUpInvalidDuplicateMachinesAndZones` discards the primary machine's `BatchPutMachine` header entry, and `PrepareBatchMachines` only applies header fields to non-primary clones. Since the coordinator clears tags/zones from the PUT body, the primary machine is created with no ownership tag and no zone.

**Impact:** Every batch's first machine has no `karpenter.azure.com_aksmachine_nodeclaim` tag (breaks GC, crash-recovery) and no zone (breaks topology spread). For single-NodeClaim tests (batch size 1), every machine is the primary — so all single-node tests fail.

**RP fix:** 

**This PR does not need changes** — the bug is on the RP side. But this PR cannot pass E2E until the RP fix is merged.

PR #1498, opened 2026-03-06
chore: remove outdated .claude/settings.json

comtalyst:
**Description**

The model config is outdated in terms of model technology, and can stealthily hold developers back
The model config is better not be in repo-level, but rather environment-level? For codespace, could consider adding it somewhere else.

* 

- [ ] Yes, issue opened: # 
- [X] No

**Release Note**

```release-note

```


.claude/settings.json
```diff
@@ -1,14 +0,0 @@
-{
-  "env": {
-    "ANTHROPIC_AUTH_TOKEN": "sk-",
```
matthchr, line 3: I think/thought that we need this file for the .devcontainer we have for claude to work.

Are you using that container @comtalyst or something else?

comtalyst: @matthchr No I was using my local environment. But from my understanding, Claude Code should generate its own config when not available.

comtalyst: In fact, the token is even empty here. This file wouldn't work automatically without dev changes.

comtalyst: @matthchr Running `claude` doesn't work on codespaces either. The risk here should be low enough, and we can revisit this later after some "enchantments".

matthchr: Token being empty here is expected/OK. It's for using claude with litellm or copilot-api-proxy, which is what is run by the claude-code-copilot-api container, see this cmd:
> "nohup bash -c '$(yarn global bin)/copilot-api start --github-token $GITHUB_TOKEN > /tmp/copilot-api.log &

So you don't need a real token because you're not using anthropics API anyway.

I'm not advocating keeping this file but I do think this file as-is makes claude-code-copilot-api container work, so if we want to drop this file from the repo we may need to also update that container to have it generate or otherwise use a correct .claude configuration for its use-case.

matthchr: (And I do occasionally use `claude-code-copilot-api` container as a nice way to have claude working in Karpenter -- actually I am curious how you are using it without it -- do you have your own claude sub you're using?

PR #1499, opened 2026-03-06
refactor: convert pure unit tests from Ginkgo to table-driven

comtalyst:
**Description**

Converts pure unit tests (no external dependencies, no provisioning, no env.Client) from Ginkgo BDD-style to standard Go table-driven tests for improved developer experience and faster execution.

Files converted:

- pkg/apis/v1beta1: hash, localdns, status tests
- pkg/apis/v1alpha2: hash, localdns, status tests
- pkg/utils: subnet parser tests
- pkg/providers/instance: helpers and utils tests
- pkg/providers/instancetype: ephemeral disk and kube reserved (extracted to separate files)
- pkg/metrics: image selection error count
- pkg/operator/options: validation, env var fallback, all option tests

E2E tests using ExpectProvisionedAndWaitForPromises or env.Client remain in Ginkgo.

* Unit tests here

- [ ] Yes, issue opened: # 
- [X] No

**Release Note**

```release-note

```


pkg/operator/options/options_test.go
```diff
@@ -17,336 +17,343 @@ limitations under the License.
 package options_test
```
comtalyst, line 17: Is it still making sense for some test files to be named `suite_test.go` if they stop using ginkgo?

comtalyst: Good catch — no, it doesn't make sense anymore. Renamed `suite_test.go` → `options_test.go` in 9e26c795, since the file has zero Ginkgo references and is now purely standard Go table-driven tests.

(The other `suite_test.go` in `pkg/providers/instancetype/` is kept as-is since it still contains the Ginkgo suite bootstrap for the remaining Ginkgo-based tests there.)

comtalyst: Are you sure about the other in instancetype package? I thought they are just testing `instancetype.List`, which could be tested without ginkgo? Not sure about the role of `ExpectApplied` either, but perhaps something worth testing? Or I could be wrong here.

comtalyst: You're right to question it further — I was too quick to dismiss. Let me clarify what's actually in `instancetype/suite_test.go` after this PR:

The extracted files (`kube_reserved_test.go`, `ephemeral_disk_test.go`) are already pure Go table-driven — no Ginkgo, no `env.Client`, just direct function calls. Those are done.

What remains in `suite_test.go` (798 lines) is all Ginkgo-based and genuinely depends on the test environment infrastructure:
- `TestAzure` → `RunSpecs` (the Ginkgo bootstrap)
- `BeforeSuite`/`AfterSuite` managing `env.Stop()`
- `BeforeEach` resetting `cluster`, `azureEnv`, `testOptions`, populating NSGs
- `AfterEach` calling `ExpectCleanedUp(ctx, env.Client)`
- The `Describe/Context/It` blocks testing filtering, LocalDNS cache invalidation, MaxPods, Encryption at Host, etc. — all calling `ExpectApplied(ctx, env.Client, ...)` and `azureEnv.InstanceTypesProvider.List(ctx, nodeClass)`

So `suite_test.go` in instancetype is still a legitimate Ginkgo suite file. The tests in it call `env.Client` (to apply CRDs into the fake kube-apiserver) and `azureEnv.InstanceTypesProvider.List()` (which depends on the wired-up fake Azure environment). They are not pure unit tests and were intentionally left unconverted, per the PR description: *"E2E tests using ExpectProvisionedAndWaitForPromises or env.Client remain in Ginkgo."*

`ExpectApplied` itself is from karpenter-core's test expectations — it does a real `client.Create`/`client.Update` against the envtest apiserver, so those tests can't trivially drop the Ginkgo harness without also replacing the environment setup.

comtalyst: Your direction on "E2E tests using ExpectProvisionedAndWaitForPromises or env.Client remain in Ginkgo." is correct, but I'm just not sure that `ExpectApplied` contribute to anything here, given what we really want to test is `InstanceType.List()` (unless I miss more things). The Before/After/"frameworks" stuff doesn't matter if the test doesn't use them. Have they really been useful? Can we test `List()` (and more validations, if any) without the "wired-up Azure environment"?
If you are not sure, go and do some experiments then get back with results/proof.

In all this, I'm talking about `instancetype/suite_test.go` by the way.

comtalyst: You're right. I went and traced the code — here's what I found:

**`InstanceTypesProvider.List()` does NOT need `env.Client` or `ExpectApplied`.**

The `DefaultProvider` struct (in `instancetypes.go`) has no `client.Client` field at all. `List(ctx, nodeClass)` reads everything from the passed-in `nodeClass` object directly (`.Spec.Kubelet`, `.Spec.ImageFamily`, `.Spec.OSDiskSizeGB`, `.GetEncryptionAtHost()`, `.IsLocalDNSEnabled()`, etc.) and gets SKU data from the in-memory `skuClient` (backed by `fake.ResourceSKUsAPI`). Zero Kubernetes API calls in the List path.

Looking at the remaining tests in `suite_test.go` (798 lines), here's the actual dependency map:

| Test block | Calls `ExpectApplied`? | Actually needs it? |
|---|---|---|
| Filtering in InstanceType | No | No |
| MaxPods | No | No |
| Basic | No | No |
| Filtering GPU SKUs AzureLinux | Yes | **No** — cargo-culted |
| Filtering by Encryption at Host | Yes | **No** — cargo-culted |
| LocalDNS DescribeTable | Yes | **No** — cargo-culted |
| Cache invalidation with LocalDNS | Yes | **No** — cargo-culted |
| Ephemeral Disk context | Empty (already extracted) | N/A |

The tests that call `ExpectApplied` before `List()` don't need to — `List` never fetches the nodeClass from the apiserver. The only real `env` dependency is `test.ApplyDefaultStatus()` which calls `env.KubernetesInterface.Discovery().ServerVersion()` to populate `nodeClass.Status.KubernetesVersion`. But that could be replaced with a hardcoded version string since it's test data anyway.

**Bottom line: all remaining tests in `instancetype/suite_test.go` could be converted to standard Go table-driven tests without Ginkgo or the envtest apiserver.** The `azureEnv.InstanceTypesProvider` itself is fully constructed from fakes (`fake.ResourceSKUsAPI`, `fake.PricingAPI`, in-memory caches) — it doesn't need `env.Client` at all.

This is a bigger refactor than what this PR scoped ("pure unit tests only"), but the answer to your question is: yes, these tests can absolutely live without the Ginkgo harness and the wired-up environment. If you'd like, I can tackle that as a follow-up, or extend this PR.

comtalyst: Hold up. If it still requires `fakes`, then this could be "something in-between". Let's keep it this way for now. Good catch.

comtalyst:
(AI-generated)

**Rebased onto updated `comtalyst/test-reunification`** (which now includes the main rebase + all fixes from PR #1456).

**Additional fix** (`b6b1fea4`): The Ginkgo→Go conversion of `aksmachineinstanceutils_test.go` was written before PR #1468 (MachineStatus migration) merged, so the converted tests still referenced the removed `KarpenterAKSMachineCreationTimestampTagKey` tag constant, causing a compile error. Fixed by:
- Removing the timestamp tag from `makeBaseMachine()` setup
- Using `Status.CreationTimestamp` directly (matching main's behavior after #1468)
- Removing the now-irrelevant "should handle missing creation time gracefully" test
- Removing unused `time` import

PR #1502, opened 2026-03-08
feat: add Custom Linux OS Configuration and failSwapOn (custom Kubelet config) support for aksmachineapi + bootstrappingclient modes

comtalyst:
**Description**

Add linuxOSConfig field to AKSNodeClassSpec, enabling users to configure sysctl parameters, transparent huge pages, and swap file size on Linux agent nodes. This implements the [AKS Custom Linux OS Configuration](https://learn.microsoft.com/en-us/azure/aks/custom-node-configuration) feature. Also add failSwapOn (Custom Kubelet Config) support, as swapFileSizeMB relies on it being set to false

* New unit tests
* Verified the parity in validation with AKS API
* Self-hosted E2E w/ https://github.com/Azure/karpenter-provider-azure/pull/1505

- [ ] Yes, issue opened: # 
- [X] No

**Release Note**

```release-note

```

### DIFF
--- a/test/suites/utilization/fips_test.go
+++ b/test/suites/utilization/fips_test.go
@@ -19,6 +19,7 @@ package utilization_test
 import (
 	"fmt"
 	"strings"
+	"time"
 
 	opstatus "github.com/awslabs/operatorpkg/status"
 	. "github.com/onsi/ginkgo/v2"
@@ -68,6 +69,11 @@ var _ = Describe("FIPS", Label("runner"), func() {
 		})
 
 		It("should provision FIPS-enabled Ubuntu nodes", func() {
+			deadline := lo.Must(time.Parse(time.RFC3339, "2027-01-15T00:00:00Z"))
+			if time.Now().Before(deadline) {
+				Skip("Ubuntu + FIPS doesn't work on 1.34 + Cilium. Should re-enable when we have Ubuntu2204 support for 1.35")
+			}
+
 			nodeClass.Spec.ImageFamily = lo.ToPtr(v1beta1.UbuntuImageFamily)
 			nodeClass.Spec.FIPSMode = lo.ToPtr(v1beta1.FIPSModeFIPS)
 


matthchr:
Can you fix the title to make it clear that this covers more than just Machine API?

edit: Actually I guess this is "just" machine API in the impl, but adds the fields in AKSNodeClass, but https://github.com/Azure/karpenter-provider-azure/pull/1504/changes includes the changes for Bootstrap API.

I sorta wonder if we should just merge both of those together?

pkg/apis/v1beta1/aksnodeclass.go
```diff
@@ -391,6 +397,203 @@ type KubeletConfiguration struct {
...
+type SysctlConfiguration struct {
+	// fsAioMaxNr specifies the maximum number of AIO (Asynchronous I/O) requests.
+	// Maps to fs.aio-max-nr.
+	// +kubebuilder:validation:Minimum=65536
+	// +kubebuilder:validation:Maximum=6553500
+	// +optional
+	FsAioMaxNr *int32 `json:"fsAioMaxNr,omitempty"`
+	// fsFileMax specifies the maximum number of file handles the kernel will allocate.
+	// Maps to fs.file-max.
+	// Note: The AKS documented maximum is 9223372036854775807 (int64 max), but the ARM API type is int32.
```
matthchr: This comment seems like we should close on with AKS RP? Should this be int64 and we got the type wrong in AKS RP?

Either way, leaving this note here seems wrong? Its not something users should be thinking/caring about

pkg/apis/v1beta1/aksnodeclass.go
```diff
@@ -391,6 +397,203 @@ type KubeletConfiguration struct {
...
+	// +optional
+	NetCoreWmemDefault *int32 `json:"netCoreWmemDefault,omitempty"`
+	// netCoreWmemMax specifies the maximum send socket buffer size in bytes.
+	// Maps to net.core.wmem_max.
+	// +kubebuilder:validation:Minimum=212992
+	// +kubebuilder:validation:Maximum=134217728
+	// +optional
+	NetCoreWmemMax *int32 `json:"netCoreWmemMax,omitempty"`
+	// netIPv4IPLocalPortRange specifies the local port range that is used by TCP and UDP traffic.
+	// Must be in the format "first last", where first is in the range 1024-60999 and last is in the range 32768-65535.
```
matthchr, line 539: This matches AKS RP but not sold this is the actual format we want to accept here?

It is what the raw Linux sysctl takes though so maybe worth keeping as-is?

We should maybe discuss?

matthchr: same with other port ranges below

pkg/apis/v1beta1/aksnodeclass.go
```diff
@@ -391,6 +397,203 @@ type KubeletConfiguration struct {
...
+	// +optional
+	Sysctls *SysctlConfiguration `json:"sysctls,omitempty"`
+	// transparentHugePageDefrag sets the kernel's transparent_hugepage/defrag behavior.
+	// +kubebuilder:validation:Enum:={always,defer,"defer+madvise",madvise,never}
+	// +optional
+	TransparentHugePageDefrag *string `json:"transparentHugePageDefrag,omitempty"`
+	// transparentHugePageEnabled sets the kernel's transparent_hugepage/enabled behavior.
+	// +kubebuilder:validation:Enum:={always,madvise,never}
+	// +optional
+	TransparentHugePageEnabled *string `json:"transparentHugePageEnabled,omitempty"`
```
matthchr: We should define actual enum consts here?

pkg/apis/v1beta1/aksnodeclass.go
```diff
@@ -391,6 +397,203 @@ type KubeletConfiguration struct {
...
+	// +kubebuilder:validation:Minimum=1
+	// +optional
+	SwapFileSizeMB *int32 `json:"swapFileSizeMB,omitempty"`
+	// sysctls specifies sysctl settings for Linux agent nodes.
+	// +optional
+	Sysctls *SysctlConfiguration `json:"sysctls,omitempty"`
+	// transparentHugePageDefrag sets the kernel's transparent_hugepage/defrag behavior.
+	// +kubebuilder:validation:Enum:={always,defer,"defer+madvise",madvise,never}
+	// +optional
+	TransparentHugePageDefrag *string `json:"transparentHugePageDefrag,omitempty"`
```
matthchr: We should define actual enum consts here?

pkg/apis/v1beta1/aksnodeclass.go
```diff
@@ -391,6 +397,203 @@ type KubeletConfiguration struct {
...
 }
 
+// LinuxOSConfiguration defines the Custom Linux OS Configuration for nodes.
+// These settings are applied at node provisioning time and map to AKS Custom Linux OS Configuration.
+// https://learn.microsoft.com/en-us/azure/aks/custom-node-configuration
+type LinuxOSConfiguration struct {
+	// swapFileSizeMB specifies the size in MB of a swap file that will be created on each node.
+	// +kubebuilder:validation:Minimum=1
+	// +optional
+	SwapFileSizeMB *int32 `json:"swapFileSizeMB,omitempty"`
```
matthchr: MB suffix anti-pattern in k8s, possibly?

I think often these are strings which user chooses the suffix for? Like for mem limits?

pkg/apis/v1beta1/aksnodeclass.go
```diff
@@ -391,6 +397,203 @@ type KubeletConfiguration struct {
...
+	// netIPv4TCPTwReuse enables/disables reuse of TIME_WAIT sockets for new connections.
+	// Maps to net.ipv4.tcp_tw_reuse.
+	// +optional
+	NetIPv4TCPTwReuse *bool `json:"netIPv4TCPTwReuse,omitempty"`
+	// netIPv4TCPKeepaliveIntvl specifies the frequency of the probes sent out. Measured in seconds.
+	// Maps to net.ipv4.tcp_keepalive_intvl.
+	// +kubebuilder:validation:Minimum=10
+	// +kubebuilder:validation:Maximum=90
+	// +optional
+	NetIPv4TCPKeepaliveIntvl *int32 `json:"netIPv4TCPKeepaliveIntvl,omitempty"`
```
matthchr, line 602: I _think_ in this case it makes sense to keep this as-is, but calling out that in k8s generally timespans are strings and they take the `s` or `ms` suffix so the user can choose the units.

In _this_ context though, I am not sure that's valuable and matching the actual sysctl is probably better.

pkg/cloudprovider/suite_aksmachineapi_features_test.go
```diff
@@ -703,5 +703,181 @@ var _ = Describe("CloudProvider", func() {
...
+				Expect(aksMachine.Properties.OperatingSystem.LinuxProfile).ToNot(BeNil())
+				linuxOSConfig := aksMachine.Properties.OperatingSystem.LinuxProfile.LinuxOSConfig
+				Expect(linuxOSConfig).ToNot(BeNil())
+
+				// Verify top-level fields
+				Expect(lo.FromPtr(linuxOSConfig.SwapFileSizeMB)).To(Equal(int32(1500)))
+				Expect(lo.FromPtr(linuxOSConfig.TransparentHugePageDefrag)).To(Equal("madvise"))
+				Expect(lo.FromPtr(linuxOSConfig.TransparentHugePageEnabled)).To(Equal("always"))
+
+				// Verify sysctl fields
```
matthchr, line 790: Not verifying all the sysctls you set?

pkg/cloudprovider/suite_aksmachineapi_features_test.go
```diff
@@ -703,5 +703,181 @@ var _ = Describe("CloudProvider", func() {
...
+				// Sysctls should be set
+				Expect(linuxOSConfig.Sysctls).ToNot(BeNil())
+				Expect(lo.FromPtr(linuxOSConfig.Sysctls.VMMaxMapCount)).To(Equal(int32(262144)))
+				Expect(lo.FromPtr(linuxOSConfig.Sysctls.VMSwappiness)).To(Equal(int32(10)))
+
+				// Other sysctls should be nil
+				Expect(linuxOSConfig.Sysctls.FsAioMaxNr).To(BeNil())
+			})
+
+			It("should create AKS machine with only THP settings when only THP is specified", func() {
```
matthchr: minor: don't abbreviate THP?

pkg/cloudprovider/suite_aksmachineapi_features_test.go
matthchr: I think gomega has structural assertions you can use here (so the types dont have to match as long as the fields do?), but if not there are various libs we could use, or we could json serialize if they're supposed to match like that... or we could just list all the fields explicitly.

pkg/controllers/nodeclass/status/validation.go
```diff
@@ -83,6 +86,16 @@ func (r *ValidationReconciler) Reconcile(ctx context.Context, nodeClass *v1beta1
 		}
 	}
 
+	// Validate LinuxOSConfig spec fields that can't be expressed as CRD markers or CEL
+	if msg := validateNetIPv4IPLocalPortRange(nodeClass); msg != "" {
```
matthchr: I dont think we want this.

I think we probably can write CEL validation for this, see https://kubernetes.io/docs/reference/using-api/cel/#cel-options-language-features-and-libraries

If not, we could change the format so that we can (see my comment above on the field)

pkg/controllers/nodeclass/status/validation.go
```diff
@@ -112,3 +125,44 @@ func (r *ValidationReconciler) validateDiskEncryptionSetRBAC(ctx context.Context
 	log.FromContext(ctx).V(1).Info("Disk Encryption Set RBAC validation passed", "desID", r.parsedDiskEncryptionSetID)
 	return nil
 }
+
+// validateNetIPv4IPLocalPortRange checks that the port range string has valid numeric bounds.
+// CRD pattern validation ensures the format is "first last", but numeric range checks require parsing.
+// AKS API enforces: first ∈ [1024, 60999], last ∈ [32768, 65535], first ≤ last.
+func validateNetIPv4IPLocalPortRange(nodeClass *v1beta1.AKSNodeClass) string {
```
matthchr: IMO remove (see above comment)

pkg/controllers/nodeclass/status/validation_test.go
```diff
@@ -113,6 +113,88 @@ var _ = Describe("Validation Reconciler", func() {
 		})
 	})
 
+	Context("netIPv4IPLocalPortRange validation", func() {
```
matthchr: IMO remove (see above)

pkg/providers/instance/aksmachineinstancehelpers.go
```diff
@@ -120,7 +120,15 @@ func (p *DefaultAKSMachineProvider) buildAKSMachineTemplate(ctx context.Context,
 				OSDiskSizeGB: nodeClass.Spec.OSDiskSizeGB, // AKS machine API defaults it if nil
 				OSDiskType:   osDiskType,
 				EnableFIPS:   enableFIPS,
-				// LinuxProfile:   nil,
+				LinuxProfile: func() *armcontainerservice.MachineOSProfileLinuxProfile {
+					linuxOSConfig := configureLinuxOSConfig(nodeClass)
```
matthchr, line 124: minor, style: Could call this `newMachineLinuxOSConfig`?

`configure` prefix makes me think it's got a parameter of the thing it is configuring, which I don't love as much. I like "makes a new thing" pattern you have here but feel the name doesn't match.

pkg/apis/v1beta1/aksnodeclass.go
comtalyst: Looks like anti-pattern so far. However, several issues on changing that:
- AKS (not K8s API obviously) only use and accept integer with assumed unit like this
  - There exist `containerLogMaxSize` here, but ended up having to convert to MB as that's what AKS want. Conversion is an additional complexity, plus potentially losing the significance if provided smaller unit.
- We've been inconsistent on this too. E.g., `containerLogMaxSize` vs `osDiskSizeGB`.

The question becomes, is K8s pattern worth that complexity given this? Assuming push for AKS change is not worth it.

pkg/apis/v1beta1/aksnodeclass.go
comtalyst: Given both AKS RP and raw Linux sysctl formats, I am leaning towards keeping as-is. 
Is bad UX your concern? Technically, some users might even prefer no extra fields from what they know.

pkg/apis/v1beta1/aksnodeclass.go
matthchr: Conversion isn't actually _that_ complex, is it?

The biggest/worst issue is rounding probably, when fractions of MB are given, as we have to translate that into an MB value, and we probably need to document that behavior...

We do conversion in this form already, look at `ConvertContainerLogMaxSizeToMB` (you must have already seen).

Ack that we already have osDiskSizeGB as well.

I _think_ I still tend towards the k8s recommended shape here, but curious if @tallaxes has thoughts

comtalyst: > ConvertContainerLogMaxSizeToMB (you must have already seen).

Yes, and I wish there is nothing more of that because of the rounding issues.

matthchr: Some relevant docs:
https://github.com/kubernetes/community/blob/main/contributors/devel/sig-architecture/api-conventions.md#units

Which isn't really that helpful because it basically says "eh do whatever", and it feels like we should have a rule ourselves.

comtalyst: I think it is bad UX either way. I _think_ I would rather choose the honestly bad (having 'MB') than implicitly bad ("we will round the number"). 
Unless we want to get serious about following K8s pattern, which, will still have inconsistency issue with existing APIs (and another timestamp issue we mentioned). And because of rounding, I think the trade-off is not worth.

pkg/apis/v1beta1/aksnodeclass.go
comtalyst: AKS RP only accept int32. But, yes, could be a follow-up.
Note removed.

pkg/apis/v1beta1/aksnodeclass.go
tallaxes: Agree on imperfect UX in either case, but think I still prefer the suffix, just because it makes the units unambiguous (though don't like it visually). This is integer, so fractions are not an option/issue, right?

comtalyst: @tallaxes do you mean prefer the suffix on the field name (`SwapFileSizeMB`) or suffix provided by the user (`SwapFileSize`)?

pkg/cloudprovider/suite_aksmachineapi_features_test.go
comtalyst: The fields is not exactly the same. E.g., `NetIPv4TcpkeepaliveIntvl` vs `NetIPv4TCPKeepaliveIntvl`, which we choose to "correct" it. I think it is fine keeping this way.

pkg/providers/instance/aksmachineinstancehelpers.go
comtalyst: I would prefer to keep it consistent for now. Could consider changing it with all others, but minor.

pkg/apis/v1beta1/aksnodeclass.go
tallaxes: Ah, my bad, actually missed that units can be provided in the value. Yes, that would be better I think, more aligned with everything else in Kubernetes, with downsides manageable. (Yet another benefit is that it makes MB/MiB distinction possible and explicit.)

comtalyst: Changed to `SwapFileSize`

pkg/apis/v1beta1/aksnodeclass.go
matthchr: I am fine keeping it as-is if we can CEL validate it. I don't want to have to do AKSNodeClass status validation just for this one field because it's wonkily formatted.

pkg/cloudprovider/suite_aksmachineapi_features_test.go
matthchr: Yeah that's fine -- we can just manually list all the fields like you did here.

pkg/apis/v1beta1/aksnodeclass.go
matthchr: Resolving this, looks like you solved the main issue I had which was the lack of validation via CEL. I agree with the general philosophy of "for things in this struct, let's just match the sysctl format as much as makes sense"

pkg/apis/v1beta1/aksnodeclass.go
matthchr: Can you start a thread (maybe in AKS RP channel?) about it, and @ the node sig and RP folks?

Let's start the conversation at least before we forget about it. We don't have to block here on it

pkg/apis/v1beta1/crd_validation_cel_test.go
```diff
@@ -861,4 +861,319 @@ var _ = Describe("CEL/Validation", func() {
...
+					LinuxOSConfig: &v1beta1.LinuxOSConfiguration{
+						Sysctls: &v1beta1.SysctlConfiguration{
+							NetIPv4IPLocalPortRange: lo.ToPtr("60000 32768"),
+						},
+					},
+				},
+			}
+			Expect(env.Client.Create(ctx, nodeClass)).ToNot(Succeed())
+		})
+		It("should accept when port range is not set", func() {
```
matthchr, line 1027: maybe consider a "should reject invalid port range" which just gives a garbage range like "7foo 8bar" or something?

pkg/providers/instance/aksmachineinstancehelpers.go
```diff
@@ -337,13 +347,76 @@ func convertContainerLogMaxSizeToMB(containerLogMaxSize string) *int32 {
 	return customscriptsbootstrap.ConvertContainerLogMaxSizeToMB(containerLogMaxSize)
 }
 
+func convertSwapFileSizeToMB(swapFileSize string) *int32 {
+	// TODO: rename the utils below.
```
matthchr, line 351: Are you planning to send a follow-up for this? (Or have an item tracking it? No need for an item necessarily just an easy fix so we might as well do it ASAP after this?)

pkg/providers/instance/aksmachineinstancehelpers.go
```diff
@@ -337,13 +347,76 @@ func convertContainerLogMaxSizeToMB(containerLogMaxSize string) *int32 {
...
+	if nodeClass == nil || nodeClass.Spec.LinuxOSConfig == nil {
+		return nil
+	}
+
+	linuxOSConfig := &armcontainerservice.LinuxOSConfig{}
+	if nodeClass.Spec.LinuxOSConfig.SwapFileSize != nil && *nodeClass.Spec.LinuxOSConfig.SwapFileSize != "" {
+		linuxOSConfig.SwapFileSizeMB = convertSwapFileSizeToMB(*nodeClass.Spec.LinuxOSConfig.SwapFileSize)
+	}
+	if nodeClass.Spec.LinuxOSConfig.TransparentHugePageDefrag != nil {
+		linuxOSConfig.TransparentHugePageDefrag = lo.ToPtr(string(*nodeClass.Spec.LinuxOSConfig.TransparentHugePageDefrag))
```
matthchr, line 372: minor: Is the intent with conversion here to avoid having the ptr point to the same object?

Technically you can just do this, I think:
```
linuxOSConfig.TransparentHugePageDefrag = (*string)(nodeClass.Spec.LinuxOSConfig.TransparentHugePageDefrag)
```

which is probably a bit simpler.

If you want to make sure we have a copy (so modifications of one don't affect the other), that's fine although it might be worth leaving a note and also is inconsistent with what we're doing below for all the `sysctlConfig`'s (they are not copies so ptrs point to same value and changing one may change the other?)

I am fine either way just seems like maybe worth being consistent and/or leaving a note as to why you chose the one you chose via comment

pkg/providers/instance/aksmachineinstancehelpers.go
comtalyst: There is a pending refactor PRs what the rename will bother. I will do that later.

PR #1504, opened 2026-03-08
feat: wire LinuxOSConfig and failSwapOn into bootstrapping client path

comtalyst:
**Description**

Thread LinuxOSConfiguration through the ImageFamily interface and resolver into the ProvisionClientBootstrap, enabling Custom Linux OS Configuration (sysctls, transparent huge pages, swap file size) for the bootstrapping client (VM-based) provisioning mode.

Stacked on #1502.

Changes:
- Add linuxOSConfig parameter to CustomScriptsNodeBootstrapping interface method
- Update all 5 image family implementations (azlinux, azlinux3, ubuntu2204, ubuntu2404, ubuntu2004) to accept the new parameter
- Add convertLinuxOSConfigToModel() and convertSysctlConfigToModel() converter functions mapping v1beta1 API types to swagger models
- Wire CustomLinuxOSConfig in ConstructProvisionValues()
- Also add failSwapOn (Custom Kubelet Config) support, as swapFileSizeMB relies on it being set to false

* (TODO) NAP E2E w/ https://github.com/Azure/karpenter-provider-azure/pull/1505

- [ ] Yes, issue opened: # 
- [X] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/providers/imagefamily/customscriptsbootstrap/utils.go
+++ b/pkg/providers/imagefamily/customscriptsbootstrap/utils.go
@@ -165,8 +165,10 @@ func convertLinuxOSConfigToModel(linuxOSConfig *v1beta1.LinuxOSConfiguration) *m
 	}
 
 	result := &models.CustomLinuxOSConfig{
-		SwapFileSizeMB: linuxOSConfig.SwapFileSizeMB,
-		Sysctls:        convertSysctlConfigToModel(linuxOSConfig.Sysctls),
+		Sysctls: convertSysctlConfigToModel(linuxOSConfig.Sysctls),
+	}
+	if linuxOSConfig.SwapFileSize != nil && *linuxOSConfig.SwapFileSize != "" {
+		result.SwapFileSizeMB = ConvertContainerLogMaxSizeToMB(*linuxOSConfig.SwapFileSize)
 	}
 	if linuxOSConfig.TransparentHugePageDefrag != nil {
 		result.TransparentHugePageDefrag = lo.ToPtr(string(*linuxOSConfig.TransparentHugePageDefrag))
--- a/pkg/providers/imagefamily/ubuntu_2204_test.go
+++ b/pkg/providers/imagefamily/ubuntu_2204_test.go
@@ -68,7 +68,7 @@ func TestUbuntu2204_CustomScriptsNodeBootstrapping(t *testing.T) {
 	localDNS := &v1beta1.LocalDNS{Mode: v1beta1.LocalDNSModeDisabled}
 	artifactStreaming := &v1beta1.ArtifactStreaming{Enabled: lo.ToPtr(false)}
 	linuxOSConfig := &v1beta1.LinuxOSConfiguration{
-		SwapFileSizeMB:             lo.ToPtr(int32(1500)),
+		SwapFileSize:               lo.ToPtr("1500Mi"),
 		TransparentHugePageEnabled: lo.ToPtr(v1beta1.TransparentHugePageEnabledMadvise),
 	}
 


.claude/settings.json
```diff
@@ -1,14 +0,0 @@
-{
-  "env": {
```
matthchr, line 2: unrelated?

pkg/apis/v1alpha2/zz_generated.deepcopy.go
```diff
@@ -148,6 +148,7 @@ func (in *AKSNodeClassSpec) DeepCopyInto(out *AKSNodeClassSpec) {
 	if in.LinuxOSConfig != nil {
 		in, out := &in.LinuxOSConfig, &out.LinuxOSConfig
 		*out = new(LinuxOSConfiguration)
+		(*in).DeepCopyInto(*out)
```
matthchr, line 151: unrelated?

why is this here?

pkg/apis/v1beta1/zz_generated.deepcopy.go
```diff
@@ -148,6 +148,7 @@ func (in *AKSNodeClassSpec) DeepCopyInto(out *AKSNodeClassSpec) {
 	if in.LinuxOSConfig != nil {
 		in, out := &in.LinuxOSConfig, &out.LinuxOSConfig
 		*out = new(LinuxOSConfiguration)
+		(*in).DeepCopyInto(*out)
```
matthchr, line 151: Same q

pkg/providers/imagefamily/customscriptsbootstrap/provisionclientbootstrap.go
```diff
@@ -136,7 +137,8 @@ func (p *ProvisionClientBootstrap) ConstructProvisionValues(ctx context.Context)
 		// AgentPoolWindowsProfile: &models.AgentPoolWindowsProfile{},               // Unsupported as of now; TODO(Windows)
 		// KubeletDiskType:         lo.ToPtr(models.KubeletDiskTypeUnspecified),    // Unsupported as of now
 		// CustomLinuxOSConfig:     &models.CustomLinuxOSConfig{},                   // Unsupported as of now (sysctl)
-		EnableFIPS: lo.ToPtr(enableFIPS),
+		CustomLinuxOSConfig: convertLinuxOSConfigToModel(p.LinuxOSConfig),
```
matthchr, line 140: Assuming you've confirmed either by code inspection or testing with NPS that this actually works on their end and flows where it needs to flow?

matthchr: (i.e. not enough to pass it to them, they need to then pass it to AgentBaker)

pkg/providers/imagefamily/customscriptsbootstrap/utils.go
```diff
@@ -157,3 +157,56 @@ func convertLocalDNSZoneOverrideToModel(override *v1beta1.LocalDNSZoneOverride)
...
+	if linuxOSConfig == nil {
+		return nil
+	}
+
+	return &models.CustomLinuxOSConfig{
+		SwapFileSizeMB:             linuxOSConfig.SwapFileSizeMB,
+		Sysctls:                    convertSysctlConfigToModel(linuxOSConfig.Sysctls),
+		TransparentHugePageDefrag:  linuxOSConfig.TransparentHugePageDefrag,
+		TransparentHugePageEnabled: linuxOSConfig.TransparentHugePageEnabled,
+		// Ulimits is not supported in our API types; leave nil
```
matthchr: Is this a sysctl we just dont support? I dont see any ref to it in aks docs either?

Possibly could just remove this comment?

pkg/providers/imagefamily/ubuntu_2004.go
```diff
@@ -130,6 +130,7 @@ func (u Ubuntu2004) CustomScriptsNodeBootstrapping(
 	fipsMode *v1beta1.FIPSMode,
 	_ *v1beta1.LocalDNS, // Ubuntu 20.04 does not support LocalDNS
 	artifactStreaming *v1beta1.ArtifactStreaming,
+	_ *v1beta1.LinuxOSConfiguration, // Ubuntu 20.04 does not support LinuxOSConfig
```
matthchr: Does this have scheduling implications that we aren't honoring correctly?

For example what happens when I ask for FIPS + LinuxConfig -- we should reject that possibly at the nodeclass level but I didn't see us doing that?

I _think_ you can't get Ubuntu 2004 any other way than FIPS, but if you can then we need to also consider those cases and block? Silently ignoring here seems wrong? I mean not that this code is wrong but wrong that we could get into a situation where user asked for Ubuntu 2004 + Sysctls and we didn't error but they just magically dont get the sysctls.

pkg/providers/imagefamily/customscriptsbootstrap/utils.go
comtalyst: We don't support now (likely AKS too). Dropped the comment.

pkg/providers/imagefamily/ubuntu_2004.go
comtalyst: This "Ubuntu 20.04 does not support" was most likely a mistake. Dropped.

PR #1505, opened 2026-03-08
test: add e2e integration tests for LinuxOSConfig

comtalyst:
**Description**

Add integration tests verifying Custom Linux OS Configuration is correctly applied to provisioned nodes.

Stacked on #1504 (which is stacked on #1502).

* Self-hosted E2E (Machine API)

- [ ] Yes, issue opened: # 
- [X] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/cloudprovider/cloudprovider.go
+++ b/pkg/cloudprovider/cloudprovider.go
@@ -235,7 +235,8 @@ func (c *CloudProvider) createAKSMachineInstance(ctx context.Context, nodeClass 
 		aksMachinePromise.AKSMachineID,
 		aksMachinePromise.VMResourceID,
 		false,
-		aksMachinePromise.AKSMachineNodeImageVersion)
+		aksMachinePromise.AKSMachineNodeImageVersion,
+		aksMachinePromise.CreationTimestamp)
 	if err != nil {
 		return nil, fmt.Errorf("failed to build NodeClaim from AKS machine template, %w", err)
 	}
--- a/pkg/fake/aksmachinesapi.go
+++ b/pkg/fake/aksmachinesapi.go
@@ -18,6 +18,7 @@ package fake
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"net/http"
 	"strings"
@@ -257,7 +258,7 @@ func (c *AKSMachinesAPI) BeginCreateOrUpdate(
 		AKSMachine:        parameters,
 		Options:           options,
 	}
-	aksMachine := input.AKSMachine
+	aksMachine := deepCopyMachine(input.AKSMachine)
 	id := MkMachineID(input.ResourceGroupName, input.ResourceName, input.AgentPoolName, input.AKSMachineName)
 	aksMachine.ID = &id
 	aksMachine.Name = &input.AKSMachineName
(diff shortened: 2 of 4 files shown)


pkg/providers/imagefamily/customscriptsbootstrap/provisionclientbootstrap.go
matthchr, line 68: minor: I think needs a rebase now after the other one merged?

test/suites/integration/linuxosconfig_test.go
```diff
@@ -0,0 +1,439 @@
...
+	// 1. We create an AKSNodeClass with linuxOSConfig (sysctls, THP, swap) and a NodePool.
+	// 2. We schedule a pod that triggers Karpenter to provision a new node via AKS.
+	// 3. Once the node is ready, we create a "verification pod" on that node to read back
+	//    the kernel settings and confirm they match what we configured.
+	//
+	// How verification pods read settings:
+	//   There's no Kubernetes API to read a node's kernel parameters. The only way is to
+	//   run a process on the node and read from the kernel's virtual filesystems:
+	//   - Sysctls: the `sysctl` command reads /proc/sys/ (e.g., /proc/sys/net/core/somaxconn)
+	//   - THP: read /sys/kernel/mm/transparent_hugepage/enabled and .../defrag
```
matthchr: minor: thp abbreviation may be worth avoiding again? Though in this context pretty clear what it means since it says so later.

test/suites/integration/linuxosconfig_test.go
```diff
@@ -0,0 +1,439 @@
...
+		// Create a privileged pod on the node to read transparent huge page settings
+		verifyPod := createTHPVerificationPod(node.Name)
+		env.ExpectCreated(verifyPod)
+		defer env.ExpectDeleted(verifyPod)
+
+		logs := eventuallyGetPodLogs(verifyPod)
+		By(fmt.Sprintf("Transparent huge page verification output:\n%s", logs))
+
+		// THP enabled file shows the active setting in brackets: always [madvise] never
+		Expect(logs).To(ContainSubstring("[madvise]"))
```
matthchr: Should we be checking for the specific setting that has this field?

If we mix them up, this test would pass still.

matthchr: Can we do similar to above where the assert is something like: `net.core.optmem_max = 40960`, rather than just `40960`?

test/suites/integration/linuxosconfig_test.go
```diff
@@ -0,0 +1,439 @@
...
+
+		// Swap should be enabled. Parse total swap from /proc/meminfo.
+		// SwapTotal line will show the total swap in kB, which should be ~1500MB.
+		// Allow some tolerance since the kernel may report slightly different values.
+		Expect(logs).To(ContainSubstring("SwapTotal:"))
+		// Verify swap is NOT zero (i.e., swap is enabled)
+		Expect(logs).ToNot(ContainSubstring("SwapTotal:           0 kB"))
+	})
+
+	It("should provision a node with full LinuxOSConfig (sysctls + THP + swap)", func() {
```
matthchr: minor: Seems somewhat duplicate with above tests?

IMO we don't need to verify this separately than the above? So either do them all in one like this one is doing or do them all separate like above but not both?

test/suites/integration/linuxosconfig_test.go
```diff
@@ -0,0 +1,439 @@
...
+	sysctlArgs := strings.Join(sysctls, " ")
+	pod := coretest.Pod(coretest.PodOptions{
+		ObjectMeta: metav1.ObjectMeta{
+			GenerateName: "sysctl-verify-",
+			Namespace:    "default",
+		},
+		Image: "mcr.microsoft.com/azurelinux/busybox:1.36",
+		Command: []string{
+			"sh", "-c",
+			fmt.Sprintf("sysctl %s && sleep 30", sysctlArgs),
```
matthchr: Why sleep 30?

test/suites/integration/linuxosconfig_test.go
```diff
@@ -0,0 +1,439 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			GenerateName: "thp-verify-",
+			Namespace:    "default",
+		},
+		Image: "mcr.microsoft.com/azurelinux/busybox:1.36",
+		Command: []string{
+			"sh", "-c",
+			"cat /sys/kernel/mm/transparent_hugepage/enabled && " +
+				"cat /sys/kernel/mm/transparent_hugepage/defrag && " +
+				"sleep 30",
```
matthchr: same q

matthchr: and below as well

test/suites/integration/linuxosconfig_test.go
comtalyst: Took this from LocalDNS E2E. I don't think we need it + tested that we don't need it

PR #1510, opened 2026-03-09
chore: improve machine label filtering

matthchr:

* Unit tests

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,4 +1,4 @@
-FROM mcr.microsoft.com/oss/go/microsoft/golang:1.25-bookworm@sha256:7f31d02dec6d256dfcfa0c7fcc54f1002f8a0c71d6a6a60740e193abce6a1181
+FROM mcr.microsoft.com/oss/go/microsoft/golang:1.26-bookworm@sha256:0c8bf18868e61d346f4315ba1ed965cc8092d168f69a9d30729be2a56c4a7b27
 
 # [Optional] Uncomment this section to install additional OS packages.
 # graphviz for pprof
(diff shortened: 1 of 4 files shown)


pkg/providers/instance/aksmachineinstancehelpers.go
```diff
@@ -255,25 +255,11 @@ func configureLabelsAndMode(nodeClaim *karpv1.NodeClaim, instanceType *corecloud
...
-		"failure-domain.beta.kubernetes.io/region",
-		"beta.kubernetes.io/os",
-		"beta.kubernetes.io/arch",
-		"failure-domain.beta.kubernetes.io/zone",
-		"topology.kubernetes.io/zone",
-		"topology.kubernetes.io/region",
-		"node.kubernetes.io/instance-type",
-		"kubernetes.io/arch",
-		"kubernetes.io/os",
-		"node.kubernetes.io/windows-build",
```
comtalyst, line 271: This `"node.kubernetes.io/windows-build"` is missing from kubelet labels list in `labels.go`, but could consider later when Karpenter support Windows?

matthchr: yeah good eye. So I checked and AKS RP doesn't actually reserve that label (Karpenter can set it). It's not in Kubelet labels list because it has `node.kubernetes.io` so it's already allowed for kubelet (see `kubeletLabelNamespaces`)

matthchr: which is basically a long-winded way of me saying: "I think we're already doing the right thing with that label. AKS doesn't set it so it's not reserved, we can set it from Karpenter.

We may need another item about if aks _should_ set it when we support windows, but that's maybe a separate discussion.

PR #1512, opened 2026-03-10
Add kuberentes.azure.com/NodeHealthy repair policy condition

mingqishao:
**Description**
This add a new condition in the RepairPolicies to integrate with [Azure's Cluster Health Monitor.](https://github.com/Azure/cluster-health-monitor/blob/5172303fe1c06b46a81dca2c33060c7174539c5c/pkg/controller/checknodehealth/controller.go#L65) 

* No test for this. E2E will be done in Cluster Health Monitor when it integrate with Karpenter.

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/cloudprovider/cloudprovider.go
+++ b/pkg/cloudprovider/cloudprovider.go
@@ -545,7 +545,7 @@ func (c *CloudProvider) RepairPolicies() []cloudprovider.RepairPolicy {
 			TolerationDuration: 10 * time.Minute,
 		},
 		{
-			ConditionType:      "kuberentes.azure.com/NodeHealthy",
+			ConditionType:      "kubernetes.azure.com/NodeHealthy",
 			ConditionStatus:    corev1.ConditionFalse,
 			TolerationDuration: 0,
 		},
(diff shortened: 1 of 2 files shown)


pkg/cloudprovider/cloudprovider.go
```diff
@@ -544,6 +544,11 @@ func (c *CloudProvider) RepairPolicies() []cloudprovider.RepairPolicy {
 			ConditionStatus:    corev1.ConditionUnknown,
 			TolerationDuration: 10 * time.Minute,
 		},
+		{
+			ConditionType:      "aks.azure.com/NodeHealthy",
```
matthchr: minor: Our labels all use `kuberentes.azure.com` -- should we do the same here? Then we could use our label const for `kubernetes.azure.com` prefix.

mingqishao: sure. I change to "kubernetes.azure.com/NodeHealthy"

pkg/cloudprovider/cloudprovider.go
```diff
@@ -544,6 +544,11 @@ func (c *CloudProvider) RepairPolicies() []cloudprovider.RepairPolicy {
 			ConditionStatus:    corev1.ConditionUnknown,
 			TolerationDuration: 10 * time.Minute,
 		},
+		{
+			ConditionType:      "kuberentes.azure.com/NodeHealthy",
```
matthchr: Looks like we have some tests here:
test/suites/integration/repair_policy_test.go

Can you add an entry for this in the `DescribeTable("Conditions", func(unhealthyCondition corev1.NodeCondition) {`

Section?

mingqishao: I added that.

PR #1515, opened 2026-03-13
fix: preserve primary machine tags and zones in batch PUT body

comtalyst:
## Summary

The batch coordinator stripped per-machine fields (tags, zones) from the PUT body for all machines, including the primary. The RP creates the primary machine from the body and only applies header fields to non-primary clones — the primary's header entry is silently filtered out. This caused the first machine in every batch to lose its ownership tag, all custom tags, and its zone assignment.

**Impact:**
- Garbage collection misidentified tagless machines as orphans
- Crash-recovery couldn't match machines to NodeClaims
- Zone topology spread constraints were violated when the primary got an arbitrary zone

**Fix:** Send the primary's original template as the PUT body (with tags and zones intact) and exclude the primary from the header's `batchMachines` array. Non-primary machines continue to receive their per-machine fields from the header, which the RP applies during cloning.

## Changes

- `coordinator.go`: Use `batch.requests[0].template` as body instead of cleared shared template. Exclude primary from `buildBatchHeader`.
- `aksmachinesapi.go` (fake): Create primary from body, then non-primary from entries — matching real RP behavior.
- `coordinator_test.go`: Updated tests to verify primary's fields are preserved in body and excluded from context entries.

## Test plan

- [x] All batch unit tests pass (22/22)
- [x] All fake package tests pass
- [x] Full project compiles
- [ ] E2E validation with batch enabled (pending — retry build in progress)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pkg/fake/aksmachinesapi.go
+++ b/pkg/fake/aksmachinesapi.go
@@ -300,40 +300,13 @@ func (c *AKSMachinesAPI) createSingleMachine(input *AKSMachineCreateOrUpdateInpu
 	})
 }
 
-// createBatchMachines handles batch creation: creates one machine from the body
-// (the primary) plus one machine per batch entry (non-primary) from the header.
-//
-// This simulates what the real Azure API does:
-//   - Primary machine: created from the PUT body (has its own tags/zones)
-//   - Non-primary machines: cloned from the body, then overwritten with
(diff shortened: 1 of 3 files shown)


comtalyst:
Closing in favor of the RP-side fix (Option B). The RP should handle all machines uniformly via the header

PR #1525, opened 2026-03-16
Design: CRD validation sync strategy with Machine API

comtalyst:


comtalyst:
(AI-generated)

The Layer 1 implementation from this design doc is now in #1545 (CRD validation sync for labels and taints). The design doc itself still serves as the architectural roadmap for Layers 2-4.

PR #1535, opened 2026-03-16
feat: add instancetype controller

matthchr:
**Description**
Moves just-in-time SKU computation out of the hot path and into its own controller.

* Unit tests
* E2E

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note
NONE
```

### DIFF
--- a/pkg/controllers/instancetype/controller.go
+++ b/pkg/controllers/instancetype/controller.go
@@ -1,0 +1,66 @@
+/*
+Portions Copyright (c) Microsoft Corporation.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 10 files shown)


pkg/controllers/instancetype/controller.go
```diff
@@ -0,0 +1,66 @@
...
+	controllerruntime "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/karpenter/pkg/operator/injection"
+
+	instancetypeprovider "github.com/Azure/karpenter-provider-azure/pkg/providers/instancetype"
+)
+
+const (
+	InstanceTypesRefreshInterval = 12 * time.Hour
```
matthchr, line 34: worth noting is is faster than it was before (it was 23h before).

@tallaxes if you think we should keep it at 23h vs 12.

PR #1538, opened 2026-03-16
test: add test for unavailableofferings checking instanceType on cp.List

matthchr:

* Unit test

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/controllers/nodeclass/hash/suite_test.go
+++ b/pkg/controllers/nodeclass/hash/suite_test.go
@@ -44,7 +44,7 @@ import (
 
 var ctx context.Context
 var env *coretest.Environment
-var awsEnv *test.Environment
+var azureEnv *test.Environment
 var hashController *hash.Controller
 
 func TestAPIs(t *testing.T) {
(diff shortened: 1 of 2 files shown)


matthchr:
No but I thought there was one so I added this test, which showed it was actually fine. The case is interesting though (sku that has since been offering restricted) so figured it was worth keeping the test regardless for regression reasons.

pkg/controllers/nodeclass/hash/suite_test.go
```diff
@@ -57,7 +57,7 @@ var _ = BeforeSuite(func() {
 	env = coretest.NewEnvironment(coretest.WithCRDs(apis.CRDs...), coretest.WithCRDs(v1alpha1.CRDs...), coretest.WithFieldIndexers(test.AKSNodeClassFieldIndexer(ctx)))
 	ctx = coreoptions.ToContext(ctx, coretest.Options())
 	ctx = options.ToContext(ctx, test.Options())
-	awsEnv = test.NewEnvironment(ctx, env)
```
comtalyst, line 60: really?

matthchr: lol yea that was my thinking too =)

PR #1539, opened 2026-03-18
feat: adding amd support and byo driver support

MaximilianoUribe:
Fixes #  [AKS Public Roadmap issue #5362](https://github.com/Azure/AKS/issues/5362)

**Description**
Adding support for AMD gpu skus, this goes hand-in-hand with support for BYO driver field into AKSNodeClass since we don't support AMD driver installation in AKS as of March 30, 2026.

* Ran GPU test suite

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


pkg/providers/instancetype/instancetype.go
```diff
@@ -230,11 +230,17 @@ func setRequirementsHyperVGeneration(requirements scheduling.Requirements, sku *
 }
 
 func setRequirementsGPU(requirements scheduling.Requirements, sku *skewer.SKU, vmsize *skewer.VMSizeType) {
-	if utils.IsNvidiaEnabledSKU(sku.GetName()) {
+	manufacturer := utils.GetGPUManufacturer(sku.GetName())
```
matthchr, line 233: We should get Wilson to start a conversation with the compute folks so we don't have to maintain this list. They should return it in the compute API

pkg/apis/v1beta1/aksnodeclass.go
```diff
@@ -277,6 +280,30 @@ const (
...
+type GPU struct {
+	// driverInstallation controls whether GPU drivers are installed on
+	// nodes with GPU-capable VM sizes. When set to Preferred (or not specified),
+	// GPU drivers are installed automatically. When set to None, GPU
+	// driver installation is skipped — use this when managing GPU drivers
+	// via a GPU Operator or other external mechanism.
+	// This field is ignored for non-GPU VM sizes.
+	// +default="Preferred"
+	// +optional
+	DriverInstallation *DriverInstallationMode `json:"driverInstallation,omitempty"`
```
matthchr: If I set this to Preferred ... does it mean that my Nodeclass can only schedule GPU nodes? I think you're saying _no_, it can schedule anything still. We just prefer to install the driver if we can on that node type. For non-GPU nodes, or GPU nodes where we can't install the driver... we just don't.

pkg/providers/instancetype/instancetype.go
```diff
@@ -230,11 +230,17 @@ func setRequirementsHyperVGeneration(requirements scheduling.Requirements, sku *
...
-		if vmsize.AcceleratorType != nil {
-			requirements[v1beta1.LabelSKUGPUName].Insert(*vmsize.AcceleratorType)
-		}
+	case "amd":
+		requirements[v1beta1.LabelSKUGPUManufacturer].Insert(v1beta1.ManufacturerAMD)
+	default:
+		return
+	}
+	if vmsize.AcceleratorType != nil {
+		requirements[v1beta1.LabelSKUGPUName].Insert(*vmsize.AcceleratorType)
```
matthchr, line 243: One interesting question... are we going to want/need a label on the node that indicates if it got the driver installed or not? Maybe a question for the GPU folks.

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -165,6 +165,7 @@ func (p *Provider) getStaticParameters(
 		GPUDriverVersion:               utils.GetGPUDriverVersion(instanceType.Name),
 		GPUDriverType:                  utils.GetGPUDriverType(instanceType.Name),
 		GPUImageSHA:                    utils.GetAKSGPUImageSHA(instanceType.Name),
+		GPUDriverInstallationEnabled:   nodeClass.IsGPUDriverInstallationEnabled(),
```
matthchr, line 173: I think that this bit needs to take into account the actual instance type, and the driver mode (preferred or not), and then set true | false based on those 2 things.

Other thing is maybe to check with AgentBaker folks if thye already do this -- i.e. if we ask in bootstrap for a driver... will they ignore it if it won't work?
Same q for machine: If we ask for a driver on a VM family + OS its not allowed on, do we get rejected? I would imagine yes?

pkg/apis/v1alpha2/aksnodeclass.go
```diff
@@ -95,6 +95,9 @@ type AKSNodeClassSpec struct {
 	// For more details see aka.ms/aks/localdns.
 	// +optional
 	LocalDNS *LocalDNS `json:"localDNS,omitempty"`
+	// gpu contains configuration for GPU-enabled nodes.
+	// +optional
+	GPU *GPU `json:"gpu,omitempty"`
```
matthchr, line 110: Possible breaking change issue:
What if NVidia has 7 GPUs, 6 allow driver install by AKS but 1 doesn't. Previously this was being filtered to 6 for scheduling considerations by Karpenter due to the check in instancetypes, but now we're going to remove that, which considers the 7th GPU... what is stopping that 7th GPU from showing up if the user is asking "just give me an nvidia GPU I don't care which one"

This can also happen in reverse if a new GPU SKU is published that we don't yet support driver install for, but it shows up in the skewer list and might be selected. edit: Ok no it won't because we hardcode the list of GPUs right now so we don't event detect it as a GPU SKU until we do a Karpenter rollout. BUT if compute gives us the mfg in the SKUs API then we might have this happen.

Possible outs:
1. Allow Preferred || Enabled -- in enabled mode, only match instance types we can install the driver on.
2. Make the driver-enablement a label and support scheduling on it.
3. Don't worry about this because core basically admits it's too risky to ask for something really broad like `GPUManufacturer: nvidia`

maybe others?

MaximilianoUribe:
There seems to be an inconsistency with us defining a GPUNode=true && ConfigGPUDriverIfNeeded=false in scheduling AgentBaker does not follow this path in provisioning due to the function [valideGPUDrivers](https://github.com/Azure/AgentBaker/blob/78c64576dd657af64eaffa4ea8fca93e3e5bda65/parts/linux/cloud-init/artifacts/cse_config.sh#L976) that gets called when ConfigGPUDriverIfNeeded=false is not made for this scenario. I found a workaround with making GPUNode=false && ConfigGPUDriverIfNeeded=false that does schedule the desired sku and also works in scheduling non-gpu workloads (we can't schedule gpu with no driver), seems to be OK since this does not affect the provisioning or functionality, but it does change the vm gpu annotation to false. @matthchr wondering what you think should be the approach here...

pkg/apis/v1alpha2/aksnodeclass.go
```diff
@@ -297,6 +300,36 @@ const (
...
+	// Use this when running a GPU Operator or managing drivers outside of AKS.
+	DriverInstallationNone DriverInstallationMode = "None"
+)
+
+// GPU contains configuration for GPU-enabled nodes.
+type GPU struct {
+	// driverInstallation controls whether GPU drivers are installed on
+	// nodes with GPU-capable VM sizes.
+	// When set to Install (or not specified), GPU drivers are always installed
+	// and only GPU SKUs with driver installation support are considered for
```
matthchr: Clarify which SKUs AKS supports here, maybe by linking to an aka.ms link or public doc? Or else just say "NVidia only" for now, if we don't have a doc.

pkg/apis/v1alpha2/aksnodeclass.go
```diff
@@ -297,6 +300,36 @@ const (
...
+	// scheduling.
+	// When set to None, GPU driver installation is skipped — use this when
+	// managing GPU drivers via a GPU Operator or other external mechanism.
+	// All GPU SKUs are available for scheduling in this mode.
+	// This field is ignored for non-GPU VM sizes.
+	// +default="Install"
+	// +optional
+	DriverInstallation *DriverInstallationMode `json:"driverInstallation,omitempty"`
```
matthchr: I thought we discussed/debated having this be something like `Mode: `Driver|FullyManaged|None`, to more easily extend to the upcoming managed GPU experience? 

I guess maybe what I'm asking is, what is the eventual shape of this going to be when ManagedGPU comes along, since it will before too long and I know that's what Node is pushing.

I _think_ I see two choices:
1. Mostly just mirror the AKS API:
    ```
    gpu:
      nvidia:
        mode: Driver|FullyManaged|None
    ```
    In this variant, you could always request AMD sizes and they just won't have drivers, because there is no AMD section and the default is no drivers. So what we're saying here is "NVIDIA GPUs can choose their mode, AMD has no choice", but users can still choose by size at the NodePool level.
2. Pull mode field up into `gpu` (similar to `driver` in existing AKS API but combined)
    ```
    gpu:
      mode: mode: Driver|FullyManaged|None
      nvidia:
        .. // additional fields to be added later such as MIG profile
    ```
    In this variant, selecting mode: Driver | FullyManaged at the top level would exclude you from getting AMD nodes at all (they'd be filtered out since we can't deliver AMD + Drivers atm).
3. (I think rejected in the meeting we had) Use labels for all of this. I think we discussed this in our meeting and determined that it didn't make sense to do this and that today there were no labels that really accurately represented these dimensions anyway.

I don't think the API proposed here as-is matches any of the above

matthchr: cc @tallaxes for his thoughts here. I think this is a bit of a special case because we know the upstream AKS API is changing so we want to come up with something that matches it in spirit but may not actually match it exactly in practice (because it is changing)

matthchr: May need to consider breaking-change-ness here too. Option 2 has the advantage of allowing us to retain our exact semantics today (we can't provision AMD GPUs and we won't even discover them unless user takes explicit action to change the driver setting away from the default). OTOH breaking here doesn't seem that risky (how many people are asking for both NVIDIA + AMD GPUs in a single NodePool? I can't imagine it's common or even anybody is doing it), and the resulting semantics of option 1 may be cleaner as it allows a single NodeClass to be used on pools for both AMD and NVIDIA GPUs.

pkg/fake/zz_generated.sku.southcentralus.go
```diff
@@ -983,5 +983,89 @@ func init() {
 			Locations:    &[]string{"southcentralus"},
 			LocationInfo: &[]compute.ResourceSkuLocationInfo{{Location: lo.ToPtr("southcentralus"), Zones: &[]string{}}},
 		},
+		{
```
matthchr: Did you add this manually?

I think you need to update the makefile that generates this to add this as an examined size? See hack/codegen.sh `skugen` -- and then re-run it to ensure that this gets set correctly?

matthchr: Same comment for the other file.

pkg/providers/imagefamily/bootstrap/aksbootstrap.go
```diff
@@ -300,12 +300,18 @@ func (a AKS) applyOptions(nbv *NodeBootstrapVariables) {
 	nbv.NetworkSecurityGroup = fmt.Sprintf("aks-agentpool-%s-nsg", a.ClusterID)
 	nbv.RouteTable = fmt.Sprintf("aks-agentpool-%s-routetable", a.ClusterID)
 
-	if a.GPUNode {
+	if a.GPUNode && a.GPUDriverInstallationEnabled {
 		nbv.GPUNode = true
```
matthchr, line 304: Just confirming, it's OK to set `GPUNode` to false if Driver is off?

It seems a bit counterintuative (it's still a GPU node, just no driver), but I think that's how AgentBaker has it set up too so I think is correct - just want to confirm (and may be worth a comment here)

pkg/providers/imagefamily/bootstrap/containerd.toml.gtpl
```diff
@@ -1,16 +1,16 @@
 version = 2
 oom_score = 0
 [plugins."io.containerd.grpc.v1.cri"]
-  sandbox_image = "mcr.microsoft.com/oss/kubernetes/pause:3.6" 
+  sandbox_image = "mcr.microsoft.com/oss/kubernetes/pause:3.6"
   [plugins."io.containerd.grpc.v1.cri".containerd]
-    {{- if .GPUNode }}
+    {{- if .ConfigGPUDriverIfNeeded }}
```
matthchr, line 6: Ahh wait you've changed this, so now [this question](https://github.com/Azure/karpenter-provider-azure/pull/1539/changes#r3047711104) is more important - what exactly does `GPUNode` do in AgentBaker?

matthchr, line 6: Does this match upstream/AgentBaker, or is this us diverging? Can you share a link?

pkg/providers/instance/aksmachineinstancehelpers.go
```diff
@@ -158,12 +158,15 @@ func (p *DefaultAKSMachineProvider) buildAKSMachineTemplate(ctx context.Context,
 	}, nil
 }
 
-func configureGPUProfile(instanceType *corecloudprovider.InstanceType) *armcontainerservice.GPUProfile {
+func configureGPUProfile(instanceType *corecloudprovider.InstanceType, nodeClass *v1beta1.AKSNodeClass) *armcontainerservice.GPUProfile {
 	// If none is specified, then that's not GPU instance, so nil is fine. Current version of AKS machine API supports this.
 	if utils.IsNvidiaEnabledSKU(instanceType.Name) {
```
matthchr: I'm wondering if we want this level of filtering here?

If we're filtering out sizes that we expect should support GPU driver in `instancetype.go`, then we shouldn't get to here with an invalid combination? So we could just pass-through the value from the `nodeClass` directly?

pkg/providers/instancetype/instancetype.go
wdarko1: Will sync with the Compute team on this

pkg/utils/gpu.go
```diff
@@ -45,6 +45,9 @@ type GPUSKUConfig map[string]GPUSKUInfo
 var (
 	nvidiaEnabledSKUs        = make(map[string]bool)
 	marinerNvidiaEnabledSKUs = make(map[string]bool)
+	amdEnabledSKUs           = make(map[string]bool)
```
matthchr, line 49: minor, style: Maybe for a separate PR as this one is pretty big already - we could put this into `pkg/utils/gpu/gpu.go` and then usage becomes `gpu.<thing>` which is nice. Also, avoids an uber-utils package which is generally considered to be bad Go practice.

pkg/utils/gpu.go
```diff
@@ -63,7 +66,11 @@ func readGPUSKUConfig() {
 	}
 
 	for sku, info := range gpuSKUConfig {
-		if info.GPU == "nvidia" {
+		allGPUSKUs[sku] = info.GPU
+		gpuSKUOSSupport[sku] = info.OS
+
+		switch info.GPU {
+		case "nvidia":
```
matthchr: Should these `nvidia` and `amd` strings be consts? We seem to use them in at least a few places?

pkg/utils/gpu.go
```diff
@@ -148,3 +157,52 @@ var ConvergedGPUDriverSizes = map[string]bool{
...
+func GetGPUManufacturer(vmSize string) string {
+	vmSize = normalizeVMSize(vmSize)
+	return allGPUSKUs[vmSize]
+}
+
+// HasDriverInstallationSupport returns true if the system knows how to install
+// GPU drivers for this VM SKU. Currently all NVIDIA SKUs have driver installation
+// support, while AMD SKUs do not. This is the single abstraction point for this
+// decision — when AMD driver support is added, only this function needs to change.
+func HasDriverInstallationSupport(vmSize string) bool {
```
matthchr: very minor, style, feel free to ignore: `IsDriverInstallSupported` would match your `Is...` pattern with the other methods that return bools.

pkg/utils/supported-gpus.yaml
```diff
@@ -266,26 +266,26 @@ standard_nd128isr_gb300_v6:
...
-#   gpu: amd
-#   os: ["ubuntu"]
-# standard_nv24ads_v710_v5:
-#   gpu: amd
-#   os: ["ubuntu"]
-# standard_nv28adms_v710_v5:
-#   gpu: amd
-#   os: ["ubuntu"]
+standard_nv4ads_v710_v5:
+  gpu: amd
```
matthchr, line 270: This file seems out of date? Needs rebase?

pkg/providers/instancetype/instancetype.go
matthchr: Please loop Max and I (and probably Alex) into that convo as well.

pkg/providers/instance/aksmachineinstancehelpers.go
MaximilianoUribe: Looks safe because of this, also this explains more of how we treat the GPU_NODE=false in AgentBaker that gets propagated from 
EnableNvidia=false

RP side:                                                                                                                                                                                        
  - IsNvidiaEnabledSKU() returns false for all AMD SKUs — this is the hard gate for NVIDIA driver installation in every code path                                                                   
  - GPUProfile{Driver: "None"} → InstallGPUDriver=false, but even if GPUProfile is nil, AMD SKUs get EnableNvidia=false because IsNvidiaEnabledSKU=false                                            
  - The RP validator explicitly allows AMD + InstallGPUDriver=false and explicitly rejects AMD + InstallGPUDriver=true                                  
  - Both nil and {Driver: "None"} produce identical behavior: EnableNvidia=false                                                                                                                    
                                                                                                                                                                                                    
  AgentBaker side:                                                                                                                                                                                  
  - Purely a pass-through — it does NOT independently detect GPU SKUs                                                                                                                               
  - GPU_NODE=false → no NVIDIA drivers, no fabric manager, containerd uses runc, cleanup runs                                                                                                       
  - Standard_NV4ads_V710_v5 doesn't appear anywhere in AgentBaker — it provisions as a completely non-GPU node
  - The EnableAMDGPU field exists in the protobuf but is dead code (not wired to CSE)

pkg/apis/v1alpha2/aksnodeclass.go
MaximilianoUribe: IMO, going with option 2 should be the safest best, since it would not cause any issues with current behavior and also allows us to add these nvidia fields in the future without changing the structure

pkg/utils/supported-gpus.yaml
MaximilianoUribe: where do you see that is out of date? its the same as in rp for what I can see

matthchr: I am wrong =)

pkg/apis/v1alpha2/aksnodeclass.go
matthchr: FWIW what I suggested for option 2 above still pulls the "managedGPUExperience" up into this field, which we don't currently have, and calls this field `mode` so that it can be expanded to cover that in a nonbreaking way. So I think we need to discuss this a bit more still because this doesn't match with that plan, though I am not 100% sure that plan is correct.

matthchr: Ahh, wait you did what I suggested above but the diff I was looking at wasnt' showing it!

pkg/apis/v1alpha2/aksnodeclass.go
```diff
@@ -297,6 +300,36 @@ const (
...
+	// mode controls GPU driver management on GPU-enabled nodes.
+	// When set to Driver (or not specified), GPU drivers are installed by AKS
+	// and only GPU SKUs with driver installation support are considered for scheduling,
+	// we are only supporting NVIDIA GPUs for driver installation via AKS.
+	// When set to None, GPU driver installation is skipped — use this when
+	// managing GPU drivers via a GPU Operator or other external mechanism.
+	// All GPU SKUs are available for scheduling in this mode.
+	// This field is ignored for non-GPU VM sizes.
+	// +default="Driver"
+	// +optional
```
matthchr, line 329:
```suggestion
	// mode controls GPU driver management on GPU-enabled nodes.
	// When set to Driver (or not specified), GPU drivers are installed by AKS
	// and only GPU SKUs with managed driver installation support.
	// Currently, only NVIDIA GPUs support managed driver installation.
	// When set to None, GPU driver installation is skipped — use this when
	// managing GPU drivers via a GPU Operator or other external mechanism.
	// All GPU SKUs are available for scheduling in this mode.
	// This field is ignored for non-GPU VM sizes.
	// +default="Driver"
	// +optional
```

matthchr: Applies to beta too.

pkg/providers/imagefamily/bootstrap/aksbootstrap.go
```diff
@@ -300,12 +300,18 @@ func (a AKS) applyOptions(nbv *NodeBootstrapVariables) {
...
 		nbv.ConfigGPUDriverIfNeeded = true
 		nbv.GPUDriverVersion = a.GPUDriverVersion
 		nbv.GPUDriverType = a.GPUDriverType
 		nbv.GPUImageSHA = a.GPUImageSHA
+	} else {
+		// For non-GPU nodes or GPU nodes with driverInstallation: None,
```
matthchr, line 310: minor: note about GPUNode being set to false as well, which is expected/required by AgentBaker?

pkg/providers/instancetype/instancetypes.go
```diff
@@ -328,6 +331,21 @@ func (p *DefaultProvider) isInstanceTypeSupportedByLocalDNS(sku *skewer.SKU, nod
...
+	// "None" mode allows all GPU SKUs.
+	if nodeClass.GetGPUMode() != v1beta1.GPUModeDriver {
+		return true
+	}
+	name := sku.GetName()
+	// Non-GPU SKUs are always allowed
+	if !utils.IsGPUSKU(name) {
+		return true
+	}
+	// In "Install" mode, only allow GPU SKUs with driver installation support
```
matthchr, line 345:
```suggestion
	// In "Driver" mode, only allow GPU SKUs with driver installation support
```

maybe?

test/suites/gpu/suite_test.go
```diff
@@ -119,6 +119,60 @@ var _ = Describe("GPU", func() {
...
+
+	It("should provision a GPU node with mode None",
+		Label("GPU"),
+		func() {
+			noneMode := v1beta1.GPUModeNone
+			nodeClass := env.DefaultAKSNodeClass()
+			nodeClass.Spec.GPU = &v1beta1.GPU{Mode: &noneMode}
+
+			nodePool := env.DefaultNodePool(nodeClass)
+			test.ReplaceRequirements(nodePool, karpv1.NodeSelectorRequirementWithMinValues{
```
matthchr, line 131: minor, maybe leave a comment as to why you're doing this? (I think it's to force it off D series so that we can actually get a GPU)

test/suites/gpu/suite_test.go
```diff
@@ -119,6 +119,60 @@ var _ = Describe("GPU", func() {
...
+			// Verify the GPU node is provisioned, initialized, and the
+			// workload pod becomes healthy. With mode: None,
+			// Karpenter provisions the node without installing GPU drivers.
+			// The node should still come up and be ready.
+			env.EventuallyExpectHealthyPodCountWithTimeout(
+				time.Minute*15,
+				labels.SelectorFromSet(deployment.Spec.Selector.MatchLabels),
+				int(*deployment.Spec.Replicas),
+			)
+			env.ExpectCreatedNodeCount("==", int(*deployment.Spec.Replicas))
```
matthchr, line 173: Should we be checking any labels or other properties of the node to ensure/config that it actually has no driver installed? I think the pod would still schedule even if the node _was_ advertising a GPU, right?

pkg/apis/v1alpha2/aksnodeclass.go
tallaxes: Option 2 (shared driver mode flag) seems both cleaner and more forward looking to me, at the expense of slight deviation from AKS API - which I think is acceptable

pkg/providers/imagefamily/customscriptsbootstrap/provisionclientbootstrap.go
```diff
@@ -194,7 +195,7 @@ func (p *ProvisionClientBootstrap) ConstructProvisionValues(ctx context.Context)
 	if utils.IsNvidiaEnabledSKU(p.InstanceType.Name) {
```
matthchr, line 195: minor: Do we need this `utils.IsNvidiaEnabledSKU(p.InstanceType.Name)` anymore?

Since we're controlling GPU driver installation elsewhere, we could possibly swap this to check the new `GPUDriverInstallationEnabled` instead?

It can't be true for AMD, right?

Feel free to ignore if you think it's risky/confusing.

pkg/providers/instancetype/suite_test.go
```diff
@@ -2225,6 +2225,77 @@ var _ = Describe("InstanceType Provider", func() {
...
+					ExpectApplied(ctx, env.Client, nodeClassInstall)
+					instanceTypes, err = azureEnv.InstanceTypesProvider.List(ctx, nodeClassInstall)
+					Expect(err).ToNot(HaveOccurred())
+				})
+
+				It("should include NVIDIA GPU SKUs", func() {
+					Expect(instanceTypes).Should(ContainElement(WithTransform(getName, Equal("Standard_NC16as_T4_v3"))))
+				})
+				// Standard_NV4ads_V710_v5 is not in the fake SKU data for southcentralus
+				PIt("should not include AMD GPU SKUs", func() {
```
matthchr, line 2268: Remove `PIt` tests? Or fix?

Or file an item tracking improvement?

Not 100% sure what the intent with these is.

matthchr (approved):

PR #1546, opened 2026-03-19
feat: add allocationstrategy provider

matthchr:
**Description**
add allocationstrategy provider

This gives us a place we can expand in the future to do more complex instance type ordering/filtering

* Unit test
* E2E: https://github.com/Azure/karpenter-provider-azure/actions/runs/23306098564

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/.github/actions/install-deps/action.yaml
+++ b/.github/actions/install-deps/action.yaml
@@ -29,14 +29,16 @@ runs:
       env:
         K8S_VERSION: ${{ inputs.k8sVersion }}
       run: make toolchain
-    # Install trivy separately since apt installs to /usr/bin which isn't cacheable
+    # Install trivy v0.69.3 with SHA256 verification (pinned binary, no apt repo trust)
     - shell: bash
       run: |
         if ! command -v trivy &> /dev/null; then
-          wget -qO - https://aquasecurity.github.io/trivy-repo/deb/public.key | gpg --dearmor | sudo tee /usr/share/keyrings/trivy.gpg > /dev/null
-          echo "deb [signed-by=/usr/share/keyrings/trivy.gpg] https://aquasecurity.github.io/trivy-repo/deb generic main" | sudo tee -a /etc/apt/sources.list.d/trivy.list
-          sudo apt-get update
-          sudo apt-get install -y trivy
+          TRIVY_VERSION="0.69.3"
+          TRIVY_SHA256="a484057aafde31089cf2558ca0f79a4bc835125a5ee6834183a5bcf0735af358"
+          wget -qO /tmp/trivy.deb "https://github.com/aquasecurity/trivy/releases/download/v${TRIVY_VERSION}/trivy_${TRIVY_VERSION}_Linux-64bit.deb"
+          echo "${TRIVY_SHA256}  /tmp/trivy.deb" | sha256sum --check --strict
+          sudo dpkg -i /tmp/trivy.deb
+          rm /tmp/trivy.deb
         fi
     # Install CGO cross-compilers for multi-arch builds (apt installs to /usr which isn't cacheable)
     - shell: bash
--- a/hack/toolchain.sh
+++ b/hack/toolchain.sh
@@ -118,10 +118,12 @@ kubebuilder() {
 
 gettrivy() {
     if ! command -v trivy &> /dev/null; then
-        wget -qO - https://aquasecurity.github.io/trivy-repo/deb/public.key | gpg --dearmor | sudo tee /usr/share/keyrings/trivy.gpg > /dev/null
-        echo "deb [signed-by=/usr/share/keyrings/trivy.gpg] https://aquasecurity.github.io/trivy-repo/deb generic main" | sudo tee -a /etc/apt/sources.list.d/trivy.list
-        sudo apt-get update
-        sudo apt-get install -y trivy
+        TRIVY_VERSION="0.69.3"
+        TRIVY_SHA256="a484057aafde31089cf2558ca0f79a4bc835125a5ee6834183a5bcf0735af358"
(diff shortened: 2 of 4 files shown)


pkg/providers/instance/aksmachineinstance.go
```diff
@@ -416,6 +423,7 @@ func (p *DefaultAKSMachineProvider) beginCreateMachine(
 	nodeClass *v1beta1.AKSNodeClass,
 	nodeClaim *karpv1.NodeClaim,
 	instanceTypes []*corecloudprovider.InstanceType,
```
comtalyst, line 421: Of, is there a good opportunity to stop passing `instanceTypes` now that we have higher-level `instanceOfferings`?

pkg/providers/instance/aksmachineinstance.go
```diff
@@ -181,9 +185,12 @@ func (p *DefaultAKSMachineProvider) BeginCreate(
 	if err != nil {
 		return nil, fmt.Errorf("failed to generate AKS machine name from NodeClaim name %q: %w", nodeClaim.Name, err)
 	}
-	instanceTypes = offerings.OrderInstanceTypesByPrice(instanceTypes, scheduling.NewNodeSelectorRequirementsWithMinValues(nodeClaim.Spec.Requirements...))
+	instanceOfferings := p.allocationStrategyProvider.FilterInstanceOfferings(
+		allocationstrategy.NewInstanceOfferings(instanceTypes),
+		scheduling.NewNodeSelectorRequirementsWithMinValues(nodeClaim.Spec.Requirements...),
+	)
```
comtalyst: Might be cleaner to do this inside `beginCreateMachine(...)` to avoid passing multiple overlapping(?) things. Thoughts?

pkg/providers/instance/vminstance.go
```diff
@@ -221,8 +225,11 @@ func (p *DefaultVMProvider) BeginCreate(
 	nodeClaim *karpv1.NodeClaim,
 	instanceTypes []*corecloudprovider.InstanceType,
 ) (*VirtualMachinePromise, error) {
-	instanceTypes = offerings.OrderInstanceTypesByPrice(instanceTypes, scheduling.NewNodeSelectorRequirementsWithMinValues(nodeClaim.Spec.Requirements...))
-	vmPromise, err := p.beginLaunchInstance(ctx, nodeClass, nodeClaim, instanceTypes)
+	instanceOfferings := p.allocationStrategyProvider.FilterInstanceOfferings(
+		allocationstrategy.NewInstanceOfferings(instanceTypes),
+		scheduling.NewNodeSelectorRequirementsWithMinValues(nodeClaim.Spec.Requirements...),
+	)
+	vmPromise, err := p.beginLaunchInstance(ctx, nodeClass, nodeClaim, instanceOfferings)
```
comtalyst: (same comments with machine variant)

pkg/providers/instance/aksmachineinstance.go
comtalyst: (Or, initailize this in `PickSkuSizePriorityAndZone`? Haven't think through yet, may continue later)

matthchr: Is your thinking here because that would avoid us needing to pass `instanceOfferings` alongside instanceTypes to beginCreateMachine? Basically you just want one to be passed? Agree that would be better - in vmInstance doing it here was fine because we didn't need instanceTypes then, but we still do for Machine (and it's not that easy/clean to get rid of it, I checked).

I'll move it in a level for both machine and VM instance.

pkg/providers/instance/aksmachineinstance.go
matthchr: Somewhat difficult to do because of shared methods in machine path that use instanceTypes - not all calls are from this path, but only this path does the `InstanceOfferings` thing (because only this path needs the filtered + ordered set)

pkg/providers/instance/aksmachineinstance.go
matthchr: can't put it in `PickSkuSizePriorityAndZone` without promoting that method to either take the `allocationStrategyProvider` or making it a method here on VMProvider (less re-use because then we'd have to duplicate it for Machine).

I think setting this up ahead of time like we were doing before, so that the structure is:
1. Build prioritized list
2. Pick an item from that list

That way in the future if we move to something like fleet API, we can still use our prioritized list ordering, but we won't be picking the 0th item only anymore we'll be picking the top 20 or something.

pkg/providers/instance/vminstance.go
matthchr: resolving this -- did as suggested, you can take a look and confirm your thoughts on it.

pkg/providers/instance/aksmachineinstance.go
```diff
@@ -441,8 +444,13 @@ func (p *DefaultAKSMachineProvider) beginCreateMachine(
 		log.FromContext(ctx).Error(err, "failed to check for existing AKS machine", "aksMachineName", aksMachineName)
 	}
 
+	instanceOfferings := p.allocationStrategyProvider.FilterInstanceOfferings(
+		allocationstrategy.NewInstanceOfferings(instanceTypes),
+		scheduling.NewNodeSelectorRequirementsWithMinValues(nodeClaim.Spec.Requirements...),
+	)
+
 	// Decide on offerings
```
comtalyst, line 452: nit: the added block above could still be under this "Decide on offerings" umbrella, not blocking

matthchr: Will fix in a followon PR

PR #1551, opened 2026-03-20
Improve unavilableofferings cache

matthchr:
**What this change does**
- Mark the whole family with >= vCPUs as unavailable.
- Updated the "low but not out" of quota case to have 10m TTL rather than 3m.
- Adds tests ensuring that we have the correct behavior when confronted with unavailable offering situations

* Unit tests

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/providers/instance/offerings/commonerrorhandlers.go
+++ b/pkg/providers/instance/offerings/commonerrorhandlers.go
@@ -36,11 +36,32 @@ var (
 	OverconstrainedAllocationFailureReason      = "OverconstrainedAllocationFailure"
 	SKUNotAvailableReason                       = "SKUNotAvailable"
 
-	LowQuotaTTL                 = 10 * time.Minute
+	// LowQuotaTTL is the TTL for offerings that return a quota error but the quota limit is not 0.
+	// This means there is still some quota available, just not enough to fulfill the current request.
+	// We set this to some value "reasonably lower" than SubscriptionQuotaReachedTTL.
+	// It needs to be long enough that we can attempt multiple sizes in the same family in descending order and still have the entry
+	// in the cache when we get to the smaller sizes, to allow failover to another family.
+	// For example, if we're trying to fit 64 cores worth of pods and we only have quota for 10 cores of Dv2:
+	// D64_v2  -> D32_v2 -> D16_v2 all fail before one D8_v2 finally succeeds, but we need more cores than that so we proceed on to D4_v2 and D2_v2 (one of which succeeds).
+	// This effectively utilizes all the quota, but requires that we don't have the D64 cache entry expire before we've tried all the sizes in the family down to the one that works,
+	// otherwise we might allocate a D8_v2 and then go back and try D64_v2 again - in the worst case we can get caught in a "loop" doing this and never actually proceed to a size
+	// that works.
+	// TODO: If/when we factor in actual quota API usage we may be able to reduce this TTL and use the quota API data to inform which sizes we try first.
+	LowQuotaTTL = 10 * time.Minute
+	// SubscriptionQuotaReachedTTL is the TTL for offerings that return a quota error with a limit of 0, meaning there is no quota available for that SKU family at all in the subscription.
+	// This is often the case if the user doesn't have any quota for that offering at all, hence the longer TTL.
+	// TODO: If/when we factor in actual quota API usage in the future we may be able to get rid of this longer TTL and just rely on LowQuotaTTL.
 	SubscriptionQuotaReachedTTL = 1 * time.Hour
-	AllocationFailureTTL        = 1 * time.Hour
-	SKUNotAvailableSpotTTL      = 1 * time.Hour
-	SKUNotAvailableOnDemandTTL  = 23 * time.Hour
+	// AllocationFailureTTL is the TTL for offerings that returned an allocation failure from Azure. AllocationFailure usually means that there is a capacity
+	// crunch of some kind (for that zone, for that offering, etc). It's unlikely that the capacity crunch resolves itself in a very short amount of time, hence the longer TTL.
+	AllocationFailureTTL = 1 * time.Hour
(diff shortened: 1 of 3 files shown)


pkg/fake/types.go
```diff
@@ -29,8 +30,25 @@ type MockedFunction[I any, O any] struct {
 	CalledWithInput AtomicPtrStack[I] // Stack used to keep track of passed input to this function
 	Error           AtomicError       // Error to return a certain number of times defined by custom error options
 
-	successfulCalls atomic.Int32 // Internal construct to keep track of the number of times this function has successfully been called
-	failedCalls     atomic.Int32 // Internal construct to keep track of the number of times this function has failed (with error)
+	successfulCalls     atomic.Int32 // Internal construct to keep track of the number of times this function has successfully been called
+	failedCalls         atomic.Int32 // Internal construct to keep track of the number of times this function has failed (with error)
+	customTransformerMu sync.RWMutex
+	customTransformer   func(*I) error // Optional hook called before the default transformer; can modify input or return an error to short-circuit
```
matthchr, line 36: I _assume_ that this is fine and we're OK with it, but want to check with @tallaxes

pkg/cache/unavailableofferings.go
```diff
@@ -115,10 +115,36 @@ func (u *UnavailableOfferings) isFamilyUnavailable(sku *skewer.SKU, zone, capaci
...
+
+// markFamilyUnavailableAtCPUCountImpl is the internal implementation that marks a VM family unavailable at a given CPU count threshold.
 // Value of -1 is used as a "wholeVMFamilyBlockedSentinel" to indicate that the entire VM family is blocked in this zone for the specified capacity type.
-// skuFamilyName e.g. "StandardDv2Family" for "Standard_D2_v2" VM SKU
-func (u *UnavailableOfferings) MarkFamilyUnavailableAtCPUCount(ctx context.Context, skuFamilyName, zone, capacityType string, cpuCount int64, ttl time.Duration) {
+func (u *UnavailableOfferings) markFamilyUnavailableAtCPUCountImpl(ctx context.Context, sku *skewer.SKU, zone, capacityType string, cpuCount int64, ttl time.Duration) {
+	skuFamilyName := sku.GetFamilyName()
+	// This is a hedge against skewer having bad data where family name is missing,
+	// If family name is missing, we won't do any family level blocking, but we'll still mark the specific offering as unavailable.
+	if skuFamilyName == "" {
```
matthchr, line 141: This data comes directly from the compute API, so I am not actually sure we need this defensiveness.

WDYT? Should I remove it?

pkg/test/expectations/expectations.go
```diff
@@ -208,6 +211,43 @@ func ExpectProvisionedAndWaitForPromises(
...
+	}
+	return nodeClaims
+}
+
+func BindPodsToNode(ctx context.Context, client client.Client, cluster *state.Cluster, scheduledClaim *scheduling.NodeClaim, node *corev1.Node) {
+	GinkgoHelper()
+
+	for _, pod := range scheduledClaim.Pods {
+		// We have to manually bind the pod to the node when using a fakeClient by setting the value for pod.Spec.NodeName
+		if strings.Contains(reflect.TypeOf(client).String(), "fake") {
```
tallaxes, line 240: nit: feels a little fragile ... is this the best/only way?

pkg/cache/unavailableofferings.go
tallaxes: I'd keep it, does not harm anything

pkg/fake/types.go
tallaxes: LGTM, and useful!

pkg/providers/instance/offerings/commonerrorhandlers.go
```diff
@@ -36,6 +36,7 @@ var (
 	OverconstrainedAllocationFailureReason      = "OverconstrainedAllocationFailure"
 	SKUNotAvailableReason                       = "SKUNotAvailable"
 
+	LowQuotaTTL                 = 10 * time.Minute
```
tallaxes: Curious how TTL was chosen ... (And let's maybe call out the change from 3 to 10 min in PR description?)

pkg/test/expectations/expectations.go
matthchr: It's what upstream does too -- this is effectively copied from their helper functions.

matthchr: I'll leave a note here to that effect.

pkg/providers/instance/offerings/commonerrorhandlers.go
matthchr: I'll update PR description and leave a comment here in the code.

matthchr: Ok here's the comment I wrote:
```
	// LowQuotaTTL is the TTL for offerings that return a quota error but the quota limit is not 0.
	// This means there is still some quota available, just not enough to fulfill the current request.
	// We set this to some value "reasonably lower" than SubscriptionQuotaReachedTTL.
	// It needs to be long enough that we can attempt multiple sizes in the same family in descending order and still have the entry
	// in the cache when we get to the smaller sizes, to allow failover to another family.
	// For example, if we're trying to fit 64 cores worth of podsand we only have quota for 10 cores of Dv2:
	// D64_v2  -> D32_v2 -> D16_v2 all fail before one D8_v2 finally succeeds, but we need more cores than that so we proceed on to D4_v2 and D2_v2 (one of which succeeds).
	// This effectively utilizes all the quota, but requires that we don't have the D64 cache entry expire before we've tried all the sizes in the family down to the one that works,
	// otherwise we might allocate a D8_v2 and then go back and try D64_v2 again - in the worst case we can get caught in a "loop" doing this and never actually proceed to a size
	// that works.
	// TODO: If/when we factor in actual quota API usage we may be able to reduce this TTL and use the quota API data to inform which sizes we try first.
	LowQuotaTTL = 10 * time.Minute
```

PR #1552, opened 2026-03-20
fix: pin trivy

tallaxes:
Fixes # 

**Description**

*

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/cloudprovider/cloudprovider.go
+++ b/pkg/cloudprovider/cloudprovider.go
@@ -544,6 +544,11 @@ func (c *CloudProvider) RepairPolicies() []cloudprovider.RepairPolicy {
 			ConditionStatus:    corev1.ConditionUnknown,
 			TolerationDuration: 10 * time.Minute,
 		},
+		{
+			ConditionType:      "kubernetes.azure.com/NodeHealthy",
+			ConditionStatus:    corev1.ConditionFalse,
+			TolerationDuration: 0,
+		},
 	}
 }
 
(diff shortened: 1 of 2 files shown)


.github/actions/install-deps/action.yaml
```diff
@@ -29,15 +29,9 @@ runs:
...
-    # Install trivy separately since apt installs to /usr/bin which isn't cacheable
-    - shell: bash
-      run: |
-        if ! command -v trivy &> /dev/null; then
-          wget -qO - https://aquasecurity.github.io/trivy-repo/deb/public.key | gpg --dearmor | sudo tee /usr/share/keyrings/trivy.gpg > /dev/null
-          echo "deb [signed-by=/usr/share/keyrings/trivy.gpg] https://aquasecurity.github.io/trivy-repo/deb generic main" | sudo tee -a /etc/apt/sources.list.d/trivy.list
-          sudo apt-get update
-          sudo apt-get install -y trivy
-        fi
+    - uses: aquasecurity/setup-trivy@3fb12ec12f41e471780db15c232d5dd185dcb514 # v0.2.6
```
matthchr: Might be good to confirm which is safer - pinned setup-trivy or pinned trivy-repo?

tallaxes: Probably about the same, though without action there is one less link in the chain; changing. (Dependabot won't be bumping it without action, but this might be for the best here.)

PR #1556, opened 2026-03-21
feat: add instanceTypes override for pinned GPU and custom SKU support

comtalyst:
**Description**

Add `instanceTypes` field to `AKSNodeClass` (v1beta1 and v1alpha2) and enhance the instance type provider to support pinning specific VM SKUs that may be filtered out by default compatibility checks or missing from pricing APIs.

This is needed for scenarios where users deploy to pinned GPU capacity (e.g. H200 SKUs) that:
- May not appear in standard Azure retail pricing APIs
- May be filtered by the default image compatibility check (NVMe-only SKUs)
- May have non-standard naming that `GetVMSize()` can't parse

### Changes

**API fields:**
- `instanceTypes []string` on `AKSNodeClass` spec — when set, only these instance types are considered for provisioning

**Instance type provider enhancements:**
- Cache the raw unfiltered SKU data (`rawSKUCache`) for fallback lookups
- Augment the filtered SKU set with explicitly requested types from the raw cache
- Force-create offerings with zero price when pricing data is missing for requested types
- Construct fallback `VMSizeType` for SKUs with non-standard naming conventions
- Skip image compatibility check in non-AKS mode (`azurevm` provision mode uses custom images)
- Filter results to only requested types when the override is set
- Include instance types filter in cache key to avoid stale results

Build verification (`go build ./pkg/...`). Hash version tests updated.

- [ ] Yes, issue opened: #
- [X] No

**Release Note**

```release-note
Add instanceTypes field to AKSNodeClass for pinning specific VM SKUs
```

### DIFF
--- a/charts/karpenter-crd/templates/karpenter.azure.com_azurenodeclasses.yaml
+++ b/charts/karpenter-crd/templates/karpenter.azure.com_azurenodeclasses.yaml
@@ -61,6 +61,15 @@ spec:
               spec is the top level specification for the Azure Karpenter Provider.
               This will contain configuration necessary to launch instances in Azure.
             properties:
+              dataDiskSizeGB:
+                description: |-
+                  dataDiskSizeGB is the size of an additional data disk in GB to attach for
+                  container runtime storage (e.g., containerd root). A Premium_LRS managed disk
+                  is attached at LUN 0. The image's bootstrap must handle mounting this disk.
+                format: int32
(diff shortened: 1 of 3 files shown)


comtalyst:
(AI-generated) Superseded by the v4 rebase. Replacement PR created on top of `comtalyst/azurenodeclass-v4` (PR #1579).

PR #1575, opened 2026-03-23
chore: 1es hosted pool

MaximilianoUribe:
Fixes # 

**Description**
Implemented self-hosted runner with 1ES hosted pools to avoid issues with github runners timing out in some of our E2E runs.

E2E tests
* https://github.com/Azure/karpenter-provider-azure/actions/runs/23554556532 (passed ACR test)
* https://github.com/Azure/karpenter-provider-azure/actions/runs/23556676364 (failed integration test, due to a flake?)
* https://github.com/Azure/karpenter-provider-azure/actions/runs/23617637819 (passed GPU test)

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


MaximilianoUribe:
I have my doubts about having the deploy scripts in here... went with approach used in ASO: https://github.com/Azure/azure-service-operator/pull/3044/changes#diff-ce0cf8d13aab17545ac63a07585756218971941a0e4ecfbc29c6c49b8f29bbf7

So imo it should be ok and good if we want to make changes in the future.

.github/workflows/ci-test.yml
```diff
@@ -11,7 +11,8 @@ jobs:
     permissions:
       contents: read
       statuses: write
-    runs-on: ubuntu-latest
+    runs-on:
+      labels: [self-hosted, 1ES.Pool=karpenter-ci-1es-pool]
```
theunrepentantgeek: Based on the onboarding docs for 1ES pools, you also need to specify a `JobID` parameter thusly:

```
runs-on: [ self-hosted, 1ES.Pool=my-github-runner-pool, "JobId=<uniquejobprefix>-${{ github.run_id }}-${{ github.run_number }}-${{ github.run_attempt }}" ]
```

Failure to do this contributed to recent widespread flakiness of 1ES pools.

Needs to be applied to all relevant workflows.

hack/1es-pool/deploy.bicep
```diff
@@ -0,0 +1,86 @@
...
+var poolSettings = {
+  maxPoolSize: 25 // 25 × 4 = 100 cores (DDSv5 family)
+  resourcePredictions: [
+    {}              // Sunday: no agents
+    {
+      '17:00': 12   // 9 AM Monday PST
+    }
+    {
+      '01:00': 0   // 5 PM Monday PST
+      '17:00': 12   // 9 AM Tuesday PST
```
theunrepentantgeek: These predictions both assume everyone is in the same timezone, and that daylight savings has no effect.

As a starting point, consider the [timings used by Azure Service Operator](https://github.com/Azure/azure-service-operator/blob/main/scripts/v2/cipool/deploy.bicep)

.github/actions/e2e/install-karpenter/action.yaml
```diff
@@ -56,5 +56,9 @@ runs:
       AZURE_LOCATION: ${{ inputs.location }}
       PROVISION_MODE: ${{ inputs.provision_mode }}
       AZURE_ACR_NAME: ${{ inputs.acr_name }}
+      GOTMPDIR: /mnt/go-tmp
```
matthchr, line 61: minor: why are these needed? Worth a comment probably?

.github/workflows/ci-test.yml
```diff
@@ -11,7 +11,8 @@ jobs:
     permissions:
       contents: read
       statuses: write
-    runs-on: ubuntu-latest
+    runs-on:
```
matthchr: What about ci.yaml? I guess it's fine as-is because it's very short/fast?

matthchr: Though strictly speaking we could leave ci-test on the GH runners too, if we wanted to? They don't take very long so don't have as much need to move?

.github/workflows/e2e.yaml
```diff
@@ -177,6 +178,10 @@ jobs:
         run: |
           curl -Lo skaffold https://github.com/GoogleContainerTools/skaffold/releases/download/v2.14.2/skaffold-linux-amd64 --fail
           install skaffold /usr/local/bin/
+      - name: generate SSH key
```
matthchr: Why is this needed? SSH key for what?

.github/actionlint.yaml
```diff
@@ -0,0 +1,3 @@
+self-hosted-runner:
```
matthchr, line 1: Is this required to inform actionlint that the self-hosted runner exists?

hack/1es-pool/deploy.bicep
```diff
@@ -0,0 +1,100 @@
...
+//   - az login
+//   - AZURE_SUBSCRIPTION_ID env var set
+
+param location string = 'westus3'
+
+var poolName = 'karpenter-ci-1es-pool'
+var sku = 'Standard_D4ds_v5'
+
+// Base 1ES Image Resource IDs
+var ubuntu2204GalleryVersionResourceId = '/subscriptions/723b64f0-884d-4994-b6de-8960d049cb7e/resourceGroups/CloudTestImages/providers/Microsoft.Compute/galleries/CloudTestGallery/images/MMSUbuntu22.04-Secure/versions/latest'
```
matthchr: ASO has 2 of these because we just moved, but you don't need both I don't think. Just pick 2404 for us.

hack/1es-pool/deploy.bicep
```diff
@@ -0,0 +1,100 @@
...
+var ubuntu2404GalleryVersionResourceId = '/subscriptions/723b64f0-884d-4994-b6de-8960d049cb7e/resourceGroups/CloudTestImages/providers/Microsoft.Compute/galleries/CloudTestGallery/images/MMSUbuntu24.04-Secure/versions/latest'
+
+var poolSettings = {
+  maxPoolSize: 25 // 25 × 4 = 100 cores (DDSv5 family)
+  resourcePredictions: [
+    {
+      '21:00': 12  // 9 AM Monday NZT
```
matthchr, line 26: I don't object to this set of predictions, but may be worth commenting why 12? In ASO we have a specific reason we picked 12.

hack/1es-pool/deploy.bicep
```diff
@@ -0,0 +1,100 @@
...
+      '05:00': 0   // 5 PM Friday NZT
+      '16:00': 12  // 9 AM Friday PST
+    }
+    {
+      '00:00': 0   // 5 PM Friday PST
+    }
+  ]
+}
+
+resource agentImage2204 'Microsoft.CloudTest/images@2020-05-07' = {
```
matthchr: minor: Drop this and delete the image via the portal too (I think dropping here won't delete it)

hack/1es-pool/deploy.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+
+set -euo pipefail # fail on...failures
+set -x # log commands as they run
+
+# Note: You must run az login before executing this script
+
+[[ -z "${AZURE_SUBSCRIPTION_ID}" ]] && echo "AZURE_SUBSCRIPTION_ID is not set" && exit 1
+
+ROOT=$(dirname "${BASH_SOURCE[0]}")
```
matthchr, line 10: So I think we're missing one thing from ASO which mitigates the "github token" timeout thing.

It _probably_ makes sense to take this PR as-is, but the `identity.bicep` in ASO assigns access to the subscription to a UMI which is assigned to the pool.

I _think_ we want to do the same thing here, so that the E2E tests can then use that identity whose token doesn't ever expire and allows tests to run longer than 1h.

matthchr: Are you also working on that, or maybe that's a second follow-up workitem?

.github/actions/e2e/install-karpenter/action.yaml
MaximilianoUribe: yes, for some reason our e2es were running out of memory capacity, but redirecting the load to this path we have more capacity

.github/workflows/ci-test.yml
MaximilianoUribe: yes, I think the tradeoff of including them would be that they would take the space of other jobs, and since they are very short I would prefer to leave them in gh runners

hack/1es-pool/deploy.bicep
MaximilianoUribe: Will leave the comment

hack/1es-pool/deploy.sh
MaximilianoUribe: I can do as a followup work item

.github/workflows/e2e.yaml
MaximilianoUribe: step guarantees that ~/.ssh/id_rsa and ~/.ssh/id_rsa.pub exist with correct permissions before any az aks create call

MaximilianoUribe: but now that I think about it, maybe not needed, since we would have a clean state after every job? theoretically

theunrepentantgeek: It's an option on the 1ES pool to reuse or repave nodes after use:
image

It's considered good practice to get a fresh node every time.

I suspect you don't need to generate an SSH key up front, as `az aks create` will do that automatically.

.github/workflows/e2e.yaml
```diff
@@ -101,7 +89,8 @@ env:
 jobs:
   run-suite:
     name: suite-${{ inputs.suite }}
-    runs-on: ubuntu-latest
+    runs-on:
+      labels: [self-hosted, 1ES.Pool=karpenter-ci-1es-pool, "JobId=e2e-${{ github.run_id }}-${{ github.run_number }}-${{ github.run_attempt }}"]
```
matthchr: I think this needs this:
> Labelling Job Ids like you suggested: "JobId=<uniquejobprefix>-${{ github.run_id }}-${{ github.run_number }}-${{ github.run_attempt }}"
finally seems to have mitigated the issue, BUT one must be careful when running pipelines that use "matrix".
> With matrix, one pipeline run will spawn multiple jobs that will have the same JobId, if you are not careful:

```
jobs:
  example:
    runs-on: [ self-hosted, 1ES.Pool=my-github-runner-pool, "JobId=example-${{ github.run_id }}-${{ github.run_number }}-${{ github.run_attempt }}" ]
    strategy:
      matrix:
        version: [12, 14, 16]
```

vs 

```
runs-on: [ self-hosted, 1ES.Pool=my-github-runner-pool, "JobId=example_v${{ matrix.version }}-${{ github.run_id }}-${{ github.run_number }}-${{ github.run_attempt }}" ]
```

.github/actions/e2e/install-karpenter/action.yaml
matthchr: Memory capacity or disk capacity?

Can you add a comment here linking to the documentation or other details about why `/mnt` solves that?

hack/1es-pool/deploy.bicep
```diff
@@ -0,0 +1,92 @@
...
+    imageType: 'SharedImageGallery'
+    resourceId: ubuntu2404GalleryVersionResourceId
+  }
+}
+
+resource hostedPool 'Microsoft.CloudTest/hostedpools@2020-05-07' = {
+  name: poolName
+  location: location
+  tags: {
+    UseScaleSet: 'true'
```
matthchr, line 79: minor: worth a comment that this forces us to the new pool allocation mode that's still somewhat experimental but avoids needing the job-id on each job?

matthchr (dismissed):

hack/1es-pool/deploy.bicep
MaximilianoUribe: yes, but more than experimental they mentioned that this will be the way forward to avoid the conflicts with the github infra changes

.github/workflows/e2e-matrix.yaml
```diff
@@ -66,5 +63,4 @@ jobs:
       azure_vm_size: ${{ inputs.azure_vm_size }}
     secrets:
       E2E_CLIENT_ID: ${{ secrets.E2E_CLIENT_ID }}
```
matthchr, line 69: Did you change this secret in GitHub? Or do we not need it anymore?
I sorta think TenantID/ClientID go hand in hand, so trying to understand why we can get rid of one but not the other.

.github/workflows/e2e.yaml
```diff
@@ -25,10 +25,12 @@ on:
     secrets:
       E2E_CLIENT_ID:
         required: true
-      E2E_TENANT_ID:
-        required: true
       E2E_SUBSCRIPTION_ID:
         required: true
+      E2E_SUBSCRIPTION_ID_TEST:
```
matthchr, line 33: Confused what this is about.

.github/workflows/e2e.yaml
```diff
@@ -163,16 +166,11 @@ jobs:
           git_ref: ${{ inputs.git_ref }}
       - uses: ./.github/actions/install-deps
       - name: az login
-        uses: azure/login@a457da9ea143d694b1b9c7c869ebb04ebe844ef5 # v2.3.0
-        with:
-          client-id: ${{ secrets.E2E_CLIENT_ID }}
-          tenant-id: ${{ secrets.E2E_TENANT_ID }}
-          subscription-id: ${{ secrets.E2E_SUBSCRIPTION_ID }}
+        shell: bash
+        run: az login --identity --client-id "$AZURE_CLIENT_ID"
```
matthchr, line 173: Can we not use the `azure/login` action anymore? Why not?

.github/workflows/e2e.yaml
matthchr: Needed for migration, is that the issue? If so, can you add a comment here and we plan on eventually unifying back onto E2E_CLIENT_ID, etc?

MaximilianoUribe: yes, need these for migration, once this lands they will be unified

.github/workflows/e2e.yaml
MaximilianoUribe: this is the simpler method to authenticate UMIs directly on the vms, we could use azure/login but this is cleaner IMO

.github/workflows/e2e.yaml
matthchr: Maybe leave a comment to that effect here?

.github/actions/e2e/create-cluster/action.yaml
```diff
@@ -45,12 +39,6 @@ runs:
     - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
       with:
         ref: ${{ inputs.git_ref }}
-    - name: az login
```
matthchr, line 48: Feel like we're being inconsistent amongst these jobs?

Some you've removed login entirely, and they don't need clientid anymore, whereas some still do login and pass clientid.

For this action to work, I would assume we do need to login - maybe it passes in current usage because something before this logs in already (so we don't _technically_ need to log in again?)

Is your thinking that that makes sense as a pre-requisite for these actions, so we don't need to pass + repeat login here?

If that's what you're going for though (expecting setup to have already happened), then maybe we should remove the `az set sub` section too? IMO it's weird to have one but not the other...

I think either we:
1. login + set sub in this action, to make it fully generic and works anywhere (even in contexts that haven't yet logged in).
2. Don't login/set sub, and add a comment that explains that for this action to work the caller must already have logged in and set the sub.

wdyt?

matthchr (approved):

PR #1369, opened 2026-01-23
feat: Poll some long running operations faster

theunrepentantgeek:
## Description

Changes polling of two long running operations to use 3s intervals instead of 30s (the default). Replaces #1222 

**How was this change tested?**

Local testing passes, E2E pending.

**Does this change impact docs?**

- [ ] Yes, PR includes docs updates
- [ ] Yes, issue opened: # 
- [x] No

## Release Note

```release-note
NONE
```

PR #1584, opened 2026-03-24
feat: Standardize all Azure long-running-operation pollers on the same 3s polling cycle

theunrepentantgeek:
## Description

Building on prior work in #1369 this sets the polling speed of _all_ Azure pollers to a 3s cycle, using a common factory method (allowing them all to be easily updated in future).

Also checked into the source for the poller (see `azure-sdk-for-go/sdk/azcore@v1.21.0/runtime/poller.go`) and verified that it correctly checks for, and responds to, standard `Retry-After` headers.

Local execution of `make test`. Also relying on CI.

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
Reduction of the polling interval for long-running operations should result in fewer delays between sequential operations.
```


pkg/providers/instance/poller.go
```diff
@@ -0,0 +1,32 @@
...
+
+package instance
+
+import (
+	"time"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
+)
+
+// createDefaultPollerOptions returns a PollUntilDoneOptions with a default polling frequency of 3 seconds.
```
tallaxes:
```suggestion
// defaultPollerOptions returns a PollUntilDoneOptions with a default polling frequency of 3 seconds.
```

matthchr:
@tallaxes asked:
> Do we actually get Retry-After in VM creation, and what kind of values? Without knowing this we don't know if this change improves things ...

We definitely do. Here's VM provisioning [logs from ASO HTTP recording](https://github.com/Azure/azure-service-operator/blob/main/v2/internal/controllers/recordings/Test_Compute_VM_20220301_CRUD.yaml)

At least based on when that was recorded (a while ago), PUT NIC has no retry-after header, but PUT VirtualMachines has:
- Retry-After: 10 on the PUT itself
- Retry-After: 35 on the first GET to the async operation URL
- No retry-after on the 2nd poll of the async operation URL (which is when it succeeds)

It's possible that we've elided some additional internal polls, so we could experiment w/ it ourselves too live, but my guess is they are setting ~35s retry-after on async-operation polling.

tallaxes:
> It's possible that we've elided some additional internal polls, so we could experiment w/ it ourselves too live, but my guess is they are setting ~35s retry-after on async-operation polling.

This would mean reduced `Frequency` value in options does not help. Beyond collecting some more data, and given that reducing this latency is an important target, our options are (likely the same for Machine API backend):
1) See if CRP can/should adjust Retry-After (improves E2E latency for everybody, at the expense of more frequent calls for all)
2) Do custom polling ourselves (at the expense of more frequent calls in Karpenter/NAP)

(There may be an "adjust SDK behavior" option too?)

pkg/providers/instance/poller.go
theunrepentantgeek: Fixed.

PR #1589, opened 2026-03-25
fix: clean up leftover E2E resources (if any)

tallaxes:
**Description**

Noticed one can't rerun E2E test if it did not cleanup the cluster (which happens if job dies in the middle); this should help.

*

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/providers/instancetype/skus.go
+++ b/pkg/providers/instancetype/skus.go
@@ -70,7 +70,10 @@ var (
 	)
 	// karpenterRestrictedVMSKUs are VMS SKUs that are known to
 	// be problematic with karpenter-provider-azure.
-	karpenterRestrictedVMSKUs = sets.New[string]()
+	karpenterRestrictedVMSKUs = sets.New[string](
+		"Standard_E64i_v3",
+		"Standard_E64is_v3",
+	)
 )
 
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e.yaml
```diff
@@ -179,6 +179,8 @@ jobs:
           ACR_NAME=$(echo "$SUITE"acr${{ github.run_id }} | awk '{print tolower($0)}')
           echo "Using acr name \"$ACR_NAME\""
           echo "ACR_NAME=$ACR_NAME" >> "$GITHUB_ENV"
+      - name: cleanup leftover resources, if any (e.g. from a previous run that failed to clean up)
```
matthchr, line 182: We use  ./.github/actions/e2e/cleanup below to do this.

What's the reasoning for not using it here?

tallaxes: Good point, should try that. Was not sure if that cleanup makes assumptions re what has been created, but if it does - we should fix it anyway.

tallaxes: On a closer look: Yes, could reuse, but will require some refactoring. Right now it does not wait and does extra checkout/login that we don't need here. Added a comment, can revisit later.

PR #1590, opened 2026-03-26
chore: remove deprecated skus

matthchr:

* Unit tests

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/providers/instancetype/skus.go
+++ b/pkg/providers/instancetype/skus.go
@@ -70,7 +70,10 @@ var (
 	)
 	// karpenterRestrictedVMSKUs are VMS SKUs that are known to
 	// be problematic with karpenter-provider-azure.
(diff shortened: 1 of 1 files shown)


tallaxes (approved):
LGTM. Just to double check expectations: won't be available for explicit selection either; is not expected to cause drift / issues with already provisioned nodes, right?

matthchr:
> LGTM. Just to double check expectations: won't be available for explicit selection either; is not expected to cause drift / issues with already provisioned nodes, right?

Correct. Will block explicit selection because List InstanceTypes is called for that and this drops the items from its response. Same for provisioning, instance types are listed so we won't provision these anymore (if say, somebody creates a NodeClaim requesting them directly), as they aren't in List InstanceTypes.

Shouldn't cause any drift, only impacts provisioning.

PR #1596, opened 2026-03-30
chore: run go fix (Go 1.26), bump golangci-lint v2.8.0 → v2.11.4

tallaxes:
**Description**

Applies Go 1.26 fixers via go fix ./... across the codebase and bumps golangci-lint from v2.8.0 to v2.11.4 so staticcheck recognizes Go 1.26's new(expr) syntax. Also updates the fake SKU generator so regenerated fake SKU files use the same new(...) pointer style as the checked-in generated outputs.

| Fixer | Count | Example |
|-------|-------|---------|
| `newexpr` | ~4200 | `lo.ToPtr("foo")` → `new("foo")` |
| `any` | 33 | `interface{}` → `any` |
| `waitgroup` | 3 | `wg.Add(1); go f(); wg.Done()` → `wg.Go(f)` |
| `fmtappendf` | 3 | `[]byte(fmt.Sprintf(...))` → `fmt.Appendf(nil, ...)` |
| `stringscut` | 2 | `strings.Index` → `strings.Cut` |
| `stringsseq` | 1 | `strings.Split` range → `strings.SplitSeq` |
| `reflecttypefor` | 1 | `reflect.TypeOf(x)` → `reflect.TypeFor[T]()` |
| `mapsloop` | 1 | explicit map loop → `maps` package |

Bumps golangci-lint from v2.8.0 to v2.11.4 so that staticcheck recognizes Go 1.26's `new(expr)` syntax — without this bump, staticcheck falsely reports SA4006 (unused value) on `new(variable)` patterns.

Adds targeted `//nolint:gosec` annotations for 14 pre-existing false positives newly surfaced by the updated gosec in v2.11.4:
- **G101** (12 sites): test structs containing `TokenAudience` or `KubeletClientTLSBootstrapToken` fields with fake test values
- **G118** (2 sites): `context.WithCancel` where the cancel func is stored in a package-level var and called in `AfterSuite`

* `make presubmit` — lint clean (0 issues), all Ginkgo tests pass

- [x] No

```release-note
NONE
```


pkg/fake/aksagentpoolsapi.go
```diff
@@ -154,13 +154,13 @@ func (c *AKSAgentPoolsAPI) BeginDeleteMachines(
...
 		if c.aksDataStorage != nil && c.aksDataStorage.AKSMachines != nil {
-			c.aksDataStorage.AKSMachines.Range(func(key, value interface{}) bool {
+			c.aksDataStorage.AKSMachines.Range(func(key, value any) bool {
 				machineID := key.(string)
 				// Check if this machine belongs to the same agent pool
 				expectedPrefix := fmt.Sprintf("/subscriptions/subscriptionID/resourceGroups/%s/providers/Microsoft.ContainerService/managedClusters/%s/agentPools/%s/machines/",
 					input.ResourceGroupName, input.ResourceName, input.AgentPoolName)
-				if strings.HasPrefix(machineID, expectedPrefix) {
-					machineName := strings.TrimPrefix(machineID, expectedPrefix)
+				if after, ok0 := strings.CutPrefix(machineID, expectedPrefix); ok0 {
```
theunrepentantgeek: Variable naming is awkward here.

pkg/fake/zz_generated.sku.southcentralus.go
theunrepentantgeek, line 1: This is a generated file - have you also updated the generator?

pkg/providers/pricing/client/pricingapi_types.go
```diff
@@ -44,7 +44,7 @@ type Item struct {
 	Type                 string    `json:"type"`
 	IsPrimaryMeterRegion bool      `json:"isPrimaryMeterRegion"`
 	ArmSkuName           string    `json:"armSkuName"`
-	EffectiveEndDate     time.Time `json:"effectiveEndDate,omitempty"`
+	EffectiveEndDate     time.Time `json:"effectiveEndDate"`
```
theunrepentantgeek: Isn't this a breaking change?

theunrepentantgeek (commented):
Overall looks good, a nice tidyup. A couple of comments and other things.

Are there any remaining uses of `github.com/samber/lo`?
I'm wondering if this needs a `go mod tidy`.

The `sync.WaitGroup` fixup is nice, any idea if we were hitting the associated race condition or not.

We should also perhaps consider whether there are any new linters we want to adopt with this version of golangci-lint - but given the size of this, a question for a later PR.

pkg/fake/zz_generated.sku.southcentralus.go
tallaxes: Good catch - and no. Originally I thought we would just add "go fix" to the workflow - in which case we would not need to update the code gen - but decided against it, and did not revisit the generator

pkg/providers/pricing/client/pricingapi_types.go
tallaxes: Good catch. Probably not breaking, because we are only consuming/deserializing - but I don't see the reason for this change in the first place, will take a look. Edit: removed

tallaxes:
> Are there any remaining uses of github.com/samber/lo?

Oh yeah, all over the place.

pkg/fake/zz_generated.sku.southcentralus.go
tallaxes: Updated

PR #1603, opened 2026-04-01
fix: add sku-hyperv-generation to CRD label allowlist

comtalyst:
**Description**

Adds `karpenter.azure.com/sku-hyperv-generation` to the CRD CEL validation allowlist in `hack/validation/requirements.sh`.

`computeRequirements()` adds `sku-hyperv-generation` to every instance type (with values `"1"`, `"2"`, or both), but the CRD allowlist was missing it. If a user references this label in a pod `nodeSelector` (e.g., to target HyperV Gen2 VMs), the label flows through the scheduler into the NodeClaim spec and the API server rejects NodeClaim creation with `"label domain karpenter.azure.com is restricted"`. The pod stays Pending with no user-facing event — the error is only visible in Karpenter controller logs.

Normal operation (no explicit reference to the label) is unaffected — instance type requirements are used for matching only and don't propagate to the NodeClaim spec.

* 

- [ ] Yes, issue opened: #
- [x] No

**Release Note**

```release-note
fix: add sku-hyperv-generation to CRD label allowlist so users can reference it in pod nodeSelectors and NodePool requirements without NodeClaim creation being rejected
```

### DIFF
--- a/charts/karpenter-crd/templates/karpenter.sh_nodeclaims.yaml
+++ b/charts/karpenter-crd/templates/karpenter.sh_nodeclaims.yaml
@@ -127,7 +127,7 @@ spec:
                           - message: label "kubernetes.io/hostname" is restricted
                             rule: self != "kubernetes.io/hostname"
                           - message: label domain "karpenter.azure.com" is restricted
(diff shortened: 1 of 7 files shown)


comtalyst:
(AI-generated)

Closing this PR. The fix was incorrect.

`sku-hyperv-generation` is intentionally in `RestrictedLabels` — it's an internal label that users are not meant to schedule on directly. The existing test `"should not allow internal labels"` explicitly verifies this restriction. Adding it to the CRD allowlist breaks that test and violates the design intent.

The original investigation was valid (the label IS missing from the allowlist), but that's by design. The issue I actually hit during E2E was a different label (`karpenter.azure.com/azurenodeclass`) which needs to go in the AzureNodeClass PR stack, not here.

If we decide users SHOULD be able to schedule on `sku-hyperv-generation`, we'd need to move it from `RestrictedLabels` to `WellKnownLabels` and update both the CRD allowlist and the test — a broader design decision.

PR #1607, opened 2026-04-02
fix: set ArtifactStreamingProfile on AKS Machine API path

comtalyst:
Fixes # 

**Description**

The artifact streaming feature (#1397) added support for the `bootstrappingclient` provision mode via labels and bootstrap scripts, but did not update the AKS Machine API (`aksmachineapi`) path. 

* Unit tests (4 new tests in `suite_aksmachineapi_features_test.go`)
* Integration E2E suite (NAP)

- [ ] Yes, issue opened: #
- [x] No

**Release Note**

```release-note
```

### DIFF
--- a/pkg/providers/instance/aksmachineinstancehelpers.go
+++ b/pkg/providers/instance/aksmachineinstancehelpers.go
@@ -142,7 +142,7 @@ func (p *DefaultAKSMachineProvider) buildAKSMachineTemplate(ctx context.Context,
 				NodeTaints:               nodeTaints,
 				MaxPods:                  nodeClass.Spec.MaxPods, // AKS machine API defaults it per network plugins if nil.
 				// WorkloadRuntime:          nil,
-				// ArtifactStreamingProfile: nil,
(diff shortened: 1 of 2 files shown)


pkg/providers/instance/aksmachineinstancehelpers.go
```diff
@@ -169,6 +169,16 @@ func configureGPUProfile(instanceType *corecloudprovider.InstanceType) *armconta
 	return nil
 }
 
+func configureArtifactStreamingProfile(nodeClass *v1beta1.AKSNodeClass, instanceType *corecloudprovider.InstanceType) *armcontainerservice.AgentPoolArtifactStreamingProfile {
+	arch := instanceType.Requirements.Get(v1.LabelArchStable).Any()
```
matthchr: I think this `Any` is actually getting you a single value because each instanceType is a single arch, but may be worth a comment? (Or possibly a swap to using something like direct indexing to avoid even the possibility of random choice here?).

MaximilianoUribe (approved):
lgtm

PR #1610, opened 2026-04-02
feat: set LocalDNSProfile on AKS Machine API path

comtalyst:
Fixes # 

**Description**

With the SDK v9 bump (#1609), `MachineProperties` now has a `LocalDNSProfile` field. Maps the Karpenter NodeClass `LocalDNS` spec to the SDK `LocalDNSProfile` when building the AKS Machine template:

- `Mode` (Preferred/Required/Disabled) → direct enum cast
- `VnetDNSOverrides`/`KubeDNSOverrides` (list keyed by zone) → `map[string]*LocalDNSOverride`
- `CacheDuration`/`ServeStaleDuration` (Go duration) → seconds (`int32`)
- All other fields (QueryLogging, Protocol, ForwardDestination, etc.) → direct enum casts

* Unit tests
* Integration E2E suite (NAP)

- [ ] Yes, issue opened: #
- [x] No

**Release Note**

```release-note
```


saewoni:
💭 Consider: test for Mode = Disabled / Preferred

  The tests only exercise Mode = Required. A quick check that Disabled and Preferred also map correctly would round out coverage, though since it's a direct string cast, the risk is low.

💭 Consider: test for KubeDNSOverrides field-level assertions

  The third test deeply asserts all fields on VnetDNSOverrides["."] but doesn't verify KubeDNSOverrides field values — only checks HaveLen(2) in the first test. This is fine since convertLocalDNSOverrides is shared, but if you want belt-and-suspenders, one  assertion on a KubeDNS entry wouldn't hurt.

saewoni:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

pkg/cloudprovider/suite_aksmachineapi_features_test.go
```diff
@@ -926,5 +926,210 @@ var _ = Describe("CloudProvider", func() {
 				Expect(aksMachine.Properties.OperatingSystem.LinuxProfile).To(BeNil())
 			})
 		})
+
+		Context("Create - LocalDNS", func() {
+			It("should set LocalDNSProfile with mode Required", func() {
+				nodeClass.Spec.LocalDNS = &v1beta1.LocalDNS{
+					Mode:             v1beta1.LocalDNSModeRequired,
```
rakechill, line 1001: Could we have this use armcontainerservice modes instead?
in the v1beta1 structs for LocalDNS I mean --> https://github.com/Azure/karpenter-provider-azure/blob/09db05d7f61d2fc8d641baff926e7e95118f052c/pkg/apis/v1beta1/aksnodeclass.go#L147-L165 

I see you using directly in line 949

rakechill: I guess, in general, how many of these custom LocalDNS fields in aks nodeclass could we replace with armcontainerservice counterparts?

rakechill (commented):
left some comments

pkg/cloudprovider/suite_aksmachineapi_features_test.go
comtalyst: For this code specifically where we are assigning to Karpenter NodeClass, given Karpenter API is also providing the constants, it must use them.

rakechill: I more meant can we make a change (maybe separate from this pr) to update karpenter API to use sdk constants directly...

comtalyst: But, I guess the question you are interested in is whether Karpenter API should define those constants in the first place, given we have `armcontainerservice` constants.

They are two separate layers owned by different entity (Karpenter maintainers vs Azure maintainers), and is allowed to have different formats, or possible values. E.g., K8s preferred format may not be what Azure API has, and Karpenter is technically able to do some defaulting/modification in between. Unless we set our philosophy for Karpenter API to map to Machine API directly, which is not the case now.

With that, I don't think Karpenter API should directly consume Azure API's constants, which is owned by a different layer. Still, when defining constants in the code you shown, it might okay setting it to `armcontainerservice` constants in some cases, as long as we own that decision.

comtalyst: In other words, those constant declarations indicate our commitment to support these values. We wouldn't want `armcontainerservice` to add/remove/edit whatever without our watch. But we could declare to reuse some constants still.

rakechill: > They are two separate layers owned by different entity (Karpenter maintainers vs Azure maintainers), and is allowed to have different formats, or possible values

Agreed, but I assume we want to mostly align with Azure APIs (even if a subset are left out for a time or we delay on picking up those changes). 

I don't think it should block this merge, but it doesn't make sense to me to have these constants (and maybe a few structs...) defined anew in Karpenter API. 

I recommend taking this as a follow-up task, either for an agent to assess or as a lower priority cleanup item. Maybe AKS LocalDNS folks could take a look, too. (cc @saewoni )

rakechill (dismissed):
approving, won't block on the localdns logic not consuming azure APIs directly

pkg/cloudprovider/suite_aksmachineapi_features_test.go
comtalyst: It is not that simple to just align with Azure APIs. LocalDNS, for example, have some conversion logic even on Karpenter side. 
Let's say, if Azure API surprise us with a new value, and we don't have this Karpenter API constants layer, then, from user's perspective, we implicitly support is, which is not the reality. 
I acknowledge that there is an extra work to add constants every time there is a new possible value. But it is more like a safeguard to ourselves to verify our support state. That, in fact, could be done with agents, but with feature owner review.

comtalyst: But if you really see the gains in total alignment with Azure API, a wider discussion should be there.

rakechill (approved):

PR #1613, opened 2026-04-02
fix: don't create empty-zone offerings for zone-restricted SKUs

### LINKED ISSUE #1384: Topology spread constraints fail with empty zone domain in counts
### Version

Karpenter version: 1.6.6, and 1.7.1
Mode: Self hosted (see below comment, someone reproduced with NAP as well).
Kubernetes version: 1.33
Region: eastus2

### Expected Behavior

With topology spread constraint `maxSkew: 1` and `whenUnsatisfiable: DoNotSchedule`, Karpenter should provision nodes in zones that satisfy the constraint.

**Current distribution:** `zone1=2, zone2=1, zone3=1`
**Expected:** Provision in zone2 or zone3 (maxSkew would be 1)

### Actual Behavior

Karpenter reports topology constraint as unsatisfiable and refuses to provision nodes. Logs show an unexpected empty string domain in topology counts:

```
counts = : 0, eastus2-1: 2, eastus2-2: 1, eastus2-3: 1
```

This appears to cause incorrect maxSkew calculation. My guess here is that the first part of that error string is missing a zone, hence ` : 0`, when others have format `$zone: $count`.

1. Deploy with topology spread constraints in a zonal region:

```yaml
topologySpreadConstraints:
  - maxSkew: 1
    topologyKey: topology.kubernetes.io/zone
    whenUnsatisfiable: DoNotSchedule
    labelSelector:
      matchLabels:
        app.kubernetes.io/name: my-name
```

2. Scale to create pending pods requiring new nodes

3. Observe Karpenter refusing to provision despite valid zones available

### Resource Specs and Logs

Note, we have two node pools, `default` and `ephemeral` (spot instances for cronjob job pods).

In both cases for us, one Deployment, and one Statefulset, both using the `default` nodepool, caused this error.

```json
{"level":"ERROR","time":"2026-01-27T18:23:45.982Z","logger":"controller","message":"could not schedule pod","commit":"28204ca","controller":"provisioner","namespace":"","name":"","reconcileID":"7b152da1-16b0-41a1-a483-025bfe3f72ca","Pod":{"name":"<REDACTED_POD_NAME>","namespace":"<REDACTED_POD_NAMESPACE>"},"taint":"dedicated=ephemeral:NoSchedule","error":"did not tolerate taint (taint=dedicated=ephemeral:NoSchedule); unsatisfiable topology constraint for topology spread, key=topology.kubernetes.io/zone (counts = : 0, eastus2-1: 1, eastus2-2: 1, eastus2-3: 1, podDomains = topology.kubernetes.io/zone Exists, nodeDomains = topology.kubernetes.io/zone Exists","errorCauses":[{"error":"did not tolerate taint (taint=dedicated=ephemeral:NoSchedule)"},{"error":"unsatisfiable topology constraint for topology spread, key=topology.kubernetes.io/zone (counts = : 0, eastus2-1: 1, eastus2-2: 1, eastus2-3: 1, podDomains = topology.kubernetes.io/zone Exists, nodeDomains = topology.kubernetes.io/zone Exists"}]}
```

### Short-term mitigation

Switching from `DoNotSchedule` to `ScheduleAnyway` allowed Karpenter to create a node, and immediately following, saw the event `Pod should schedule on: nodeclaim/$name`.

I include this because it seems to me that this setting is only really for the kube scheduler, and since karpenter _still_ needed to create a new node to fit, it probably should have been able to do so in the earlier case with `DoNotSchedule`. In fact, we rely on `DoNotSchedule` to require karpenter to schedule a new node in another zone, so that we maintain zone spread.

This is what leads me to believe this is a bug in the implementation somewhere.

### Community Note

k24dizzle:
## Problem

Fixes #1384

When using `TopologySpreadConstraints` with `whenUnsatisfiable: DoNotSchedule`, Karpenter refuses to provision new nodes even when valid zones are available.

The root cause is that some VM SKUs have `NotAvailableForSubscription` zone restrictions covering all zones in a region. `skewer` correctly strips those blocked zones, leaving an empty zone set. `instanceTypeZones()` then falls through to `sets.New("")` — a fallback originally intended for truly non-zonal regions — causing these SKUs to appear as available with `zone=""`.

Karpenter core treats `""` as a valid topology domain, so instead of seeing zone counts `[2, 1, 1]` across 3 real zones, the scheduler sees `[0, 2, 1, 1]` across 4 domains. With `maxSkew: 1`, this makes the constraint unsatisfiable even though valid zones exist.

See https://github.com/kubernetes-sigs/karpenter/pull/2864 for an attempt to fix this in the upstream karpenter, this PR is an attempt to fix this here in the azure provider by not offering these instances in the first place.

## Fix

Add `hasRawZonesInRegion()` to check `locationInfo` directly (before restriction filtering). If a SKU has zones defined in the region but none are available after restrictions are applied, return an empty set instead of `sets.New("")`. This causes the SKU to produce no offerings and be dropped from the instance type list. The `sets.New("")` path for truly non-zonal regions is preserved.

## Verification

In affected subscriptions, SKUs like `Standard_L48s_v3` report zone support in `locationInfo` but carry `NotAvailableForSubscription` restrictions covering all zones. After the fix, instance type count in westus2 drops from 507 → 496, `zone=""` offerings drop to 0, and TSC scheduling with `whenUnsatisfiable: DoNotSchedule` works correctly.

<details>
<summary>Inspecting a zone-restricted SKU with <code>az vm list-skus</code></summary>

```bash
az vm list-skus \
  --location westus2 \
  --resource-type virtualMachines \
  --query "[?name=='Standard_L48s_v3'].{name:name,locationInfo:locationInfo,restrictions:restrictions}" \
  --output json
```

```json
{
  "locationInfo": [{ "zones": ["1", "2", "3"] }],
  "restrictions": [{
    "reasonCode": "NotAvailableForSubscription",
    "type": "Zone",
    "restrictionInfo": { "zones": ["1", "2", "3"] }
  }]
}
```

</details>

### DIFF
--- a/pkg/providers/instancetype/instancetypes.go
+++ b/pkg/providers/instancetype/instancetypes.go
@@ -204,6 +204,21 @@ func (p *DefaultProvider) Get(ctx context.Context, instanceType string) (*skewer
 	return nil, fmt.Errorf("instance type %s not found", instanceType)
 }
 
+// hasRawZonesInRegion returns true if the SKU has zones defined in locationInfo for the given region,
+// regardless of subscription restrictions. This distinguishes zone-restricted SKUs (zones exist but
+// blocked by NotAvailableForSubscription) from truly non-zonal SKUs (no zones defined at all).
+func hasRawZonesInRegion(sku *skewer.SKU, region string) bool {
+	if sku.LocationInfo == nil {
+		return false
+	}
+	for _, li := range *sku.LocationInfo {
+		if li.Location != nil && strings.EqualFold(*li.Location, region) {
+			return li.Zones != nil && len(*li.Zones) > 0
+		}
+	}
+	return false
+}
+
 // instanceTypeZones generates the set of all supported zones for a given SKU
 // The strings have to match Zone labels that will be placed on Node
 func (p *DefaultProvider) instanceTypeZones(sku *skewer.SKU) sets.Set[string] {
@@ -217,7 +232,15 @@ func (p *DefaultProvider) instanceTypeZones(sku *skewer.SKU) sets.Set[string] {
 			return utils.MakeAKSLabelZoneFromARMZone(p.region, zone)
 		})...)
 	}
-	return sets.New("") // empty string means non-zonal offering
+	// If the SKU has zones defined in this region but they're all restricted for this subscription
+	// (NotAvailableForSubscription), return an empty set so no offerings are created.
+	// The sets.New("") fallback was intended for truly non-zonal regions (e.g. westcentralus),
+	// not for zone-restricted SKUs in zonal regions — those cannot be provisioned and should
+	// not appear as available offerings.
+	if hasRawZonesInRegion(sku, p.region) {
+		return sets.New[string]()
+	}
+	return sets.New("") // truly non-zonal region: no zones defined at all
 }
 
 // TODO: review; switch to controller-driven updates
--- a/pkg/providers/instancetype/instancetypes_test.go
+++ b/pkg/providers/instancetype/instancetypes_test.go
@@ -1,0 +1,87 @@
+/*
+Portions Copyright (c) Microsoft Corporation.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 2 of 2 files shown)


k24dizzle:

tallaxes:
Closing in favor of https://github.com/Azure/karpenter-provider-azure/pull/1615

PR #1615, opened 2026-04-02
fix: use zone "0" for regional VMs to match AKS node labels

### LINKED ISSUE #1384: Topology spread constraints fail with empty zone domain in counts

tallaxes:
Fixes https://github.com/Azure/karpenter-provider-azure/issues/1384

**Description**

Regional (non-zonal) VM SKUs were internally represented with `topology.kubernetes.io/zone=""` (empty string). The empty string caused two problems. 
* First, it was inserted into topology-spread domain counts as a zero-count domain, which could poison maxSkew calculations even when users never explicitly targeted it. 
* Second, Karpenter core also uses "" as the internal sentinel for "no valid domain selected" in nextDomainTopologySpread, so if the scheduler ever did select the empty domain as the best candidate, it converted that selection into DoesNotExist and treated scheduling as unsatisfiable.

Switching regional VM zones to "0" fixes the sentinel collision and matches AKS's regional node label, but it does not by itself fix every case where an extra regional domain participates in topology-spread skew math.

Key changes:
- Add `RegionalZone` constant (`"0"`) for the regional VM zone label value
- Add `MakeAKSLabelZoneFromARMZones()` that takes an ARM zones array directly, eliminating intermediate empty-string zone values
- Update all zone functions (`MakeAKSLabelZoneFromVM`, `GetAKSLabelZoneFromAKSMachine`, etc.) to return `RegionalZone` instead of `""`
- Refactor `MakeAKSLabelZoneFromARMZone()` to no longer accept or return empty strings
- Remove all empty-string zone handling throughout the codebase

* Unit tests for zone utility functions (`pkg/utils/zone_test.go`)
* Unit tests for instance type provider zone-aware provisioning, including:
  - Regional instance types get zone label `"0"` in zonal regions
  - No empty zone domain in any instance type offering (regression guard)
  - Non-zonal SKUs excluded via zone NodePool requirements (single zone and all zones)
  - Topology spread constraints work correctly when non-zonal SKUs exist (reproduces #1384)
  - Provisioning in non-zonal regions still works
* Unit tests for AKS machine instance utils (`BuildNodeClaimFromAKSMachine`, `GetAKSLabelZoneFromAKSMachine`)
* Unit tests for cloud provider (`pkg/cloudprovider/`) including AKS machine API offerings
* Full test suites pass: `pkg/utils`, `pkg/providers/instance`, `pkg/providers/instancetype`, `pkg/cloudprovider`
* `make presubmit` passes

- [ ] Yes, issue opened: #
- [x] No

**Release Note**

```release-note
fix: Regional (non-zonal) VM SKUs now use zone value "0" instead of "" (empty string) for topology.kubernetes.io/zone, matching the label AKS places on non-zonal nodes. This fixes topology spread constraint deadlocks when non-zonal SKUs exist in a zonal region (issue #1384). Users can now filter out regional SKUs via NodePool requirements on topology.kubernetes.io/zone.
```


pkg/cloudprovider/suite_aksmachineapi_offerings_test.go
```diff
@@ -343,7 +343,7 @@ var _ = Describe("CloudProvider", func() {
 				ExpectProvisionedAndWaitForPromises(ctx, env.Client, cluster, cloudProvider, coreProvisioner, azureEnv, pod)
 
 				node := ExpectScheduled(ctx, env.Client, pod)
-				Expect(node.Labels).To(HaveKeyWithValue(v1.LabelTopologyZone, ""))
+				Expect(node.Labels).To(HaveKeyWithValue(v1.LabelTopologyZone, utils.RegionalZone))
```
matthchr: minor, feel free to ignore: bare utils packages and `utils.Xyz` make me sad.

`utils/zone/...` and usage `zone.Regional` would read better.

pkg/utils/zone.go
```diff
@@ -23,39 +23,53 @@ import (
...
-	}
-	if len(vm.Zones) == 1 {
-		if vm.Location == nil {
-			return "", fmt.Errorf("virtual machine is missing location")
-		}
-		return MakeAKSLabelZoneFromARMZone(*vm.Location, *(vm.Zones)[0]), nil
-	}
-	if len(vm.Zones) > 1 {
-		return "", fmt.Errorf("virtual machine has multiple zones")
+	location := ""
```
matthchr: minor: This is `lo.FromPtr` behavior, AFAIK.

matthchr (dismissed):
LGTM, 2 small nits

pkg/cloudprovider/suite_aksmachineapi_offerings_test.go
tallaxes: Done; used utils/zones/... (because zone as a variable is just too convenient, and `zones.Regional` still works, I think)

matthchr (dismissed):

matthchr (approved):

PR #1628, opened 2026-04-08
chore: add design for provisioning regional VMs

tallaxes:
**Description**

Proposed design for provisioning regional VMs

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/designs/0009-regional-vm-provisioning.md
+++ b/designs/0009-regional-vm-provisioning.md
@@ -493,44 +493,6 @@ case RegionWideForCapacityType:
 
 The exact mapping from Azure error families to these scopes can be finalized later, but handlers that currently iterate all offerings for an instance type or capacity type need review before regional offerings are added.
 
-#### Option C: AKSNodeClass-Owned Envelope
-
-Generate the full set of zonal and regional offerings, then filter them according to the NodeClass policy before they are exposed to scheduling:
-
-```go
-func filterOfferingsByPlacementPolicy(
-    offerings cloudprovider.Offerings,
-    policy *v1beta1.ZonePlacementPolicy,
-) cloudprovider.Offerings {
-  effectivePolicy := lo.FromPtr(policy)
-  if effectivePolicy == "" {
-    return applyOmittedPlacementPolicy(offerings)
-  }
-  switch effectivePolicy {
-  case v1beta1.ZonePlacementPolicyZonal:
-        return lo.Filter(offerings, func(o *cloudprovider.Offering, _ int) bool {
-            return o.Requirements.Get(corev1.LabelTopologyZone).Any() != zones.Regional
-        })
-    case v1beta1.ZonePlacementPolicyRegional:
-        return lo.Filter(offerings, func(o *cloudprovider.Offering, _ int) bool {
-            return o.Requirements.Get(corev1.LabelTopologyZone).Any() == zones.Regional
-        })
-    case v1beta1.ZonePlacementPolicyAny:
-        return offerings
-    default:
-        return offerings
-    }
-}
-```
-
-`applyOmittedPlacementPolicy` is a placeholder for the product choice described in **Default Policy**: omission can mean `Any`, omission can mean `Compatible`, or omission can be rejected by validation if the field is required.
-
-If the API chooses to expose a named compatibility-preserving mode, it should be called `Compatible` rather than `Default`, and it should preserve today's zonal-only behavior for zone-capable SKUs while keeping regional-only SKUs eligible.
-
-This keeps the compatibility envelope, or any explicit policy, provider-owned. NodePool and pod requirements continue to narrow the filtered offering set in the normal way.
-
-If omission is used as the compatibility-preserving default in a given API version, newly created AKSNodeClasses in that version inherit that same behavior until the default changes or the field is set explicitly.
-
 ### Implementation Notes
 
 - **Dependency: zone `"0"` normalization first.** This design depends on [PR #1615](https://github.com/Azure/karpenter-provider-azure/pull/1615) (fixing [#1384](https://github.com/Azure/karpenter-provider-azure/issues/1384)), which establishes `"0"` as the canonical non-zonal zone value.
@@ -576,16 +538,16 @@ This design uses two different kinds of scheduling labels, and they should not b
 | Label | Present on offerings? | Present on launched nodes? | Workload-schedulable? |
 |---|---|---|---|
 | `topology.kubernetes.io/zone` | Yes | Yes | Yes |
-| `karpenter.azure.com/zone-placement` | Yes, if the chosen design adds it | Only if explicit propagation is implemented | Not by default |
+| `karpenter.azure.com/zone-placement` | Yes | Yes — propagated from the selected offering | Yes, once propagated |
 
 - **`topology.kubernetes.io/zone` is fully workload-facing.** Pods may reference it in `nodeSelector`, `nodeAffinity`, and `topologySpreadConstraints`. For regional nodes, the value is `"0"`.
-- **`karpenter.azure.com/zone-placement` is NodePool/NodeClass-facing only** in the first version. Workload use should be a separate later decision.
+- **`karpenter.azure.com/zone-placement` is propagated to nodes** from the selected offering. NodePools use it to control placement; workloads can also reference it in `nodeSelector` or `nodeAffinity` if needed.
 
 Zone `"0"` is not new to AKS — non-zonal agent pools already produce nodes with this value. Regional Karpenter nodes join an existing topology label space. What changes is the frequency of `"0"` nodes, especially for topology spread. See [Appendix: Azure Zone and Placement Reference](#appendix-azure-zone-and-placement-reference).
 
 ### Observability
 
-- **Placement outcome should be visible through one stable operator-facing surface in the first version.** If `zone-placement` is not projected to nodes, document the canonical place to inspect instead, such as NodeClaims, controller events, or provider logs.
+- **Propagate `zone-placement` to launched nodes.** The selected offering's `zone-placement` value (`zonal` or `regional`) should be set as a label on the node, so operators can inspect placement outcome directly via `kubectl get nodes --show-labels`.
 - **Minimum useful signal:** operators should be able to tell whether a launched node was effectively zonal or regional.
 - **Useful additional signal when practical:** distinguish between explicitly requested regional placement and regional placement reached because compatible zonal offerings were unavailable.
 
--- a/designs/0009-regional-vm-provisioning.md
+++ b/designs/0009-regional-vm-provisioning.md
@@ -1,0 +1,814 @@
+# Regional (Non-Zonal) VM Provisioning Support
+
+## Table of Contents
+
+- [Summary](#summary)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Motivation](#motivation)
+  - [Problem](#problem)
+  - [Known Issues](#known-issues)
+  - [Background: Azure Zonal vs Regional VMs](#background-azure-zonal-vs-regional-vms)
+  - [Current State](#current-state)
+- [Design Constraints](#design-constraints)
+  - [Placement Envelope vs Allocation Strategy](#placement-envelope-vs-allocation-strategy)
+  - [Zone as Placement Instruction and Topology Dimension](#zone-as-placement-instruction-and-topology-dimension)
+  - [Backward Compatibility](#backward-compatibility)
+- [Design Options](#design-options)
+  - [Options Summary](#options-summary)
+  - [Option A: Zone-Only Offering Model](#option-a-zone-only-offering-model)
+  - [Option B: Explicit Placement Controls](#option-b-explicit-placement-controls)
+  - [Option C: Provider-Owned Placement Policy](#option-c-provider-owned-placement-policy)
+  - [Option D: Silent Fallback in the Provisioning Path](#option-d-silent-fallback-in-the-provisioning-path)
+  - [Preference Pattern: Weighted NodePools](#preference-pattern-weighted-nodepools)
+- [Recommendation](#recommendation)
+  - [Option B Implementation Checklist](#option-b-implementation-checklist)
+  - [Option C: When to Choose Instead](#option-c-when-to-choose-instead)
+  - [Conceptual Code Change](#conceptual-code-change)
+  - [Implementation Notes](#implementation-notes)
+  - [Design Constraints on Future Strategy](#design-constraints-on-future-strategy)
+  - [Validation Matrix](#validation-matrix)
+  - [Rollout Guidance](#rollout-guidance)
+- [Behavioral and Operational Considerations](#behavioral-and-operational-considerations)
+  - [Schedulability](#schedulability)
+  - [Observability](#observability)
+  - [AKS Agent Pool Parity and Mapping](#aks-agent-pool-parity-and-mapping)
+  - [Compatibility Modes and Regional-Only SKUs](#compatibility-modes-and-regional-only-skus)
+  - [Storage and Zone Topology](#storage-and-zone-topology)
+  - [Machine API and Provisioning Path Impact](#machine-api-and-provisioning-path-impact)
+  - [Drift and Upgrade Implications](#drift-and-upgrade-implications)
+- [Open Questions](#open-questions)
+- [References](#references)
+- [Appendix: Azure Zone and Placement Reference](#appendix-azure-zone-and-placement-reference)
+
+## Summary
+
+Support provisioning regional (non-zonal) VMs in Karpenter on Azure for zone-capable SKUs, and provide user-facing controls for regional placement. This enables fallback when zonal offerings are exhausted or restricted, and explicit regional or mixed-mode eligibility when the placement envelope permits it.
+
+The recommended first version widens the default envelope so both zonal and regional offerings are eligible. It pairs that with launch-side selection of one concrete cheapest compatible offering, plus an internal price-first ranking that prefers zonal offerings on equal-price ties. This ranking is the **zonal-preferred tie-break**. The design also requires regional-aware unavailability handling so zonal failures do not accidentally suppress regional fallback.
+
+> **TL;DR:** Add `karpenter.azure.com/zone-placement` as an offering label (Option B), widen the default envelope to include regional offerings for zone-capable SKUs, and apply zonal-preferred tie-breaking so the common case stays zonal-first. This is a behavioral compatibility change for existing unconstrained NodePools.
+
+### Goals
+
+- Enable provisioning of regional (non-zonal) VMs for zone-capable SKUs when all availability zones are exhausted, restricted, or otherwise unavailable
+- Provide explicit user-facing controls for regional placement behavior for zone-capable SKUs, including regional-only and mixed-mode eligibility where appropriate
+- Minimize unexpected behavior changes on upgrade; exact post-upgrade behavior may still change if the resulting placement remains predictable and acceptably close to today's zonal-first behavior
+- Ensure regional offerings work correctly with the existing Machine API and VM provisioning paths
+- Use zone `"0"` consistently to match AKS node labels set by cloud-provider-azure for non-zonal VMs
+
+### Non-Goals
+
+- VMSS-based regional support — Karpenter creates standalone VMs only; VMSS fault domain distribution is out of scope
+- Topology spread constraint automation — users who mix zonal and regional nodes are responsible for configuring topology spread constraints appropriately
+- Proactive zonal capacity or quota management — this design does not add quota-aware scheduling, capacity-aware zone choice, or other predictive handling of Azure zone-specific state; it continues to react to provisioning failures via the UnavailableOfferings cache
+- Changing the zone label semantics for non-zonal VMs — the `topology.kubernetes.io/zone = "0"` value is determined by cloud-provider-azure, not Karpenter
+
+## Motivation
+
+### Problem
+
+This design addresses two distinct problems:
+
+1. **Zonal exhaustion with no regional fallback.** On Azure, a VM SKU that appears zonal (zones listed in the Resource SKUs API) can fail provisioning in every supported zone — due to quota exhaustion, capacity shortage, or subscription restrictions — yet succeed when provisioned without a zone (regional/non-zonal). Regional placement removes the zone constraint, giving Azure access to a larger pool of hardware — including racks not assigned to any availability zone — that zonal requests cannot reach. Today, Karpenter only generates **zonal offerings** for zone-capable SKUs. If all zones are marked unavailable (via the UnavailableOfferings cache), the SKU becomes entirely unprovisionable — even though Azure could place the VM regionally. This is distinct from zone *restrictions* in the Resource SKUs API: when all zones are restricted, skewer returns an empty zone set and the SKU falls through to a regional (`zone="0"`) offering. The gap is specifically for runtime capacity/quota failures where zonal offerings exist but are all marked unavailable.
+
+2. **No user control over regional placement.** There is no mechanism for users to opt into regional placement for zone-capable SKUs — even when they explicitly prefer non-zonal VMs (e.g., for cost-insensitive batch workloads where availability matters more than zone pinning).
+
+### Known Issues
+
+- [Azure/karpenter-provider-azure#1002](https://github.com/Azure/karpenter-provider-azure/issues/1002) — Feature request: run VMs without redundancy options (no availability zone)
+- [Azure/karpenter-provider-azure#1384](https://github.com/Azure/karpenter-provider-azure/issues/1384) — Non-zonal zone label uses `""` instead of `"0"`, breaking topology spread constraints (prerequisite fix, implemented in [PR #1615](https://github.com/Azure/karpenter-provider-azure/pull/1615))
+- [Azure/AKS#5574](https://github.com/Azure/AKS/issues/5574) — NAP cannot assign regional (non-AZ) nodes. Users requesting `zone: 0` only get regional-only SKUs (e.g., `Standard_DS3`), because zone-capable SKUs don't generate regional offerings
+- Reported KAITO request — a GPU SKU cannot currently be provisioned through Karpenter/NAP because zonal placement is blocked in the target environment even though regional placement remains viable
+
+### Background: Azure Zonal vs Regional VMs
+
+Azure VMs can be created **with a zone** (pinned to specific AZ hardware) or **without a zone** (regional — Azure chooses any available hardware in the region, including non-zonal racks). The regional capacity pool is a superset of zonal capacity. A SKU listing zones in the Resource SKUs API means it is zone-capable, not that capacity is guaranteed in those zones — and zone-scoped failures or restrictions do not imply regional restrictions.
+
+Karpenter uses `"0"` as the zone value for non-zonal offerings because it must match the actual AKS node label: cloud-provider-azure reads the fault domain number from ARM (always `0` for standalone VMs) and sets it as `topology.kubernetes.io/zone`. See [#1384](https://github.com/Azure/karpenter-provider-azure/issues/1384) for the prerequisite fix from `""` to `"0"`.
+
+For detailed Azure zone mechanics, the Resource SKUs API, common failure modes, and zone label semantics, see [Appendix: Azure Zone and Placement Reference](#appendix-azure-zone-and-placement-reference).
+
+### Current State
+
+Today, for a zone-capable SKU (e.g., `Standard_NC24ads_A100_v4` in eastus with zones [1,2,3]), Karpenter generates only zonal offerings:
+
+```
+Offerings:
+  (Standard_NC24ads_A100_v4, eastus-1, on-demand)
+  (Standard_NC24ads_A100_v4, eastus-2, on-demand)
+  (Standard_NC24ads_A100_v4, eastus-3, on-demand)
+  (Standard_NC24ads_A100_v4, eastus-1, spot)
+  (Standard_NC24ads_A100_v4, eastus-2, spot)
+  (Standard_NC24ads_A100_v4, eastus-3, spot)
+```
+
+There is **no regional/non-zonal offering**. Zone `"0"` is only generated for SKUs where the SKU API returns empty zones (SKUs that don't support zones at all, or non-zonal regions).
+
+The relevant code is in `pkg/providers/instancetype/instancetypes.go`:
+
+```go
+func (p *DefaultProvider) instanceTypeZones(sku *skewer.SKU) sets.Set[string] {
+    skuZones := lo.Keys(sku.AvailabilityZones(p.region))
+    if len(skuZones) > 0 {
+        return sets.New(lo.Map(skuZones, func(zone string, _ int) string {
+            return zones.MakeAKSLabelZoneFromARMZone(p.region, zone)
+        })...)
+    }
+    return sets.New(zones.Regional)  // "0" — only for non-zonal SKUs
+}
+```
+
+## Design Constraints
+
+Three constraints shape the design space: the distinction between which offerings are eligible and which one wins, the dual role of zone values, and the backward compatibility impact of widening the placement envelope. Regional VMs also interact with zone-pinned storage; see [Storage and Zone Topology](#storage-and-zone-topology).
+
+### Placement Envelope vs Allocation Strategy
+
+This design separates two questions:
+
+1. **Which offerings are eligible at all?** This is the **placement envelope** — defined by hard constraints such as `topology.kubernetes.io/zone`, `karpenter.azure.com/zone-placement`, or an AKSNodeClass placement policy.
+2. **If multiple offerings are eligible, which one should win?** This is the **allocation strategy** — a ranking policy applied after the eligible set is determined.
+
+The distinction matters because these look similar but represent different contracts:
+
+- **Envelope-only zonal:** only zonal offerings are eligible
+- **Envelope allowing both (`Any`):** zonal and regional offerings are both eligible and may be chosen interchangeably
+- **`Any` envelope plus ranking:** both are eligible, but zonal wins on equal-price ties
+
+An envelope that allows both is **not** the same as "zonal fallback to regional."
+
+> **Zonal-preferred tie-break (definition):** The recommended internal default ranking for the first version. Within an `Any` envelope, keep normal price ordering but prefer a compatible zonal offering when zonal and regional offerings are price-equivalent; fall back to regional when the best available price tier has no compatible zonal offering.
+
+Where different policy types belong:
+
+- If only one placement mode is acceptable → express it in the **placement envelope** (`Zonal`, `Regional`, `zone NotIn [0]`, etc.)
+- If multiple placement modes are acceptable and the question is which wins first → that is an **allocation strategy** problem
+- **Weighted NodePools** are a coarse-grained, core-visible preference across separate envelopes; **allocation strategy** is fine-grained ranking within one envelope
+
+This framework applies to Options **A**, **B**, and **C** whenever the envelope can expose both zonal and regional offerings. It does not apply to **D**, because silent fallback is not represented in the offering model.
+
+### Zone as Placement Instruction and Topology Dimension
+
+The zone value participates in two independent mechanisms at different stages:
+
+1. **Before VM creation — Karpenter's scheduling simulation** selects an offering, which includes a zone value (e.g., `eastus-1` or `"0"`). This zone determines where the VM will be placed: Karpenter translates it into the ARM `zones` property on the VM creation request. At this stage, the zone is a **placement instruction**.
+
+2. **After VM creation — Kubernetes scheduling** uses the same zone value, now a label on the node (`topology.kubernetes.io/zone`), as a **topology dimension** for topology spread constraints, pod affinity, and other zone-aware scheduling decisions.
+
+Regional VMs create a tension between these two roles: they have no real availability zone, but this provider path still needs a consistent zone label for post-launch topology semantics, and cloud-provider-azure sets that label to the fault domain number (always `"0"` for standalone VMs). This `"0"` value is semantically different from `eastus-1` — it represents a fault domain, not an availability zone, and provides no indication of where Azure physically placed the VM. Yet topology spread constraints treat it as just another zone domain.
+
+### Backward Compatibility
+
+Existing users expect:
+- Zone-capable SKUs produce zonal nodes
+- Topology spread on zone works correctly
+- No surprise non-zonal nodes appearing
+
+Most compatibility questions in this design are about the **placement envelope** for existing objects, not about ranking. If an upgrade widens the envelope from zonal-only to `Any`, then regional nodes become possible outcomes for launches, consolidation, and topology interactions. That is a real behavioral compatibility change, even if a zonal-preferred strategy keeps zonal as the common case. By contrast, changing ranking inside the same envelope is usually a smaller change: it affects which acceptable offering wins first, not which outcomes are allowed at all.
+
+This gives a useful rule of thumb:
+
+- If the question is **"can regional ever happen after upgrade?"**, that is an **envelope** question.
+- If the question is **"when zonal and regional are both allowed, which should win first?"**, that is a **ranking** question.
+
+Full preservation is one option, but not the only acceptable one. A smaller behavioral compatibility change may still be acceptable if it is predictable and still leans toward today's zonal-first intent.
+
+This gives a compatibility spectrum:
+
+- **Strict preservation:** existing objects keep an effectively zonal-only envelope unless users opt in to something broader
+- **Softened preservation:** existing objects may include regional placement in the envelope, but the common case stays close to today's zonal-first behavior through internal ranking or similar safeguards
+- **Explicit behavioral change:** omission or lack of requirements means `Any`, with no implied zonal-only behavior beyond whatever ranking the implementation applies
+
+The choice between those cases is a product decision. Strict preservation requires a provider-owned compatibility envelope (e.g., AKSNodeClass-owned filtering). Softened preservation makes options that widen eligibility but prefer zonal placement in common cases more viable. The working recommendation is to accept softened preservation (see [Recommendation](#recommendation)).
+
+## Design Options
+
+### Options Summary
+
+The options below compare different ways to define the placement envelope. Throughout this section, "behavioral compatibility change" means a change in observable placement behavior, not an API/schema change. Within **Option B**, the remaining sub-choice is the placement vocabulary: two-valued (`zonal`, `regional`) or three-valued (adding `regional-only`). **Weighted NodePools** are a preference pattern that layers on top of **A**, **B**, or **C** (not **D**); see [Preference Pattern: Weighted NodePools](#preference-pattern-weighted-nodepools).
+
+| Option | Approach | Backward compatible? | Default semantics | User control | Key tradeoff |
+|---|---|---|---|---|---|
+| **A** | Zone is the only placement control surface | No | Unconstrained NodePool = zonal and regional eligible | `zone` requirements only | Smallest implementation, but overloads topology with placement intent |
+| **B** | Placement mode is modeled explicitly via `zone-placement` | No, though a price-first default that prefers zonal on equal-price ties softens the common-case impact | No `zone-placement` requirement = `Any`; recommended first-version behavior is price-first with zonal preference on equal-price ties | `zone-placement` plus `zone` requirements | Primary recommendation; clean explicit model, but unconstrained existing NodePools widen after upgrade |
+| **C** | AKSNodeClass owns the placement envelope before scheduling | Yes, if omission or the defaulted policy preserves current behavior | Compatibility envelope when needed; otherwise a provider-owned default such as `Any` is separate | Optional AKSNodeClass policy plus `zone` requirements | Strongest alternative for stricter compatibility or provider-owned defaults, but adds a second control surface and defaulting complexity |
+| **D** | Regional retry happens outside the offering model | At the API surface only | Automatic fallback after zonal failure | None | Smallest user-facing change, but hides placement changes from the scheduler and user |
+
+Under the working recommendation, prefer **B** with the two-value variant; weighted NodePools remain an optional explicit preference pattern on top. Prefer **C** only if a provider-owned default envelope or stricter compatibility path is needed.
+
+### Option A: Zone-Only Offering Model
+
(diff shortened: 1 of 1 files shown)


designs/0009-regional-vm-provisioning.md
```diff
@@ -0,0 +1,814 @@
...
+
+## Summary
+
+Support provisioning regional (non-zonal) VMs in Karpenter on Azure for zone-capable SKUs, and provide user-facing controls for regional placement. This enables fallback when zonal offerings are exhausted or restricted, and explicit regional or mixed-mode eligibility when the placement envelope permits it.
+
+The recommended first version widens the default envelope so both zonal and regional offerings are eligible. It pairs that with launch-side selection of one concrete cheapest compatible offering, plus an internal price-first ranking that prefers zonal offerings on equal-price ties. This ranking is the **zonal-preferred tie-break**. The design also requires regional-aware unavailability handling so zonal failures do not accidentally suppress regional fallback.
+
+> **TL;DR:** Add `karpenter.azure.com/zone-placement` as an offering label (Option B), widen the default envelope to include regional offerings for zone-capable SKUs, and apply zonal-preferred tie-breaking so the common case stays zonal-first. This is a behavioral compatibility change for existing unconstrained NodePools.
```
matthchr: I assume this will get answered later but what do example values for this label look like?

designs/0009-regional-vm-provisioning.md
```diff
@@ -0,0 +1,814 @@
...
+
+- Enable provisioning of regional (non-zonal) VMs for zone-capable SKUs when all availability zones are exhausted, restricted, or otherwise unavailable
+- Provide explicit user-facing controls for regional placement behavior for zone-capable SKUs, including regional-only and mixed-mode eligibility where appropriate
+- Minimize unexpected behavior changes on upgrade; exact post-upgrade behavior may still change if the resulting placement remains predictable and acceptably close to today's zonal-first behavior
+- Ensure regional offerings work correctly with the existing Machine API and VM provisioning paths
+- Use zone `"0"` consistently to match AKS node labels set by cloud-provider-azure for non-zonal VMs
+
+### Non-Goals
+
+- VMSS-based regional support — Karpenter creates standalone VMs only; VMSS fault domain distribution is out of scope
```
matthchr: This non-goal is a bit confusing to me? I mean I guess yes it's true we're not adding VMSS support here it also mentions fault domain, which maybe is what it's really trying to talk about?

designs/0009-regional-vm-provisioning.md
```diff
@@ -0,0 +1,814 @@
...
+- Topology spread constraint automation — users who mix zonal and regional nodes are responsible for configuring topology spread constraints appropriately
```
matthchr: very minor:
```suggestion
- Topology spread constraint automation — users who mix zonal and regional nodes are responsible for configuring topology spread constraints appropriately on their workloads
```

designs/0009-regional-vm-provisioning.md
```diff
@@ -0,0 +1,814 @@
...
+- Reported KAITO request — a GPU SKU cannot currently be provisioned through Karpenter/NAP because zonal placement is blocked in the target environment even though regional placement remains viable
+
+### Background: Azure Zonal vs Regional VMs
+
+Azure VMs can be created **with a zone** (pinned to specific AZ hardware) or **without a zone** (regional — Azure chooses any available hardware in the region, including non-zonal racks). The regional capacity pool is a superset of zonal capacity. A SKU listing zones in the Resource SKUs API means it is zone-capable, not that capacity is guaranteed in those zones — and zone-scoped failures or restrictions do not imply regional restrictions.
+
+Karpenter uses `"0"` as the zone value for non-zonal offerings because it must match the actual AKS node label: cloud-provider-azure reads the fault domain number from ARM (always `0` for standalone VMs) and sets it as `topology.kubernetes.io/zone`. See [#1384](https://github.com/Azure/karpenter-provider-azure/issues/1384) for the prerequisite fix from `""` to `"0"`.
```
matthchr:
```suggestion
Karpenter uses `"0"` as the zone value for non-zonal offerings because it must match the actual AKS node label: cloud-provider-azure reads the fault domain number from ARM (always `0` for non-zonal standalone VMs) and sets it as `topology.kubernetes.io/zone`. See [#1384](https://github.com/Azure/karpenter-provider-azure/issues/1384) for the prerequisite fix from `""` to `"0"`.
```

designs/0009-regional-vm-provisioning.md
```diff
@@ -0,0 +1,814 @@
...
+- [Azure/AKS#5574](https://github.com/Azure/AKS/issues/5574) — NAP cannot assign regional (non-AZ) nodes. Users requesting `zone: 0` only get regional-only SKUs (e.g., `Standard_DS3`), because zone-capable SKUs don't generate regional offerings
+- Reported KAITO request — a GPU SKU cannot currently be provisioned through Karpenter/NAP because zonal placement is blocked in the target environment even though regional placement remains viable
+
+### Background: Azure Zonal vs Regional VMs
+
+Azure VMs can be created **with a zone** (pinned to specific AZ hardware) or **without a zone** (regional — Azure chooses any available hardware in the region, including non-zonal racks). The regional capacity pool is a superset of zonal capacity. A SKU listing zones in the Resource SKUs API means it is zone-capable, not that capacity is guaranteed in those zones — and zone-scoped failures or restrictions do not imply regional restrictions.
```
matthchr, line 85: Is there any indication in the instance types API if a VM supports regional in addition to zonal, or do they _always_ support regional?

Possibly this is answered in the appendix I haven't gotten there yet.

designs/0009-regional-vm-provisioning.md
```diff
@@ -0,0 +1,814 @@
...
+    }
+    return sets.New(zones.Regional)  // "0" — only for non-zonal SKUs
+}
+```
+
+## Design Constraints
+
+Three constraints shape the design space: the distinction between which offerings are eligible and which one wins, the dual role of zone values, and the backward compatibility impact of widening the placement envelope. Regional VMs also interact with zone-pinned storage; see [Storage and Zone Topology](#storage-and-zone-topology).
+
+### Placement Envelope vs Allocation Strategy
```
matthchr, line 125: Yes, I think generally this is very important to think about -- thanks for calling it out very clearly here.

designs/0009-regional-vm-provisioning.md
```diff
@@ -0,0 +1,814 @@
...
+
+This design separates two questions:
+
+1. **Which offerings are eligible at all?** This is the **placement envelope** — defined by hard constraints such as `topology.kubernetes.io/zone`, `karpenter.azure.com/zone-placement`, or an AKSNodeClass placement policy.
+2. **If multiple offerings are eligible, which one should win?** This is the **allocation strategy** — a ranking policy applied after the eligible set is determined.
+
+The distinction matters because these look similar but represent different contracts:
+
+- **Envelope-only zonal:** only zonal offerings are eligible
+- **Envelope allowing both (`Any`):** zonal and regional offerings are both eligible and may be chosen interchangeably
```
matthchr: interchangeably also means possibly "randomly" or "arbitrarily" in this context, is that right?

designs/0009-regional-vm-provisioning.md
```diff
@@ -0,0 +1,814 @@
...
+- **`Any` envelope plus ranking:** both are eligible, but zonal wins on equal-price ties
+
+An envelope that allows both is **not** the same as "zonal fallback to regional."
```
matthchr: Maybe needs an explanation of why (I think the point is that if they're both considered equally then there would possibly be cases where regional wins over zonal, so it's not a _fallback_).

designs/0009-regional-vm-provisioning.md
```diff
@@ -0,0 +1,814 @@
...
+
+> **Zonal-preferred tie-break (definition):** The recommended internal default ranking for the first version. Within an `Any` envelope, keep normal price ordering but prefer a compatible zonal offering when zonal and regional offerings are price-equivalent; fall back to regional when the best available price tier has no compatible zonal offering.
```
matthchr: Is `Any` here an actual value users are passing somewhere, or just a logical construct in this design? Not clear at this point.

designs/0009-regional-vm-provisioning.md
```diff
@@ -0,0 +1,814 @@
...
+
+Where different policy types belong:
+
+- If only one placement mode is acceptable → express it in the **placement envelope** (`Zonal`, `Regional`, `zone NotIn [0]`, etc.)
+- If multiple placement modes are acceptable and the question is which wins first → that is an **allocation strategy** problem
+- **Weighted NodePools** are a coarse-grained, core-visible preference across separate envelopes; **allocation strategy** is fine-grained ranking within one envelope
+
+This framework applies to Options **A**, **B**, and **C** whenever the envelope can expose both zonal and regional offerings. It does not apply to **D**, because silent fallback is not represented in the offering model.
```
matthchr: minor: I think this is talking about the options we're about to get to in the doc, but we haven't gotten to them yet. May be worth including the word "below" or similar in here somewhere just for clarity.

designs/0009-regional-vm-provisioning.md
```diff
@@ -0,0 +1,814 @@
...
+
+1. **Before VM creation — Karpenter's scheduling simulation** selects an offering, which includes a zone value (e.g., `eastus-1` or `"0"`). This zone determines where the VM will be placed: Karpenter translates it into the ARM `zones` property on the VM creation request. At this stage, the zone is a **placement instruction**.
+
+2. **After VM creation — Kubernetes scheduling** uses the same zone value, now a label on the node (`topology.kubernetes.io/zone`), as a **topology dimension** for topology spread constraints, pod affinity, and other zone-aware scheduling decisions.
+
+Regional VMs create a tension between these two roles: they have no real availability zone, but this provider path still needs a consistent zone label for post-launch topology semantics, and cloud-provider-azure sets that label to the fault domain number (always `"0"` for standalone VMs). This `"0"` value is semantically different from `eastus-1` — it represents a fault domain, not an availability zone, and provides no indication of where Azure physically placed the VM. Yet topology spread constraints treat it as just another zone domain.
```
matthchr, line 158: I assume there's no fixing this, at this point? (Does it even _need_ fixing?)

designs/0009-regional-vm-provisioning.md
```diff
@@ -0,0 +1,814 @@
...
+- No surprise non-zonal nodes appearing
+
+Most compatibility questions in this design are about the **placement envelope** for existing objects, not about ranking. If an upgrade widens the envelope from zonal-only to `Any`, then regional nodes become possible outcomes for launches, consolidation, and topology interactions. That is a real behavioral compatibility change, even if a zonal-preferred strategy keeps zonal as the common case. By contrast, changing ranking inside the same envelope is usually a smaller change: it affects which acceptable offering wins first, not which outcomes are allowed at all.
+
+This gives a useful rule of thumb:
+
+- If the question is **"can regional ever happen after upgrade?"**, that is an **envelope** question.
+- If the question is **"when zonal and regional are both allowed, which should win first?"**, that is a **ranking** question.
+
+Full preservation is one option, but not the only acceptable one. A smaller behavioral compatibility change may still be acceptable if it is predictable and still leans toward today's zonal-first intent.
```
matthchr: Full preservation of what? I think you're talking about "existing behavior" here - though I am not 100% squaring that with the ask, because I think we're saying that we want to expand the envelope, as we're not currently considering regional placement at all.. though actualyl as I  write that I realize maybe what you're saying is that we have sorta 3 cases:

1. user didn't explicitly ask for zonal only (no regional), but since that's our behavior today for the VM sizes they're using they're effectively getting it, and that's what they want too.
2. User did explicitly ask for zonal (zones in 1,2,3 or similar?)
3. User didn't explicitly ask for zonal only, and would be open to regional as a fallback.
4. User wants to ask regional-only to start with

If this is correct it may be worth calling out the user use-cases somewhere (apologies if you already do, Im not done with the doc yet).

Actually this is very close/related to your compatibility spectrum, in that it's just the user-facing side of it "what they want vs what they get" - with the added capabilities that this design is proposing included.

designs/0009-regional-vm-provisioning.md
```diff
@@ -0,0 +1,814 @@
...
+| **C** | AKSNodeClass owns the placement envelope before scheduling | Yes, if omission or the defaulted policy preserves current behavior | Compatibility envelope when needed; otherwise a provider-owned default such as `Any` is separate | Optional AKSNodeClass policy plus `zone` requirements | Strongest alternative for stricter compatibility or provider-owned defaults, but adds a second control surface and defaulting complexity |
+| **D** | Regional retry happens outside the offering model | At the API surface only | Automatic fallback after zonal failure | None | Smallest user-facing change, but hides placement changes from the scheduler and user |
+
+Under the working recommendation, prefer **B** with the two-value variant; weighted NodePools remain an optional explicit preference pattern on top. Prefer **C** only if a provider-owned default envelope or stricter compatibility path is needed.
```
matthchr, line 208: I mean there is always a provider-owned default envelope w.r.t. zonal vs regional, isn't there? It's either what we (explicitly or implicitly) default on AKSNodeClass, or it's what we do in the absence of any label, isn't it?

designs/0009-regional-vm-provisioning.md
```diff
@@ -0,0 +1,814 @@
...
+**User controls via `topology.kubernetes.io/zone` requirement:**
+
+| Intent | NodePool requirement |
+|---|---|
+| Only zonal | `zone NotIn [0]` or `zone In [eastus-1, eastus-2, eastus-3]` |
+| Only regional | `zone In [0]` |
+| Specific zonal zones | `zone In [eastus-1, eastus-2]` |
+| Any | No zone requirement |
+
+This is the smallest code change and composes naturally with existing exact-zone requirements. Like **Option B**, it widens the default envelope for unconstrained NodePools. The main drawback is semantic overload: zone becomes both topology and placement mode, making abstract intent such as "zonal anywhere in the region" less portable across clusters or regions.
```
matthchr, line 234: Worth noting that this semantic overload is unavoidable if the user is looking at topology spread constraints, right?

designs/0009-regional-vm-provisioning.md
```diff
@@ -0,0 +1,814 @@
...
+
+**Verdict:** technically simple but weaker than **Option B** due to semantic overload of the zone label.
+
+### Option B: Explicit Placement Controls
+
+Introduce an Azure-specific offering label such as `karpenter.azure.com/zone-placement`, and treat placement mode as an orthogonal dimension from zone topology.
```
matthchr: minor, but maybe important: `zone-placement` makes me think it's similar/the same as `zone`, whereas what we're talking about is "are we putting it in a zone at all".

I don't have a better name but maybe just highlighting that `zone-placement=regional` sorta implies that `regional` is a type of zone, or something, when it's really the absence.

I don't love suggesting something like `karpenter.azure.com/zone-aware=true|false` (or zonal?) but it's another possible path. maybe we can come up something with a better name?

designs/0009-regional-vm-provisioning.md
```diff
@@ -0,0 +1,814 @@
...
+
+The critical design rule is:
+
+> If placement is controlled only through NodePool requirements, then **absence of a `zone-placement` requirement means `Any`**, not "default zonal".
+
+The recommended first version pairs that `Any` default with launch-side selection of a concrete compatible offering from the sorted set, plus an internal price-first ranking that prefers zonal offerings on equal-price ties, so the wider envelope still behaves zonal-first in the common case. Changing offering sort order alone is not sufficient.
+
+This model composes cleanly with normal zone requirements because `zone-placement` filters the placement envelope, while `topology.kubernetes.io/zone` continues to select concrete topology domains within that envelope.
+
+Users can still control placement via `zone` alone when the zone value already identifies the intended mode (e.g., `zone In [0]` effectively means regional only). However, `zone-placement` expresses abstract intent like "zonal anywhere in the region" more cleanly and portably across clusters and regions.
```
matthchr: "zonal anywhere in the region" means "any zone is fine but must be a zone", right?

designs/0009-regional-vm-provisioning.md
```diff
@@ -0,0 +1,814 @@
...
+| Only regional | `zone-placement In [regional]` | `zone In [0]` | Clean zone-only form because `0` already identifies regional placement |
+| Only regional, explicit zone | `zone-placement In [regional]` + `zone In [0]` | `zone In [0]` | Equivalent in practice; keeping both can make intent more obvious |
+| Specific zonal zones | `zone-placement In [zonal]` + `zone In [eastus-1, eastus-2]` | `zone In [eastus-1, eastus-2]` | Natural when the user really wants named zones |
+| Any | No `zone-placement` requirement, or `zone-placement In [zonal, regional]` | No `zone` requirement | A zone-only include list would require naming all currently eligible zones |
+
+**Pros:**
+- Clean separation of concerns — zone topology stays pure, placement mode is an orthogonal dimension
+- Composes naturally with exact zone requirements
+- Lets users say `zonal only`, `regional only`, or `any` without overloading zone semantics
+- Makes NodePool intent more portable across clusters and regions by avoiding the need to name or enumerate concrete zones for common placement modes
```
matthchr, line 278: Option A technically doesn't require this either, does it?

designs/0009-regional-vm-provisioning.md
```diff
@@ -0,0 +1,814 @@
...
+
+**Verdict:** the recommended primary model when an explicit, requirement-based experience is desired and some behavioral widening is acceptable.
+
+#### Placement Vocabulary
+
+Two variants for the `zone-placement` label values:
+
+- **Two-value model (recommended):** `zonal` and `regional`. All regional offerings, including regional-only SKUs, are tagged `regional`. This is the simplest design and the recommended form for the first version.
+
+- **Three-value model (optional):** Add `regional-only` to distinguish inherently regional SKUs from zone-capable SKUs placed regionally. Not required for correctness — only useful if the design wants finer-grained observability or future policy controls.
```
matthchr, line 294: Yeah this feels like it adds some confusion (IMO?). If the user wants to select specific regional (non-zonal) SKUs, they can just select the SKU? Do we really have/see a need for users who may want an _arbitrary_ (or at least somewhat arbitrary) SKU as long as it's regional and not zonal? Seems a very niche case.

designs/0009-regional-vm-provisioning.md
```diff
@@ -0,0 +1,814 @@
...
+
+**Pros:**
+- Supports a real compatibility-preserving envelope when required
+- Also supports provider-owned defaults and managed-product behavior when strict preservation is relaxed
+- Keeps exact zone selection orthogonal and intact
+- Makes the enforcement point explicit: the provider filters offerings before scheduling
+
+**Cons:**
+- Adds a second control surface
+- If `Compatible` behavior is retained, it can codify current behavior even where current behavior is not desirable
```
matthchr, line 325: Yeah I think the fact we own the allocation policy today (it's price, and maybe _assumed_ to be price, but we have quite a lot of leeway in practice, I think?) means we have the flexibility to change things and I don't know if we want to open the door to some `Compatible` variant in perpetuity

designs/0009-regional-vm-provisioning.md
```diff
@@ -0,0 +1,814 @@
...
+  weight: 1
+  template:
+    spec:
+      requirements:
+        - key: karpenter.azure.com/zone-placement
+          operator: In
+          values: [regional]
+```
+
+**Verdict:** recommended when users need explicit or stronger ordered preference than the provider default.
```
matthchr, line 399: Though strictly speaking there's not much practical difference between within-envelope ordering versus this, is there?

Since with in the envelope if we still require this ordering the two experiences become:
**Single NodePool**
- First, we must exhaust all zonal options
- Then, we try the non-zonal options

**Two Nodepools**
- First, we must exhaust all the zonal options (which allows us to fall through to the lower weight pool)
- Then, we try the non-zonal options.

Main caveat there being that assumes zonal + non-zonal have the same price (so tie is preferring zonal). If for some reason zonal is more expensive, non-zonal would win. Though I guess from the way Azure does pricing that's not possible.

designs/0009-regional-vm-provisioning.md
```diff
@@ -0,0 +1,814 @@
...
+
+If the project requires a provider-owned compatibility envelope or provider-owned defaults, use **Option C** instead; see [Option C](#option-c-provider-owned-placement-policy).
+
+### Option B Implementation Checklist
+
+1. **Add `karpenter.azure.com/zone-placement`** with values `zonal` and `regional`
+2. **Generate regional offerings for zone-capable SKUs** with `zone="0"` for capacity types that support regional placement
+3. **Tag all regional offerings**, including regional-only SKUs, as `regional`
+4. **Do not infer a default** from the absence of a NodePool requirement; no `zone-placement` requirement means `Any`
```
matthchr: Possibly we should refer to `Any` as `zone-placement In [zonal, regional]`, as that's more correct given the semantics we've chosen? Or else call out that you're using any as a shorthand for that.

designs/0009-regional-vm-provisioning.md
```diff
@@ -0,0 +1,814 @@
...
+5. **Apply the zonal-preferred tie-break** within the `Any` envelope, and make the launch path select one concrete offering using that ranking rather than selecting an arbitrary zone within the chosen capacity type
+6. **Make error handling and `UnavailableOfferings` marking regional-aware**, so zonal-scoped failures do not automatically mark the regional fallback unavailable
```
matthchr, line 414: Is this actually required or can we just use the fake zone 0 for this already?

designs/0009-regional-vm-provisioning.md
```diff
@@ -0,0 +1,814 @@
...
+7. **Use weighted NodePools only as an optional explicit preference pattern** when users want stronger or different precedence than the provider default
```
matthchr: minor: Not really an item that makes sense on an implementation checklist

designs/0009-regional-vm-provisioning.md
```diff
@@ -0,0 +1,814 @@
...
+
+If the project requires a provider-owned compatibility envelope or provider-owned defaults, use **Option C** instead; see [Option C](#option-c-provider-owned-placement-policy).
+
+### Option B Implementation Checklist
+
+1. **Add `karpenter.azure.com/zone-placement`** with values `zonal` and `regional`
+2. **Generate regional offerings for zone-capable SKUs** with `zone="0"` for capacity types that support regional placement
+3. **Tag all regional offerings**, including regional-only SKUs, as `regional`
+4. **Do not infer a default** from the absence of a NodePool requirement; no `zone-placement` requirement means `Any`
+5. **Apply the zonal-preferred tie-break** within the `Any` envelope, and make the launch path select one concrete offering using that ranking rather than selecting an arbitrary zone within the chosen capacity type
```
matthchr: Not sure how specific you want to get here, but if this is for AI, may want to point at where this should happen or say the allocation strategy name here (assuming that's the codepath we want to use to do this, which I think it is).

designs/0009-regional-vm-provisioning.md
```diff
@@ -0,0 +1,814 @@
...
+6. **Make error handling and `UnavailableOfferings` marking regional-aware**, so zonal-scoped failures do not automatically mark the regional fallback unavailable
+7. **Use weighted NodePools only as an optional explicit preference pattern** when users want stronger or different precedence than the provider default
+
+### Option C: When to Choose Instead
+
+Choose **Option C** when provider-owned defaults or a strict compatibility envelope justify the extra control-surface complexity. The implementation checklist for that path:
```
matthchr: Probably cut this, no need for it given the plan is option B - can revise the document to talk about impl path for option C if we change to that as the option (but I like B more too).

designs/0009-regional-vm-provisioning.md
```diff
@@ -0,0 +1,814 @@
...
+        - key: karpenter.azure.com/zone-placement
+          operator: In
+          values: [regional]
+```
+
+**Verdict:** recommended when users need explicit or stronger ordered preference than the provider default.
+
+## Recommendation
+
+We recommend **Option B**: add `karpenter.azure.com/zone-placement` as an offering label, widen the default envelope to include regional, and apply the zonal-preferred tie-break. This accepts a behavioral compatibility change for existing unconstrained NodePools — regional nodes become possible outcomes — but the tie-break keeps zonal placement as the common case. For the full rationale, see [Option B](#option-b-explicit-placement-controls). For behavioral implications such as storage interactions, consolidation, and topology spread, see [Behavioral and Operational Considerations](#behavioral-and-operational-considerations).
```
matthchr: Are we going to have issues where we end up with an ordering that's effectively:
D16 - zonal
D16 - regional
D8 - zonal
D8 - regional

and then we hit say a capacity error on D16 zonal, so we fall back to... D16 (regional?) whereas we might otherwise have fallen back to 2x D8-zonal (one of which might work?)

designs/0009-regional-vm-provisioning.md
```diff
@@ -0,0 +1,814 @@
...
+Introduce an Azure-specific offering label such as `karpenter.azure.com/zone-placement`, and treat placement mode as an orthogonal dimension from zone topology.
+
+Example offerings:
+
+```text
+Offerings:
+  (NC24ads, eastus-1, on-demand, zone-placement=zonal)
+  (NC24ads, eastus-2, on-demand, zone-placement=zonal)
+  (NC24ads, eastus-3, on-demand, zone-placement=zonal)
+  (NC24ads, 0,        on-demand, zone-placement=regional)
```
matthchr: Something to consider - adding an option here to try has runtime impact on capacity/quota constraints. We need to be careful as some of UnavailableOfferings cache is tuned to assume certain things about errors and their impact on availability at the zone level (if you look at the code you'll see what I mean). 

If those assumptions are wrong for regional VMs, it means we need to:
1. Complicate that code further to handle those cases
2. Try more VM sizes, which could have fallback performance implications.

I don't have a solution to this "problem", but it something we should keep carefully in mind when modifying the unavailable offerings cache.

designs/0009-regional-vm-provisioning.md
```diff
@@ -0,0 +1,814 @@
...
+    Available: available,
+}
+```
+
+In `computeRequirements`, derive the instance type's `zone-placement` requirement from available offerings so NodePool requirements can filter on it. Absence of a NodePool requirement does **not** imply zonal-only behavior.
+
+In launch selection, derive both capacity type and zone from one concrete offering rather than from the cheapest capacity type plus an arbitrary zone within that capacity type. A conservative sketch is:
+
+```go
+func PickInstanceTypeAndOffering(
```
matthchr, line 462: allocation strategy rather than here? (Though the actual snippet here doesn't have any logic in it and looks cleaner than what we have now so maybe what this is actually saying is waht I am thinking that we pull the ordering logic into allocationStrategy and here we're effectively picking the 0th item always).

Can do away with this even, I think?
```
	// The cheapest offering determines both the capacity type (priority) and zone
	cheapest := best.Offerings.Cheapest()
	if cheapest == nil {
		return nil, "", ""
	}
```

Since we already ordered by price... (and other considerations)

designs/0009-regional-vm-provisioning.md
```diff
@@ -0,0 +1,814 @@
...
+case RegionalOnly:
+  markRegionalOfferingUnavailable(...)
+case RegionWideForCapacityType:
+  markZonalAndRegionalOfferingsUnavailable(...)
+}
+```
+
+The exact mapping from Azure error families to these scopes can be finalized later, but handlers that currently iterate all offerings for an instance type or capacity type need review before regional offerings are added.
+
+#### Option C: AKSNodeClass-Owned Envelope
```
matthchr: Probably clearer to just remove this section if we're not gonna do it? (I like the investigation of it in the design doc above but no point in impl notes for it if we don't wanna implement it?)

designs/0009-regional-vm-provisioning.md
```diff
@@ -0,0 +1,814 @@
...
+- **Regional-aware unavailability handling:** handlers that currently mark all offerings for an instance type or capacity type unavailable must distinguish zonal-scoped from regional-scoped failures once `zone="0"` offerings exist. Zone `"0"` works as a cache key, but the scope of which offerings get marked unavailable must change.
+
+### Design Constraints on Future Strategy
+
+- **No natural zonal preference in Azure pricing.** Any zonal-first default is an implementation choice, not a price signal from Azure. Weighted NodePools can express preference today; a future `allocationStrategy` could express configurable ranking.
+- **Consolidation and allocation strategy interaction.** Equal-price tie-breaking between zonal and regional is safe because it does not create a cost-model mismatch with core. However, a future ranking strategy that prefers different-price or non-price outcomes without reflecting that in offering requirements or price would not automatically influence consolidation decisions. This constrains future `allocationStrategy` design.
```
matthchr, line 511: Is there an upstream issue discussing this at all?

Maybe we should open one for community discussion if not?

designs/0009-regional-vm-provisioning.md
```diff
@@ -0,0 +1,814 @@
...
+- **Managed defaults:** managed products may stamp zonal-only defaults onto managed default NodePools or equivalent provider-owned envelopes, but that should be documented as a platform policy rather than implied as a provider-wide default.
+- **Conflicting controls:** if NodeClass policy and NodePool requirements narrow to an empty envelope, prefer validation or explicit status over leaving operators to infer the problem from a generic scheduling failure.
+
+## Behavioral and Operational Considerations
+
+### Schedulability
+
+This design uses two different kinds of scheduling labels, and they should not be treated as interchangeable.
+
+| Label | Present on offerings? | Present on launched nodes? | Workload-schedulable? |
```
matthchr, line 539: What does `workload-schedulable` mean? Why isn't zone-placement workload schedulable? If it's in compute requirements, it would be, wouldn't it?

matthchr: I guess I am not following this section.

designs/0009-regional-vm-provisioning.md
```diff
@@ -0,0 +1,814 @@
...
+| `karpenter.azure.com/zone-placement` | Yes, if the chosen design adds it | Only if explicit propagation is implemented | Not by default |
+
+- **`topology.kubernetes.io/zone` is fully workload-facing.** Pods may reference it in `nodeSelector`, `nodeAffinity`, and `topologySpreadConstraints`. For regional nodes, the value is `"0"`.
+- **`karpenter.azure.com/zone-placement` is NodePool/NodeClass-facing only** in the first version. Workload use should be a separate later decision.
+
+Zone `"0"` is not new to AKS — non-zonal agent pools already produce nodes with this value. Regional Karpenter nodes join an existing topology label space. What changes is the frequency of `"0"` nodes, especially for topology spread. See [Appendix: Azure Zone and Placement Reference](#appendix-azure-zone-and-placement-reference).
+
+### Observability
+
+- **Placement outcome should be visible through one stable operator-facing surface in the first version.** If `zone-placement` is not projected to nodes, document the canonical place to inspect instead, such as NodeClaims, controller events, or provider logs.
```
matthchr: minor: probably cleaner here again to stick with the option we picked, place a line in the sand and say the label is propagated to nodes, to avoid these "if x then y..." situations.

We chose an option, so we need to observe that option?

designs/0009-regional-vm-provisioning.md
```diff
@@ -0,0 +1,814 @@
...
+
+**Pros:**
+- Clean separation of concerns — zone topology stays pure, placement mode is an orthogonal dimension
+- Composes naturally with exact zone requirements
+- Lets users say `zonal only`, `regional only`, or `any` without overloading zone semantics
+- Makes NodePool intent more portable across clusters and regions by avoiding the need to name or enumerate concrete zones for common placement modes
+- Fits Karpenter's existing offering and requirement model
+
+**Cons:**
+- If this is the only control surface, it is **not strictly backward compatible**: no `zone-placement` requirement means regional is eligible. The zonal-preferred tie-break softens the common-case impact, but the widened envelope is a real behavioral compatibility change
```
matthchr: We could phrase the control-surface differently (as a boolean? with DoesNotExist as the alternative?), which means we could default a label such as `karpenter.azure.com/is-regional: true`, with default DoesNotExist if we wanted to exactly preserve the old behavior, couldn't we?

matthchr (commented):
Basically LGTM but have some questions.

designs/0009-regional-vm-provisioning.md
tallaxes: `zonal` and `regional`, and Yes answered later but adding here as well. (There is also a possibility of third value of `regional-only`, but seems like an overkill.)

designs/0009-regional-vm-provisioning.md
tallaxes: This is about not trying to replicate VMSS-specific placement behaviours (like FD spreading), rewording for clarity

designs/0009-regional-vm-provisioning.md
tallaxes: No, there isn't; AFAIK any SKU (that's not location-restricted), can be deployed regionally - and Yes covered in the appendix

designs/0009-regional-vm-provisioning.md
tallaxes: Yes; adding wording to clarify

designs/0009-regional-vm-provisioning.md
tallaxes: Adding

designs/0009-regional-vm-provisioning.md
tallaxes: Logical construct; adding clarification

designs/0009-regional-vm-provisioning.md
tallaxes: I do think it is a bit broken. From topology spread perspective, domain values should represent non-overlapping failure boundaries, and mixing zones and FDs in the same domain space breaks this assumption. The best we can do at this point though (in the presence of these cases) is to try compensate with further constratins in workload topology spread (like affinity + `nodeAffinityPolicy: Honor`, spread only across zones the pod can actually land on). There is a bit more on this in the appendix, but, if mixing is common, we should probably provide explicit guidance for workload topology spread constraints (and/or warnings).

designs/0009-regional-vm-provisioning.md
tallaxes: Yes, this talks about preservation of existing behavior - but argues that maybe we don't have to preserve it exactly. Good point on explicit use cases, adding.

designs/0009-regional-vm-provisioning.md
tallaxes: Yes unavoidable from topology spread perspective - because, without additional effort (like `nodeAffinityPolicy: Honor` etc.), it naturally considers all existing domain values - but avoidable from placement control surface (NodePool/NodeClass definition) perspective, because we have options for separating placement intent semantically into a different knob.

designs/0009-regional-vm-provisioning.md
tallaxes: Yeah, `zone-placement=regional` does feel slightly odd since regional is the absence of a zone. The alternatives have their own issues though: remember we also want to be able to express "both" (I think), and a boolean like `zone-aware=true|false` doesn't compose as cleanly with In/NotIn (e.g., `In [zonal, regional]` to express Any is natural, whereas `In [true, false]` is awkward). `placement-mode` loses the zone context - but could work? I am thinking think `zone-placement` is the best compromise - the "placement" suffix signals it's about where to place rather than which zone domain, and it works naturally in requirement expressions. But open to better ideas ...

designs/0009-regional-vm-provisioning.md
tallaxes: Yes - "zonal anywhere in the region" means "must be placed in a zone, but I don't care which one." With `zone-placement In [zonal]`, you express that without enumerating `zone In [eastus-1, eastus-2, eastus-3]`.

designs/0009-regional-vm-provisioning.md
tallaxes: I guess; though for some of the placement modes Option A is harder to make portable. E.g. for zonal only in Option A you would have to either enumerate all zones (natural, but not portable) or use `zone NotIn [0]` ...

designs/0009-regional-vm-provisioning.md
tallaxes: The difference is that the single-NodePool tie-break is price-tier local, not global envelope precedence. It prefers zonal over regional only among otherwise equally competitive offerings, so it can choose regional before exhausting all zonal options. Two weighted NodePools are stronger: they prefer any viable zonal envelope over regional, even if the surviving zonal option is a different and more expensive SKU.

designs/0009-regional-vm-provisioning.md
tallaxes: Adding callout (and maybe should stop using `Any`, which looks like a value?)

designs/0009-regional-vm-provisioning.md
tallaxes: Not 100% sure without diving into code, but worth keeping the callout I think. It is possible that once we have zone 0 offering this comes out naturally, but it is also possible that we need to carve out exception for regional if we do something like "mark all offerings as unsavable" ...

designs/0009-regional-vm-provisioning.md
tallaxes: Fair, removing

designs/0009-regional-vm-provisioning.md
tallaxes: Ack. Not trying to make this fully implementable yet, but adding a bit about allocation strategy

designs/0009-regional-vm-provisioning.md
tallaxes: Yes, though I don't see this as an issue. As long as regional is allowed, it should pick D16 regional if D16 zonal is not available. (2x D8 is a separate packing decision, right? And if D8-zonal were cheaper and still compatible with that same NodeClaim it would sort ahead of D16 ...)

designs/0009-regional-vm-provisioning.md
tallaxes: Agreed - and there is an implementation checklist callout for regional-aware unavailability handling

designs/0009-regional-vm-provisioning.md
tallaxes: I am not aware of any, and Yes good idea to open one - though maybe better to wait a bit until we have a good example of a concreted relevant allocation strategy?

designs/0009-regional-vm-provisioning.md
tallaxes: I am not sure this works ... How do you express "zonal only"? And the default - which presumably would mean "both zonal and regional" - would not preserve the old behavior exactly, since the old behavior does not include regional for zonal-capable SKUs ...

designs/0009-regional-vm-provisioning.md
tallaxes: For the "any" case, we still need to propagate the actual value explicitly, I think? We should - the current wording hedges a bit - and also tries to leave space for Option C ... I'll see if I can reword it for clarity, or maybe remove altogether ...

designs/0009-regional-vm-provisioning.md
matthchr: > so it can choose regional before exhausting all zonal options.

This would be in the case where there was a VM size that was regional only that got picked? Otherwise I am not sure I follow how we end up not exhausting all zonal options first?

designs/0009-regional-vm-provisioning.md
matthchr: Yeah especially in backticks it makes me think it's referring to some actual value

designs/0009-regional-vm-provisioning.md
matthchr: Yeah you're right I don't think this works for a couple reasons. I was thinking our default today was totally zonal but that's not quite true we support regional in a few cases so "totally zonal" as the default is not accurate. Also this doesn't easily allow zonal + regional which is something we probably want to support.

designs/0009-regional-vm-provisioning.md
tallaxes: It will exhaust all zonal options _at the same price_, but then it can try the same zonal-capable SKU as regional, before trying other more expensive zonal options. So it is not exhausting _all_ compatible zonal before trying regional. At least that how I see the suggested allocation strategy, with tie-break for the same price ...

designs/0009-regional-vm-provisioning.md
tallaxes: Yes - with a caveat that there are limits to what we can do in the absence of any label. With the NodePool-only shape, omission naturally means no additional narrowing, so the effective default is the widest envelope the provider exposes. An AKSNodeClass-owned policy does not have that constraint: it can apply a narrower compatibility envelope.

designs/0009-regional-vm-provisioning.md
tallaxes: Removed/reworded everywhere except in relation to Option C, where it is a real value https://github.com/Azure/karpenter-provider-azure/pull/1628/commits/38da81013bdff5affea864d7c84c928a419b585c

designs/0009-regional-vm-provisioning.md
tallaxes: Yes, at minimum, the ordering logic should move into allocation strategy, and we should stop deriving capacity type from Cheapest() and then picking an arbitrary zone. I’d also like to leave open whether allocation strategy should own the final concrete pick as well, versus returning an ordered candidate set and letting the caller take the first one; WDYT? Updating the text to make that boundary explicit without locking us into one implementation yet https://github.com/Azure/karpenter-provider-azure/pull/1628/commits/f7ccfba3f3175e73493661c7ce57fa002c8a5c1a

designs/0009-regional-vm-provisioning.md
matthchr: Is there a world where we could "fix" this?

Should we pursue that? For example: could we convince the CloudProvider folks to not set the zone label at all (set some other label?) in the regional case, starting in Kubernetes 1.36 or something? Obviously this is breaking but possibly we could get data on how many regional (non-zonal) VMs are actually allocated and get a sense of scope here? I would assume that there's sorta two cases:
1. User using regional scope on purpose, but likely not using topology spread or depending on the label because it doesn't make a lot of sense to do so.
2. User using only zonal - no change
3. User using a mix (on purpose) - _probably_ more disruption due to zonal spread not working correctly in these cases, so this is almost a bugfix for these users.
4. User using a mix (on accident) - If the user is using topology spread or relying on the label they're probably only doing so for the zonal VMs... again, almost a bugfix?

It sorta seems like we should have a talk with the CloudProvider folks, to me... (and get some data)

designs/0009-regional-vm-provisioning.md
matthchr: Ah yes we'll delete the nodeClaim and recreate 2x of them if we go to a D8 packing.

matthchr: You already called this out but I guess the main thing is going to be making sure that we manage the unavailable offerings cache (and future capacity API) property so we don't add too many attempts for things that won't work.

designs/0009-regional-vm-provisioning.md
matthchr: Yeah ok, ack I see your point.

designs/0009-regional-vm-provisioning.md
matthchr: AI thoughts:

Here are alternative suggestions, ordered by my recommendation:

1. placement-scope — values: zonal, regional
Best overall. Reads naturally in both directions: "the placement scope is zonal" and "the placement scope is regional." Neither value implies it's a type of zone — "scope" clearly asks "at what level is placement happening?" The concern about losing zone context (tallaxes raised this about placement-mode) is largely mitigated because the values zonal/regional are inherently about the zone dimension. Composes cleanly: In [zonal, regional] for any, In [zonal] for zone-only.

2. zone-scope — values: zonal, regional
Keeps "zone" explicitly in the label name. "Scope" frames the question as "what is the scope of zone selection?" — zonal = a specific zone, regional = the whole region. The slight awkwardness is that zone-scope=regional could read as "the zone's scope is regional" which is a bit odd, but it's better than zone-placement=regional because "scope" doesn't imply that regional is a zone.

3. zone-mode — values: zonal, regional
Very concise. "Mode" is neutral and doesn't imply regional is a type of zone — it's the mode of zone handling. zone-mode=regional reads as "the zone mode is regional," meaning "we're operating in regional mode with respect to zones." Slightly less self-documenting than placement-scope but shorter.

4. zone-binding — values: zonal, regional
Borrows from Kubernetes' PV binding concept. zone-binding=zonal means "bound to a specific zone," zone-binding=regional means "not bound to any zone." The metaphor of binding/unbound works well for conveying presence/absence. Potential risk: people might confuse it with actual PV binding semantics.

5. availability-scope — values: zonal, regional
Ties directly to Azure's "availability zone" terminology. More descriptive and Azure-native, but longer and possibly too Azure-specific for a label that's already under karpenter.azure.com/.

Why not booleans
Agreed with tallaxes' point — zone-aware=true|false or zonal=true|false are semantically clearest but In [true, false] is ugly for "any," and it loses the natural readability of In [zonal, regional].

tallaxes: I like `placement-scope` or `availability-scope` (or maybe placement/availability-domain?)

designs/0009-regional-vm-provisioning.md
tallaxes: From cloud provider "contract" perspective - Yes, I think that's possible; cloud provider can return empty zone (and the label won't be set) and can return additional labels as well (relevant node controller [code](https://github.com/kubernetes/cloud-provider/blob/1722119cb3f9a9ed49205f1dd5f95212a2576339/controllers/node/node_controller.go#L526-L549))
From backward compatibility perspective - That might be a challenge, since Yes a breaking change and there are likely a lot of existing regional nodes with that label ...

Edit: Started a conversation with cloud provider, but resolving here. Will add anything we learn later.

designs/0009-regional-vm-provisioning.md
matthchr: Yes I think placement-scope sounds good (or at least slightly better than `zone-placement`, to me?

matthchr (dismissed):
Can re-approve when we decide between amended label name: placement-scope and availability-scope

designs/0009-regional-vm-provisioning.md
tallaxes: Went with `placement-scope` (over `availability-scope`, to focus on placement and avoid appearance of making statements about availability / redundancy) https://github.com/Azure/karpenter-provider-azure/pull/1628/commits/fd6b090c810893d89fada07a488035c481c24513

matthchr (approved):

PR #1640, opened 2026-04-15
feat: handle AKS Machine API sync validation errors in offerings cache

comtalyst:
**Description**

Handle AKS Machine API sync validation errors in the offerings cache. When `BeginCreateOrUpdate` returns a validation error synchronously (before async provisioning), Karpenter now classifies it and marks the SKU unavailable — matching the behavior of the existing async provisioning error path.

Without this, sync errors were treated as opaque failures, causing retries on the same doomed SKU instead of falling back to alternatives.

Handled errors (all map to SKU-not-available cache action):
- `VMSizeNotSupported`: VM size not in SKU store for subscription/location
- `BadRequest` + "not supported for subscription": main vmsize validator variant

* New unit + acceptance tests
* Manual run w/ overridden VM size. Persistent E2E will be considered separately.
```
{
  "level": "ERROR",
  "message": "Reconciler error",
  "error": "launching nodeclaim, creating nodeclaim, creating AKS machine failed, failed to begin create AKS machine \"test-sync-err-mcs47\", handled sync error: the requested SKU is unavailable for instance type Standard_D2_v3 in zone westus3-3 with capacity type on-demand, for more details please visit: https://aka.ms/azureskunotavailable"
}
```

- [ ] Yes, issue opened: #
- [x] No

**Release Note**

```release-note
NONE
```


pkg/providers/instance/offerings/aksmachinebegincreateerrorhandlers.go
```diff
@@ -0,0 +1,105 @@
...
+	"strings"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"github.com/Azure/karpenter-provider-azure/pkg/cache"
+	"github.com/Azure/skewer"
+	corecloudprovider "sigs.k8s.io/karpenter/pkg/cloudprovider"
+)
+
+// HandlableError is a code + message pair extracted from an API error response.
+// It is intentionally minimal and API-agnostic. This module decides how to interpret it.
```
matthchr, line 32: minor: this file has a weird name. `handlableerrorhanders`

Is there a better name?

`machineerrorhandlers`? `containerservicerrorhandlers`?

pkg/providers/instance/offerings/aksmachinebegincreateerrorhandlers.go
```diff
@@ -0,0 +1,105 @@
...
+type HandlableError struct {
```
matthchr, line 33: minor, spelling (here and elsewhere). I think this should be `HandleableError`, right?

matthchr, line 33: Is the need/desire for this that we have some errors that come via the async path (not `azcore.ResponseError`), and some that come via the sync path, and we want to share error handling logic as much as we can?

Put another way: Why the need for this abstraction?

pkg/providers/instance/offerings/handlableerrorhandlers.go
```diff
@@ -0,0 +1,105 @@
...
+
+type handlableErrorHandlerEntry struct {
+	match  func(*HandlableError) bool
+	handle errorHandle
+}
+
+// HandlableErrorHandler classifies HandlableErrors and takes appropriate offerings cache actions.
+// TODO: consider replacing other error handlers with this one, which is more generic and serves the purpose. Can consider this when moving towards handling AKS errors more.
+type HandlableErrorHandler struct {
+	UnavailableOfferings *cache.UnavailableOfferings
```
matthchr: minor: Do these two fields actually need to be exported?

I observe that the existing handlers have them exported but I am not sure they needed to be?

pkg/providers/instance/offerings/handlableerrorhandlers.go
```diff
@@ -0,0 +1,105 @@
...
+	HandlerEntries       []handlableErrorHandlerEntry
+}
+
+// NewMachineBeginCreateErrorHandler creates a handler for AKS Machine API sync-phase errors.
+// TODO: consider sharing this on azure-sdk-for-go-extensions like other error handlers.
+func NewMachineBeginCreateErrorHandler(unavailableOfferings *cache.UnavailableOfferings) *HandlableErrorHandler {
```
matthchr: minor: mismatch in create function here and return type name? Why not either call `HandlableErrorHandler` `MachineBeginCreateErrorHandler` or call this `NewHandlableErrorHandler`?

Note: I tend to prefer `MachineBeginCreateErrorHandler` (this is just for that, right?)

pkg/providers/instance/aksmachineinstance.go
```diff
@@ -573,6 +579,20 @@ func (p *DefaultAKSMachineProvider) handleMachineProvisioningError(ctx context.C
 	return fmt.Errorf("failed to create AKS machine %q during %s, unhandled provisioning error: code=%s, message=%s", aksMachineName, phase, lo.FromPtr(innerError.Code), lo.FromPtr(innerError.Message))
 }
 
+func (p *DefaultAKSMachineProvider) handleMachineBeginCreateError(ctx context.Context, aksMachineName string, instanceType *corecloudprovider.InstanceType, zone string, capacityType string, he *offerings.HandlableError) error {
+	sku, skuErr := p.instanceTypeProvider.Get(ctx, instanceType.Name)
+	if skuErr != nil {
+		return fmt.Errorf("failed to get instance type %q: %w, begin create error left unhandled: %w", instanceType.Name, skuErr, he)
+	}
+
+	handledError := p.beginCreateErrorHandling.Handle(ctx, sku, instanceType, zone, capacityType, he)
```
matthchr: minor, style: 
```suggestion
	err := p.beginCreateErrorHandling.Handle(ctx, sku, instanceType, zone, capacityType, he)
```

I think this is just a plain `err`, right?

pkg/providers/instance/offerings/handlableerrorhandlers.go
```diff
@@ -0,0 +1,105 @@
...
+	return &HandlableErrorHandler{
+		UnavailableOfferings: unavailableOfferings,
+		HandlerEntries: []handlableErrorHandlerEntry{
+			{
+				match:  isSKUNotAvailableForSubscription,
+				handle: handleSKUNotAvailableError,
+			},
+			{
+				match:  isSKUNotAvailableForSubscriptionBadRequest,
+				handle: handleSKUNotAvailableError,
```
matthchr: This handler only marks a single capacity type (spot | ondemand) as unavailable, but I think both of these errors we may want to mark both?

There is another helper, `markAllZonesUnavailableForBothCapacityTypes`, used in `handleAllocationFailureError`, which marks both?

Can you confirm which is intended/ideal?

pkg/providers/instance/offerings/aksmachinebegincreateerrorhandlers.go
comtalyst: - The format of error returned by the batch API cannot be converted to SDK error easily. It returns multiple errors embedded in the main error. So we better off with a handler that support this format.
- Error handling logic don't really need anything beyond `code` and `message`. Given multiple forms of error we are having (e.g., SDK error, machine async ProvisioningError, batch error), this `HandlableError` is what every type can be abstracted to. As a result, yes, we can share logic as much as we can.
- If we want to separate the expected errors and the unexpected errors. Before this, we are leaning towards using `error` type as a vehicle for both. But I would argue that `error` type should be reserved for real, unexpected errors. These handlable errors are not always the case, as Karpenter may expect to handle it.

matthchr: > The format of error returned by the batch API cannot be converted to SDK error easily. It returns multiple errors embedded in the main error. So we better off with a handler that support this format.

Not 100% sure I am following. Isn't the usage/conversion here just:
```
		he := offerings.ErrorToHandlableError(err)
		if he != nil {
			return nil, p.handleMachineBeginCreateError(ctx, aksMachineName, instanceType, zone, capacityType, he)
		}
```

which looks to me like we don't do anything fancy processing? Or are you talking about the async path (which AFAIK doesn't take this codepath today?)

> Error handling logic don't really need anything beyond code and message. Given multiple forms of error we are having (e.g., SDK error, machine async ProvisioningError, batch error), this HandlableError is what every type can be abstracted to. As a result, yes, we can share logic as much as we can.

This is a goal of yours and not in this PR, right? Since as far as I can tell here this error handler is specifically for machine sync errors, so there's not anything shared currently, right?

> If we want to separate the expected errors and the unexpected errors. Before this, we are leaning towards using error type as a vehicle for both. But I would argue that error type should be reserved for real, unexpected errors. These handlable errors are not always the case, as Karpenter may expect to handle it.

I don't think I agree with this. There are already tons of `errors` managed by Karpenter that may be retryable - for example look at the SDKs, we get errors for things like HTTP 500 or connection failures, even though the appropriate response to those errors is usually to retry.

comtalyst: <blockquote>
Not 100% sure I am following. Isn't the usage/conversion here just:

```
		he := offerings.ErrorToHandlableError(err)
		if he != nil {
			return nil, p.handleMachineBeginCreateError(ctx, aksMachineName, instanceType, zone, capacityType, he)
		}
```

which looks to me like we don't do anything fancy processing? Or are you talking about the async path (which AFAIK doesn't take this codepath today?)
</blockquote>

I am talking batch API, sync path. Not here, but in batch PR.
Code: https://github.com/Azure/karpenter-provider-azure/blob/b79933e69708781f471181f3b3231c2fefe2336c/pkg/providers/azclient/aksmachinesheaderbatch/apiresponseutils.go
API expectations: https://github.com/Azure/karpenter-provider-azure/blob/b79933e69708781f471181f3b3231c2fefe2336c/designs/0009-aks-machines-batch-creation.md#error-handling

Let's say, alternatively, we force convert these to `error` type, and extend/reuse existing [ResponseErrorHandler](https://github.com/Azure/karpenter-provider-azure/blob/main/pkg/providers/instance/offerings/resperrorhandlers.go) ([matcher in azure-sdk-for-go-extensions](https://github.com/Azure/azure-sdk-for-go-extensions/blob/main/pkg/errors/armerrors.go))
- It expects: the error being SDK's `ResponseError` type underneath
- So, for our batch error case to work, converting to `error` is not enough. It has to be `ResponseError`. Not only SDK didn't really provide a good tool for that, it does not really make sense because these batch errors are not `ResponseError`. 
- Conversion would be best-effort, in which in the end, we can, but it adds to cross-module assumption that is not compile time
- However, the role of `ResponseError` type in error handling is not as important. We only need code and message. This goes to my point on vision for shared error handler, which can still divide by what code/message format it is responsible for, but not error type.

Another alternative: if we extend/reuse existing [ErrorDetailHandler](https://github.com/Azure/karpenter-provider-azure/blob/main/pkg/providers/instance/offerings/errordetailhandlers.go)
- Its interface accepts `ErrorDetail` rather than `error` then reflect later. So, ideally, we convert batch error to `ErrorDetail`
- The type, however, is more fair. `ErrorDetail` is AKS type, similar to extracted batch error.
- However, it is not guaranteed that the API error we get is AKS error. Things can fail before it reaches AKS RP (e.g., NotFound), and the error returned won't be `ErrorDetail` type
- Same arguments with `ResponseErrorHandler` on best-effort conversion applies.

Another alternative: we still create a new handler, just not with `HandlableError`.
- `error` type doesn't allow the separation of code and message easily. Ideally, we want to move to rely more on whole error code than error message (or its substring). AKS errors seem to support that so far(?), once we do.
- `ResponseError` doesn't make sense, per arguments against `ResponseErrorHandler`
- `ErrorDetail` neglects ARM errors, per arguments against `ErrorDetailHandler`


<blockquote>
This is a goal of yours and not in this PR, right? Since as far as I can tell here this error handler is specifically for machine sync errors, so there's not anything shared currently, right?
</blockquote>

Yes, not in this PR. There is a TODO though:

```go
// TODO: consider replacing other error handlers with this one, which is more generic and serves the purpose. Can consider this when moving towards handling AKS errors more.
```

<blockquote>
I don't think I agree with this. There are already tons of errors managed by Karpenter that may be retryable - for example look at the SDKs, we get errors for things like HTTP 500 or connection failures, even though the appropriate response to those errors is usually to retry.
</blockquote>

The way to separate expected/unexpected is based on the prerequisite that all services are up and performing correctly per contract. 500 and connection failures are in unexpected category. However, client layer doesn't need to act fatal on every unexpected errors they see. Retrying from these as a result is appropriate.

Quota/capacity errors, on the other hand, is utilized by Karpenter by-design. Karpenter fully expect these errors, and use it as a part of its functionality to filter out unavailable offerings, then keep going. I would argue that these are the expected flow of things rather than failure modes.

By the time we have `HandlableError`, even though Karpenter expects, it is not necessary that the thing inside is utilizable/expected. Internal server error could still be there, but the error handlers will decide the final category. But I don't think that we should declare that it is erroneous until proven so.

I acknowledge that the framework outside of our codebase this is not perfectly suitable for this model. For example, Karpenter core still reads error type returned by `Create()` to determine its nature. At our layer, however, this could be an opportunity to clarify. Thoughts?

matthchr: I mean I think it's worth separating two topics here:
1. Does `HandlableError` make sense here, given our unique requirements and the fact that some errors are not coming from an `error` today (they come from GET on the machine). I think this is a somewhat special case, and the answer to this question probably is yes it makes sense here given our requirements.
2. _In general_, should we avoid `error` for retryable cases. IMO no. We should do what Go wants us to do and what matches most with the ecosystem that we're in, and IMO that is that `error` can be fatal or non-fatal, there is no implication that `error` is always something fatal and I don't think we should try to make our code (in general) follow that pattern, even if in this particular instance primarily due to re-suing `error` and non-error paths, it makes sense for this specific case. It is not a general pattern that is correct AFAIK and we should continue (elsewhere) to treat errors as possibly fatal, or possibly retryable, depdending on what they are.

pkg/providers/instance/offerings/handlableerrorhandlers.go
comtalyst: Marking both makes more sense, given the assumption that SKU not available applies to both. Updated.

pkg/providers/instance/offerings/handlableerrorhandlers.go
comtalyst: (renamed)

pkg/providers/instance/offerings/handlableerrorhandlers.go
comtalyst: They don't need to be. Unexported.

pkg/providers/instance/offerings/aksmachinebegincreateerrorhandlers.go
comtalyst: (renamed)

pkg/providers/instance/offerings/aksmachinebegincreateerrorhandlers.go
comtalyst: So far, we have been naming these files "<error type> + handlers" rather than where the error comes from. Which means CRP sync/async shares the same file. Machine API async is on its own, and the only one that uses `ErrorDetail`, so `ErrorDetailHandler`.
In longer term, for clarity, I think we might better shift towards categorizing it by source rather than type.

But for this case, I think it is fine to get started here, naming it `AKSMachineErrorHandlers`

pkg/providers/instance/offerings/aksmachinebegincreateerrorhandlers.go
```diff
@@ -52,54 +52,74 @@ func ErrorToHandlableError(err error) *HandlableError {
...
+	ctx context.Context,
+	unavailableOfferings *cache.UnavailableOfferings,
+	sku *skewer.SKU,
+	instanceType *corecloudprovider.InstanceType,
+	zone,
+	capacityType,
+	errorCode,
+	errorMessage string,
+) error {
+	markAllZonesUnavailableForBothCapacityTypes(ctx, unavailableOfferings, sku, instanceType, SKUNotAvailableReason, SKUNotAvailableOnDemandTTL)
```
matthchr, line 107: @tallaxes be aware of this in your outstanding PR as this will likely merge first, and you'll want to correct this handling for machine API at least in your PR on merge.

matthchr (approved):

pkg/providers/instance/offerings/aksmachinebegincreateerrorhandlers.go
comtalyst: On (2), matching the ecosystem is fair. 
Overall, let's revisit this when we work on AKS errors

PR #1641, opened 2026-04-15
feat: set Spotmaxprice for Spot VMs

wdarko1:
Fixes #1631 

**Description**

Adds a user-configurable spotMaxPrice field to AKSNodeClass for controlling the max hourly USD price for spot VMs. If prices exceed this hourly rate, spot VMs are voluntarily disrupted. [Spot Max Price documentation](https://learn.microsoft.com/azure/aks/spot-node-pool#max-price-for-a-spot-pool)

Valid values are any positive decimal (up to 5 decimal places) or -1 (no price-based eviction, default).

## API
- Added `SpotMaxPrice *float64` to `AKSNodeClassSpec` in both `v1beta1` and `v1alpha2`
- Default: `-1` via `+kubebuilder:default=-1`
- CEL validation: `spotMaxPrice == -1.0 || spotMaxPrice > 0.0` — rejects `0` and any negative other than `-1`
- Field is silently ignored for on-demand capacity type
- Bumped `AKSNodeClassHashVersion` v3 → v4 (new field has non-zero default; triggers re-reconcile of existing NodeClasses)

```yaml
spec:
  spotMaxPrice: "0.98765"   # max $0.98765/hr; use "-1" to disable price-based eviction
```

## Fixed-point helper
New `ParseSpotMaxPrice` / `SpotMaxPriceFixed()` in `pkg/apis/v1beta1/spotmaxprice.go` (mirrored in `v1alpha2`) parses the string to `int64` (USD × 100000) using string splitting + `strconv` — no `float`, `big.Float`, or decimal libs:

```
nil       → nil
"-1"      → -1     (sentinel: no price-based eviction)
"2"       → 200000
"0.98765" → 98765
"0"       → error  (must be > 0)
"1.123456"→ error  (> 5 decimal places)
```

## Provisioning
- **VM path** (`vminstance.go`): `setVMPropertiesBillingProfile` now reads `nodeClass.Spec.SpotMaxPrice`, falls back to `-1` if nil
- **AKS Machine API path** (`aksmachineinstancehelpers.go`): New `configureSpotBilling` helper sets `MachineBillingProfile.SpotMaxPrice` (`float64` → `float32` conversion at call site) for Spot; returns `nil` for on-demand
- `setVMPropertiesBillingProfile` and `configureSpotBilling` convert the validated string to float only at the Azure API boundary (`strconv.ParseFloat`), which is safe since the CRD pattern has already validated the format.

## Generated / CRD
- Updated `zz_generated.deepcopy.go` in both API versions
- Updated `karpenter.azure.com_aksnodeclasses.yaml` with the new field schema and CEL validation rule in both version schemas

* Unit tests for `setVMPropertiesBillingProfile` and `configureSpotBilling` covering nil, -1, positive values, and on-demand no-op
* CEL validation table tests: valid (-1, 0.5, 0.98765, 100.0) and invalid (0, -0.5, -2)
Hash test entries for SpotMaxPrice mutation; currentHashVersion updated to "v5"

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
Spot Max Price can now be set to allow users to set a maximum allowed spot VM hourly price, and evict spot nodes once they exceed that value. [Spot Max Price documentation](https://learn.microsoft.com/azure/aks/spot-node-pool#max-price-for-a-spot-pool)
```


tallaxes (commented):
Some thoughts (focusing on the intent/design for starters):
* It would be better (in Kubernetes API) to represent "no limit" by omitting the setting (making field optional), rather than -1. (It would also simplify constraints/parsing, make it backward compatible, and likely get rid of hash version bump ...)
* Should we filter out the offerings that are known to be more expensive than max price, i.e. not even try provisioning them? (I don't think there is an obvious answer here, given the cached price may not be authoritative ...)
* This introduces new possible provisioning failure reasons, that need to be handled. Docs say `MaxPriceValueInvalid` (which we can probably make sure does not happen) and `SpotPriceGreaterThanProvidedMaxPrice` - which we need to think carefully about how to handle. (Would want to avoid trying and failing to provision the same spot SKU again and again ... May want to stop considering temporarily - but only for relevant AKSNodeClasses?)
* Would likely need both unit/acceptance tests for this (including simulated failures), and real E2E tests

PR #1645, opened 2026-04-24
feat: add support for kubernetes.azure.com/priority label

matthchr:
This should be preferred over the scalesetpriority variant

**Description**

* Unit tests
* E2E tests: https://github.com/Azure/karpenter-provider-azure/actions/runs/24908854155 - I believe all 3 of these flakes are known issues

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


tallaxes (approved):
LGTM. Something to consider for later (nothing blocking):

- Let's maybe update the 0006-requirements-and-labels.md with the new label? (And let's not forget the user-facing docs.)
- Do we need guidance on that `capacity-type`, `scalesetpriority`, and `priority` should not be mixed (unless the values agree) - or is it obvious? And maybe a preference? (`capacity-type` as canonical in Karpenter, unless one of the others needed for compatibility?)
- Wondering if we should look into surfacing “no instance types available” better (maybe with reason?) - as we introduce (more?) ways to have conflicting requirements ...

comtalyst (approved):

matthchr:
> Wondering if we should look into surfacing “no instance types available” better (maybe with reason?) - as we introduce (more?) ways to have conflicting requirements ...

Yeah this would be good to do for sure. I've had issues in the past where we ask for conflicting requirements and it's not that easy to figure out. I am not sure that this is really going to increase that chance much. There are lots of other ways we already have that result in this situation, but filed an issue for this here: https://github.com/Azure/karpenter-provider-azure/issues/1657

> Do we need guidance on that capacity-type, scalesetpriority, and priority should not be mixed (unless the values agree) - or is it obvious? And maybe a preference? (capacity-type as canonical in Karpenter, unless one of the others needed for compatibility?)

I will add it in the public docs.

> Let's maybe update the 0006-requirements-and-labels.md with the new label? (And let's not forget the user-facing docs.)

Sent a PR for this

PR #1659, opened 2026-04-28
feat: support regional (non-zonal) VM provisioning

### LINKED ISSUE #1002: Run VMs without redundancy options
### Version

Hi guys,

Since 1.4.0 `topology.kubernetes.io/zone` can be used.
Can somebody help me to understand how can I use `topology.kubernetes.io/*` requirements to create VMs with NO availability options ("No infrastructure redundancy required" as per portal saying)?
In language of default AKS autoscaler through scalesets with no availability zone these labels are set as following:
```
metadata:
  labels:
    ...
    topology.kubernetes.io/region: northeurope
    topology.kubernetes.io/zone: '0'
    ...
```
But trying to set this requirements like that fails to provision the node further.

My setup:
k8s version: 1.31.2
karpenter (both crd chart and controller chart: 1.5.0

Thanks in advance!

### Expected Behavior

Karpenter can create VMs without redundancy options (no availability zone)

### Actual Behavior

Either it doesn't work or created VM uses the availability zone.

Using karpenter >1.4.0 try to create VMs without redundancy options.

### Resource Specs and Logs

NodePool spec:
```
    spec:
      expireAfter: 24h
      nodeClassRef:
        group: karpenter.azure.com
        kind: AKSNodeClass
        name: default-karpenter-nodeclass
      requirements:
        - key: kubernetes.io/arch
          operator: In
          values:
            - amd64
        - key: kubernetes.io/os
          operator: In
          values:
            - linux
          operator: In
          values:
            - on-demand
        - key: karpenter.azure.com/sku-name
          operator: In
          values:
            - Standard_D64s_v3
            - Standard_D48s_v3
            - Standard_D32s_v3
            - Standard_D16s_v3
            - Standard_D8s_v3
            - Standard_D4s_v3
        - key: gpupool
          operator: In
          values:
            - workloads
        - key: topology.kubernetes.io/zone
          operator: In
          values:
            - '0'
        - key: topology.kubernetes.io/region
          operator: In
          values:
            - 'northeurope'
```
AKSNodeClass spec:
```
spec:
  imageFamily: Ubuntu2204
  osDiskSizeGB: 100
```
Error I saw:
`Cannot disrupt NodeClaim: not all pods would schedule, karpenter/inflate-f77fb7b66-zld2r => incompatible with nodepool "workload-karpenter-nodepool", daemonset overhead={"cpu":"915m","memory":"2038922537","pods":"16"}, no instance type satisfied resources {"cpu":"1915m","memory":"10628857129","pods":"17"} and requirements gpupool In [workloads], karpenter.azure.com/sku-name In [Standard_D16s_v3 Standard_D32s_v3 Standard_D48s_v3 Standard_D4s_v3 Standard_D64s_v3 and 1 others], karpenter.sh/capacity-type In [on-demand], karpenter.sh/nodepool In [workload-karpenter-nodepool], kubernetes.io/arch In [amd64], kubernetes.io/os In [linux], node-restriction.aurora.equinor.com/purpose In [workloads], topology.kubernetes.io/region In [northeurope], topology.kubernetes.io/zone In [0] (no instance type met the scheduling requirements or had a required offering); incompatible with nodepool "spotworkload-karpenter-nodepool", daemonset overhead={"cpu":"915m","memory":"2038922537","pods":"16"}, no instance type satisfied resources {"cpu":"1915m","memory":"10628857129","pods":"17"} and requirements gpupool In [workloads], karpenter.azure.com/sku-name In [Standard_D16s_v3 Standard_D32s_v3 Standard_D48s_v3 Standard_D4s_v3 Standard_D64s_v3 and 1 others], karpenter.sh/capacity-type In [spot], karpenter.sh/nodepool In [spotworkload-karpenter-nodepool], kubernetes.io/arch In [amd64], kubernetes.io/os In [linux], node-restriction.aurora.equinor.com/purpose In [workloads], topology.kubernetes.io/region In [northeurope], topology.kubernetes.io/zone In [0] (no instance type met the scheduling requirements or had a required offering) karpenter/inflate-f77fb7b66-6mwr4 => incompatible with nodepool "workload-karpenter-nodepool", daemonset overhead={"cpu":"915m","memory":"2038922537","pods":"16"}, no instance type satisfied resources {"cpu":"1915m","memory":"10628857129","pods":"17"} and requirements gpupool In [workloads], karpenter.azure.com/sku-name In [Standard_D16s_v3 Standard_D32s_v3 Standard_D48s_v3 Standard_D4s_v3 Standard_D64s_v3 and 1 others], karpenter.sh/capacity-type In [on-demand], karpenter.sh/nodepool In [workload-karpenter-nodepool], kubernetes.io/arch In [amd64], kubernetes.io/os In [linux], node-restriction.aurora.equinor.com/purpose In [workloads], topology.kubernetes.io/region In [northeurope], topology.kubernetes.io/zone In [0] (no instance type met the scheduling requirements or had a required offering); incompatible with nodepool "spotworkload-karpenter-nodepool", daemonset overhead={"cpu":"915m","memory":"2038922537","pods":"16"}, no instance type satisfied resources {"cpu":"1915m","memory":"10628857129","pods":"17"} and requirements gpupool In [workloads], karpenter.azure.com/sku-name In [Standard_D16s_v3 Standard_D32s_v3 Standard_D48s_v3 Standard_D4s_v3 Standard_D64s_v3 and 1 others], karpenter.sh/capacity-type In [spot], karpenter.sh/nodepool In [spotworkload-karpenter-nodepool], kubernetes.io/arch In [amd64], kubernetes.io/os In [linux], node-restriction.aurora.equinor.com/purpose In [workloads], topology.kubernetes.io/region In [northeurope], topology.kubernetes.io/zone In [0] (no instance type met the scheduling requirements or had a required offering) karpenter/inflate-f77fb7b66-v9cmm => incompatible with nodepool "workload-karpenter-nodepool", daemonset overhead={"cpu":"915m","memory":"2038922537","pods":"16"}, no instance type satisfied resources {"cpu":"1915m","memory":"10628857129","pods":"17"} and requirements gpupool In [workloads], karpenter.azure.com/sku-name In [Standard_D16s_v3 Standard_D32s_v3 Standard_D48s_v3 Standard_D4s_v3 Standard_D64s_v3 and 1 others], karpenter.sh/capacity-type In [on-demand], karpenter.sh/nodepool In [workload-karpenter-nodepool], kubernetes.io/arch In [amd64], kubernetes.io/os In [linux], node-restriction.aurora.equinor.com/purpose In [workloads], topology.kubernetes.io/region In [northeurope], topology.kubernetes.io/zone In [0] (no instance type met the scheduling requirements or had a required offering); incompatible with nodepool "spotworkload-karpenter-nodepool", daemonset overhead={"cpu":"915m","memory":"2038922537","pods":"16"}, no instance type satisfied resources {"cpu":"1915m","memory":"10628857129","pods":"17"} and requirements gpupool In [workloads], karpenter.azure.com/sku-name In [Standard_D16s_v3 Standard_D32s_v3 Standard_D48s_v3 Standard_D4s_v3 Standard_D64s_v3 and 1 others], karpenter.sh/capacity-type In [spot], karpenter.sh/nodepool In [spotworkload-karpenter-nodepool], kubernetes.io/arch In [amd64], kubernetes.io/os In [linux], node-restriction.aurora.equinor.com/purpose In [workloads], topology.kubernetes.io/region In [northeurope], topology.kubernetes.io/zone In [0] (no instance type met the scheduling requirements or had a required offering) karpenter/inflate-f77fb7b66-rcpcg => incompatible with nodepool "workload-karpenter-nodepool", daemonset overhead={"cpu":"915m","memory":"2038922537","pods":"16"}, no instance type satisfied resources {"cpu":"1915m","memory":"10628857129","pods":"17"} and requirements gpupool In [workloads], karpenter.azure.com/sku-name In [Standard_D16s_v3 Standard_D32s_v3 Standard_D48s_v3 Standard_D4s_v3 Standard_D64s_v3 and 1 others], karpenter.sh/capacity-type In [on-demand], karpenter.sh/nodepool In [workload-karpenter-nodepool], kubernetes.io/arch In [amd64], kubernetes.io/os In [linux], node-restriction.aurora.equinor.com/purpose In [workloads], topology.kubernetes.io/region In [northeurope], topology.kubernetes.io/zone In [0] (no instance type met the scheduling requirements or had a required offering); incompatible with nodepool "spotworkload-karpenter-nodepool", daemonset overhead={"cpu":"915m","memory":"2038922537","pods":"16"}, no instance type satisfied resources {"cpu":"1915m","memory":"10628857129","pods":"17"} and requirements gpupool In [workloads], karpenter.azure.com/sku-name In [Standard_D16s_v3 Standard_D32s_v3 Standard_D48s_v3 Standard_D4s_v3 Standard_D64s_v3 and 1 others], karpenter.sh/capacity-type In [spot], karpenter.sh/nodepool In [spotworkload-karpenter-nodepool], kubernetes.io/arch In [amd64], kubernetes.io/os In [linux], node-restriction.aurora.equinor.com/purpose In [workloads], topology.kubernetes.io/region In [northeurope], topology.kubernetes.io/zone In [0] (no instance type met the scheduling requirements or had a required offering) karpenter/inflate-f77fb7b66-2gh7z => incompatible with nodepool "workload-karpenter-nodepool", daemonset overhead={"cpu":"915m","memory":"2038922537","pods":"16"}, no instance type satisfied resources {"cpu":"1915m","memory":"10628857129","pods":"17"} and requirements gpupool In [workloads], karpenter.azure.com/sku-name In [Standard_D16s_v3 Standard_D32s_v3 Standard_D48s_v3 Standard_D4s_v3 Standard_D64s_v3 and 1 others], karpenter.sh/capacity-type In [on-demand], karpenter.sh/nodepool In [workload-karpenter-nodepool], kubernetes.io/arch In [amd64], kubernetes.io/os In [linux], node-restriction.aurora.equinor.com/purpose In [workloads], topology.kubernetes.io/region In [northeurope], topology.kubernetes.io/zone In [0] (no instance type met the scheduling requirements or had a required offering); incompatible with nodepool "spotworkload-karpenter-nodepool", daemonset overhead={"cpu":"915m","memory":"2038922537","pods":"16"}, no instance type satisfied resources {"cpu":"1915m","memory":"10628857129","pods":"17"} and requirements gpupool In [workloads], karpenter.azure.com/sku-name In [Standard_D16s_v3 Standard_D32s_v3 Standard_D48s_v3 Standard_D4s_v3 Standard_D64s_v3 and 1 others], karpenter.sh/capacity-type In [spot], karpenter.sh/nodepool In [spotworkload-karpenter-nodepool], kubernetes.io/arch In [amd64], kubernetes.io/os In [linux], node-restriction.aurora.equinor.com/purpose In [workloads], topology.kubernetes.io/region In [northeurope], topology.kubernetes.io/zone In [0] (no instance type met the scheduling requirements or had a required offering) and 0 other(s)`

Disregard the similarity of the errors, I had almost identical nodepool for spot VMs.

The only correct option that worked for me now didn't solve my task (omitting these reqs also works):
```
    - key: topology.kubernetes.io/zone
      operator: In
      values:
        - northeurope-1
    - key: topology.kubernetes.io/region
      operator: In
      values:
        - northeurope
```

### Community Note

tallaxes:
Fixes #1002

**Description**

Adds the `karpenter.azure.com/placement-scope` offering label (values `zonal`, `regional`) and widens the default placement envelope to include regional offerings for zone-capable SKUs. NodePools can opt into either placement explicitly via this label or via the `topology.kubernetes.io/zone In ["0"]` shorthand. Unconstrained NodePools now have regional offerings as a fallback when zonal offerings are exhausted, addressing the long-standing zonal-exhaustion gap (Azure/AKS#5574, KAITO).

Key implementation points:

- Offering generation in `instancetypes.go` emits a `zone="0"` regional offering alongside zonal offerings for zone-capable SKUs; regional-only SKUs are unchanged.
- Offering selection moved into `allocationstrategy.Provider.Allocate`, which returns a `*Selection` wrapping the chosen `*InstanceType` and `*Offering`. The launch helpers consume the `Selection` directly instead of re-deriving capacity type and zone.
- Equal-price ranking in `allocationstrategy/stages/offering_rank.go` is price → capacity type (spot before on-demand) → placement scope (zonal before regional). Zone-only ties are randomized before stable sort to avoid concentrating launches in one zone. The visible consequence at equal price is that a regional spot offering can rank ahead of a zonal on-demand offering, matching Karpenter core's spot-first convention.
- `UnavailableOfferings` handlers are placement-scope aware: zonal-scoped failures (`AllocationFailed`, `OverconstrainedAllocationFailed`, `SkuNotAvailable`) only block offerings within the attempted placement scope, leaving regional fallback intact, and vice versa.
- The `placement-scope` label is propagated to launched nodes via both the VM and AKS Machine API paths.

Behavioral compatibility note: existing unconstrained NodePools may now receive regional nodes when zonal capacity is exhausted. The capacity-first zonal-preferred tie-break keeps the common case zonal-first.

Design doc: [`designs/0009-regional-vm-provisioning.md`](designs/0009-regional-vm-provisioning.md)

* Unit tests across:
  - Offering generation and `placement-scope` requirement projection (`pkg/providers/instancetype/suite_test.go`)
  - Allocation strategy ranking (price/capacity/scope/zone-shuffle, including `TestFilterInstanceOfferings_SpotRegionalOfferingBeforeOnDemandZonalAtSamePrice`, `TestFilterInstanceOfferings_ZonalOfferingsBeforeRegionalAtSamePriceAndCapacityType`, `TestFilterInstanceOfferings_ZoneTiesAreShuffled`, and `TestAllocate`)
  - Placement-scoped unavailability handlers (`pkg/providers/instance/offerings/commonerrorhandlers_test.go`)
  - `vmInstanceToNodeClaim` placement-scope projection (`pkg/cloudprovider/cloudprovider_test.go`)
  - Zone helpers and `PlacementScopeForOffering` (`pkg/utils/zones/zones_test.go`)
  - AKS Machine API parity tests (`pkg/cloudprovider/suite_aksmachineapi_offerings_test.go`)
* One end-to-end integration test exercising explicit regional placement (`test/suites/integration/zone_constraints_test.go`).
* E2E: 
  * https://github.com/Azure/karpenter-provider-azure/actions/runs/25031099853
  * https://github.com/Azure/karpenter-provider-azure/actions/runs/25305955581 - Machine mode

- [ ] Yes, issue opened: #
- [ ] No

(README adds a behavioral-change paragraph and an example NodePool requirement; design doc `designs/0009-regional-vm-provisioning.md` is included.)

**Release Note**

```release-note
Adds support for provisioning regional (non-zonal) VMs for zone-capable SKUs via the new karpenter.azure.com/placement-scope requirement (values: `zonal`, `regional`). Unconstrained NodePools may now receive regional nodes when zonal capacity is exhausted; see 0009-regional-vm-provisioning.md for the placement and ranking semantics.
```


matthchr:
Do we want to have a discussion with the cloudprovider folks before we merge this, on the `0` behavior?

I am guessing that whatever changes/fixes they may it won't happen for a while and we want this as-is anyway, but may be useful to know where we're going regardless?

examples/v1/general-purpose.yaml
```diff
@@ -43,6 +43,11 @@ spec:
       - key: karpenter.azure.com/sku-family
         operator: In
         values: [D]
+      # Optional: Uncomment if you want to constrain Azure placement to zonal or regional capacity.
+      #           Regional capacity is labeled with topology.kubernetes.io/zone=0.
+      # - key: karpenter.azure.com/placement-scope
+      #   operator: In
+      #   values: ["regional"]
```
matthchr: minor: I _assume_ the more likely preference would be `zonal` rather than regional? Do we want to use that as the example?

pkg/cloudprovider/suite_aksmachineapi_offerings_test.go
matthchr, line 158: Says zone here and in assertion but test name says offering? Seems wrong?

pkg/cloudprovider/suite_aksmachineapi_offerings_test.go
```diff
@@ -160,8 +160,8 @@ var _ = Describe("CloudProvider", func() {
 				Expect(node.Labels[v1.LabelTopologyZone]).ToNot(Equal(initialZone))
 			})
 
-			// Ported from VM test: "should fail to provision when OverconstrainedAllocation errors are hit, then switch capacity type and succeed"
-			It("should fail to provision when OverconstrainedAllocation errors are hit, then switch capacity type and succeed", func() {
+			// Ported from VM test: "should fail to provision when OverconstrainedAllocation errors are hit, then switch offering and succeed"
+			It("should fail to provision when OverconstrainedAllocation errors are hit, then switch offering and succeed", func() {
```
matthchr: same q here - swapped title to offering but we _are_ swapping capacity type here, why the change?

pkg/cloudprovider/suite_aksmachineapi_offerings_test.go
```diff
@@ -183,20 +183,25 @@ var _ = Describe("CloudProvider", func() {
...
-				Expect(node.Labels[karpv1.CapacityTypeLabelKey]).To(Equal(karpv1.CapacityTypeOnDemand))
+				Expect(node.Labels[v1.LabelInstanceTypeStable] == vmSize &&
+					node.Labels[karpv1.CapacityTypeLabelKey] == initialCapacityType &&
+					node.Labels[v1.LabelTopologyZone] == zone).To(BeFalse())
 			})
 
-			// Ported from VM test: "should fail to provision when AllocationFailure errors are hit, then switch VM size and succeed"
-			It("should fail to provision when AllocationFailure errors are hit, then switch VM size and succeed", func() {
+			// Ported from VM test: "should fail to provision when AllocationFailure errors are hit, then switch placement and succeed"
+			It("should fail to provision when AllocationFailure errors are hit, then switch placement and succeed", func() {
```
matthchr, line 210: Just making sure we kept a test where we swap VM size because that whole family doesn't work. Where is that test?

pkg/providers/allocationstrategy/stages/instance_offering.go
```diff
@@ -33,5 +33,9 @@ type InstanceOffering struct {
 }
 
 type Stage interface {
+	// Process applies the stage's transformation to the provided slice. Stages
```
matthchr, line 36: Thanks, this is very useful

pkg/providers/allocationstrategy/allocation_strategy.go
```diff
@@ -19,12 +19,29 @@ package allocationstrategy
...
+	// Allocate selects a single instance type and offering for a NodeClaim.
+	// Returns nil when no compatible offering is available.
+	//
+	// This interface models client-side allocation: the provider chooses one
+	// concrete offering and the caller is responsible for provisioning. It
+	// accommodates future client-side strategies that consult external Azure
+	// APIs (e.g. placement-score, capacity advice) to inform the decision,
+	// since the contract here is just "given candidates, return a choice."
+	//
+	// It does NOT accommodate future server-side APIs that combine decision
```
matthchr, line 40: This is what I sorta liked about the `[]InstanceOffering` return -- it separated scoring and allocation so if there was a provider like Fleet we could pass the first N into that API and let it "finish picking".

Maybe premature though as we definitely don't need it now

README.md
```diff
@@ -258,6 +258,11 @@ spec:
         - key: karpenter.azure.com/sku-family
           operator: In
           values: [D]
+        # Optional: constrain Azure placement to zonal or regional capacity.
+        # Regional capacity is labeled with topology.kubernetes.io/zone=0.
+        # - key: karpenter.azure.com/placement-scope
+        #   operator: In
+        #   values: ["regional"]
```
matthchr: Same q here about showcasing lock to zonal (old behavior) rather than lock to regional (rare?)

pkg/providers/allocationstrategy/stages/offering_rank_test.go
```diff
@@ -0,0 +1,49 @@
...
+import (
+	"testing"
+
+	corev1 "k8s.io/api/core/v1"
+	karpv1 "sigs.k8s.io/karpenter/pkg/apis/v1"
+	corecloudprovider "sigs.k8s.io/karpenter/pkg/cloudprovider"
+	"sigs.k8s.io/karpenter/pkg/scheduling"
+)
+
+func TestCompareOfferings_DoesNotRankEquivalentZonalOfferingsByZoneName(t *testing.T) {
```
matthchr, line 34: minor: I somewhat intentionally didn't write tests at this level, because I felt they were maybe misleading. Since stages are always used in the overall strategy, I think we should put the tests into allocation_strategy_test.go rather than here.

Since you're not actually testing the whole stage here though and are unit testing a helper I do think this makes sense where it is, but maybe just worth noting or being aware of that we should continue to keep the actual "does the stage do what I want" tests as part of the allocation strategy level rather than testing here.

pkg/providers/allocationstrategy/stages/offering_rank_test.go
```diff
@@ -0,0 +1,49 @@
...
+	westus1 := testOffering(0.1, karpv1.CapacityTypeOnDemand, "westus-1")
+	westus2 := testOffering(0.1, karpv1.CapacityTypeOnDemand, "westus-2")
+
+	if comparison := compareOfferings(westus1, westus2); comparison != 0 {
+		t.Fatalf("expected equivalent offerings in different zones to compare equal, got %d", comparison)
```
matthchr: minor, use gomega for asserts rather than raw `t.Fatalf`? I think that's our standard style.

pkg/providers/allocationstrategy/allocation_strategy.go
matthchr: although we do have azureProvider + possibly fleet coming soon... aren't we going to need at least some/all of this provider ordering for it too? For example "prefer spot over non-spot" and "prefer zonal over regional" I assume are not going the be inherent to Fleet, Karpenter will need to tell Fleet that, won't it?

pkg/providers/allocationstrategy/stages/offering_rank.go
```diff
@@ -0,0 +1,130 @@
...
+		return 1
+	}
+	if iCapacityRank, jCapacityRank := capacityTypeRank(i), capacityTypeRank(j); iCapacityRank != jCapacityRank {
+		return iCapacityRank - jCapacityRank
+	}
+	if iScopeRank, jScopeRank := placementScopeRank(i), placementScopeRank(j); iScopeRank != jScopeRank {
+		return iScopeRank - jScopeRank
+	}
+	return 0
+}
```
matthchr, line 104: minor, style:
may be worth comments here like so:

```suggestion
func compareOfferings(i, j *corecloudprovider.Offering) int {
	if i == nil && j == nil {
		return 0
	}
	if i == nil {
		return 1
	}
	if j == nil {
		return -1
	}
	// Prefer the lower priced offering
	if i.Price < j.Price {
		return -1
	}
	if i.Price > j.Price {
		return 1
	}
	// Prefer capacity-type spot over dedicated
	if iCapacityRank, jCapacityRank := capacityTypeRank(i), capacityTypeRank(j); iCapacityRank != jCapacityRank {
		return iCapacityRank - jCapacityRank
	}
	// Prefer zonal over regional
	if iScopeRank, jScopeRank := placementScopeRank(i), placementScopeRank(j); iScopeRank != jScopeRank {
		return iScopeRank - jScopeRank
	}
	return 0
}
```

pkg/providers/allocationstrategy/stages/offering_rank.go
```diff
@@ -0,0 +1,130 @@
...
+
+func placementScopeRank(offering *corecloudprovider.Offering) int {
+	switch zones.PlacementScopeForOffering(offering) {
+	case v1beta1.PlacementScopeZonal:
+		return 0
+	case v1beta1.PlacementScopeRegional:
+		return 1
+	default:
+		return 2
```
matthchr, line 115: Can this ever happen in practice? Do we generate any offerings without a specific placement scope?

pkg/providers/allocationstrategy/allocation_strategy.go
matthchr: Or is your thinking in that instance that Fleet would select the specific stages it needs directly rather than using what we call now "allocation strategy"?

pkg/providers/allocationstrategy/allocation_strategy_test.go
```diff
@@ -313,12 +314,174 @@ func TestFilterInstanceOfferings_SpotOfferingsBeforeOnDemandAtSamePrice(t *testi
...
+	filtered := provider.FilterInstanceOfferings(context.Background(), allocationstrategy.NewInstanceOfferings(instanceTypes), requirements)
+	g.Expect(filtered).To(HaveLen(2))
+	g.Expect(filtered[0].InstanceType.Name).To(Equal("Standard_Zonal"))
+}
+
+// TestFilterInstanceOfferings_ZoneTiesAreShuffled verifies that offerings tied
+// on every comparison dimension except zone do not always pick the same zone.
+// The test runs many iterations and asserts that every eligible zone shows up
+// at index 0 at least once. With a uniform shuffle the probability of any zone
+// being missed across 200 trials is vanishingly small.
```
matthchr, line 398: very minor (worth a TODO at most): This is a property-based test and there are frameworks like rapidtest that make this easier, if we want more tests in this style.

pkg/providers/instance/offerings/commonerrorhandlers.go
```diff
@@ -194,6 +198,13 @@ func handleZonalAllocationFailureError(
 }
 
 // AllocationFailure means that VM allocation to the dedicated host has failed. But it can also mean "Allocation failed. We do not have sufficient capacity for the requested VM size in this region."
+//
+// Scope: this error is treated as exhaustion across the *attempted placement
```
matthchr, line 232: This comment is in a confusing location?

There doesn't seem to be a scope parameter here, and we're talking about specific error codes like `ZonalAllocationFailed` or `AllocationFailed` when this is a general helper `handleAllocationFailureError`.

matthchr: Ahh I think I see what you're saying (though I still think the comment is confusing).

The zone isn't used for zone... it is just used to infer placement scope and block the whole scope.

pkg/providers/instance/offerings/commonerrorhandlers.go
```diff
@@ -66,42 +67,46 @@ var (
 
 type errorHandle func(ctx context.Context, unavailableOfferings *cache.UnavailableOfferings, sku *skewer.SKU, instanceType *corecloudprovider.InstanceType, zone, capacityType, errorCode, errorMessage string) error
 
-// markOfferingsUnavailableForCapacityType marks all offerings of the specified capacity type as unavailable
-func markOfferingsUnavailableForCapacityType(
+func markOfferingsUnavailableForCapacityTypeAndPlacement(
```
matthchr, line 74: Make sure the machine API variants that call these helpers (I think they exist?) are also doing the right thing. I think they will be because you're changing the helpers but just worth confirming.

pkg/providers/instance/vminstance.go
```diff
@@ -923,7 +926,10 @@ func (p *DefaultVMProvider) getLaunchTemplate(
 	additionalLabels := lo.Assign(
 		claimLabels,
 		labels.GetAllSingleValuedRequirementLabels(instanceType.Requirements),
-		map[string]string{karpv1.CapacityTypeLabelKey: capacityType},
+		map[string]string{
+			karpv1.CapacityTypeLabelKey: capacityType,
+			v1beta1.LabelPlacementScope: placementScope,
```
matthchr, line 931: We don't set the zone label via this mechanism because it's automatically set by agentbaker/kubelet bootstrap already? (I hadn't noticed it not flowing through here before, but it doesn't it seems?)

pkg/providers/instancetype/instancetypes.go
```diff
@@ -225,9 +227,10 @@ func (p *DefaultProvider) instanceTypeZones(sku *skewer.SKU) sets.Set[string] {
 // offering, you can do the following thanks to this invariant:
 //
 //	offering.Requirements.Get(v1.TopologyLabelZone).Any()
-func (p *DefaultProvider) createOfferings(sku *skewer.SKU, zones sets.Set[string]) cloudprovider.Offerings {
+func (p *DefaultProvider) createOfferings(sku *skewer.SKU, zoneSet sets.Set[string]) cloudprovider.Offerings {
```
matthchr: minor, style: I didn't really find `zones` to be confusingly named before =)

pkg/providers/instancetype/suite_test.go
```diff
@@ -403,18 +403,21 @@ var _ = Describe("InstanceType Provider", func() {
...
 				zone, err := zones.MakeAKSLabelZoneFromVM(&vm)
 				Expect(err).ToNot(HaveOccurred())
-				ExpectUnavailable(azureEnv, fake.MakeSKU(string(initialVMSize)), zone, karpv1.CapacityTypeSpot)
+				ExpectUnavailable(azureEnv, fake.MakeSKU(initialVMSize), zone, karpv1.CapacityTypeSpot)
 
 				azureEnv.VirtualMachinesAPI.VirtualMachineCreateOrUpdateBehavior.BeginError.Set(nil)
 				ExpectProvisionedAndWaitForPromises(ctx, env.Client, cluster, cloudProvider, coreProvisioner, azureEnv, pod)
 				node := ExpectScheduled(ctx, env.Client, pod)
-				Expect(node.Labels[v1.LabelTopologyZone]).ToNot(Equal(zone))
+				Expect(node.Labels[v1.LabelInstanceTypeStable] == initialVMSize &&
```
matthchr: Relaxed assertion here compared to before, harder to debug (which one failed?) and also shouldn't it be deterministic which one of these happens for the given test?

AI flagged it too:
```
 3. Weakened test assertions in error-handler integration tests
 - Priority: Medium
 - Skeptic: Tests changed from asserting specific capacity-type switches to asserting "the triple changed" (Expect(a
 == x && b == y && c == z).To(BeFalse())). This would pass even with incorrect behavior as long as any dimension
differs.
 - Advocate: With regional offerings, the exact switch target is no longer deterministic (could be regional-spot or
zonal-OD depending on ranking). The assertions correctly verify retry-with-different-offering behavior.
 - Architect: The weaker assertions are a pragmatic concession. Better assertions could verify the new expected
ranking outcome.
 - Recommendation: Consider strengthening to assert the expected fallback per the ranking rules (e.g., regional-spot
 before zonal-OD at equal price), rather than just "something changed."
```

pkg/utils/zones/zones.go
```diff
@@ -30,6 +33,26 @@ import (
...
+// reads the karpenter.azure.com/placement-scope requirement when present and
+// otherwise falls back to inferring the scope from the offering's zone.
+func PlacementScopeForOffering(offering *corecloudprovider.Offering) string {
+	if offering == nil {
+		return ""
+	}
+	if scope := offering.Requirements.Get(v1beta1.LabelPlacementScope).Any(); scope != "" {
+		return scope
+	}
+	return PlacementScopeForZone(offering.Requirements.Get(corev1.LabelTopologyZone).Any())
```
matthchr: Is the fallback actually necessary here? We build the collection of offerings, right?

worth at least a note that we don't expect it can ever happen but we have it just as a guard? I mean in some ways it seems a weak guard because we build the set of offering requirements and if we don't trust that zone might not be there either...?

pkg/cloudprovider/suite_aksmachineapi_offerings_test.go
tallaxes: Fixed naming (and added zonal placement to constrain behavior)

pkg/cloudprovider/suite_aksmachineapi_offerings_test.go
tallaxes: Added "should fail to provision when AllocationFailure errors are hit and all placements for the VM size are unavailable, then switch VM size and succeed"

pkg/cloudprovider/suite_aksmachineapi_offerings_test.go
tallaxes: Fixed (the name was wrong)

pkg/utils/zones/zones.go
tallaxes: Agreed, too defensive; removed

pkg/providers/instancetype/instancetypes.go
tallaxes: We are now importing zones package, so can't shadow it. Renamed to offeringZones (still not great, but better I think)

pkg/providers/instancetype/suite_test.go
tallaxes: Ack, restored

pkg/providers/instance/offerings/commonerrorhandlers.go
tallaxes: Confirmed; they are wired via NewErrorDetailHandler

pkg/providers/instance/offerings/commonerrorhandlers.go
tallaxes: Added a comment for clarity

pkg/providers/allocationstrategy/stages/offering_rank.go
tallaxes: No, added a comment for clarity. Not sure we can do anything better

pkg/providers/instance/vminstance.go
tallaxes: No, I think it is just left to CCM (Node Controller), so we are not passing it to kubelet. And IIRC another consideration was (at the time) that we may have cases where zone selection is left to the platform and thus has to be determined from the VM _after_ provisioning (we do that in vmInstanceToNodeClaim) - probably no longer the case.

pkg/providers/allocationstrategy/stages/offering_rank_test.go
tallaxes: Agree, added a short comment on this unit test

pkg/providers/allocationstrategy/allocation_strategy.go
tallaxes: Ack on the challenges here, and like the idea of reuse. The short answer is that I was not sure, and opted to err on the side of cleaner fit to the immediate (client-side) needs - and not including the "pick" stage in the "allocator" felt wrong. It definitely seems like some scoring mechanics could be reusable here - but not yet sure how it would compose with API capabilities (whatever allocation strategies it might support, to what extent it can respect the passed-in order, etc.) Let's maybe wait to see what capabilities & needs materialize there and revisit this area then?

tallaxes:
> Do we want to have a discussion with the cloudprovider folks before we merge this, on the 0 behavior?

Good idea @matthchr; agree it won't come soon enough, but good to know where we are going

tallaxes:
@comtalyst PTAL at how this affects recent Machine changes, in https://github.com/Azure/karpenter-provider-azure/pull/1659/commits/dcbd2caa9ad632f1845ea7fd72993c99071fbf81

pkg/providers/instance/aksmachineinstancehelpers_test.go
comtalyst, line 494: minor: consider expecting this new label

pkg/providers/instance/aksmachineinstanceutils.go
```diff
@@ -105,6 +105,7 @@ func BuildNodeClaimFromAKSMachineTemplate(
 	}
 	if zone != nil {
 		labels[corev1.LabelTopologyZone] = *zone
+		labels[v1beta1.LabelPlacementScope] = zones.PlacementScopeForZone(*zone)
```
comtalyst, line 108: This nodeclaim label assignment shouldn't be conditional on zone's existence(?)

comtalyst (commented):
(For Machine API implications only)

Let's also run the E2E w/ Machine API.

pkg/providers/allocationstrategy/allocation_strategy.go
matthchr: Yes I agree that designing for now and we can fix later, later, makes sense.

pkg/providers/instance/offerings/commonerrorhandlers.go
```diff
@@ -66,42 +67,76 @@ var (
 
 type errorHandle func(ctx context.Context, unavailableOfferings *cache.UnavailableOfferings, sku *skewer.SKU, instanceType *corecloudprovider.InstanceType, zone, capacityType, errorCode, errorMessage string) error
 
-// markOfferingsUnavailableForCapacityType marks all offerings of the specified capacity type as unavailable
-func markOfferingsUnavailableForCapacityType(
+// markOfferingsUnavailableForCapacityTypeAndPlacement marks every offering for
+// the attempted capacity type within the attempted placement scope. The zone
+// parameter identifies the selected offering and is used to infer placement
+// scope; for a zonal offering this intentionally includes sibling zones.
+func markOfferingsUnavailableForCapacityTypeAndPlacement(
```
matthchr, line 74: OK so my thinking is:
- You didn't change quota handling
- You changed handling for `handleOverconstrainedAllocationFailureError`, `handleSKUNotAvailableError`, and `handleAllocationFailureError`

I _think_ that's it?

So in terms of picking errors to investigate via data to make sure this is safe, we should look at these, rpobably?

pkg/providers/instance/aksmachineinstanceutils.go
tallaxes: Technically no, but right now zone is always there (it will just be Regional="0" for non-zonal placement), and we are using it as a proxy for placement

pkg/providers/instance/aksmachineinstancehelpers_test.go
tallaxes: Already expected in [this test](https://github.com/Azure/karpenter-provider-azure/pull/1659/changes/BASE..dcbd2caa9ad632f1845ea7fd72993c99071fbf81#diff-40224d85460b765460356654de264732acd7db15adde2305ebb5c2892e9dfd49R457-R461), but added here as well https://github.com/Azure/karpenter-provider-azure/pull/1659/commits/0e80d2106ebd5883629a5d4e6761ce2d3701cc9a

pkg/providers/instance/aksmachineinstanceutils.go
comtalyst: I mean, for the machine/node-via-kubelet, it unconditionally assigned in aksmachineinstacehelpers.go/`configureLabelsAndMode()`, like capacity label.
```go
	nodeLabels := lo.Assign(nodeClaim.Labels, claimLabels, labels.GetAllSingleValuedRequirementLabels(instanceType.Requirements), map[string]string{
		karpv1.CapacityTypeLabelKey: capacityType,
		v1beta1.LabelPlacementScope: placementScope,
	})
```

Here, capacity label is unconditionally assigned, but not placement scope.

tallaxes: I was going to say "it is essentially unconditional here as well, because zone is always there" - but now I am seeing zone field is marked as optional here, why is that?

comtalyst: It is most likely to mimic this function with [VM instance counterpart](https://github.com/Azure/karpenter-provider-azure/blob/924d83fff91538832d6bf7dc3c77f40ba1bee9ea/pkg/cloudprovider/cloudprovider.go#L519-L523).
But, yes, in practice, it is not really possible for this path, and the validation is done before calling this. We could remove this "zone being optional" here or follow-up PR to be clean.

tallaxes: In which case - "it is essentially unconditional here as well, because zone is always there", and mimics the corresponding assignment logic in [VM instance counterpart](https://github.com/Azure/karpenter-provider-azure/pull/1659/changes#diff-a6af1de71edb43b701d83b9ee10e345cf3c6ef3fc738ee3132c3309b6958896dR638-R642). 🙂

pkg/providers/instance/offerings/commonerrorhandlers.go
tallaxes: As discussed offline, focusing on gathering stats on regional-only (either no-zone/non-zonal or all-zones-restricted) VM SKUs in regions, because, where those already exist, the effect of adding more to the mix is significantly reduced. 

One datapoint: Based on two subscriptions tested (if a region is not location-restricted for the subscription) Karpenter should generally observe regional-only VM SKU options in every tested Azure region except centraluseuap.

matthchr (approved):

PR #1664, opened 2026-05-01
test: batch + machine cache in acceptance tests

comtalyst:
Fixes # 

**Description**

*

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


.vscode/settings.json
```diff
@@ -1,5 +1,5 @@
 {
-    "go.testTimeout": "240s",
```
PabloTriv: Is there a reason we need longer timeouts?

pkg/test/environment.go
```diff
@@ -266,8 +267,8 @@ func NewRegionalEnvironment(ctx context.Context, env *coretest.Environment, regi
 		testOptions.NodeResourceGroup,
 		clusterName,
 		testOptions.AKSMachinesPoolName,
-		machinecache.WithTTL(0),
-		machinecache.WithCacheDisabled(),
+		machinecache.WithTTL(1*time.Second),
```
PabloTriv, line 270: Is this really doing anything different than disabling? I'd assume if the cache was actually being used, tests would fail because the expected number of calls vs actual number of calls to machine api would be different. I suspect such a short ttl is causing gets to just fall back

.vscode/settings.json
comtalyst: It took 269s locally.
Most likely coming from batch/cache reliance on time.

pkg/test/environment.go
comtalyst: > expected number of calls vs actual number of calls to machine api would be different

Some of that has to be changed already, but good point on possible hidden assumption. We can try different TTL and see how it does?
Though, I doubt 1s is "short". From the test's perspective where things execute in less than, or few seconds, it is quite long.

Note that this is not enough to cover every possible case w/ cache, but better than nothing.

PabloTriv (dismissed):

PR #1680, opened 2026-05-10
fix: validate Azure NodePool requirement values

### LINKED ISSUE #1679: Add known-value validation for Azure provider well-known NodePool requirement labels
## Summary

Azure registers several provider-specific labels as Karpenter well-known labels, but they do not currently participate in Karpenter core's `WellKnownValuesForRequirements` validation. As a result, invalid values for Azure-specific requirement keys can pass CRD admission and runtime validation, then only surface later as scheduling incompatibility or no matching offerings.

## Proposed Scope

Add `WellKnownValuesForRequirements` entries for Azure labels with small, stable value domains:

- `karpenter.azure.com/placement-scope`: `zonal`, `regional`
- `karpenter.azure.com/sku-networking-accelerated`: `true`, `false`
- `karpenter.azure.com/sku-storage-premium-capable`: `true`, `false`
- `kubernetes.azure.com/mode`: `system`, `user`
- `kubernetes.azure.com/scalesetpriority`: `spot`, `regular`
- `kubernetes.azure.com/priority`: `spot`, `regular`
- `kubernetes.azure.com/os-sku`: `Ubuntu`, `AzureLinux` - TBD, could be wider or need to be skipped
- `kubernetes.azure.com/fips_enabled`: `true`

Avoid static known-value validation for SKU name, family, series, GPU name, and numeric labels because those are inventory-derived or better handled through scheduling compatibility.

## Expected Behavior

NodePools using only invalid values for these labels should get `ValidationSucceeded=False` with a clear validation message, matching Karpenter core's existing behavior for invalid `karpenter.sh/capacity-type` values.

tallaxes:
Fixes #1679 

**Description**

This PR registers Azure-supported known value domains for NodePool requirement runtime validation.

It adds provider-owned value validation for:

- `karpenter.sh/capacity-type`
- `karpenter.azure.com/sku-accelerated-networking`
- `karpenter.azure.com/sku-storage-premium-capable`
- `karpenter.azure.com/sku-gpu-manufacturer`
- `karpenter.azure.com/placement-scope`
- `kubernetes.azure.com/mode`
- `kubernetes.azure.com/scalesetpriority`
- `kubernetes.azure.com/priority`
- `kubernetes.azure.com/os-sku`
- `kubernetes.azure.com/fips_enabled`

For capacity type, Azure now overrides the Karpenter core known values to the currently supported values: `on-demand` and `spot`. `reserved` is intentionally treated as unsupported until Azure reserved capacity support is implemented.

For `kubernetes.azure.com/os-sku`, validation uses the coarse AKS node label values: `Ubuntu` and `AzureLinux`. (Versioned OS values such as `Ubuntu2204`, `Ubuntu2404`, `AzureLinux2`, and `AzureLinux3` are not valid values for this requirement label; they are used by image-family resolution, Machine API/bootstrap OSSKU handling, or the materialized node labels `kubernetes.azure.com/os-sku-requested` and `kubernetes.azure.com/os-sku-effective`.)

* `go test ./pkg/apis/v1beta1 ./pkg/apis/v1alpha2`
* [E2EMatrixTrigger · Azure/karpenter-provider-azure@3b38db0](https://github.com/Azure/karpenter-provider-azure/actions/runs/25686790464)

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
Validate Azure-specific NodePool requirement values at runtime, including capacity type, placement scope, SKU capability labels, GPU manufacturer, AKS mode/priority labels, OS SKU, and FIPS labels. Azure capacity type validation now accepts only on-demand and spot until reserved capacity support is added.
```

### DIFF
--- a/pkg/apis/v1alpha2/crd_validation_cel_test.go
+++ b/pkg/apis/v1alpha2/crd_validation_cel_test.go
@@ -778,6 +778,7 @@ var _ = Describe("CEL/Validation", func() {
 			v1beta1.AKSLabelMode,
 			v1beta1.AKSLabelScaleSetPriority,
 			v1beta1.AKSLabelPriority,
+			v1beta1.AKSLabelOSSKU,
 			v1beta1.AKSLabelFIPSEnabled,
 		)
 
@@ -885,6 +886,8 @@ var _ = Describe("CEL/Validation", func() {
 			Entry("AKS mode", v1beta1.AKSLabelMode, v1beta1.ModeSystem, "control-plane"),
 			Entry("AKS scale set priority", v1beta1.AKSLabelScaleSetPriority, v1beta1.ScaleSetPriorityRegular, "low"),
 			Entry("AKS priority", v1beta1.AKSLabelPriority, v1beta1.PriorityRegular, "low"),
+			Entry("AKS OS SKU Ubuntu", v1beta1.AKSLabelOSSKU, v1beta1.OSSKUUbuntu, v1beta1.Ubuntu2204ImageFamily),
+			Entry("AKS OS SKU AzureLinux", v1beta1.AKSLabelOSSKU, v1beta1.OSSKUAzureLinux, "AzureLinux3"),
 			Entry("AKS FIPS enabled", v1beta1.AKSLabelFIPSEnabled, "true", "false"),
 		)
 	})
--- a/pkg/apis/v1alpha2/crd_validation_cel_test.go
+++ b/pkg/apis/v1alpha2/crd_validation_cel_test.go
@@ -768,6 +768,8 @@ var _ = Describe("CEL/Validation", func() {
 	})
 
 	Context("Requirements", func() {
+		// Labels with registered WellKnownValuesForRequirements reject arbitrary In values.
+		// Exclude them from the generic well-known label test below, which uses "test" as a placeholder value.
 		knownValueRequirementLabels := sets.New(
 			karpv1.NodePoolLabelKey,
 			karpv1.CapacityTypeLabelKey,
@@ -839,6 +841,7 @@ var _ = Describe("CEL/Validation", func() {
 		DescribeTable("should validate Azure known requirement values", func(key, validValue, invalidValue string) {
 			oldNodePool := nodePool.DeepCopy()
 
+			By("rejecting a requirement with only invalid values")
 			test.ReplaceRequirements(nodePool, karpv1.NodeSelectorRequirementWithMinValues{
 				Key:      key,
 				Operator: corev1.NodeSelectorOpIn,
@@ -848,6 +851,7 @@ var _ = Describe("CEL/Validation", func() {
 			Expect(nodePool.RuntimeValidate(ctx)).ToNot(Succeed())
 			Expect(env.Client.Delete(ctx, nodePool)).To(Succeed())
 
+			By("accepting a requirement with only valid values")
 			nodePool = oldNodePool.DeepCopy()
 			test.ReplaceRequirements(nodePool, karpv1.NodeSelectorRequirementWithMinValues{
 				Key:      key,
@@ -858,6 +862,7 @@ var _ = Describe("CEL/Validation", func() {
 			Expect(nodePool.RuntimeValidate(ctx)).To(Succeed())
 			Expect(env.Client.Delete(ctx, nodePool)).To(Succeed())
 
+			By("failing open when valid and invalid values are both present")
 			nodePool = oldNodePool.DeepCopy()
 			test.ReplaceRequirements(nodePool, karpv1.NodeSelectorRequirementWithMinValues{
 				Key:      key,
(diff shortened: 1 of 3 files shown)


pkg/apis/v1alpha2/crd_validation_cel_test.go
```diff
@@ -767,9 +768,23 @@ var _ = Describe("CEL/Validation", func() {
 	})
 
 	Context("Requirements", func() {
+		knownValueRequirementLabels := sets.New(
```
matthchr, line 773: minor: maybe leave a comment on this explaining exactly what it is?

I _think_ it's the set of labels that have a known set of allowed values (they do not allow arbitrary strings). Is that right?

pkg/apis/v1alpha2/crd_validation_cel_test.go
```diff
@@ -20,6 +20,7 @@ import (
 	"strings"
 
 	"github.com/Azure/karpenter-provider-azure/pkg/apis/v1alpha2"
+	"github.com/Azure/karpenter-provider-azure/pkg/apis/v1beta1"
```
matthchr, line 23: It's been on my list for a while to move the label definitions someplace else, but this is a great example of why we probably should...

Though I guess really we should just drop v1alpha2 at some point.

pkg/apis/v1alpha2/crd_validation_cel_test.go
```diff
@@ -803,18 +821,75 @@ var _ = Describe("CEL/Validation", func() {
...
+		DescribeTable("should validate Azure known requirement values", func(key, validValue, invalidValue string) {
+			oldNodePool := nodePool.DeepCopy()
+
+			test.ReplaceRequirements(nodePool, karpv1.NodeSelectorRequirementWithMinValues{
+				Key:      key,
+				Operator: corev1.NodeSelectorOpIn,
+				Values:   []string{invalidValue},
+			})
+			Expect(env.Client.Create(ctx, nodePool)).To(Succeed())
+			Expect(nodePool.RuntimeValidate(ctx)).ToNot(Succeed())
```
matthchr: We allow NotIn invalidValue?

I mean I guess it makes sense.

pkg/apis/v1alpha2/crd_validation_cel_test.go
```diff
@@ -803,18 +821,75 @@ var _ = Describe("CEL/Validation", func() {
...
 			nodePool = oldNodePool.DeepCopy()
-		})
+		},
+			Entry("unknown capacity type", "xspot"),
+			Entry("reserved capacity type", karpv1.CapacityTypeReserved),
+		)
+		DescribeTable("should validate Azure known requirement values", func(key, validValue, invalidValue string) {
+			oldNodePool := nodePool.DeepCopy()
+
+			test.ReplaceRequirements(nodePool, karpv1.NodeSelectorRequirementWithMinValues{
```
matthchr, line 851: minor: consider test comments or logging statements to clarify the intent of each assertion?

```
// valid value works

// invalid value doesn
```

pkg/apis/v1alpha2/crd_validation_cel_test.go
```diff
@@ -803,18 +821,75 @@ var _ = Describe("CEL/Validation", func() {
...
+			Expect(env.Client.Delete(ctx, nodePool)).To(Succeed())
+
+			nodePool = oldNodePool.DeepCopy()
+			test.ReplaceRequirements(nodePool, karpv1.NodeSelectorRequirementWithMinValues{
+				Key:      key,
+				Operator: corev1.NodeSelectorOpIn,
+				Values:   []string{strings.ToUpper(validValue)},
+			})
+			Expect(env.Client.Create(ctx, nodePool)).To(Succeed())
+			Expect(nodePool.RuntimeValidate(ctx)).ToNot(Succeed())
```
matthchr: Do we expect a particular error? Should we assert on that here?

pkg/apis/v1alpha2/crd_validation_cel_test.go
matthchr: note: All these comments apply to the other file too

pkg/apis/v1beta1/labels.go
```diff
@@ -32,6 +32,17 @@ func init() {
 	// computeRequirements in pkg/providers/instancetype/instancetype.go, because (as far as I can tell)
 	// Karpenter core expects that WellKnownLabels are mapped to requirements.
 	karpv1.WellKnownLabels = karpv1.WellKnownLabels.Union(AzureWellKnownLabels)
+	// Register exact, stable Azure-supported value domains for Karpenter core runtime requirement validation.
+	karpv1.WellKnownValuesForRequirements[karpv1.CapacityTypeLabelKey] = sets.New(karpv1.CapacityTypeOnDemand, karpv1.CapacityTypeSpot)
```
matthchr, line 36: Doh! I didn't realize this was a thing. Good catch.

matthchr (dismissed):
LGTM, had a few minor comments primarily about test formatting/clarity.

pkg/apis/v1alpha2/crd_validation_cel_test.go
tallaxes: Ack. +1 on just dropping v1alpha2

pkg/apis/v1alpha2/crd_validation_cel_test.go
tallaxes: Yes, only 'In' lists are validated

pkg/apis/v1alpha2/crd_validation_cel_test.go
tallaxes: No specific type. Could still check the error content / substrings, but this seems fragile. So inclined to leave as is. WDYT?

matthchr: My concern with this type of "it fails" assertion is always that it starts failing for some other reason but we count it as success even when we shouldn't. If this were production code I'd agree that parsing strings for error is a smell, but for a test I think it's better than not?

Worst case is, the error changes and we realize it (because we have a test), and we just have to fix the test (easy?). Best case is, something happens and this is failing for some other reason, and we notice and realize we need to do something because that "other reason" is unexpected.

I mean it's minor either way but yeah I'd probably personally say assert on substring here.

tallaxes: Ack, good points, adding substring-based checks. Edit: Done

matthchr (approved):

PR #1684, opened 2026-05-16
fix: make parsing cluster ID more reliable

matthchr:

* Unit tests
* E2E: https://github.com/Azure/karpenter-provider-azure/actions/runs/25955064764

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -160,6 +165,15 @@ func (p *Provider) getStaticParameters(
 
 	// ATTENTION!!!: changes here will NOT be effective on AKS machine nodes (ProvisionModeAKSMachineAPI); See aksmachineinstance.go/aksmachineinstancehelpers.go.
 	// Refactoring for code unification is not being invested immediately.
+
+	nsg, err := p.nsgProvider.ManagedNetworkSecurityGroup(ctx)
```
tallaxes, line 169: This is cached internally and thus cheap to do every time, right?

pkg/providers/launchtemplate/launchtemplate.go
```diff
@@ -160,6 +165,15 @@ func (p *Provider) getStaticParameters(
 
 	// ATTENTION!!!: changes here will NOT be effective on AKS machine nodes (ProvisionModeAKSMachineAPI); See aksmachineinstance.go/aksmachineinstancehelpers.go.
```
tallaxes, line 166: I presume cluster id is not needed on Machine API path, right?

tallaxes (approved):

pkg/providers/launchtemplate/launchtemplate.go
matthchr: Correct, because it's only needed for AgentBaker parameters which we don't need to fill out for Machine path.

pkg/providers/launchtemplate/launchtemplate.go
matthchr: Correct in fact we never re-look it up today. It's looked up once and we hold that original lookup in memory for the life of the process. This is safe because (AFAIK) it never changes. It's what we've been doing and should be safe for this case too (the name can't change anyway and that's all we need).

Issue #1625, opened 2026-04-07
Pods in E2E tests going to `Failed` status due to scheduling race


matthchr:
1. I have a pending pod and one node that is coming up.
2. Node comes up and goes ready, but has node.cilium.io/agent-not-ready:NoExecute taint. Pod can't be scheduled yet because taint is not removed yet.
3. Now, taint is removed. I think the way this happens is that daemonset/service on node calls APIServer and edits the node object to remove the taint. Doing so changes the state in etcd but neither scheduler nor node know about this change yet because their watches haven't synced yet.
4. Scheduler watch notices the taint is gone and says cool now I can schedule, and does so.
5. Kubelet watch hasn't actually seen the taint removal yet (slower than scheduler) and so pod gets rejected and goes to failed with time=2026-04-07T00:27:32Z type=Warning reason=TaintToleration from=kubelet message=Predicate TaintToleration failed: node(s) had taints that the pod didn't tolerate

I think the same thing can in theory happen with any of the startup taints, but I am seeing it most frequently with Cilium specifically (not sure if that's because it's last, or some other issue...)

matthchr:
Possibly we can move to use deployment or job  rather than `coretest.UnschedulablePod`.

rakechill:
The BYOK e2e suite has a ~78% pass rate on main (32/41 runs) since 4/25. The 9 failures break down into a few categories, one of which seems to manifest as this exact issue.

_**Pod timeout with NodeClassResolutionError -- 4 failures (recurring):**_ 
This is the main unresolved issue. Pods get scheduled onto Karpenter-provisioned nodes, immediately enter Failed phase, get deleted, and then Eventually() spins for 16 minutes polling for a pod that no longer exists. The NodeClassResolutionError ("Failed resolving NodeClass") appears on the nodepool in every case. Likely root cause: a race condition where the test creates NodeClass + NodePool + Pod atomically, and karpenter provisions a node before the NodeClass status controller finishes validation.

tallaxes:
Ran into this when deflaking Storage E2E. Thinking we can switch to Deployment where possible, and use Pod with taint toleration (bounded via tolerationSeconds) where direct Pod is important for some reason. Trying this for Storage in #1685, with wider experiment in #1686.

matthchr:
Closing this based on the expectation above that it was fixed by @tallaxes - Alex feel free to reopen if you disagree, but I haven't seen any of these flakes lately.

PR #1685, opened 2026-05-17
test(e2e): deflake storage suite

tallaxes:
**Description**

Deflakes the Storage E2E suite on AKS by making the storage workload setup more resilient to Azure/Cilium timing, storage cleanup behavior, and zonal Azure disk availability.

Changes:
- Adds `env.Pod(...)` as an Azure E2E wrapper around `coretest.Pod(...)` for tests that should apply Azure environment defaults.
- Gives pod-based tests a bounded Cilium startup-taint toleration to avoid the known startup-taint race.
- Uses hostPath-backed static PVs instead of the fake CSI driver in real clusters, avoiding orphaned/problematic fake CSI `VolumeAttachment` behavior.
- Runs PVC-backed storage test workloads as `Deployment`s so transient pod failures can be retried by the workload controller.
- Adds common Deployment health expectation helpers for Deployment-backed E2E workloads.
- Pins single-zone Storage topology tests to a stable ARM zone instead of shuffling subscription availability-zone mappings, which can include zones where the default `Standard_LRS` node OS disk is unavailable.
- Leaves generic ephemeral volume and emptyDir tests pod-based.

* E2E: https://github.com/Azure/karpenter-provider-azure/actions/runs/25981042380 ✅

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
```


test/pkg/environment/azure/environment.go
```diff
@@ -253,3 +255,17 @@ func (env *Environment) AZLinuxNodeClass() *v1beta1.AKSNodeClass {
 	nodeClass.Spec.ImageFamily = lo.ToPtr(v1beta1.AzureLinuxImageFamily)
 	return nodeClass
 }
+
+// Pod wraps coretest.Pod for Azure E2E tests; use it instead of coretest.Pod when the test should apply Azure environment defaults.
```
matthchr, line 259: When should the test apply Azure env defaults though? can we be more clear in the comment here?

And/or do we want a reference to the actual issue that discusses this more in-depth?
https://github.com/Azure/karpenter-provider-azure/issues/1625

I am not sure if linking the issue or a larger comment here is best, but I think one of them makes sense?

matthchr (dismissed):

test/pkg/environment/azure/environment.go
tallaxes: Added a bit of both

matthchr (approved):

PR #1690, opened 2026-05-18
fix: batcher could sometimes wait til idle timeout even when full

matthchr:
Fixes a unit test flake that was perioidcally happening in CI as well.

**Description**

* Unit tests
* Ran UTs locally under `-race` until I reproed the failure, added fix, confirmed no repo anymore

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/utils/batcher/batcher.go
+++ b/pkg/utils/batcher/batcher.go
@@ -201,6 +201,12 @@ func (b *Batcher[RequestPayload, ResponsePayload]) run() {
 //  2. maxTimeout passes (latency SLA)
 //  3. Any batch reaches maxBatchSize (full batch)
 func (b *Batcher[RequestPayload, ResponsePayload]) waitForIdle() {
+	// Check immediately in case all requests were enqueued before we started
+	// listening on the trigger channel (their signals may have been coalesced/dropped).
+	if b.anyBatchFull() {
+		return
+	}
+
 	maxTimer := time.NewTimer(b.opts.MaxTimeout)
 	idleTimer := time.NewTimer(b.opts.IdleTimeout)
 	defer maxTimer.Stop()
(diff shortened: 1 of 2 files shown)


pkg/utils/batcher/batcher.go
comtalyst, line 214: Technically there might still be a tiny window of race around here(?), but I suppose the risk is neglectable, especially that we have timeout?

comtalyst (approved):

pkg/utils/batcher/batcher.go
matthchr: I don't think so, because at this point the core `run` loop has called `waitForIdle` (which means it emptied the trigger channel). If we get to the start of `waitForIdle` there are three main cases:

1. Something in the meantime has written the trigger channel again, so it's full.
2. Nothing has (but a batch is full)
3. Nothing has (but a batch is not full)

Case 1, channel is already full so read from it on like 220 will trigger and we're happy.
Case 2, batch is full is caught by the newly added code.
Case 3, batch is not full yet so we're fine to wait.

In case 2 case, if we get full AFTER the added lines but before ln 220, then we must have written to trigger again too (it was empty when waitForIdle was called in the run loop), so we'll trigger the check there and are good.

I confirmed with AI which said:
```
Good question, but actually no — that window is safe. Here's why:

When waitForIdle() is entered, the trigger channel is guaranteed empty (it was just drained by run()'s [case <-b.trigger:](vscode-file://vscode-app/c:/Users/matthchr/AppData/Local/Programs/Microsoft%20VS%20Code/0958016b2a/resources/app/out/vs/code/electron-browser/workbench/workbench.html)). So between the initial anyBatchFull() check and the select:

If new requests arrive and fill the batch, the first one's trigger send will succeed (the buffer has room since the channel is empty)
The remaining ones may be dropped, but that's fine
When select starts, it immediately receives that buffered signal
anyBatchFull() checks actual batch state → sees it's full → returns
The one-signal-is-enough guarantee holds because anyBatchFull() inspects actual batch contents rather than counting signals. As long as we receive any trigger after the batch becomes full, we'll detect it — and the empty channel guarantees at least one signal gets through.

The original bug was different: the channel could be full (from a signal sent before run() drained it and before waitForIdle() existed), causing ALL signals from the 3 concurrent enqueues to be dropped — leaving waitForIdle() with nothing to receive.
```

PR #1691, opened 2026-05-18
fix: cap machine name at 34 chars (instead of 35) to fit 63-char node name limit

comtalyst:
Fixes # 

**Description**

Machine API-side fix will be separate, but this should resolve the issue.

* Unit tests (`go test ./pkg/providers/instance/...`)

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
```


tallaxes (approved):

tallaxes:
@comtalyst I assume this can't be a breaking change, because it would not have worked to begin with?

comtalyst:
@tallaxes correct

theunrepentantgeek (approved):

PR #1692, opened 2026-05-18
fix: type-safety issues in fakes packages

matthchr:
Dropped sync.Map in favor a well-typed approach with RWLock and an underlying map[K]V.

* Unit tests

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


matthchr:
The context here is that in another PR I noticed that we were mixing `*string` and `string` as keys in these fake maps, which means that while `List` calls would work (because List just iterates all the values and doesn't care about the keys), Get calls don't work.

Given `sync.Map` has very specific use-cases anyway, and we're not really making use of those here, swapped to a simple RWMutex-backed map with a similar interface to `sync.Map`.

comtalyst (approved):

PR #1694, opened 2026-05-21
chore: block setting kubernetes.azure.com taints

matthchr:
* The only allowed kubernetes.azure.com taint is kubernetes.azure.com/scalesetpriority=spot.

* Unit tests
* Investigation of existing clusters via Kusto metrics

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/charts/karpenter-crd/templates/karpenter.sh_nodeclaims.yaml
+++ b/charts/karpenter-crd/templates/karpenter.sh_nodeclaims.yaml
@@ -214,6 +214,10 @@ spec:
                         type: string
                         minLength: 1
                         pattern: ^([a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*(\/))?([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9]$
+                        maxLength: 316
+                        x-kubernetes-validations:
+                          - message: taint domain "kubernetes.azure.com" is restricted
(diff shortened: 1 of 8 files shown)


charts/karpenter-crd/templates/karpenter.sh_nodeclaims.yaml
```diff
@@ -214,6 +214,10 @@ spec:
                         type: string
                         minLength: 1
                         pattern: ^([a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*(\/))?([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9]$
+                        maxLength: 316
+                        x-kubernetes-validations:
+                          - message: taint domain "kubernetes.azure.com" is restricted
+                            rule: '!self.startsWith("kubernetes.azure.com/") || self == "kubernetes.azure.com/scalesetpriority"'
```
tallaxes, line 220: In several other places we use "regex for everything before first /".endsWith(...) to catch subdomains as well. I presume we don't care about subdomains here?

tallaxes (approved):

charts/karpenter-crd/templates/karpenter.sh_nodeclaims.yaml
matthchr: yeah, even AKS RP doesn't block subdomains here today so this matches RP exactly.

comtalyst (approved):
Did we decided to not label this as breaking?
(https://xkcd.com/1172/)

PR #1703, opened 2026-06-02
fix: refactor instancetypes cache key

matthchr:
* Include fips mode in the instance type cache.
* Make the cache key harder to get wrong in the future.

* Unit tests
* E2E tests: https://github.com/Azure/karpenter-provider-azure/actions/runs/26797869240

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/providers/instancetype/instancetypes.go
+++ b/pkg/providers/instancetype/instancetypes.go
@@ -56,6 +56,20 @@ const (
 	InstanceTypesCacheTTL = 23 * time.Hour
 )
 
+// instanceTypeParameters contains the resolved set of AKSNodeClass fields that affect
+// instance-type construction. The instance-type cache key is derived by hashing this
+// struct; adding a new field here automatically incorporates it into the key.
+type instanceTypeParameters struct {
+	ImageFamily              string
+	OSDiskSizeGB             int32
+	MaxPods                  int32
+	EncryptionAtHost         bool
+	GPUMode                  v1beta1.GPUMode
+	ArtifactStreamingEnabled bool
+	FIPSMode                 v1beta1.FIPSMode
+	LocalDNSEnabled          bool
+}
+
 type Provider interface {
 	LivenessProbe(*http.Request) error
 	List(context.Context, *v1beta1.AKSNodeClass) ([]*cloudprovider.InstanceType, error)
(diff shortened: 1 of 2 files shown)


pkg/providers/instancetype/instancetypes.go
```diff
@@ -56,6 +56,32 @@ const (
 	InstanceTypesCacheTTL = 23 * time.Hour
 )
 
+// instanceTypeInput contains the resolved set of AKSNodeClass fields that affect
+// instance-type construction. The instance-type cache key is derived by hashing this
+// struct; adding a new field here automatically incorporates it into the key.
+type instanceTypeInput struct {
```
comtalyst: nit: "InstanceTypeProperties" might be more clear?
Thought the caveat is this is not formally a part of `cloudprovider.InstanceType`

pkg/providers/instancetype/instancetypes.go
```diff
@@ -56,6 +56,32 @@ const (
...
+	Spec   instanceTypeInputSpec
+	Status instanceTypeInputStatus
```
comtalyst: Should we distinguish these in this layer? It mimics NodeClass, but not sure if there are benefits in this layer specifically, compared to a flat list of significant properties (which is more clear IMO)

pkg/providers/instancetype/instancetypes.go
```diff
@@ -56,6 +56,32 @@ const (
...
+}
+
+// instanceTypeInputSpec contains the resolved subset of AKSNodeClassSpec fields that influence
+// cached instance types.
+type instanceTypeInputSpec struct {
```
comtalyst: Is there a good way to prevent us from forgetting to add a field here when we add an instancetype label?

comtalyst (dismissed):

pkg/providers/instancetype/instancetypes.go
matthchr: Label is not so much the issue, but anything that `computeRequirements` relies on has to be added here because it uses this type rather than the NodeClass directly., so I think this change accomplishes what is needed which is to ensure we don't get wrong cache entry.

pkg/providers/instancetype/instancetypes.go
matthchr: Not sure properties is really any better than input.

WDYT about `instanceTypeParameters`? Is it any better?

matthchr: Or `resolveInstanceTypeParameters`/`resolveInstanceTypeInput`?

matthchr: Changed to `instanceTypeParameters`

pkg/providers/instancetype/instancetypes.go
matthchr: Flattened.

pkg/providers/instancetype/instancetypes.go
comtalyst: Parameters is more clear, yes

comtalyst (approved):

Issue #119, opened 2024-01-27
Enable UltraSSD when deploying VM that supports it


rverma-dev:
### Tell us about your request

Trying to enabled and use ultra-ssd for nodepool created by karpenter

### Tell us about the problem you're trying to solve. What are you trying to do, and why is it hard?

converting below to Nodepool

```
az aks nodepool add --name tikv \
    --cluster-name  \
    --resource-group ${resourceGroup} \
    --node-vm-size ${nodeType} \
    --zones 1 2 3 \
    --aks-custom-headers EnableAzureDiskFileCSIDriver=true \
    --node-count 3 \
    --labels dedicated=tikv \
    --node-taints dedicated=tikv:NoSchedule \
    --enable-ultra-ssd
```

### Are you currently working around this issue?

No 

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

Bryce-Soghigian:
Thanks for creating an issue! 

Leaving some notes as Ultra SSD and my inital thoughts on how it conflicts with karpenter today(not to say we will never implement)

1. Ultra disks can't be used as OS disks, they can only be created as empty data disks. Currently we only are creating the os disk in current karpenter implementation. 
2. [Ultra Disks require zonal knowledge to know if its available](https://learn.microsoft.com/en-us/azure/virtual-machines/disks-enable-ultra-ssd?tabs=azure-portal#vms-using-availability-zones).  This introduces a new dimension to karpenter's requirements

We need to design and think through data disks and requirements being zonal aware.

rverma-dev:
Yeah the usecase is not for OS disks for sure. However I believe to start with we need to find a way to modify AdditionalCapabilities.UltraSSDEnabled = true.

tallaxes:
We should add enabling UltraSSD by default when using VM SKU that supports it, in supported zone - we have this information from SKU API.

PabloTriv:
Merged PR

PR #1686, opened 2026-05-17
test(e2e): deflake workload provisioning

tallaxes:
Related to #1625

**Description**

Deflake Azure E2E workload provisioning by using Deployments for workload-oriented tests and keeping direct Pods only where pod-level semantics are part of the test.

This PR is stacked on `tallaxes/deflake-storage`; the Storage-suite-specific deflake work is covered in that base branch. This PR applies the same workload resilience pattern across the remaining E2E suites.

This change:
- Adds Azure pod defaults for Cilium startup taints where direct Pods are still required.
- Adds common Deployment health expectation helpers.
- Converts workload-style E2E cases from direct Pods to Deployments.
- Derives node and VM assertions from healthy Deployment pods instead of assuming the first created node is the workload node.
- Keeps direct Pods for naked-pod coverage, pod-scoped storage, one-shot node diagnostics, termination behavior, and scheduling trigger cases.
- Adds comments explaining why the remaining direct Pod usages are intentionally not Deployments.

Test changes unrelated to the above (can be extracted / used on their own):
- (Utilization) Flip polarity of checking for SIG https://github.com/Azure/karpenter-provider-azure/pull/1686/changes/b2ea607375e5627454e9f29c399564abd64fa860
- (Scheduling) Flip polarity of checking for SIG https://github.com/Azure/karpenter-provider-azure/pull/1686/commits/856f3e8870b95a7c9f5e55597b00c2eb3d7aa889 
- (Scheduling) Include scheduling on placement-scope https://github.com/Azure/karpenter-provider-azure/pull/1686/commits/cbb7a91c56cd5ee4ad8f406bbfa3ac6279047ce8
- (Consolidation) Use more targeted ConsistentlyExpectNodesNotDisrupted in consolidation tests https://github.com/Azure/karpenter-provider-azure/pull/1686/commits/1eaca562c79e7b3fd075de3375fce40313f274ce
- (Integration) Ensure DaemonSet with LimitRange is observed before test https://github.com/Azure/karpenter-provider-azure/pull/1686/commits/ef7e88033cc2efc713fe0b4bccf93fe954d625a3

**How was this change tested?**

* E2E:
  * [E2EMatrixTrigger · Azure/karpenter-provider-azure@ef7e880](https://github.com/Azure/karpenter-provider-azure/actions/runs/26003100880) 26003100880 - everything except Consolidation passed
[...]

PR #1704, opened 2026-06-02
feat: support for ultrassd

PabloTriv:
Fixes #  #119 

**Description**
This change introduces support for ultra ssd. Included is a design doc that talks more about the implementation. UltraSSD will be enabled/disabled via a setting on the AKSNodeClass. The implementation here should be as close to AKS as possible.

New E2E: https://github.com/Azure/karpenter-provider-azure/actions/runs/27987096466/job/82831071878
*

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


designs/0013-ultra-ssd-support.md
```diff
@@ -0,0 +1,156 @@
...
+- express Ultra SSD as part of node configuration,
+- filter out VM sizes and zonal offerings that do not support Ultra SSD,
+- set the correct downstream API fields when creating capacity
+
+This document proposes how to complete that work for NAP.
+
+### Goals
+
+- Add support for enabling Ultra SSD on dynamically provisioned nodes.
+- Support both VM provisioning mode and AKS Machine API mode.
```
rakechill, line 26: Concerned about defining this on the AKSNodeClass for machine api, when it will also be configurable on the MC. Might lead to unexpected states unless this field is truly only configurable at cluster creation time.

designs/0012-ultra-ssd-support.md
```diff
@@ -0,0 +1,156 @@
...
+- it is a provisioning feature, not just a schedulable label,
+- it aligns with the current `AKSNodeClass` design pattern
+
+The user expectation of “default false” is still satisfied. If `spec.ultraSSD` or `spec.ultraSSD.enabled` is omitted, the effective value is disabled.
+
+### Decision 2: How should we filter for compatible Instances?
+
+#### Offerings Filtering
+
+Ultra SSD is only available in regions and zones that support it, and only by specific SKUs. Therefore, we need to check availability for each zone when creating Offerings for InstanceTypes.
```
rakechill: Is this just a matter of checking one more field on the SKU object during offerings filtering, or is there complex logic here that AKS RP already has a mapping for?

designs/0012-ultra-ssd-support.md
```diff
@@ -0,0 +1,156 @@
...
+spec:
+	ultraSSD:
+		enabled: true
+```
+
+Semantics:
+
+- default is disabled when omitted,
+- enabling it opts the node class into Ultra SSD-capable capacity only,
+- changing it triggers node replacement through drift.
```
rakechill: is this something that you can set/unset after cluster creation? do we want to support enabling/disabling this from karpenter or is there a way we can detect this from MC object/helmvalues?

rakechill (commented):
I think my main concern (reflected in a few different comments) is if we should treat the source of truth for a NAP cluster as the MachineAPI/ManagedCluster state OR the aksnodeclass fields? 

Need a few things confirmed:
1. Is this only set during cluster creation? or can it be enabled/disabled through the cluster lifecycle?
2. Is it set at the cluster level? agentpool level? or both?
3. Will/can/should Machine API report this as a DriftAction when the MC field for this changes?

If this is only enabled at cluster creation for AKS, I think this should instead be set via helm  values -> env vars/flags -> opts.

If this can be disabled/enabled after cluster creation, I think this nodeclass field should be limited to self-hosted and machineapi should handle setting this for machines based on MC object.

designs/0012-ultra-ssd-support.md
PabloTriv: This check is just a matter of calling sku.IsUltraSSDAvailableInAvailabilityZone(zone) for each zone. RP logic does the same.

designs/0012-ultra-ssd-support.md
PabloTriv: [UltraSSD can only be set during cluster creation or nodepool creation](https://learn.microsoft.com/en-us/azure/aks/use-ultra-disks#:~:text=This%20feature%20can%20only%20be%20set%20at%20cluster%20creation%20or%20when%20creating%20a%20node%20pool.). For that reason, I don't think we should or even can support disabling/enabling.

designs/0013-ultra-ssd-support.md
PabloTriv: It's only configurable at cluster creation (which makes the initial system pool have it enabled) and at nodepool creation. You can't disable it after you created it and you can't update a pool to have it

PabloTriv: Note that when you create a cluster with --enable-ultra-ssd only the initial pool gets the functionality. Subsequent pools added to the cluster still need the flag to be explicitly enabled.

PabloTriv:
> Is this only set during cluster creation? or can it be enabled/disabled through the cluster lifecycle?

This is set during cluster creation and nodepool creation. It can't be changed once a pool has been created. Enabling at the cluster means the system pool gets the UltraSSD capabilities, not that all nodepools/nodes in the cluster will have the functionality by default.

>Is it set at the cluster level? agentpool level? or both?

It's agentpool level. When you pass it during cluster create you're just enabling it on the initial create request. And other nodepools you add need to opt in. It's also immutable.

>Will/can/should Machine API report this as a DriftAction when the MC field for this changes?

The field can't change once the cluster/pool is created, so I don't think we'd have to worry about this.

>If this is only enabled at cluster creation for AKS, I think this should instead be set via helm values -> env vars/flags -> opts.

I think this would make sense if it was a global field that controlled all nodepools. But today each pool needs to opt into it, which is why I think it would make sense to stick it into either the NodePool or AKSNodeClass CRDs. Clusters can have a mix of ultrassd enabled/disabled pools on the same cluster.

rakechill:
> ...I think this would make sense if it was a global field that controlled all nodepools. But today each pool needs to opt into it, which is why I think it would make sense to stick it into either the NodePool or AKSNodeClass CRDs. Clusters can have a mix of ultrassd enabled/disabled pools on the same cluster.

I see, didn't realize that enabling for the cluster == enabling for the system nodepool. In that case, it makes sense to me for this to be something that gets enabled from the Karpenter side. Will customers be able to change this after the fact then? That would be different behavior than what AKS currently provides. 

I'd expect there to be some validation if a customer tries to switch from Enabled to Disabled.

Or, if we don't plan to mimic RP behavior, can we get clarity on _why_ RP doesn't allow disablement after enablement?

designs/0012-ultra-ssd-support.md
```diff
@@ -0,0 +1,156 @@
...
+
+func (in *AKSNodeClass) IsUltraSSDEnabled() bool {
+		return in.Spec.UltraSSD != nil &&
+				in.Spec.UltraSSD.Enabled != nil &&
+				*in.Spec.UltraSSD.Enabled
+}
+```
+
+This matches the existing API style for feature toggles such as `artifactStreaming`, `security.encryptionAtHost`, and `localDNS`.
+Ultra SSD should be configured as a strongly typed `AKSNodeClass` feature, not as a raw requirement.
```
rakechill: Do we know if these other features are also expected to be per-agentpool vs. per-cluster in AKS + we diverged for Karpenter?

My concern is that we're forcing all Nodepools to use a feature that is usually a per-agentpool opt-in.

rakechill (commented):
Your comments from the previous review make sense + I better understand the agentpool<>cluster relationship wrt this feature. 

A few remaining thoughts:
1. Is there any way to make this per-nodepool? I expect many clusters will have different Nodepools for different use cases + may prefer that level of granularity. Maybe checking with @wdarko1 based on current customer requirements could be useful.
2. Are the other features we have on AKSNodeClass _also_ configurable per-agentpool in AKS? Or are they cluster-wide?
3. Should we block disablement after enablement to align with current AKS behavior?
4. Should we **only** allow ultra SSD to be enabled in Karpenter if it's enabled on the cluster? Basically introducing a validation here based on cluster state.

PabloTriv:
>Is there any way to make this per-nodepool? I expect many clusters will have different Nodepools for different use cases + may prefer that level of granularity. Maybe checking with @wdarko1 based on current customer requirements could be useful.

I feel like using the AKSNodeClass is the pattern that's already established. NodePools seem to be more about generic Karpenter scheduling specs and NodeClass about provider specific stuff. The NodePool also references the NodeClass so cx still has some level of granularity by swapping around the NodeClass that their NodePool points to. Will ask @wdarko1 for his thoughts.

>Are the other features we have on AKSNodeClass also configurable per-agentpool in AKS? Or are they cluster-wide?

[FIPS](https://learn.microsoft.com/en-us/azure/aks/enable-fips-nodes?pivots=azure-cli), [LocalDNS](https://learn.microsoft.com/en-us/azure/aks/localdns-custom?tabs=enable), and [Artifact Streaming](https://learn.microsoft.com/en-us/azure/aks/artifact-streaming) all operate at nodepool level in AKS. In particular, FIPS seems the most similar, since you can enable it at cluster creation time for the system pool but requires opt-in for subsequent nodepools.

>Should we block disablement after enablement to align with current AKS behavior?

AKS RP does reject attempts to change the UltraSSD settings on an existing pool today. In theory, if someone tries to edit the field on the NodeClass CR, we should consider the node as drifted and replace it. 

>Should we only allow ultra SSD to be enabled in Karpenter if it's enabled on the cluster? Basically introducing a validation here based on cluster state.

I think No. Today in AKS you can make a cluster without enabling ultra ssd at creation and add an ultra-ssd enabled nodepool to it.

designs/0012-ultra-ssd-support.md
PabloTriv: Those features are enabled at agentpool level. I think the idea would be to have an extra field on the AKSNodeClass that defaults to false, and if the customer wants it they can set it to true.

PabloTriv:
I think that the proposed implementation would be pretty consistent with AKS. If you disable the field on the NodeClass, all the NodeClaims belonging to that NodePool/Class would be considered drifted and replaced, the same way you'd have to replace a nodepool in AKS.

Interestingly, while AKS nodepools do not let you update the ultra ssd field, standalone VMs [can actually enable/disable it](https://learn.microsoft.com/en-us/azure/virtual-machines/disks-enable-ultra-ssd?utm_source=chatgpt.com&tabs=azure-cli), but require being stopped and deallocated first.

rakechill (commented):
design lgtm, let's just add a customer experience section to the design doc that makes the enablement/disablement conditions clear + adds support/reasoning from the docs + current AKS behavior.

designs/0012-ultra-ssd-support.md
```diff
@@ -0,0 +1,156 @@
...
+
+Reasons:
+
+- it is a provisioning feature, not just a schedulable label,
```
comtalyst:
```suggestion
- it is a provisioning feature, not a schedulable label,
```

designs/0012-ultra-ssd-support.md
comtalyst: How does AKS API validates UltraSSD enablement? 
Worth referencing that here and "prove" that Karpenter obliges it, especially if that is determined by how Karpenter select SKUs. 
This is so that we can avoid provisioning time failures or accidentally having to support what AKS doesn't.

designs/0012-ultra-ssd-support.md
comtalyst: > UltraSSD can only be set during cluster creation or nodepool creation

That's quite unusual for a pool-level feature. Could you investigate why?

If it is justifiable and we want to pair with AKS, not supporting post-creation changes need extra work.

pkg/providers/instance/aksmachineinstancehelpers.go
```diff
@@ -369,6 +370,13 @@ func configureLabelsAndMode(nodeClaim *karpv1.NodeClaim, instanceType *corecloud
 	return nodeLabelPtrs, modePtr
 }
 
+func configureUltraSSDEnabled(nodeClass *v1beta1.AKSNodeClass) *bool {
+	if nodeClass == nil || nodeClass.IsUltraSSDEnabled() == false {
+		return nil
+	}
+	return lo.ToPtr(nodeClass.IsUltraSSDEnabled())
+}
+
```
comtalyst: Given this field affects instance filtering owned by Karpenter, Karpenter wants to own it and give Machine API a clear `true` or `false`.

Karpenter wouldn't want to give Machine API an unclear `nil` if Karpenter already assumes that it is `false` through instance filtering.

comtalyst: Let's also make this point of defaulting in each layer clear in the design doc.

designs/0012-ultra-ssd-support.md
```diff
@@ -0,0 +1,156 @@
...
+
+**Author:** @pablotrivino
+
+**Last updated:** June 1, 2026
+
+**Status:** Proposed
+
+## Overview
+
+AKS supports Azure Ultra Disks by enabling Ultra SSD on the cluster or on a node pool at creation time with `--enable-ultra-ssd`. Nodes created from that cluster or node pool can then attach Persistent Volumes backed by the `UltraSSD_LRS` storage class.
```
comtalyst: Worth clarifying in the design doc the differences between cluster-level and pool-level API in AKS, and how Karpenter support here build on top of that.
E.g., if it is enabled at cluster-level, what will happen to the pools that don't enable it, vice-versa? If it is really just systempool and likely doesn't affect anything in Karpenter layer, then still worth a note.

wdarko1:
> > Is there any way to make this per-nodepool? I expect many clusters will have different Nodepools for different use cases + may prefer that level of granularity. Maybe checking with @wdarko1 based on current customer requirements could be useful.
> 
> I feel like using the AKSNodeClass is the pattern that's already established. NodePools seem to be more about generic Karpenter scheduling specs and NodeClass about provider specific stuff. The NodePool also references the NodeClass so cx still has some level of granularity by swapping around the NodeClass that their NodePool points to. Will ask @wdarko1 for his thoughts.
> 
> > Are the other features we have on AKSNodeClass also configurable per-agentpool in AKS? Or are they cluster-wide?
> 
> [FIPS](https://learn.microsoft.com/en-us/azure/aks/enable-fips-nodes?pivots=azure-cli), [LocalDNS](https://learn.microsoft.com/en-us/azure/aks/localdns-custom?tabs=enable), and [Artifact Streaming](https://learn.microsoft.com/en-us/azure/aks/artifact-streaming) all operate at nodepool level in AKS. In particular, FIPS seems the most similar, since you can enable it at cluster creation time for the system pool but requires opt-in for subsequent nodepools.
> 
> > Should we block disablement after enablement to align with current AKS behavior?
> 
> AKS RP does reject attempts to change the UltraSSD settings on an existing pool today. In theory, if someone tries to edit the field on the NodeClass CR, we should consider the node as drifted and replace it.
> 
> > Should we only allow ultra SSD to be enabled in Karpenter if it's enabled on the cluster? Basically introducing a validation here based on cluster state.
> 
> I think No. Today in AKS you can make a cluster without enabling ultra ssd at creation and add an ultra-ssd enabled nodepool to it.

Agree that customers should be able to select UltraSSD at the nodepool level. Customers can use multiple `AKSNodeClass`s IIRC so customers can just point their nodepools to a specific AKSNodeClass when needed and still achieve a node/nodepool level UltraSSD enablement.  @PabloTriv @rakechill

designs/0012-ultra-ssd-support.md
PabloTriv: Added a note to the design doc in the cx experience section

designs/0012-ultra-ssd-support.md
PabloTriv: I added a reference to this doc https://learn.microsoft.com/en-us/azure/aks/use-ultra-disks. AKS will reject requests for cluster or nodepool creations that have `--enable-ultra-ssd` but whose SKUs don't have UltraSSD available in the specified regions

pkg/providers/instance/aksmachineinstancehelpers.go
PabloTriv: Updated it for machine, which mirrors AKS behavior. Today in AKS though, vmss where UltraSSD == false leave the property as nil, so I'm thinking we do the same for the bootstrappingclient path.

rakechill (commented):
design lgtm, will approve the PR once CI/E2E is complete.

pkg/apis/v1alpha2/aksnodeclass.go
```diff
@@ -116,6 +129,9 @@ type AKSNodeClassSpec struct {
 	// https://learn.microsoft.com/en-us/azure/aks/custom-node-configuration
 	// +optional
 	LinuxOSConfig *LinuxOSConfiguration `json:"linuxOSConfig,omitempty"`
+	// ultraSSD enables Ultra SSD for the provisioned nodes.
+	// +optional
+	UltraSSD *UltraSSD `json:"ultraSSD,omitempty"`
```
matthchr: We didn't put `OSDiskSizeGB` there but I wonder if we want a `Disk` or `Storage` section here, so we can group storage-related capabilities in the same way we do for GPU/Linux/etc, rather than `UltraSSD` which is too narrowly scoped (probably) to ever get new fields:

```
storage:
  enableUltraSSD:
```

or somthing?

pkg/providers/instancetype/instancetypes.go
```diff
@@ -238,9 +240,19 @@ func (p *DefaultProvider) instanceTypeZones(sku *skewer.SKU) sets.Set[string] {
 // offering, you can do the following thanks to this invariant:
 //
 //	offering.Requirements.Get(v1.TopologyLabelZone).Any()
-func (p *DefaultProvider) createOfferings(sku *skewer.SKU, offeringZones sets.Set[string]) cloudprovider.Offerings {
+func (p *DefaultProvider) createOfferings(sku *skewer.SKU, offeringZones sets.Set[string], params *instanceTypeParameters) cloudprovider.Offerings {
 	offerings := []*cloudprovider.Offering{}
 	for zone := range offeringZones {
+		if params.UltraSSDEnabled {
+			if zone == "0" && !sku.IsUltraSSDAvailableWithoutAvailabilityZone() {
```
matthchr: Do we have examples where ultrassd is enabled in some zones but not others? I assume this must happen given skewer supports it but might be good to understand what exactly that looks like.

designs/0012-ultra-ssd-support.md
```diff
@@ -0,0 +1,170 @@
...
+### Non-Goals
+
+- Adding provider-managed scheduling controls beyond offerings filtering, such as automatic Requirements, labels, taints, or tolerations.
+- Automatically steering Ultra SSD workloads onto Ultra SSD-capable nodes.
+
+## Decisions
+
+### Decision 1: Where should Ultra SSD be configured?
+
+#### Add a strongly typed field to `AKSNodeClass`
```
matthchr: Did you consider a label?

```
		scheduling.NewRequirement(v1beta1.LabelSKUStoragePremiumCapable, corev1.NodeSelectorOpIn, fmt.Sprint(sku.IsPremiumIO())),
		scheduling.NewRequirement(v1beta1.LabelSKUAcceleratedNetworking, corev1.NodeSelectorOpIn, fmt.Sprint(sku.IsAcceleratedNetworkingSupported())),
```

Are labels, and feel _very_ similar to me to ultraSSD. I also think doing a label would make the implementation a bit cleaner (because `computeRequirements` already takes `params *instanceTypeParameters` and the SKU, so we could do all of the filtering for ultraSSD+ Zones there).

It also means that a workload could ask for ultraSSD on the workload, rather than having to have a totally separate NodePool.

So I am wondering why not a well known label rather than a field on AKSNodeClass.

PabloTriv: I wrote a summary of the points we discussed in the labels vs nodeclass file. My biggest gripe with UltraSSD being a label is the precedent for filtering vs actively enabling. With this approach, we add the label to offerings that support UltraSSD, like we do with other well-known labels.

The issue is that, just because an offering supports UltraSSD, it does not necessarily mean we want to enable it. This is different from other features handled by Well-Known Labels, which either only advertise an offering as capable of something (like v1beta1.LabelSKUStoragePremiumCapable) or just always enable it by default if the offering supports it (like v1beta1.LabelSKUAcceleratedNetworking).

We can't just enable it by default since this feature specifically will [affect pricing] (https://learn.microsoft.com/en-us/azure/virtual-machines/disks-types?utm_source=chatgpt.com#:~:text=Ultra%20Disks%20VM%20reservation%20fee), so we need to check if the nodeclaim is actually requesting it, not the selection.

In other words, the label on the offering means "this offering supports Ultra SSD" whereas on the nodeclaim it means "this nodeclaim wants nodes with it enabled". When the node gets labels from the offering, it's going to be unclear if Ultra SSD is on or not. 

The implementation for actually enabling it isn't too bad since we just check if the nodeclaim has it but it does feel slightly messier to me than just adding it to the nodeclass. In the doc, I propose that we make something a label if it affects scheduling requirements AND is something we can just always enable by default on the node if supported.

PabloTriv: https://github.com/Azure/karpenter-provider-azure/pull/1775 <- nodeclass implementation for ref

.github/workflows/e2e.yaml
```diff
@@ -7,6 +7,10 @@ on:
       suite:
         type: string
         required: true
+      focus:
```
matthchr: minor: Consider splitting this into separate PR

designs/0012-(tmp)-ultra-ssd-support.md
```diff
@@ -0,0 +1,169 @@
+# Ultra SSD Support for NAP
```
matthchr: minor: presumably will remove the tmp in this file name? Meaning to split to diff PR or?

.github/workflows/e2e.yaml
matthchr: Ah you siad you'd delete this. That's fine too

pkg/apis/v1alpha2/crd_validation_cel_test.go
```diff
@@ -953,8 +955,22 @@ var _ = Describe("CEL/Validation", func() {
 	})
 	Context("Labels", func() {
 		It("should allow well known label exceptions", func() {
+			knownValueLabelLabels := sets.New(
```
matthchr: I dont think this is right? Prior to your PR this test actually passes with all `karpv1.WellKnownLabels` except NodePool - so you shouldn't need to exclude any here, AFAIK?

pkg/providers/azclient/aksmachinesheaderbatch/batchkey.go
```diff
@@ -31,7 +31,7 @@ func determineBatchKey(item *aksMachineCreatePayload) (string, error) {
 		return "", fmt.Errorf("nil payload, machine body, or properties")
 	}
 
-	template := buildSharedAKSMachineTemplate(*item.machineBody.Properties)
+	template := buildSharedAKSMachineTemplate(*item.machineBody.Properties, item.machineBody.Zones)
```
matthchr: minor, style: would it make sense to just pass `*item.machineBody` here rather than 2 parts of it?

pkg/providers/azclient/aksmachinesheaderbatch/batchkey.go
```diff
@@ -69,8 +69,15 @@ func buildSharedAKSMachineTemplate(aksMachineProperties armcontainerservice.Mach
...
 	aksMachineProperties.Status = nil
 
-	return armcontainerservice.Machine{
+	machine := armcontainerservice.Machine{
 		// Dropping all fields outside of properties. They are per-machine/read-only by default.
 		Properties: &aksMachineProperties,
 	}
+
+	if aksMachineProperties.Hardware != nil {
+		if ultraSSD := aksMachineProperties.Hardware.UltraSsdEnabled; ultraSSD != nil && *ultraSSD {
```
matthchr: minor, style: I think there's a `lo` helper to unwrap `ultraSSD` that might save you a nil check.

`lo.FromPtr(aksMachineProperties.Hardware.UltraSsdEnabled)`

maybe?

pkg/providers/instance/vminstance.go
```diff
@@ -731,6 +742,15 @@ func (p *DefaultVMProvider) createVirtualMachine(ctx context.Context, opts *crea
 	return &createResult{Poller: poller, VM: vm}, nil
 }
 
+func isUltraSSDRequested(nodeClaim *karpv1.NodeClaim) bool {
+	for _, requirement := range nodeClaim.Spec.Requirements {
```
matthchr: I think there's a better way to check this, w/ something like: `Requirements.Get(v1beta1.LabelUltraSSD)` at least? And you may be able to use `Any()` to help extract yet further.

pkg/providers/instance/vminstance.go
```diff
@@ -551,6 +551,7 @@ type createVMOptions struct {
 	UseSIG              bool
 	DiskEncryptionSetID string
 	NodePoolName        string
+	UltraSsdEnabled     bool
```
matthchr: very minor nit: 
```suggestion
	UltraSSDEnabled     bool
```

pkg/providers/instance/vminstance.go
```diff
@@ -731,6 +742,15 @@ func (p *DefaultVMProvider) createVirtualMachine(ctx context.Context, opts *crea
 	return &createResult{Poller: poller, VM: vm}, nil
 }
 
+func isUltraSSDRequested(nodeClaim *karpv1.NodeClaim) bool {
+	for _, requirement := range nodeClaim.Spec.Requirements {
+		if requirement.Key == v1beta1.LabelUltraSSD && requirement.Operator == v1.NodeSelectorOpIn && len(requirement.Values) == 1 && requirement.Values[0] == "true" {
```
matthchr: AI:

•  vminstance.go:745-751  enables Ultra SSD only for  Operator=In  with exactly one value  "true" . A semantically-equivalent  NotIn ["false"]  requirement makes the scheduler land only on ultra-capable offerings (offering filtering excludes  false ), yet provisioning returns  false  and the node is created without Ultra SSD. Silent mismatch. Documented contract is  In ["true"]  (design 0012), so it's an edge case, but deriving enablement from the selected offering's label would be more robust. Advocate conceded this as undefendable.

designs/0012-ultra-ssd-support.md
matthchr: > In other words, the label on the offering means "this offering supports Ultra SSD" whereas on the nodeclaim it means "this nodeclaim wants nodes with it enabled". When the node gets labels from the offering, it's going to be unclear if Ultra SSD is on or not.

This makes a lot of sense to me as a gripe and I agree it is not right. I was expecting that it would be:

1. If zone does not support ultraSSD, `ultrassd in [false]`
2. If zone does support it, `ultraSSD in [true, false]`.

Since we are reporting what is possible not what is _chosen_? Does that not work for some reason?

pkg/providers/instancetype/instancetypes.go
```diff
@@ -254,6 +255,7 @@ func (p *DefaultProvider) createOfferings(sku *skewer.SKU, offeringZones sets.Se
 				scheduling.NewRequirement(v1beta1.AKSLabelPriority, corev1.NodeSelectorOpIn, v1beta1.PriorityRegular),
 				scheduling.NewRequirement(corev1.LabelTopologyZone, corev1.NodeSelectorOpIn, zone),
 				scheduling.NewRequirement(v1beta1.LabelPlacementScope, corev1.NodeSelectorOpIn, placementScope),
+				scheduling.NewRequirement(v1beta1.LabelUltraSSD, corev1.NodeSelectorOpIn, fmt.Sprint(isUltraSSDAvailable(sku, zone))),
```
matthchr: I observe that right now, every one of these labels is also specified at the instanceType level in `computeRequirements`, but now `LabelUltraSSD` isn't.

I am not sure if it's actually _required_ for correctness but it seems maybe goodness to propagate it there too?

designs/0012-ultra-ssd-support.md
matthchr: Or rather, _at the instancetypes/offerings level_ we are reporting what is possible not what is chosen.

In Machine/VM provider codepath (when we have a realized NodeClaim), a value has _been_ chosen, so there we should just read what was chosen.

Then the question maybe becomes "well what is the default if we present both options to core and the user didn't pick one" - to which I think the answer is, _we_ can pick one when we provision if one has not been picked. We'll be able to tell this case because the NodeClaim will have `in [true, false]` still I _think_. It won't be copied by any of the default copies because they do `labels = labelspkg.GetAllSingleValuedRequirementLabels(instanceType.Requirements)` and the like. So we can narrow it on the claim ourselves during provisoning (somewhat like you are already).

I do think you end up needing to inform core that you have chosen a default value when you do that, by setting it on the NodeClaim in `vmInstanceToNodeClaim` and the machine equivalent.

See what we do for zone/placementscope here, where there are "options" when the claim initially comes to us, we realize a particular option, and we inform core of that realization. I think this is the same thing because it's on the offering rather than the higher level instanceType.

matthchr: Though actually one problem with what I am saying here: maybe _offerings_ are supposed to not have choices, the choices are supposed to be the offerings themselves.

If that is the case, it would mean rather than `ultraSSD in [true, false]` we would return 1 offering `ultraSSD in [true]` and one `ultraSSD in [false]`. This _works_ but it means extra objects created all the time and there is a perf cost. How many VM sizes support ultraSSD? Like what % of the fleet? If we're talking basically doubling the number of offerings and that's the only/correct way to do it (I am not sure I will need to look more and/or talk with Alex), that may be an argument for NodeClass in this case.

I hadn't really realized this, the "zonal-affinity" of this feature means it's more like zone than it is like other labels.

pkg/apis/v1alpha2/aksnodeclass.go
PabloTriv: Moved to well-known label

designs/0012-ultra-ssd-support.md
PabloTriv: Made the Offerings support [true, false] and derive the final value from the NodeClaim itself. Ambiguity will just lead to us using false.

It does differ slightly from other Offering-level settings like you pointed out. Values for Zone and Capacity, for example, get derived from the Selection and only come with one choice. UltraSSD would be the first deriving it from the NodeClaim since the Offering would leave the choice up to the user.  I don't think this is a bad thing though, but it does change the idea of what an Offering represents a little.

I think this deviation is necessary though, especially if we set this precedent for future features. Otherwise each feature will greatly expand the Offerings list by `len * <number of feature permutations>` and degrade perf.

designs/0013-ultra-ssd-support.md
tallaxes, line 1: It may be worth capturing why we even need to opt in explicitly, rather than just enabling UltraSSD compatibility unconditionally for SKUs/zones that support it. I was wondering whether the latter has any downsides, and it seems the answer is Yes, and it is [Ultra Disks VM reservation fee](https://learn.microsoft.com/en-us/azure/virtual-machines/disks-types#ultra-disks-vm-reservation-fee). Is this it, or is there anything else?

designs/0013-ultra-ssd-support.md
```diff
@@ -0,0 +1,164 @@
...
+Arguments for this option:
+
+- it is a provisioning feature, not a schedulable label,
+- it aligns with the current `AKSNodeClass` design pattern.
+
+The user expectation of “default false” is still satisfied. If `spec.ultraSSD` or `spec.ultraSSD.enabled` is omitted, the effective value is disabled.
+
+We are not choosing this option because it makes Ultra SSD a NodeClass-level pool shape even though availability is offering-specific and users need to express the requirement through scheduling constraints.
+
+#### Option B: Use a Well-Known Label
```
tallaxes, line 80: What exactly happens when the corresponding requirement is added/changed/removed? I think drift, but may be worth thinking through exactly what kind.

Also, worth checking that just adding support for this requirement (without adding requirement itself to a NodePool) won't cause drift of existing nodes. I don't think it will ... The existing Nodes won't have the label, but the new ones will?

designs/0012-ultra-ssd-support.md
tallaxes: Was thinking about multi-valued vs separate offerings, and it seems like multi-valued should be fine - even though a bit unusual. Separate offerings would be preferable if UltraSSD affects price, has independently failing availability, needs distinct capacity accounting, or should participate in generic ranking/topology logic. I don't think any of this applies (though it may be worth capturing the tradeoffs in the design).

test/suites/integration/ultrassd_test.go
```diff
@@ -0,0 +1,104 @@
...
+)
+
+var _ = Describe("UltraSSD", func() {
+	It("should have UltraSSD disabled by default", func() {
+		deployment := coretest.Deployment(coretest.DeploymentOptions{Replicas: 1})
+		env.ExpectCreated(nodeClass, nodePool, deployment)
+		pods := env.EventuallyExpectHealthyDeployment(deployment)
+
+		env.EventuallyExpectInitializedNodeCount("==", 1)
+		node := env.GetNode(pods[0].Spec.NodeName)
```
tallaxes: EventuallyExpectInitializedNodeCount returns the Node, so could use that instead of looking up from the pod?

test/suites/integration/ultrassd_test.go
```diff
@@ -0,0 +1,104 @@
...
+	karpv1 "sigs.k8s.io/karpenter/pkg/apis/v1"
+	coretest "sigs.k8s.io/karpenter/pkg/test"
+
+	"github.com/Azure/karpenter-provider-azure/pkg/apis/v1beta1"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("UltraSSD", func() {
+	It("should have UltraSSD disabled by default", func() {
```
tallaxes: "... even on UltraSSD-capable SKU"? Not sure which SKU we land on, but may need to test both? And possibly the same for explicit disablement. Though maybe much of this is already covered by acceptance tests.

test/suites/integration/ultrassd_test.go
```diff
@@ -0,0 +1,104 @@
...
+	corev1 "k8s.io/api/core/v1"
+	karpv1 "sigs.k8s.io/karpenter/pkg/apis/v1"
+	coretest "sigs.k8s.io/karpenter/pkg/test"
+
+	"github.com/Azure/karpenter-provider-azure/pkg/apis/v1beta1"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("UltraSSD", func() {
```
tallaxes: We should probably also have an actual UltraSSD_LRS PVC test ... Edit: experimenting with this in https://github.com/Azure/karpenter-provider-azure/pull/1797

go.sum
```diff
@@ -369,12 +369,14 @@ golang.org/x/crypto v0.0.0-20220722155217-630584e8d5aa/go.mod h1:IxCIyHEi3zRg3s0
 golang.org/x/crypto v0.17.0/go.mod h1:gCAAfMLgwOJRpTjQ2zCCt2OcSfYMTeZVSRtQlPC7Nq4=
 golang.org/x/crypto v0.53.0 h1:QZ4Muo8THX6CizN2vPPd5fBGHyogrdK9fG4wLPFUsto=
 golang.org/x/crypto v0.53.0/go.mod h1:DNLU434OwVakk9PzuwV8w62mAJpRJL3vsgcfp4Qnsio=
+golang.org/x/crypto v0.53.0 h1:QZ4Muo8THX6CizN2vPPd5fBGHyogrdK9fG4wLPFUsto=
+golang.org/x/crypto v0.53.0/go.mod h1:DNLU434OwVakk9PzuwV8w62mAJpRJL3vsgcfp4Qnsio=
```
tallaxes: Duplicate?

test/suites/integration/ultrassd_test.go
tallaxes, line 1: Maybe under storage, rather than integration (which is already very large)?

tallaxes (dismissed):
LGTM. Left some questions and ideas re a bit more E2E test coverage, but nothing blocking.

designs/0013-ultra-ssd-support.md
PabloTriv: The list of UltraSSD limitations is here. From https://learn.microsoft.com/en-us/azure/virtual-machines/disks-types#ultra-disk-limitations:

The following list contains Ultra Disk's limitations:
- Ultra Disks can't be used as an OS disk or with Azure Compute Gallery.
- Currently, Ultra Disks only support Single VM and Availability zone infrastructure options as locally redundant storage (LRS). Ultra Disks don't support zone redundant storage (ZRS).
- Ultra Disks don't support availability sets.
- Ultra Disks don't support disk caching.

These shouldn't pose a problem for an always-opt-in approach, but I still think we should go for explicit enablement. There is that small reservation fee you linked above, and explicit enablement would also be more aligned with what AKS does. It's also less likely we'll encounter some kind of undocumented compatibility issue, and from a user perspective enabling it via a label is a very accessible approach.

designs/0012-ultra-ssd-support.md
PabloTriv: The only one that might apply is price because of the reservation fee. But an Offering having [true, false] just means that offering is capable of enabling it, not that it will be enabled as that is left up to the user. So in that sense it does not affect price. In other words I think not having separate Offerings would be fine.

designs/0012-(tmp)-ultra-ssd-support.md
PabloTriv: removed the temp

.github/workflows/e2e.yaml
PabloTriv: removed it

designs/0013-ultra-ssd-support.md
PabloTriv: Changing the requirement on the NodePool will cause NodePoolDrift to occur. Core will detect that the NodePool hash does not match what the nodeclaim has and mark it as drifted. Changing on the workload won't have an effect though.

go.sum
PabloTriv: De-duped

designs/0013-ultra-ssd-support.md
tallaxes: Hmm, so just adding it to the workload won't drift/redeploy the nodes with UltraSSD?

Otherwise sounds like the only potentially "unexpected" scenario we should watch out for is that adding actual requirement with "false" (not terribly likely) would still cause drift. And the (temporary) labeling inconsistency between existing and new nodes (which, if true, is probably the case for any new requirement we add).

tallaxes (dismissed):
Additional E2E tests in https://github.com/Azure/karpenter-provider-azure/pull/1797 tests passed, can decide to include them now or in followup

designs/0013-ultra-ssd-support.md
PabloTriv: Note: Since the well-known label is defined in the nodeSelector, it should trigger the deployment to recreate the pod if edited, making the new pod go through scheduling again and either land on an ultrassd-enabled node or provision one. With a standalone pod (without a replicaset or a deployment) then yes I think pod might get stuck on its existing node even if we change the requirement.

Adding the false requirement would cause drift, yes. And we will have some inconsistent labeling for a bit. (Although we could just not label nodes as false if we want to change it).

tallaxes (dismissed):

designs/0013-ultra-ssd-support.md
matthchr: Won't adding it to the workload result in the workload pod being rescheduled, which may match a new node (if the requirement is changed?)

not specified to false will result in possibly just rescheduling on the same node.
not specified (or false) to true will result in it not matching the existing node (right?), so it will have to make a new node?

matthchr: Oh sorry for some reason my GitHub hadn't shown Pablo's comment above -- yes he is saying the same thing.

designs/0013-ultra-ssd-support.md
```diff
@@ -0,0 +1,195 @@
...
+
+This is a relatively straightforward solution. Server-side, the zones get replaced by the header zones anyway. The drawback with this approach is that it could lead to unexpected behavior if the server-side behavior ever changes.
+
+##### Option B: Group Requests Into Batches With the Same Zone
+Instead of just populating the shared machine template with a zone, we edit the function that builds the template to include the zone when Ultra SSD is enabled. This means that machine requests with Ultra SSD enabled will be placed in batches with requests that also have Ultra SSD enabled and are in the same zone.
+
+This approach removes uncertainty about specifying the zone, as the shared template will have the same zone as all requests in the batch. The drawback to this approach is we might end up with a lot of batches, effectively negating the advantages we gain from batching. This can happen if, for some reason, we enable Ultra SSD and end up picking multiple zones.
+
+##### Decision: Option B
+Option B essentially adds the zone to the shared template, allowing us to both consider the zone when grouping requests and specify a zone for validation. This option is good because it:
```
matthchr, line 178: Will we take this option for all requests, or only if ultraSSD is enabled?

I assume for all? (I guess I will see in impl later).

matthchr: No it looks like it's just for UltraSSD - that's probably safer anyway.

pkg/providers/azclient/aksmachinesheaderbatch/batchkey.go
```diff
@@ -46,7 +47,7 @@ func determineBatchKey(item *aksMachineCreatePayload) (string, error) {
 // buildSharedAKSMachineTemplate returns a Machine containing only the fields shared across all
 // machines in a batch, with per-machine and read-only fields zeroed.
 // Takes MachineProperties by value so the caller's original is never mutated.
```
matthchr: minor: This comment isn't quite accurate anymore.

Also since now we're:
1. copying the properties ourselves.
2. constructing a new machine and never changing the input parameter.

We can (I think?) just pass a `*armcontainerservice.Machine` here rather than by value.

pkg/providers/instance/aksmachineinstanceutils.go
```diff
@@ -108,6 +108,7 @@ func BuildNodeClaimFromAKSMachineTemplate(
 	labels[karpv1.CapacityTypeLabelKey] = capacityType
 	labels[v1beta1.AKSLabelScaleSetPriority] = KarpCapacityTypeToScaleSetPriorityLabel[capacityType]
 	labels[v1beta1.AKSLabelPriority] = KarpCapacityTypeToPriorityLabel[capacityType]
+	labels[v1beta1.LabelUltraSSD] = fmt.Sprint(aksMachineTemplate.Properties != nil && aksMachineTemplate.Properties.Hardware != nil && lo.FromPtr(aksMachineTemplate.Properties.Hardware.UltraSsdEnabled))
```
matthchr: minor, style: does this deserve a helper similar to `instance.GetUltraSSDEnabled`?

pkg/providers/instance/vminstance.go
```diff
@@ -678,6 +680,15 @@ func setVMPropertiesSecurityProfile(vmProperties *armcompute.VirtualMachinePrope
 	}
 }
 
+func setVMPropertiesAdditionalCapabilities(vmProperties *armcompute.VirtualMachineProperties, ultraSsdEnabled bool) {
```
matthchr: very minor, style: 
```suggestion
func setVMPropertiesAdditionalCapabilities(vmProperties *armcompute.VirtualMachineProperties, ultraSSDEnabled bool) {
```

To be consistent with how we're casing it elsewhere (see `UltraSSDEnabled` above ln 554 for example)

pkg/providers/instancetype/ultrassd_test.go
```diff
@@ -0,0 +1,105 @@
...
+			zone: "westus3-1",
+			want: []string{"false"},
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			t.Parallel()
+			if diff := cmp.Diff(test.want, ultraSSDOptions(test.sku, test.zone)); diff != "" {
+				t.Errorf("unexpected UltraSSD options (-want, +got):\n%s", diff)
```
matthchr: minor: should maybe use gomega for assertions, given we generally do everywhere else (even in non-ginkgo tests?)

pkg/providers/instancetype/instancetypes.go
```diff
@@ -545,3 +548,18 @@ func nvmeDiskSizeInMiB(s *skewer.SKU) (int64, error) {
 	const selector = "NvmeDiskSizeInMiB"
 	return s.GetCapabilityIntegerQuantity(selector)
 }
+
+func ultraSSDOptions(sku *skewer.SKU, zone string) []string {
+	z := zones.MakeARMZonesFromAKSLabelZone(zone)
+	switch len(z) {
+	case 0:
```
matthchr, line 580: curious: Do we have a test for this regional case?

test/suites/integration/ultrassd_test.go
tallaxes: Deferring (though may still be good to check if this is needed @PabloTriv)

test/suites/storage/ultrassd_test.go
```diff
@@ -0,0 +1,192 @@
...
+		"kind":              "managed",
+		"skuName":           "UltraSSD_LRS",
+	}
+	pvc := coretest.PersistentVolumeClaim(coretest.PersistentVolumeClaimOptions{
+		StorageClassName: &storageClass.Name,
+		Resources: corev1.VolumeResourceRequirements{Requests: corev1.ResourceList{
+			corev1.ResourceStorage: resource.MustParse("4Gi"),
+		}},
+	})
+	pod := env.Pod(coretest.PodOptions{
```
matthchr: https://github.com/Azure/karpenter-provider-azure/pull/1686

Can we not use deployment here?

pkg/providers/instancetype/instancetypes.go
```diff
@@ -545,3 +548,18 @@ func nvmeDiskSizeInMiB(s *skewer.SKU) (int64, error) {
...
+		if sku.IsUltraSSDAvailableWithoutAvailabilityZone() {
```
matthchr, line 581: 🤖 _Automated review comment (bot)_

**[High] Regional UltraSSD is advertised, but the AKS Machine API path creates a machine with no zone.**

When `zone` maps to regional (`len(z)==0`), `ultraSSDOptions` returns `["true","false"]` if `IsUltraSSDAvailableWithoutAvailabilityZone()`. A NodeClaim with `sku-storage-ultra-ssd In ["true"]` + `placement-scope In ["regional"]` can then schedule onto a regional offering. In Machine API mode, `buildAKSMachineTemplate` sets `Hardware.UltraSsdEnabled=true` with empty `Zones`, and the zone-preservation fix in `batchkey.go` only helps when zones exist. Per the design doc (§Batching), Machine API validation rejects UltraSSD without at least one zone — so this passes scheduling and fails at create time.

Please confirm whether Machine API accepts regional UltraSSD without zones. If not, don't advertise regional UltraSSD in Machine-API mode (or supply a zone), and add a regional Machine-API test — the E2E regional case appears to exercise VM mode only.

pkg/providers/azclient/aksmachinesheaderbatch/batchkey.go
```diff
@@ -69,8 +71,17 @@ func buildSharedAKSMachineTemplate(aksMachineProperties armcontainerservice.Mach
...
-	return armcontainerservice.Machine{
+	machine := armcontainerservice.Machine{
 		// Dropping all fields outside of properties. They are per-machine/read-only by default.
 		Properties: &aksMachineProperties,
 	}
+
+	// If UltraSSD is enabled, we need to preserve the zones for the batch key to be correct.
+	// This is because UltraSSD requires Zones to be specified during validation, so leaving this out
+	// will cause the shared machine to be rejected as invalid.
+	if lo.FromPtr(aksMachineProperties.Hardware.UltraSsdEnabled) {
```
matthchr: 🤖 _Automated review comment (bot)_

**[Medium] `Hardware` is dereferenced without a nil check.**

`lo.FromPtr(aksMachineProperties.Hardware.UltraSsdEnabled)` panics if `Hardware == nil`. The callers (`determineBatchKey`, `executor.executeBatch`) only guard `Properties`, not `Hardware`. It's not reachable via the normal create path today (`buildAKSMachineTemplate` always sets `Hardware`), but this is an unguarded invariant inside a shared utility. Suggest `lo.FromPtr(aksMachineProperties.Hardware).UltraSsdEnabled`-style nil-safety or an explicit `Hardware != nil` check.

pkg/providers/azclient/aksmachinesheaderbatch/batchkey.go
```diff
@@ -46,7 +47,7 @@ func determineBatchKey(item *aksMachineCreatePayload) (string, error) {
 // buildSharedAKSMachineTemplate returns a Machine containing only the fields shared across all
 // machines in a batch, with per-machine and read-only fields zeroed.
 // Takes MachineProperties by value so the caller's original is never mutated.
-func buildSharedAKSMachineTemplate(aksMachineProperties armcontainerservice.MachineProperties) armcontainerservice.Machine {
+func buildSharedAKSMachineTemplate(aksMachine armcontainerservice.Machine) armcontainerservice.Machine {
```
matthchr: 🤖 _Automated review comment (bot)_

**[Architectural] Generic batch-key layer now carries UltraSSD-specific knowledge.**

Changing the signature from `MachineProperties` to full `Machine` and special-casing `Hardware.UltraSsdEnabled` to preserve top-level `Zones` couples this generic batching abstraction to one feature's validation quirk. Elsewhere the abstraction treats zones as header/per-machine data (see `header.go`), and `executor_test.go` asserts body zones are normally nil. Correctness looks fine, but consider a generic "requires zone in shared template" extension point rather than baking the feature in here.

Also: the doc comment just above still says _"Takes MachineProperties by value so the caller's original is never mutated"_ — now stale since the param is a `Machine`.

pkg/providers/instancetype/instancetypes.go
```diff
@@ -69,6 +69,7 @@ type instanceTypeParameters struct {
 	ArtifactStreamingEnabled bool
 	FIPSMode                 v1beta1.FIPSMode
 	LocalDNSEnabled          bool
+	UltraSSDEnabled          bool
```
matthchr: 🤖 _Automated review comment (bot)_

**[Low] `UltraSSDEnabled` added to `instanceTypeParameters` but never set or read.**

This field is part of the cache-key struct but `List` never populates it and nothing reads it — it's always the zero value. Looks like a leftover from the rejected AKSNodeClass-field design (UltraSSD is NodeClaim-driven instead). Harmless today, but it can confuse future cache-invalidation work. Suggest removing it.

designs/README.md
```diff
@@ -8,6 +8,8 @@ These designs are written but not implemented.
 
 <!-- Please add newer designs at the top of this list -->
 
+- [Ultra SSD support](./0012-ultra-ssd-support.md)
```
matthchr: 🤖 _Automated review comment (bot)_

**[Low] Design index link and section are both wrong.**

This links to `./0012-ultra-ssd-support.md`, but the file added in this PR is `0013-ultra-ssd-support.md` (broken link). It's also under _"written but not implemented"_, while the doc itself has `Status: Implemented`. Please point the link at `./0013-ultra-ssd-support.md` and move it under the **Completed** section.

pkg/providers/azclient/aksmachinesheaderbatch/batchkey.go
matthchr: > but consider a generic "requires zone in shared template" extension point rather than baking the feature in here.

TBH not 100% sure what it means by this... possibly can just close this one.

pkg/providers/instancetype/instancetypes.go
tallaxes: There is a regional E2E now (and it passed on regular provisioning, did not try with Machine)

pkg/providers/instancetype/ultrassd_test.go
tallaxes: Fixed

test/suites/storage/ultrassd_test.go
tallaxes: We can; changed to Deployment

pkg/providers/instancetype/instancetypes.go
PabloTriv: RP does indeed reject validation if zones are empty. Although some docs point to it being regional https://learn.microsoft.com/en-us/azure//virtual-machines/disks-enable-ultra-ssd?utm_source=chatgpt.com&tabs=azure-portal

pkg/providers/instancetype/instancetypes.go
PabloTriv: Alex added one in ultrassd_test.go

pkg/providers/azclient/aksmachinesheaderbatch/batchkey.go
PabloTriv: Added in a generic needs zone check since I didn't see your follow up comment. I'll leave it since it could be useful if some feature down the line also needs zones

pkg/providers/instancetype/instancetypes.go
PabloTriv: Ah didn't refresh so I missed Alex's comment that said the same thing

matthchr (approved):

PR #1705, opened 2026-06-02
chore: sync GPU CUDA and GRID driver versions from AgentBaker

ganeshkumarashok:
Fixes # 

**Description**

The hardcoded NVIDIA GPU driver constants in `pkg/utils/gpu.go` had drifted behind AgentBaker, the upstream source of truth (per the existing `// TODO: Get these from agentbaker` comment). This PR syncs both the CUDA and GRID driver versions (and their `aks-gpu` image suffixes) to AgentBaker's current `parts/common/components.json`:

- **CUDA:** `550.144.03-20250328201547` → `580.126.09-20260126030251`
- **GRID:** `570` series — `550.144.06-20250512225043` → `570.211.01-20260522192315`

These constants feed `GetGPUDriverVersion` and `GetAKSGPUImageSHA`, which drive the GPU driver version and `aks-gpu` image SHA used by the custom-scripts bootstrap path. The reference permalink in the comment is also updated to the current AgentBaker commit.

**How was this change tested?**

* `go build ./pkg/utils/...` — passes
* `go test ./pkg/utils/` — passes (`ok github.com/Azure/karpenter-provider-azure/pkg/utils`)

**Does this change impact docs?**
- [ ] Yes, PR includes docs updates
- [ ] Yes, issue opened: #
- [x] No

**Release Note**

```release-note
Update bundled NVIDIA GPU driver versions to match AgentBaker: CUDA 580.126.09 and GRID 570.211.01.
```

PR #1706, opened 2026-06-02
feat: adding support for RTX PRO 6000 BSE v6 SKUs

MaximilianoUribe:
Fixes # 

**Description**

Adds RTX PRO 6000 BSE v6 GPU support for Karpenter by adding the new SKU inventory/GPU classification and routing driver-enabled RTX PRO SKUs to the NVIDIA GRID v20 driver image path.

Changes included:
- Adds GRID v20 driver selection for RTX PRO 6000 BSE v6 SKUs.
- Returns `GPU_DRIVER_TYPE=grid-v20` for these SKUs so bootstrap resolves `mcr.microsoft.com/aks/aks-gpu-grid-v20`.
- Sets GRID v20 to `NvidiaGridV20DriverVersion = 595.58.03` and `AKSGPUGridV20VersionSuffix = 20260609172331`.
- Adds RTX PRO 6000 BSE v6 SKUs to GPU support metadata as Ubuntu-only NVIDIA GPU SKUs.
- Adds RTX PRO 6000 BSE v6 SKUs to the known SKU inventory so they can be considered during instance type discovery.
- Keeps Karpenter instance type discovery best-effort when `skewer` cannot parse a new Azure VM size shape; the SKU is retained and only parsed labels such as `sku-series`, `sku-version`, and `sku-gpu-name` are omitted.
- Preserves existing CUDA and GRID behavior already synced in #1705.
- Adds targeted unit coverage for driver type, driver version, image suffix, SKU recognition, manufacturer, OS support, and driver-install support.

RTX PRO SKUs added:
- `standard_nc128ds_xl_rtxpro6000bse_v6`
- `standard_nc128lds_xl_rtxpro6000bse_v6`
- `standard_nc256ds_xl_rtxpro6000bse_v6`
- `standard_nc256lds_xl_rtxpro6000bse_v6`
- `standard_nc320ds_xl_rtxpro6000bse_v6`
- `standard_nc320lds_xl_rtxpro6000bse_v6`

Unit/package validation:

```bash
go test ./pkg/providers/instancetype ./pkg/utils
```

OSS Karpenter end-to-end validation (self-hosted build, `aksscriptless`, `westus2`), both GPU modes for `Standard_NC128lds_xl_RTXPRO6000BSE_v6`. In both cases `skewer` still cannot parse the VM size, and Karpenter retains the SKU (omitting only parsed labels) instead of filtering it out.

Without driver (`spec.gpu.mode: None`): NodeClaim created/launched, VM `Succeeded`, node registered `Ready`, CPU-only pod `Running`.

```text
kubectl get pod rtx-pro-provision-only -o wide
NAME                     READY   STATUS    NODE
rtx-pro-provision-only   1/1     Running   aks-rtx-pro-nodriver-test-xbwt5
```

With driver (`spec.gpu.mode: Driver`): Karpenter resolved `GPU_DRIVER_TYPE=grid-v20`, the node pulled `aks-gpu-grid-v20:595.58.03-...`, and the GRID v20 driver installed and loaded (verified on the node via `az vm run-command`).

```text
=====SMI=====
NVIDIA RTX Pro 6000 Blackwell DC-4-96Q, 595.58.03
=====GRIDV20=====
aks-gpu-grid-v20:595.58.03-20260609172331
=====DRIVERTYPE=====
GPU_DRIVER_TYPE=grid-v20
```

Net: Karpenter provisions the RTX PRO node and installs the correct GRID v20 driver (`nvidia-smi` reports `595.58.03`); skewer parses the SKU natively (`sku-gpu-name=RTXPRO6000BSE`). Full Kubernetes GPU scheduling additionally requires cluster-side setup (documented below) — once in place, the NodeClaim reaches `Ready=True` and a GPU pod runs `nvidia-smi` on the device.

Making GPU scheduling work (cluster-side prerequisite, not part of this change):

These SKUs come up with the GPU in MIG mode (`nvidia-smi -L` reports `MIG 4g.96gb`; `nvidia-smi -q` shows `MIG Mode Current: Enabled`), licensed as NVIDIA RTX Virtual Workstation. To advertise `nvidia.com/gpu` the only requirement is a **device-plugin version new enough to support Blackwell RTX PRO 6000**: `v0.16.2` fails (`Insufficient Permissions` with the default `migStrategy: none`, and `no matching profile IDs found` with `single`), while `v0.17.4` works. With a current-enough version, `migStrategy: none` (the default) works — no MIG-specific configuration is required.

With a Blackwell-aware device plugin, the node advertises `nvidia.com/gpu=1`, the NodeClaim `Initialized`/`Ready` conditions clear, and a GPU pod runs `nvidia-smi` on the device:

```text
node:      allocatable nvidia.com/gpu = 1
nodeclaim: rtx-pro-driver-test-...  READY=True
pod:       nvidia-smi -> NVIDIA RTX Pro 6000 Blackwell, Driver 595.58.03, CUDA 13.2 (MIG enabled)
```

On managed AKS/NAP the platform provides a compatible GPU device-plugin stack, so this cluster-side setup is handled there.

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
Add NVIDIA GRID v20 driver selection support for RTX PRO 6000 BSE v6 GPU SKUs.
```


ganeshkumarashok:
Want to test it before merging and stay in sync with my changes? I'm making further changes and testing the Agentbaker PR: https://github.com/Azure/AgentBaker/pull/8666/changes

pkg/utils/supported-gpus.yaml
```diff
@@ -265,6 +265,26 @@ standard_nd128isr_gb300_v6:
...
+  os: ["ubuntu"]
+standard_nc256lds_xl_rtxpro6000bse_v6:
+  gpu: nvidia
+  os: ["ubuntu"]
+standard_nc320ds_xl_rtxpro6000bse_v6:
+  gpu: nvidia
+  os: ["ubuntu"]
+standard_nc320lds_xl_rtxpro6000bse_v6:
+  gpu: nvidia
+  os: ["ubuntu"]
```
tallaxes, line 286: Is there a reason smaller sizes are not in the list? Judging just by generated pricing, there are Standard_NC16lds, Standard_NC32ds/lds and Standard_NC64ds/lds ...

pkg/providers/imagefamily/customscriptsbootstrap/provisionclientbootstrap.go
```diff
@@ -193,8 +193,15 @@ func (p *ProvisionClientBootstrap) ConstructProvisionValues(ctx context.Context)
 	}
 
 	if utils.IsNvidiaEnabledSKU(p.InstanceType.Name) {
+		driverType := models.DriverTypeCUDA
+		if utils.UseGridDrivers(p.InstanceType.Name) {
+			driverType = models.DriverTypeGRID
+		} else if utils.UseGridV20Drivers(p.InstanceType.Name) {
+			driverType = models.DriverTypeUnspecified
```
tallaxes: How does this work? Is there something that translates unspecified to the right driver on the service side?

tallaxes (dismissed):
LGTM, see a couple of comments/questions.

Can we also add a note on how Machine API is affected, maybe in the description? (Looks like only through SKU recognition/filtering? Though I am not quite sure why we don't populate GPUProfile.DriverType/Nvidia ... Is it all determined service-side? @comtalyst)

comtalyst:
@tallaxes Yes, GPU driver type determined server-side based on SKU.
AKS machine's `GPUProfile.DriverType` is for Windows only.

pkg/providers/imagefamily/customscriptsbootstrap/provisionclientbootstrap.go
matthchr: Agree this deserves at least a comment

pkg/providers/imagefamily/customscriptsbootstrap/provisionclientbootstrap.go
```diff
@@ -193,8 +193,15 @@ func (p *ProvisionClientBootstrap) ConstructProvisionValues(ctx context.Context)
...
+		}
+
 		provisionProfile.GpuProfile = &models.GPUProfile{
-			DriverType:       lo.ToPtr(lo.Ternary(utils.UseGridDrivers(p.InstanceType.Name), models.DriverTypeGRID, models.DriverTypeCUDA)),
```
matthchr, line 197: So we now never pass CUDA? Doesn't look like this is equivalent to what we had before?

pkg/providers/instancetype/instancetype.go
```diff
@@ -201,9 +201,11 @@ func computeRequirements(
 	// composites
 	requirements[v1beta1.LabelSKUName].Insert(sku.GetName())
 
-	// size parts
-	requirements[v1beta1.LabelSKUFamily].Insert(vmsize.Family)
-	requirements[v1beta1.LabelSKUSeries].Insert(vmsize.Series)
+	if vmsize != nil {
```
matthchr: Do we even allow nil VM size?

pkg/providers/instancetype/instancetype.go
```diff
@@ -242,7 +244,7 @@ func setRequirementsGPU(requirements scheduling.Requirements, sku *skewer.SKU, v
 	default:
 		return
 	}
-	if vmsize.AcceleratorType != nil {
+	if vmsize != nil && vmsize.AcceleratorType != nil {
```
matthchr: same q

pkg/providers/instancetype/instancetype.go
matthchr: maybe worth a comment on `computeRequirements` to explain the expectation/reality that vmsize can be nil sometimes and it's allowed/ok?

Were we panicing before?

pkg/providers/instancetype/instancetypes.go
```diff
@@ -165,8 +165,7 @@ func (p *DefaultProvider) List(
 	for _, sku := range p.instanceTypesInfo {
 		vmsize, err := sku.GetVMSize()
 		if err != nil {
-			log.FromContext(ctx).Error(err, "parsing VM size", "vmSize", *sku.Size)
```
matthchr, line 168: Is this actually expected? Can you expand on why? It feels like an error to me -- same with below

pkg/providers/imagefamily/customscriptsbootstrap/provisionclientbootstrap.go
```diff
@@ -193,8 +193,15 @@ func (p *ProvisionClientBootstrap) ConstructProvisionValues(ctx context.Context)
 	}
 
 	if utils.IsNvidiaEnabledSKU(p.InstanceType.Name) {
+		driverType := models.DriverTypeCUDA
```
matthchr: Unrelated to this PR and one that I probably would have put on @ganeshkumarashok's PR earlier, but do we need a test that ensures that the GRID/CUDA driver versions are in-sync with AgentBaker?

Maybe doesn't matter as much when fully on machine API...

pkg/utils/gpu.go
```diff
@@ -159,6 +175,20 @@ var ConvergedGPUDriverSizes = map[string]bool{
 	"standard_nc32ads_a10_v4":  true,
 }
 
+/* RTXPro6000GPUDriverSizes : NC_RTXPRO6000BSE_v6 (RTX PRO 6000 Blackwell Server
+Edition) SKUs require the GRID v20 (595.x) driver, published as the
+aks-gpu-grid-v20 image. All other GRID SKUs continue to use aks-gpu-grid.
+*/
+//nolint:gochecknoglobals
+var RTXPro6000GPUDriverSizes = map[string]bool{
```
matthchr: any reason this needs to be exported?

matthchr: I think `gochecknoglobals` is only needed because this is exported so you could just not export it and remove that

pkg/utils/supported-gpus.yaml
MaximilianoUribe: https://learn.microsoft.com/en-us/azure/virtual-machines/sizes/gpu-accelerated/nc-rtxpro6000-bse-v6-series?tabs=sizebasicgp%2Csizebasicco these are the ones listed in the azure page

---
**(AI-generated) update:** The BSE v6 family is actually larger than this docs page — `az vm list-skus` shows NC24/36/72/128/144/288/320 (ds/lds, 1–2 GPUs) across westus2 / southeastasia / CentralUSEUAP. With the skewer `v0.0.23` fix these all parse natively; the initial set ships in this PR, and adding more requires matching AgentBaker support per SKU.

pkg/providers/instancetype/instancetypes.go
MaximilianoUribe: the sku api cannot parse the name of these new skus, but you are right maybe this is not the best approach ... I also had an option to hardcode the parsing of this specific sku to not rely on the api, but this looks a little forced to me and not sure if this will be the only sku where this happens

---
**(AI-generated) update:** Approach changed after review. The underlying issue — `Azure/skewer` could not parse the new `NC..._xl_RTXPRO6000BSE_v6` size shape — was fixed upstream (Azure/skewer#34, released `v0.0.23` / `v2.0.3`). This PR bumps skewer to `v0.0.23` and **removes** the local parse fallback and nil-checks; these SKUs now parse natively (`sku-family/series/version/gpu-name` populated). Validated end-to-end on OSS Karpenter (`aksscriptless`) in both `gpu.mode: None` and `Driver`; PR merged in `09401f25`.

pkg/providers/instancetype/instancetype.go
MaximilianoUribe: we were indeed panicking before, this is for the case where the sku cannot parse the size and we rely on the name of the sku for provisioning 

---
**(AI-generated) update:** Approach changed after review. The underlying issue — `Azure/skewer` could not parse the new `NC..._xl_RTXPRO6000BSE_v6` size shape — was fixed upstream (Azure/skewer#34, released `v0.0.23` / `v2.0.3`). This PR bumps skewer to `v0.0.23` and **removes** the local parse fallback and nil-checks; these SKUs now parse natively (`sku-family/series/version/gpu-name` populated). Validated end-to-end on OSS Karpenter (`aksscriptless`) in both `gpu.mode: None` and `Driver`; PR merged in `09401f25`.

MaximilianoUribe: actually yeah, that seems like the right approach, just adding a fallback for this sku size, seems like this will be a trend from now on haha

---
**(AI-generated) update:** Approach changed after review. The underlying issue — `Azure/skewer` could not parse the new `NC..._xl_RTXPRO6000BSE_v6` size shape — was fixed upstream (Azure/skewer#34, released `v0.0.23` / `v2.0.3`). This PR bumps skewer to `v0.0.23` and **removes** the local parse fallback and nil-checks; these SKUs now parse natively (`sku-family/series/version/gpu-name` populated). Validated end-to-end on OSS Karpenter (`aksscriptless`) in both `gpu.mode: None` and `Driver`; PR merged in `09401f25`.

MaximilianoUribe: I added a fallback specific for these skus (that the sku api can't parse), the nil check has to stay for this case, but if not in this case it panics like before, so no change in behaviour

---
**(AI-generated) update:** Approach changed after review. The underlying issue — `Azure/skewer` could not parse the new `NC..._xl_RTXPRO6000BSE_v6` size shape — was fixed upstream (Azure/skewer#34, released `v0.0.23` / `v2.0.3`). This PR bumps skewer to `v0.0.23` and **removes** the local parse fallback and nil-checks; these SKUs now parse natively (`sku-family/series/version/gpu-name` populated). Validated end-to-end on OSS Karpenter (`aksscriptless`) in both `gpu.mode: None` and `Driver`; PR merged in `09401f25`.

pkg/utils/supported-gpus.yaml
MaximilianoUribe: Also, for these to be added, they would need to be support in Agent Baker as well.

pkg/providers/instancetype/instancetype.go
matthchr: Do you know is this a size/pattern that is public? Should we maybe just fix skewer?

See https://github.com/Azure/skewer/blob/main/vmsize.go#L19 and https://github.com/Azure/skewer/blob/main/v2/vmsize.go#L19

matthchr (approved):

ganeshkumarashok:
>These SKUs expose the GPU as a MIG-backed vGPU (nvidia-smi -L reports MIG 4g.96gb), licensed as NVIDIA RTX Virtual Workstation. To advertise nvidia.com/gpu:
> Run the NVIDIA device plugin with migStrategy: single — the GPU is MIG-partitioned, so none fails with NVML Insufficient Permissions.

@MaximilianoUribe device plugin deployment without any MIG strategy setting should just work for the default case

PR #1720, opened 2026-06-10
design: Node Image Version Pinning in Karpenter

### LINKED ISSUE #1220: Add support for setting ImageID for nodeClass

Add option to set ImageID in for nodeClass to a specific version or ID, like 202510.09.0

Currently only ImageFamily can be set which is limits control over the what image version to use. This causes issues when new images are broken and there is no option to pin the imageID to a known working version.

There is no option to work around this issue, I have in the past had to create an autoscaling group in AKS and disable Karpenter while the new imageID becomes available or a fixed version is released. 

### Additional Context

This should be set like the imageFamily 
```
nodeClasses:
  - name: default
    spec:
      imageFamily: Ubuntu2204
      imageID: 202510.09.0
or 
     imageVersion: 202510.09.0
```

### Attachments

_No response_

### Community Note

rakechill:
## Summary

Design document for adding `imageVersion` pinning support to `AKSNodeClass`, allowing customers to pin node image selection to a specific AKS node image version (e.g., `202604.24.0`).

## Motivation

Karpenter users can set `imageFamily` but cannot currently:
- Pin to a known-good version when the latest image has a regression
- Decouple node image updates from Kubernetes version upgrades
- Stay on a stable image version for a bounded period

The only workaround today is disabling Karpenter entirely and falling back to AKS autoscaling groups.

## Key Design Points

- **New `imageVersion` field** on `AKSNodeClassSpec`:
  - Pins node image selection to a specific version
  - Validated at admission via regex `^\d{6}\.\d{2}\.\d+$` (current `YYYYMM.DD.patch` format only)
- **`imageCreateDate` per image** in `status.images[]`:
  - Immutable release date stored once at reconciliation time
  - Consumers compute age on read, avoiding daily status churn
- **`ImageWithinSupportWindow` status condition**:
  - Warning-only (never blocks provisioning)
  - Not a readiness-dependent condition
- **`karpenter_image_age_days` gauge metric**:
  - Computed at scrape time from `imageCreateDate`
- **Cache key fix**:
  - `imageVersion` is included in `cacheKey()`
  - Pin changes force a fresh lookup so status reflects `.spec.imageVersion` correctly
- **Maintenance window behavior with pinning**:
  - When `imageVersion` is set, pin is authoritative
  - Reconciler skips maintenance-window update gating/override logic for latest selection
- **Drift detection**:
  - Continues to work via existing `status.images[]` vs node image comparison
- **`AKSNodeClassSpec.ImageID` cleanup**:
  - Existing hidden/unused stub is removed as part of this design
- **Alternatives considered**:
  - `do-not-disrupt` annotation
  - EKS-style `amiSelectorTerms` pattern
  - `NodePool.expireAfter`
  - All evaluated against rollback, k8s/image decoupling, and bounded stability-window goals

## Open Questions (pending Node SIG response)

- Is there a supported compatibility contract between Kubernetes version and node image version?
- If a user pins a node image version and upgrades Kubernetes, should pinning across that upgrade be allowed, blocked, or explicitly validated?
- On Kubernetes upgrades that trigger image refresh (`ImagesReady=False`), should pinned versions be re-validated against the new Kubernetes version?
- Is `ListNodeImageVersions` presence the intended compatibility signal, or is there another authoritative source?

## Related

- Closes https://github.com/Azure/karpenter-provider-azure/issues/1220


designs/0012-node-image-version-pinning.md
```diff
@@ -0,0 +1,744 @@
...
+- AKS support guidance already steers customers toward upgrading as a
+  prerequisite for case resolution when running stale images.
+
+The _anticipated_ policy model is:
+
+| Concept | Definition |
+|---|---|
+| **In-window** | Image age ≤ 90 days from release date. No restrictions; full support. |
+| **Outside-supported-window** | Image age > 90 days. Support may require upgrading to a newer image as a prerequisite for resolution. |
+| **Unsupported** | Potential future enforcement state where specific operations (e.g., scale-up) could be blocked. Not part of initial rollout. |
```
matthchr, line 32: This is something we need to talk/work with node sig on, but I saw Jorge pushing back on this and I agree with that.

I guess not critical here as we're punting on it though.

designs/0012-node-image-version-pinning.md
```diff
@@ -0,0 +1,744 @@
...
+
+### Key Policy Direction
+
+1. AKS ships node images frequently: **weekly for Linux, monthly for Windows**.
+2. Customers are expected to update regularly — via auto-upgrade channels or manual upgrades.
+3. **Rollback and pinning are supported flexibility mechanisms**, but they **do not override lifecycle expectations**. They are meant for short-term operational recovery, not long-term avoidance of updates.
+4. Within the supported window, scale-up is expected to be **unconditionally allowed** regardless of cluster state.
```
matthchr: and at least for now, outside of the supported window scale up is also allowed?

designs/0012-node-image-version-pinning.md
```diff
@@ -0,0 +1,744 @@
...
+90-day window.
+
+### What This Means for Karpenter
+
+Karpenter's node image selection should be **designed with a support
+window in mind**, even though the formal policy is not yet published.
+If we expose a version-pinning mechanism, it should:
+
+- Allow customers to pin within the supported window.
+- Surface image age as observable state (e.g., status conditions).
```
matthchr, line 60: Age, or publish date?

Is there a pattern for computing an always-increasing counter in k8s resources? I think usually you see something like: `lastUpdatedAt`, rather than `timeSinceLastUpdated`.

designs/0012-node-image-version-pinning.md
```diff
@@ -0,0 +1,744 @@
...
+- Warn — but not block — when a pinned image exceeds 90 days, based on
```
matthchr, line 61: Possibly answered later but what mechanism in particular are we warning on? Conditions are either healthy or not, right?

designs/0012-node-image-version-pinning.md
```diff
@@ -0,0 +1,744 @@
...
+        │                                    Requirements: [amd64, HyperV-v2] }
+        ▼
+NodeImageProvider.List()             (queries SIG or CIG for latest version per definition)
+        │
+        │  SIG: nodeImageVersions.List() → match by SKU → BuildImageIDSIG(..., version)
+        │  CIG: latestNodeImageVersionCommunity() → pick newest by PublishedDate → BuildImageIDCIG(..., version)
+        │
+        ▼
+AKSNodeClass.status.images[]         (reconciled by nodeclass status controller)
+        │                             e.g. [{ ID: "/CommunityGalleries/.../versions/202604.24.0",
```
matthchr, line 88: Note that this is resolved from the public API https://learn.microsoft.com/en-us/rest/api/aks/container-service/list-node-image-versions?view=rest-aks-2026-04-02-preview&viewFallbackFrom=rest-aks-2026-03-02-preview&tabs=HTTP

designs/0012-node-image-version-pinning.md
```diff
@@ -0,0 +1,744 @@
...
+   is populated after VM creation from the VM's storage profile or the
+   AKS Machine's `NodeImageVersion`. These are two different fields on
+   two different objects.
+
+2. **Version is the last segment** of every image ID:
+   - CIG: `/CommunityGalleries/{gallery}/images/{def}/versions/{version}`
+   - SIG: `/subscriptions/{sub}/.../images/{def}/versions/{version}`
+
+3. **Version format**: `YYYYMM.DD.patch` (e.g., `202604.24.0`). This is
+   the same string customers see in `az aks nodepool show` and AgentBaker
```
matthchr: Ooh where do we show this on az aks nodepool show?

designs/0012-node-image-version-pinning.md
```diff
@@ -0,0 +1,744 @@
...
+   release notes.
+
+4. **All architecture variants share the same version** within a release.
+   AKS releases all variants (gen1, gen2, arm64) together within 24 hours.
+
+5. The **`ImageDefinition`** (e.g., `2204gen2containerd`) is derived from
+   `imageFamily` + instance type requirements. It is not user-facing.
+
+6. Image data is **cached for 3 days** (`ImageExpirationInterval`), with
```
matthchr, line 123: but refreshed on pod restart, which as @tallaxes pointed out may result in faster than 3d in practice... Might be nice to have Kusto data on that although it's somewhat a separate concern.

designs/0012-node-image-version-pinning.md
```diff
@@ -0,0 +1,744 @@
...
+│    • Should be within AKS support window                 │
+│    • Applied as version filter in SIG/CIG resolution     │
+│                                                          │
+│  (future) imageRef: <resource-id>  ◄── stretch            │
+│    • References customer Prepared Image resource         │
+│    • Bypasses gallery resolution entirely                │
+│    • Orthogonal to imageFamily/imageVersion              │
+│                                                          │
+│  Precedence: imageRef > imageVersion                     │
+│  AKSNodeClassSpec.ImageID is an unused stub (json:"-"). │
```
matthchr: Maybe let's just remove it entirely if we're not doing anything with it? (Maybe covered in section 6.4)

designs/0012-node-image-version-pinning.md
matthchr: Ahh I see you propose `imageAgeDays`, which we could do, but I'd suggest doing a date (2026-06-12) or datetime (2026-06-12 00:00:00).

I _think_ that's more standard?

designs/0012-node-image-version-pinning.md
```diff
@@ -0,0 +1,744 @@
...
+
+---
+
+## 1. Background: The Node Image Support Window
+
+AKS is formalizing its **support window for node images**. While no
+official policy has been published on Microsoft Learn yet, there are
+already public signals pointing in this direction:
+
+- **AppLens** already surfaces a diagnostic insight for clusters running
```
matthchr, line 15: Does this work for NAP nodes?

designs/0012-node-image-version-pinning.md
```diff
@@ -0,0 +1,744 @@
...
+Add a new optional field to `AKSNodeClassSpec`:
+
+```go
+type AKSNodeClassSpec struct {
+    // ...existing fields...
+
+    // imageVersion pins node image selection to a specific AKS node image
+    // version (e.g., "202604.24.0"). When set, Karpenter will use this
+    // version instead of automatically selecting the latest available
+    // image. The version must exist in the image gallery and should be
```
matthchr, line 212: is "should be within the AKS node image support window" actually enforced anywhere? AFAIK no, is that right?

(Still good/fine to document this, just trying to wrap my head around the proposal).

designs/0012-node-image-version-pinning.md
```diff
@@ -0,0 +1,744 @@
...
+    // +optional
+    ImageVersion *string `json:"imageVersion,omitempty"`
+}
+```
+
+**Field behavior:**
+
+| `imageVersion` | `imageFamily` | Behavior |
+|---|---|---|
+| unset | set | **Today's behavior.** Latest version is auto-selected for the given image family. |
```
matthchr, line 224: Clarify what users see on the spec field when the do a GET, it's empty right?

designs/0012-node-image-version-pinning.md
```diff
@@ -0,0 +1,744 @@
...
+| set | set | Version is pinned. `imageFamily` determines which image definitions (arch/gen variants) are used; `imageVersion` determines which version of those definitions is resolved. |
+| set | unset | Version is pinned. `imageFamily` defaults to `Ubuntu` (existing default behavior). |
+
+**Regex validation**: `^\d{6}\.\d{2}\.\d+$` matches the `YYYYMM.DD.patch`
+format (e.g., `202604.24.0`). This is enforced at admission time via the
+CRD schema.
+
+### 5.2 Image Version Existence Validation
+
+When `imageVersion` is set, the **status reconciler** must validate that
```
matthchr, line 260: Note that for SIG I think we just use the ListNodeImageVersions API and _assume_ it's there in SIG. That is (AFAIK) the contract with that API, but it's slightly different than what you're saying here.

designs/0012-node-image-version-pinning.md
```diff
@@ -0,0 +1,744 @@
...
+```yaml
+status:
+  conditions:
+    - type: ImagesReady
+      status: "False"
+      reason: ImageVersionNotFound
+      message: "imageVersion 202604.24.0 not found in gallery for image family Ubuntu2204"
+```
+
+**Why status-time, not admission-time**: Validating at admission would
```
matthchr: This is technically not true, because the webhooks run in the same pod as the operator and this information can be cached.

I think the main argument against this is probably just that we dont even have a webhook right now and we'd need one to make this work. The standard pattern in upstream for this type of stuff does seem to be NodeClass status conditions (which you also call out)

designs/0012-node-image-version-pinning.md
matthchr: AI agrees:
🤖
 ## Image Creation Date vs. Image Age in Days

**Option 1: `imageCreateDate: 2026-06-12`** is the better choice for a CRD field.

### Why Creation Date wins:

| Aspect | Creation Date | Age in Days |
|--------|--------------|-------------|
| **Staleness** | Always accurate — immutable fact | Stale immediately after being written; requires continuous reconciliation to stay correct |
| **Reconciliation cost** | No updates needed | Must be updated daily or becomes wrong |
| **Consistency** | Same value regardless of when you read it | Value depends on when it was last reconciled vs. when you read it |
| **Conflict/churn** | No status churn | Daily updates cause unnecessary resource version bumps, watch events, and potential conflicts |
| **Expressiveness** | Consumer can compute age, compare dates, bucket by month, etc. | Loses the original date; can't derive it back without knowing when the field was last updated |
| **Time zones / clock skew** | Standard `metav1.Time` or date string — well-understood | Integer that's only meaningful relative to "now", which differs across controllers |
| **Kubernetes convention** | Matches patterns like `creationTimestamp`, `lastTransitionTime` | No precedent in upstream APIs for "age in days" fields |

### When age *might* be preferable:

- If you want a **user-facing display hint** (like `kubectl get` columns showing "3d") — but that's better done in a printer column computed from the date.
- If the source data genuinely only provides age without an absolute timestamp (rare).

### Recommendation:

Store the **creation date** (`metav1.Time` or a date string). Let consumers (controllers, CEL validation, printer columns) compute age on read. This follows the same principle as `ConditionTypeConsistentStateFound` in the file you're looking at — conditions store `lastTransitionTime`, not "duration since last transition."

designs/0012-node-image-version-pinning.md
```diff
@@ -0,0 +1,744 @@
...
+AKSNodeClass.spec.imageVersion       (NEW — optional pin)
+        │
+        ▼
+ImageFamily.DefaultImages()          (unchanged — still resolves image definitions)
+        ▼
+NodeImageProvider.List()             (MODIFIED)
+        │
+        │  If imageVersion is SET:
+        │    SIG: BuildImageIDSIG(..., imageVersion) directly — skip latest lookup
+        │    CIG: BuildImageIDCIG(..., imageVersion) directly — skip latest lookup
```
matthchr: What about the validation that the date specified is valid? Where does that go? I don't think you can skip the latest lookup if you want to set `ImageVersionNotFound`?

rakechill:
## Finding: `imageVersion` regex should accept both AKS version formats

While cross-referencing the design against the existing codebase, we found that AKS image versions exist in **two distinct formats**, not just one:

| Format | Example | Used for |
|---|---|---|
| `YYYYMM.DD.patch` | `202512.06.0` | Current AKS releases (since ~2023) |
| `YYYY.MM.DD` | `2022.10.03` | Legacy AKS releases |

The original design had a single-format regex (`^\d{6}\.\d{2}\.\d+$`) that would silently **reject legacy version strings** as valid pin targets — for example, a customer attempting to rollback to an older image using a `YYYY.MM.DD` version would get an admission error.

**Evidence in the codebase:**
- `isNewerVersion()` in [`pkg/providers/imagefamily/nodeimageversionsclient.go`](../pkg/providers/imagefamily/nodeimageversionsclient.go#L94) was introduced in PR #526 (commit `3a9cbc0d`) specifically to handle both formats, with a comment: *"the new versioning scheme is yearmm.dd.build, previously it was yy.mm.dd without the build id"*.
- Tests in [`pkg/providers/imagefamily/nodeimageversionsclient_test.go`](../pkg/providers/imagefamily/nodeimageversionsclient_test.go#L25) cover cross-format comparisons like `202411.12.0` vs `2022.12.15`.
- The fake API in `pkg/fake/nodeimageversionsapi.go` still emits `2022.10.03`-style versions.

**Fix applied:** Updated the admission regex to an OR pattern:
```
!!:s^(\d{6}\.\d{2}\.\d+|\d{4}\.\d{2}\.\d{2})$
```

This matches both formats at admission time. Gallery existence validation in the status reconciler remains the authoritative check regardless of format.

designs/0012-node-image-version-pinning.md
```diff
@@ -0,0 +1,744 @@
...
+set ImagesReady = True (or False if version not found)
+```
+
+#### 5.5.3 Kubernetes upgrade interaction
+
+When the K8s version reconciler detects a control plane upgrade, it sets
+`ImagesReady` to `False`, which triggers a full image refresh
+(Scenario A, case 2 in the existing code).
+
+**When `imageVersion` is set**, this refresh should still resolve to the
```
matthchr, line 522: I _tend_ to agree with this decision but probably worth calling out that (up til here) the behavior is very similar to what AKS itself does, but this is a pretty big difference from what AKS is doing today and we should get Node SIGs explicit sign-off on it. The 90d thing is optional when k8s version doesn't change but my guess is that it's _required_ (or maybe even <90d?) when k8s version changes...

We may want to explore alternatives where k8s upgrade doesn't allow you to pin across it.

designs/0012-node-image-version-pinning.md
```diff
@@ -0,0 +1,805 @@
...
+   two different objects.
+
+2. **Version is the last segment** of every image ID:
+   - CIG: `/CommunityGalleries/{gallery}/images/{def}/versions/{version}`
+   - SIG: `/subscriptions/{sub}/.../images/{def}/versions/{version}`
+
+3. **Version format**: AKS image versions are observed in at least two
+   formats:
+   - Current format: `YYYYMM.DD.patch` (e.g., `202604.24.0`)
+   - Legacy format: `YYYY.MM.DD` (e.g., `2022.10.03`)
```
matthchr: How old is this legacy format? Do we actually have any clusters using it? Can we just not support it?

designs/0012-node-image-version-pinning.md
```diff
@@ -0,0 +1,805 @@
...
+schema. Gallery existence validation in the status reconciler remains
+authoritative.
+
+### 5.2 Image Version Existence Validation
+
+When `imageVersion` is set, the **status reconciler** must validate that
+the specified version exists in the image gallery (SIG or CIG) for the
+resolved image family's definitions.
+
+Each `imageFamily` resolves to multiple **image definitions** — one per
```
matthchr, line 264: What happens when imageFamily is Ubuntu, imageVersion is pinned, and there's a k8s upgrade. I guess it changes the underlying image and all the nodes roll? Though all the nodes have to roll anyway...

designs/0012-node-image-version-pinning.md
```diff
@@ -0,0 +1,805 @@
...
+### 5.6 Drift Detection
+
+When `imageVersion` is **set**, drift detection should compare the
+running node's image version against the pinned version. If they match,
+the node is not drifted.
+
+When `imageVersion` is **unset** (default), drift detection works as
+today — comparing against the latest available version.
+
+No changes are needed to `isImageVersionDrifted()` itself — it already
```
matthchr, line 581: We will need to make sure that the Status controller takes the .spec field for pinning into account when doing the 72h cache thing it does.

designs/0012-node-image-version-pinning.md
```diff
@@ -0,0 +1,805 @@
...
+| **Karpenter integration** | Version is a filter within existing image family resolution | Entirely different image source; bypasses gallery resolution |
+
+The two features are **orthogonal but share the API surface**:
+
+- `imageVersion` operates **within** the AKS-managed image family
+  (SIG/CIG) — it's a version pin.
+- Prepared Image Spec operates **outside** the AKS-managed galleries —
+  it references a customer-owned image resource.
+
+A well-shaped API keeps these as distinct, non-overlapping fields
```
matthchr: Agree

designs/0012-node-image-version-pinning.md
matthchr: We could require user to explicitly select an OS to avoid that (2204 or 2404, not Ubuntu), although it doesn't really solve the problem it just moves it because then theyll pin 2204 and it won't work on k8s 1.37 and then we have to figure out what happens then.

designs/0012-node-image-version-pinning.md
```diff
@@ -0,0 +1,805 @@
...
+set ImagesReady = True (or False if version not found)
+```
+
+#### 5.5.3 Kubernetes upgrade interaction
+
+When the K8s version reconciler detects a control plane upgrade, it sets
+`ImagesReady` to `False`, which triggers a full image refresh
+(Scenario A, case 2 in the existing code).
+
+**When `imageVersion` is set**, this refresh should still resolve to the
```
matthchr, line 522: Does this mean the nodes roll or they don't roll?

I think they _have_ to roll? Technically they can roll to the same image but they have to roll.

designs/0012-node-image-version-pinning.md
```diff
@@ -0,0 +1,805 @@
...
+
+---
+
+## 6. Future Extensibility: Prepared Image Specification
+
+This section is out of scope for the `imageVersion` design but is
+included to demonstrate that the API shape in section 5 is
+forward-compatible with upcoming AKS capabilities.
+
+### 6.0 Background: Prepared Image Specification
```
matthchr: Good we have this -- just noting I didn't review it as closely.

designs/0012-node-image-version-pinning.md
```diff
@@ -0,0 +1,805 @@
...
+   unnecessary.
+
+---
+
+## 7. Open Questions
+
+Remaining questions not covered by the design sections above:
+
+### Operational
+1. Should we emit Kubernetes events when a pinned version crosses the
```
matthchr: Where/on what resource?

designs/0012-node-image-version-pinning.md
```diff
@@ -0,0 +1,805 @@
...
+**Authors:** @rakechill
+**Related:**
+- [AKS Prepared Image Specification (AKS#4704)](https://github.com/Azure/AKS/issues/4704)
+- [Add support for setting ImageID for nodeClass](https://github.com/Azure/karpenter-provider-azure/issues/1220)
+
+---
+
+## 1. Background: The Node Image Support Window
+
+AKS is formalizing its **support window for node images**. While no
```
matthchr, line 11: after review: We also don't discuss different options here. I personally tend towards pinning as you have proposed here, because I think it solves the problem as well as some other user complaints as well, but we should at least have some discussion/investigation about what other possible solutions look like, where we enable the user to control when the node image version is applied, but:
- Don't let them pick the specific version
- Don't let them pin across k8s version updates

Maybe in the appendix/at the end we can discuss those options at a high level and why we feel that they do not suit?

designs/0012-node-image-version-pinning.md
rakechill: Yes, I already discussed with node sig folks and PMs (including Jorge). They are still ironing out the exact long-term policy, but are aligned with adding the 90 days warning at the moment.

designs/0012-node-image-version-pinning.md
rakechill: Correct

designs/0012-node-image-version-pinning.md
rakechill: Yeah that makes sense

designs/0012-node-image-version-pinning.md
rakechill: So, you can add a status condition that does not filter up to readiness and instead just adds informational signalling. I think I reference it in detail later.

designs/0012-node-image-version-pinning.md
rakechill: ack, will reflect

designs/0012-node-image-version-pinning.md
rakechill: It's part of the agentpool object that gets returned. You can also pass in the optional arg `--query nodeImageVersion` to specifically output the image version.

https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/containerservice/armcontainerservice#ManagedClusterAgentPoolProfileProperties

```
// READ-ONLY; The version of node image
	NodeImageVersion *[string](https://pkg.go.dev/builtin#string) `json:"nodeImageVersion,omitempty" azure:"ro"`
```

designs/0012-node-image-version-pinning.md
rakechill: todo: add note on pod restart cache refresh

designs/0012-node-image-version-pinning.md
rakechill: AppLens? No idea haha

designs/0012-node-image-version-pinning.md
rakechill: No enforcement yet, but leaving flexibility for when enforcement is added in the future.

designs/0012-node-image-version-pinning.md
rakechill: TODO: add example for UX today

designs/0012-node-image-version-pinning.md
rakechill: Yeah, we'll have to use ListNodeImageVersions API + check to see if it shows up in the list, otherwise we'll pass an invalid version to Machine API.

designs/0012-node-image-version-pinning.md
rakechill: todo: clarify language in this section

designs/0012-node-image-version-pinning.md
rakechill: TODO: update this to be a static date rather than calculated days since

designs/0012-node-image-version-pinning.md
rakechill: So, we discussed k8s version + node image compatibility. I think the consensus was that we don't have a clear mapping between which node image versions work for which k8s versions. 

I'll check with @djsly to see if he has thoughts on this.

designs/0012-node-image-version-pinning.md
rakechill: So the initial lookup still has to happen. The intention with "skip lookup" will be in terms of drift. Basically, we can call list node image versions less frequently when image ID is set since we won't be drifting every time there's a new node image version.

TODO: take another pass at the flow diagrams + make sure it makes sense in the different contexts

TODO: clarify first image resolution vs. drift

designs/0012-node-image-version-pinning.md
rakechill: TODO: chat w/ node SIG to understand how this overlaps with k8s upgrades

designs/0012-node-image-version-pinning.md
rakechill: They'll still roll, but node image (likely) shouldn't change -- still need to check w/ node sig on this

TODO: add that the nodes still get upgraded, but same image version

designs/0012-node-image-version-pinning.md
rakechill: TODO: add this detail into the design

designs/0012-node-image-version-pinning.md
rakechill: Yeah, I'm still deciding if adding this as a secondary (related) design might be better. Maybe a different PR left in draft until we return to this. I think @comtalyst's meeting this week will help inform/shape this design, too.

designs/0012-node-image-version-pinning.md
rakechill: Probably AKSNodeclass.

designs/0012-node-image-version-pinning.md
matthchr: Can you clarify that here in the doc?

matthchr:
Closing this as it was superseded by #1756

PR #1723, opened 2026-06-11
fix: align OS disk sizing with AKS when osDiskSizeGB is unset

### LINKED ISSUE #5503: [Feature] OSDiskSizeGB set to max available by default in NAP
Desired Behavior: For NAP managed nodes, AKS by default provisions an Ephemeral OS disk with a size that scales according to the total temp storage of the VM SKU so long as the temp is at least 128GiB, even if no size was specified. (see [OSDiskSizeGB documentation] (https://learn.microsoft.com/azure/aks/concepts-storage#ephemeral-os-disks) for more details)

ETA: backlog

Current behavior: Right now NAP uses partial ephemeral storage, not necessarily the full available storage by default. NAP can set the OS disk size to exactly the requested OSDiskSizeGB in the [AKSNodeClass] (https://learn.microsoft.com/azure/aks/node-auto-provisioning-aksnodeclass) custom resource definition, but will not default to the max unless specified. If not specified in the AKSNodeClass, the default is 128GB

hakman:
Fixes https://github.com/Azure/AKS/issues/5503

**Description**

When osDiskSizeGB is omitted, prefer an ephemeral OS disk at the SKU-supported size (capped at 2040GB).
SKUs without ephemeral support fall back to managed disk defaults by vCPU count (128/256/512/1024GB).
VM and AKS Machine API paths consume the same resolved profile, and reported ephemeral-storage capacity matches the provisioned size.
Explicit values remain bounded to 30-2048GB.

Official reference:
https://learn.microsoft.com/en-us/azure/aks/concepts-storage#default-os-disk-sizing

> If you select a virtual machine (VM) SKU that supports Ephemeral OS disks but don't specify an OS disk size, AKS by default provisions an Ephemeral OS disk with a size that scales according to the total temp storage of the VM SKU so long as the temp is at least 128 GiB. 

> When you create a new cluster or add a new node pool to an existing cluster, the number for vCPUs determines the OS disk size by default.
> image

* private AKS cluster, with multiple SKU types

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
Aligns OS disk sizing with AKS when osDiskSizeGB is unset
```


pkg/providers/instance/aksmachineinstancehelpers.go
```diff
@@ -133,7 +133,7 @@ func (p *DefaultAKSMachineProvider) buildAKSMachineTemplate(ctx context.Context,
 			OperatingSystem: &armcontainerservice.MachineOSProfile{
 				OSType:       lo.ToPtr(armcontainerservice.OSTypeLinux),
 				OSSKU:        osSku,
-				OSDiskSizeGB: nodeClass.Spec.OSDiskSizeGB, // AKS machine API defaults it if nil
```
comtalyst, line 137: ~~For AKS Machine code path, this "AKS machine API defaults it if nil" should work and give us the parity with AKS.
I realized that it didn't previously given the value was always defaulted.~~

~~For context, AKS Machine API integration is an upcoming internal rearchitecture. NAP will soon provision nodes with AKS API (via this code path) instead of Compute API (via VM instance code path). This defaulting is one of the parities we are getting.~~ (not doing this due to scheduling calculations)

hakman: @comtalyst That would be a valid approach. I could remove some/most of the changes in this file. Though, if anything changes on AKS side, scheduling and the `sku-storage-ephemeralos-maxsize` label will be out of sync. With the current changes, values should be consistent between scheduling and runtime values.

hakman: @comtalyst Done, the machine path now sends only the resolved OS disk type and leaves osDiskSizeGB to the API when unset. You were right about the label, my actual concern was the advertised ephemeral-storage capacity, which on this path is now a best-effort mirror of AKS defaulting.

pkg/providers/instancetype/instancetypes.go
```diff
@@ -55,14 +55,20 @@ import (
 
 const (
 	InstanceTypesCacheTTL = 23 * time.Hour
+
+	// maxEphemeralOSDiskSizeGB is the maximum size (in GB) Azure supports for ephemeral OS disks.
+	maxEphemeralOSDiskSizeGB = 2040
```
comtalyst, line 63: Should it be 2040 or 2048?

AKS is 2048 though.

pkg/apis/v1alpha2/aksnodeclass.go
```diff
@@ -54,7 +54,8 @@ type AKSNodeClassSpec struct {
 	// +optional
 	VNETSubnetID *string `json:"vnetSubnetID,omitempty"`
 	// osDiskSizeGB is the size of the OS disk in GB.
-	// +default=128
+	// If not specified, it is auto-sized per instance type: an ephemeral OS disk at the largest
+	// SKU-supported size (max 2040 GB) when possible, otherwise a managed OS disk sized by vCPU count.
```
comtalyst, line 58: See https://github.com/Azure/karpenter-provider-azure/pull/1723/changes#r3826583645

pkg/apis/v1beta1/labels.go
```diff
@@ -139,7 +139,7 @@ var (
 	LabelSKUAcceleratedNetworking = Group + "/sku-networking-accelerated" // sku.AcceleratedNetworkingEnabled
 
 	LabelSKUStoragePremiumCapable     = Group + "/sku-storage-premium-capable"     // sku.IsPremiumIO
-	LabelSKUStorageEphemeralOSMaxSize = Group + "/sku-storage-ephemeralos-maxsize" // calculated as max(sku.CachedDiskBytes, sku.MaxResourceVolumeMB)
+	LabelSKUStorageEphemeralOSMaxSize = Group + "/sku-storage-ephemeralos-maxsize" // calculated as min(selected SKU-supported ephemeral OS disk size, 2040 GiB)
```
comtalyst, line 142: See https://github.com/Azure/karpenter-provider-azure/pull/1723/changes#r3826583645

pkg/providers/instancetype/instancetype.go
```diff
@@ -222,7 +222,8 @@ func computeRequirements(
 func setRequirementsEphemeralOSDiskSupported(requirements scheduling.Requirements, sku *skewer.SKU) {
 	sizeGB, _ := FindMaxEphemeralSizeGBAndPlacement(sku)
 	if sizeGB > 0 {
-		requirements[v1beta1.LabelSKUStorageEphemeralOSMaxSize].Insert(fmt.Sprint(sizeGB))
+		// Cap the advertised size at Azure's ephemeral OS disk limit to match provisioning.
```
comtalyst: Good catch. And would be great to add more detailed context, e.g., referencing what AKS supports: https://learn.microsoft.com/en-us/rest/api/aks/agent-pools/create-or-update?view=rest-aks-2026-04-01&tabs=HTTP

pkg/providers/instancetype/instancetype.go
```diff
@@ -222,7 +222,8 @@ func computeRequirements(
...
+		requirements[v1beta1.LabelSKUStorageEphemeralOSMaxSize].Insert(fmt.Sprint(min(sizeGB, maxEphemeralOSDiskSizeGB)))
```
comtalyst: Given this `min(sizeGB, maxEphemeralOSDiskSizeGB))` logic is in multiple places in this module, it might be better to be in `FindMaxEphemeralSizeGBAndPlacement`? And given that this knowledge moves a layer down, it might be fair to rename the function to `FindMaxEphemeralSizeGBAndPlacementForAKS`.

pkg/providers/instancetype/instancetypes.go
```diff
@@ -579,9 +587,65 @@ func supportsNVMeEphemeralOSDisk(sku *skewer.SKU) bool {
...
+// IsEphemeral reports whether the resolved OS disk is ephemeral.
+func (p OSDiskProfile) IsEphemeral() bool {
+	return p.Placement != nil
+}
+
+// ResolveOSDiskProfile resolves the OS disk size and type for the given SKU: an explicit
+// osDiskSizeGB is used as-is (ephemeral when it fits); nil auto-sizes to an ephemeral disk at
+// the SKU-supported size, or a vCPU-based managed default below the ephemeral threshold.
+func ResolveOSDiskProfile(sku *skewer.SKU, osDiskSizeGB *int32) OSDiskProfile {
+	maxEphemeralSizeGB, placement := FindMaxEphemeralSizeGBAndPlacement(sku)
```
comtalyst:
```suggestion
	skuMaxEphemeralOSDiskSizeGB, skuDefaultPlacement := FindMaxEphemeralSizeGBAndPlacement(sku)
```

pkg/providers/instancetype/instancetypes.go
```diff
@@ -579,9 +587,65 @@ func supportsNVMeEphemeralOSDisk(sku *skewer.SKU) bool {
...
+
+// IsEphemeral reports whether the resolved OS disk is ephemeral.
+func (p OSDiskProfile) IsEphemeral() bool {
+	return p.Placement != nil
+}
+
+// ResolveOSDiskProfile resolves the OS disk size and type for the given SKU: an explicit
+// osDiskSizeGB is used as-is (ephemeral when it fits); nil auto-sizes to an ephemeral disk at
+// the SKU-supported size, or a vCPU-based managed default below the ephemeral threshold.
+func ResolveOSDiskProfile(sku *skewer.SKU, osDiskSizeGB *int32) OSDiskProfile {
```
comtalyst:
```suggestion
func ResolveOSDiskProfile(sku *skewer.SKU, requstedOSDiskSizeGB *int32) OSDiskProfile {
```

pkg/providers/instancetype/instancetypes.go
```diff
@@ -579,9 +587,65 @@ func supportsNVMeEphemeralOSDisk(sku *skewer.SKU) bool {
...
+	maxEphemeralSizeGB, placement := FindMaxEphemeralSizeGBAndPlacement(sku)
+	maxEphemeralSizeGB = min(maxEphemeralSizeGB, maxEphemeralOSDiskSizeGB)
+	if osDiskSizeGB != nil {
+		if maxEphemeralSizeGB > 0 && int64(*osDiskSizeGB) <= maxEphemeralSizeGB {
+			return OSDiskProfile{SizeGB: *osDiskSizeGB, Placement: placement}
```
comtalyst: nit: one line per property is more clear

pkg/providers/instancetype/instancetypes.go
```diff
@@ -579,9 +587,65 @@ func supportsNVMeEphemeralOSDisk(sku *skewer.SKU) bool {
...
+}
+
+// IsEphemeral reports whether the resolved OS disk is ephemeral.
+func (p OSDiskProfile) IsEphemeral() bool {
+	return p.Placement != nil
+}
+
+// ResolveOSDiskProfile resolves the OS disk size and type for the given SKU: an explicit
+// osDiskSizeGB is used as-is (ephemeral when it fits); nil auto-sizes to an ephemeral disk at
+// the SKU-supported size, or a vCPU-based managed default below the ephemeral threshold.
```
comtalyst: Might be more ergonomic to put the process details next to the line of code that does it. Then the header comment here can abstract more things away, like "Resolves usable OS disk size and type, given SKU and requested OS disk size. Note that the requested size is always honored."

pkg/providers/instancetype/instancetypes.go
```diff
@@ -579,9 +587,65 @@ func supportsNVMeEphemeralOSDisk(sku *skewer.SKU) bool {
...
+		return OSDiskProfile{SizeGB: *osDiskSizeGB}
+	}
+	if maxEphemeralSizeGB >= minEphemeralOSDiskSizeGB {
+		return OSDiskProfile{SizeGB: int32(maxEphemeralSizeGB), Placement: placement} //nolint:gosec // G115: value bounded to [128,2040], safe int32 conversion
+	}
+	return OSDiskProfile{SizeGB: defaultManagedOSDiskSizeGB(sku)}
+}
+
+// ResolveOSDiskProfileFor resolves the OS disk profile for the named instance type.
+func ResolveOSDiskProfileFor(ctx context.Context, provider Provider, instanceTypeName string, osDiskSizeGB *int32) (OSDiskProfile, error) {
```
comtalyst:
```suggestion
func ResolveOSDiskProfileFromInstanceType(ctx context.Context, provider Provider, instanceTypeName string, osDiskSizeGB *int32) (OSDiskProfile, error) {
```

or something similar that distinguishes this and the one above (formally "ResolveOSDiskProfileFromSKU") more clearly.

pkg/providers/instancetype/instancetypes.go
```diff
@@ -579,9 +587,65 @@ func supportsNVMeEphemeralOSDisk(sku *skewer.SKU) bool {
...
+	sku, err := provider.Get(ctx, instanceTypeName)
+	if err != nil {
+		return OSDiskProfile{}, err
+	}
+	return ResolveOSDiskProfile(sku, osDiskSizeGB), nil
+}
+
+// defaultManagedOSDiskSizeGB returns the managed OS disk size by vCPU count, mirroring AKS defaulting.
```
comtalyst, line 642: If there are links to AKS defaulting references, that would help in maintenance. We are trying to catch whatever pieces that is still relying on manual parity.

test/suites/machine/machine_test.go
```diff
@@ -116,6 +116,45 @@ var _ = Describe("Machine Tests", func() {
 		}))
 	})
 
+	It("should auto-size an ephemeral OS disk when osDiskSizeGB is unset", func() {
```
comtalyst: This is not specific to machines and might better be in `nodeclaim` suite where relevant tests are in. 
I think it is okay to inspect the underlying VM object rather than the machine as well.

comtalyst (commented):
Could you run E2E tests via https://github.com/Azure/karpenter-provider-azure/actions/workflows/e2e-matrix-trigger.yaml on `aksscriptless` and `aksmachineapi` mode and include results in the PR description as well?

test/suites/machine/machine_test.go
hakman: done

pkg/providers/instancetype/instancetypes.go
hakman: done

pkg/providers/instancetype/instancetypes.go
hakman: done

pkg/providers/instancetype/instancetypes.go
hakman: done

pkg/providers/instancetype/instancetypes.go
hakman: done

pkg/providers/instancetype/instancetypes.go
hakman: done

pkg/providers/instancetype/instancetypes.go
hakman: My understanding is that the limit is 2048 for managed disks, but 2040 for ephemeral disks:
https://learn.microsoft.com/en-gb/azure/virtual-machines/ephemeral-os-disks
> Cache, temp, or NVMe disk size for the VM size or 2,040 GiB, whichever is smaller.

pkg/apis/v1alpha2/aksnodeclass.go
hakman: added link

pkg/providers/instancetype/instancetypes.go
hakman: added link

pkg/apis/v1beta1/labels.go
hakman: replied to the referenced comment

pkg/providers/instancetype/instancetype.go
hakman: added context distinguishing the AKS API’s 2048 field limit from Compute’s 2040 GiB ephemeral disk limit.

pkg/providers/instancetype/instancetype.go
hakman: I think that's fair, done

pkg/providers/instancetype/instancetypes.go
```diff
@@ -55,14 +55,21 @@ import (
 
 const (
 	InstanceTypesCacheTTL = 23 * time.Hour
+
+	// maxEphemeralOSDiskSizeGB is the maximum size (in GB) Azure supports for ephemeral OS disks.
+	// https://learn.microsoft.com/azure/virtual-machines/ephemeral-os-disks
+	maxEphemeralOSDiskSizeGB = 2040
+	// minEphemeralOSDiskSizeGB is AKS's threshold for auto-selecting an ephemeral OS disk;
+	// below it, auto-sizing falls back to vCPU-based managed defaults.
+	minEphemeralOSDiskSizeGB = 128
```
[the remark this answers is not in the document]
hakman: done

test/suites/nodeclaim/eph_osdisk_test.go
```diff
@@ -22,11 +22,41 @@ import (
...
+		Expect(string(lo.FromPtr(vm.Properties.StorageProfile.OSDisk.DiffDiskSettings.Option))).To(Equal("Local"))
+
+		const expectedDiskSizeGB int32 = 1600
+		Expect(lo.FromPtr(vm.Properties.StorageProfile.OSDisk.DiskSizeGB)).To(Equal(expectedDiskSizeGB))
+
+		advertised := nodeClaim.Status.Capacity[corev1.ResourceEphemeralStorage]
+		Expect(advertised.Cmp(resource.MustParse("1600Gi"))).To(Equal(0))
+
+		actual := env.GetNode(pods[0].Spec.NodeName).Status.Capacity[corev1.ResourceEphemeralStorage]
+		Expect(actual.Cmp(resource.MustParse("128Gi"))).To(BeNumerically(">", 0))
```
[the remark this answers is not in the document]
hakman: done

PR #1726, opened 2026-06-11
feat: support Windows node provisioning via the AKS Machine API

MarcPow:
## Description

Enable the Azure Karpenter provider to provision **Windows Server 2022 and Windows Server 2025** workers through the AKS Machine API (`aksmachineapi` and `aksmachineapiheaderbatch`). The controller itself remains a Linux workload; only provisioned workers are Windows.

Rebased onto current `main`. This revision adopts @comtalyst's guidance: the Windows node image is now pinned at create time and the `UseWindowsGen2VM` header is gone.

### Scheduling and instance types

- Derive `kubernetes.io/os` and `kubernetes.azure.com/os-sku` from the AKSNodeClass image family.
- Support `Windows2022` and `Windows2025`; deprecated Windows2019 and WindowsAnnual are intentionally excluded.
- Restrict Windows offerings to `amd64` before scheduling, so ARM64 SKUs are never advertised for a Windows NodeClass.
- Keep the Karpenter controller Deployment pinned to Linux nodes in mixed-OS clusters.

### AKSNodeClass and images

- Add separate Windows2022 and Windows2025 image families backed by the AKSWindows shared image gallery.
- Windows2022 is FIPS-off. Windows2025 is effectively FIPS-on and requires Kubernetes 1.32+.
- Windows2025 Trusted Launch uses the dedicated TL image; Kata is Linux-only, so Windows advertises no images for it.
- Reject Linux-only settings on Windows NodeClasses through CRD validation: `linuxOSConfig`, artifact streaming, and LocalDNS.

### Image pinning replaces the Gen2 header

Windows now pins `NodeImageVersion` at create time exactly like Linux. This was previously deferred because deployed Machine API parsers rejected hyphenated AKSWindows names; [aks-rp PR #16801377](https://msazure.visualstudio.com/CloudNativeCompute/_git/aks-rp/pullrequest/16801377) fixed that and has merged.

The resolved image definition carries its Hyper-V generation — `windows-2022-containerd-gen2` versus `windows-2022-containerd` — and on `PutMachine` the RP consumes the pinned value verbatim rather than resolving an image from OSSKU. Choosing the image therefore chooses the generation, so the separate `UseWindowsGen2VM` create header is redundant and has been removed along with its options plumbing and batch-key participation. Karpenter already filters images by `karpenter.azure.com/sku-hyperv-generation`, so a Gen2-only SKU gets the Gen2 image and a Gen1-only SKU gets the Gen1 image.

This also restores the drift signal: a Windows node now drifts on a new node image, which was impossible while the RP chose the image.

### Machine names

- Managed NAP (`aksmanagedap`): `<6-char NodePool-prefix hash>-<5-char NodeClaim suffix>`.
- Custom/self-hosted pools: the five-character NodeClaim suffix, to satisfy the RP-composed NetBIOS budget.

### Unit, CEL, and acceptance

- Image-family, OS/OSSKU, FIPS, Trusted Launch, Kata exclusion, image conversion, naming, and CRD/CEL coverage.
- Instance-type tests prove ARM64 stays available for Linux and is excluded for both Windows families.
- Machine API acceptance tests provision Windows2022 and Windows2025 and assert OSType, OSSKU, FIPS, absence of Linux-only profiles, the `os`/`arch`/`os-sku` labels, **and that a Windows `NodeImageVersion` is pinned with the expected image definition**.
- Drift acceptance covers both directions: a newly created Windows machine is not drifted, and a Windows node drifts when its image version changes.
- Full suite green locally: 40 Ginkgo suites passed, `golangci-lint` 0 issues, `actionlint` clean, no generated-file drift.

### E2E — Windows as an OS dimension

Per review feedback, Windows is layered onto existing suites rather than living only in its own:

- `test/suites/drift`: a Windows node drifts when the NodePool template changes, and is left alone when nothing changes.
- `test/suites/consolidation`: a Windows node is not consolidated while a do-not-disrupt pod holds it, and is consolidated once empty.
- `test/suites/windows`: the dedicated Windows2022/Windows2025 smoke suite.

These share `env.SkipIfNotWindowsCapable()` and `env.WindowsDeployment()`, and skip automatically on clusters that cannot provision Windows, so they add no cost to non-Windows runs.

Previously recorded live Azure results for the smoke suite: Windows2022 passed in 356.40s and Windows2025 passed in 335.33s with `kubernetes.azure.com/fips_enabled=true` asserted.

## Merge sequencing — please read before merging

The pinning change depends on the aks-rp parser fix being **deployed**, not merged. I verified the current rollout directly by issuing `PutMachine` with the exact image string this PR sends:

| Region | `AKSWindows-2022-containerd-gen2-<version>` | Interpretation |
|---|---|---|
| `westcentralus` | **200 OK**, Machine created | fix deployed |
| `westus2` | 400 `invalid node image version` | pre-fix parser |

A malformed control (`AKSWindows-totally-bogus-xx`) was rejected in both regions, so the `westcentralus` success is the fixed parser accepting a valid name rather than validation being skipped.

So the rollout is still in progress. In a region that has not yet received it, a pinned Windows image returns 400 and there is no longer a header fallback, so Windows provisioning fails outright. **This PR should merge only once the RP fix has reached the regions where Windows NAP will run.** Happy to re-run the live Windows E2E in a confirmed region — `westcentralus` works today — or in the standalone AKS E2E region.

## Scope

- Machine API provisioning only. Scriptless and bootstrapping-client paths continue to reject Windows.
- GPU-on-Windows and Windows-specific max-pods policy remain follow-up items.

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -7,6 +7,14 @@
 1. Create your new E2E testing suite `<test-suite-name>` within the `test/suites/` package. See: `test/README.md`
 2. Update the `workflows/e2e-matrix.yaml` workflow to include your E2E test case: `suite: [Utilization, GPU, ...]` - add in the name of your folder within the `test/suites/` package to the comma separated list. Casing does not matter.
 
+> **Note — suites that need a non-default cluster:** most suites run on the shared CI cluster
+> (`ci-mkcluster-all`, Azure CNI overlay + Cilium, in the matrix's `provision_mode`). The `Windows`
+> suite is special-cased in `workflows/e2e.yaml`: it always runs in `aksmachineapi` mode (Windows is
+> only provisionable via the AKS Machine API) on a dedicated cluster (`ci-mkcluster-all-windows`,
+> `az-mkaks-windows`) because Windows does not support the Cilium dataplane, and it uses a machines
+> pool name `<= 6` chars (`winmp`) to satisfy the Windows machine-name limit. Follow that pattern if a
+> new suite needs its own cluster shape or provisioning mode.
+
 ### Running the test case
 
 (temporary workflow until we re-enable automation)
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -7,6 +7,14 @@
 1. Create your new E2E testing suite `<test-suite-name>` within the `test/suites/` package. See: [test/README.md](../../test/README.md)
 2. Update the `workflows/e2e-matrix.yaml` workflow to include your E2E test case: `suite: [Utilization, GPU, ...]` - add in the name of your folder within the `test/suites/` package to the comma separated list. Casing does not matter.
 
+> **Note — suites that need a non-default cluster:** most suites run on the shared CI cluster
+> (`ci-mkcluster-all`, Azure CNI overlay + Cilium, in the matrix's `provision_mode`). The `Windows`
+> suite is special-cased in `workflows/e2e.yaml`: it always runs in `aksmachineapi` mode (Windows is
+> only provisionable via the AKS Machine API) on a dedicated cluster (`ci-mkcluster-all-windows`,
+> `az-mkaks-windows`) because Windows does not support the Cilium dataplane, and it uses a machines
+> pool name `<= 6` chars (`winmp`) to satisfy the Windows machine-name limit. Follow that pattern if a
+> new suite needs its own cluster shape or provisioning mode.
+
 ### Running the test case
 
 (temporary workflow until we re-enable automation)
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -51,6 +51,7 @@ on:
           - Storage
           - Subnet
           - Utilization
+          - Windows
       location:
         type: choice
         description: "Azure location to run the e2e test in"
@@ -216,9 +217,14 @@ jobs:
           git_ref: ${{ inputs.git_ref }}
           location: ${{ inputs.location }}
           k8s_version: ${{ inputs.k8s_version }}
-          provision_mode: ${{ inputs.provision_mode }}
+          # Windows is only provisionable via the AKS Machine API, so force that mode for the Windows
+          # suite regardless of the matrix's provision_mode (it would otherwise skip under aksscriptless).
+          provision_mode: ${{ inputs.suite == 'Windows' && 'aksmachineapi' || inputs.provision_mode }}
           identity_type: ${{ inputs.suite == 'Machine' && 'UserAssigned' || 'SystemAssigned' }}
           azure_vm_size: ${{ inputs.azure_vm_size }}
+          # Windows needs a dedicated (non-Cilium) cluster and a machines pool name <= 6 chars.
+          windows: ${{ inputs.suite == 'Windows' }}
+          aks_machines_pool_name: ${{ inputs.suite == 'Windows' && 'winmp' || 'testmpool' }}
       - name: install karpenter
         uses: ./.github/actions/e2e/install-karpenter
         with:
@@ -228,7 +234,8 @@ jobs:
           acr_name: ${{ env.ACR_NAME }}
           git_ref: ${{ inputs.git_ref }}
           location: ${{ inputs.location }}
-          provision_mode: ${{ inputs.provision_mode }}
+          provision_mode: ${{ inputs.suite == 'Windows' && 'aksmachineapi' || inputs.provision_mode }}
+          aks_machines_pool_name: ${{ inputs.suite == 'Windows' && 'winmp' || 'testmpool' }}
       - name: run the ${{ inputs.suite }} test suite
         env:
           AZURE_CLUSTER_NAME: ${{ env.CLUSTER_NAME }}
@@ -239,7 +246,8 @@ jobs:
           AZURE_CLIENT_ID: ${{ secrets.E2E_CLIENT_ID_TEST }}
           TEST_SUITE: ${{ inputs.suite }}
           GIT_REF: ${{ github.sha }}
-          PROVISION_MODE: ${{ inputs.provision_mode }}
+          PROVISION_MODE: ${{ inputs.suite == 'Windows' && 'aksmachineapi' || inputs.provision_mode }}
+          AKS_MACHINES_POOL_NAME: ${{ inputs.suite == 'Windows' && 'winmp' || 'testmpool' }}
         run: |
           make az-creds
           make e2etests
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -51,6 +51,7 @@ on:
           - Storage
           - Subnet
           - Utilization
+          - Windows
       location:
         type: choice
         description: "Azure location to run the e2e test in"
@@ -217,9 +218,14 @@ jobs:
           git_ref: ${{ inputs.git_ref }}
           location: ${{ inputs.location }}
           k8s_version: ${{ inputs.k8s_version }}
-          provision_mode: ${{ inputs.provision_mode }}
+          # Windows is only provisionable via the AKS Machine API, so force that mode for the Windows
+          # suite regardless of the matrix's provision_mode (it would otherwise skip under aksscriptless).
+          provision_mode: ${{ inputs.suite == 'Windows' && 'aksmachineapi' || inputs.provision_mode }}
           identity_type: ${{ inputs.suite == 'Machine' && 'UserAssigned' || 'SystemAssigned' }}
           azure_vm_size: ${{ inputs.azure_vm_size }}
+          # Windows needs a dedicated (non-Cilium) cluster and a machines pool name <= 6 chars.
+          windows: ${{ inputs.suite == 'Windows' }}
+          aks_machines_pool_name: ${{ inputs.suite == 'Windows' && 'winmp' || 'mpool' }}
       - name: install karpenter
         uses: ./.github/actions/e2e/install-karpenter
         with:
@@ -229,7 +235,8 @@ jobs:
           acr_name: ${{ env.ACR_NAME }}
           git_ref: ${{ inputs.git_ref }}
           location: ${{ inputs.location }}
-          provision_mode: ${{ inputs.provision_mode }}
+          provision_mode: ${{ inputs.suite == 'Windows' && 'aksmachineapi' || inputs.provision_mode }}
+          aks_machines_pool_name: ${{ inputs.suite == 'Windows' && 'winmp' || 'mpool' }}
       - name: run the ${{ inputs.suite }} test suite
         env:
           AZURE_CLUSTER_NAME: ${{ env.CLUSTER_NAME }}
@@ -240,7 +247,8 @@ jobs:
           AZURE_CLIENT_ID: ${{ secrets.E2E_CLIENT_ID_TEST }}
           TEST_SUITE: ${{ inputs.suite }}
           GIT_REF: ${{ github.sha }}
-          PROVISION_MODE: ${{ inputs.provision_mode }}
+          PROVISION_MODE: ${{ inputs.suite == 'Windows' && 'aksmachineapi' || inputs.provision_mode }}
+          AKS_MACHINES_POOL_NAME: ${{ inputs.suite == 'Windows' && 'winmp' || 'mpool' }}
         run: |
           make az-creds
           make e2etests
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ Karpenter provider for AKS can be used in two modes:
 ## Known limitations
 
 The following AKS features are not supported:
-* Windows nodes.
+* Windows nodes in self-hosted mode. Windows node provisioning is supported only with the AKS Machine API provision mode (used by Node Auto Provisioning); the self-hosted (scriptless) provision mode does not support Windows.
 * Kubenet and Calico.
 * IPv6 clusters.
 * [Service Principal](https://learn.microsoft.com/azure/aks/kubernetes-service-principal) based clusters. A system-assigned or user-assigned managed identity must be used.
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ Karpenter provider for AKS can be used in two modes:
 ## Known limitations
 
 The following AKS features are not supported:
-* Windows nodes.
+* Windows nodes require the AKS Machine API provision mode. This mode is used by Node Auto Provisioning and can also be used with self-hosted Karpenter; the scriptless and bootstrapping-client provision modes do not support Windows. Windows2025 requires Kubernetes 1.32 or newer and AKS Windows2025 availability for the subscription and region.
 * Kubenet and Calico.
 * IPv6 clusters.
 * [Service Principal](https://learn.microsoft.com/azure/aks/kubernetes-service-principal) based clusters. A system-assigned or user-assigned managed identity must be used.
--- a/charts/karpenter-crd/templates/karpenter.azure.com_aksnodeclasses.yaml
+++ b/charts/karpenter-crd/templates/karpenter.azure.com_aksnodeclasses.yaml
@@ -1005,6 +1005,8 @@ spec:
                 - Ubuntu2204
                 - Ubuntu2404
                 - AzureLinux
+                - Windows2022
+                - Windows2025
                 type: string
               kubelet:
                 description: |-
@@ -1680,6 +1682,12 @@ spec:
               rule: 'has(self.fipsMode) && self.fipsMode == ''FIPS'' ? (has(self.imageFamily)
                 && self.imageFamily != ''Ubuntu2204'' && self.imageFamily != ''Ubuntu2404'')
                 : true'
+            - message: FIPS is not supported for Windows image families
+              rule: '!has(self.fipsMode) || self.fipsMode != ''FIPS'' || !has(self.imageFamily)
+                || !(self.imageFamily in [''Windows2022'',''Windows2025''])'
+            - message: linuxOSConfig is not supported for Windows image families
+              rule: '!has(self.linuxOSConfig) || !has(self.imageFamily) || !(self.imageFamily
+                in [''Windows2022'',''Windows2025''])'
             - message: kubelet.failSwapOn must be set to false when linuxOSConfig.swapFileSize
                 is specified
               rule: '!has(self.linuxOSConfig) || !has(self.linuxOSConfig.swapFileSize)
--- a/charts/karpenter-crd/templates/karpenter.azure.com_aksnodeclasses.yaml
+++ b/charts/karpenter-crd/templates/karpenter.azure.com_aksnodeclasses.yaml
@@ -1031,6 +1031,8 @@ spec:
                 - Ubuntu2204
                 - Ubuntu2404
                 - AzureLinux
+                - Windows2022
+                - Windows2025
                 type: string
               kubelet:
                 description: |-
@@ -1724,14 +1726,35 @@ spec:
                 && ((has(self.security.trustedLaunch.vtpm) && self.security.trustedLaunch.vtpm)
                 || (has(self.security.trustedLaunch.secureBoot) && self.security.trustedLaunch.secureBoot)))))
                 : true'
-            - message: TrustedLaunch with FIPSMode FIPS is only supported for Ubuntu
-                and Ubuntu2204
+            - message: TrustedLaunch with FIPSMode FIPS is only supported for Ubuntu,
+                Ubuntu2204, and Windows2025
               rule: 'has(self.fipsMode) && self.fipsMode == ''FIPS'' && has(self.security)
                 && has(self.security.trustedLaunch) && ((has(self.security.trustedLaunch.vtpm)
                 && self.security.trustedLaunch.vtpm) || (has(self.security.trustedLaunch.secureBoot)
                 && self.security.trustedLaunch.secureBoot)) ? (!has(self.imageFamily)
-                || self.imageFamily == ''Ubuntu'' || self.imageFamily == ''Ubuntu2204'')
-                : true'
+                || self.imageFamily == ''Ubuntu'' || self.imageFamily == ''Ubuntu2204''
+                || self.imageFamily == ''Windows2025'') : true'
+            - message: FIPS is not supported for Windows2022
+              rule: '!has(self.fipsMode) || self.fipsMode != ''FIPS'' || !has(self.imageFamily)
+                || self.imageFamily != ''Windows2022'''
+            - message: fipsMode must be FIPS or omitted for Windows2025
+              rule: '!has(self.imageFamily) || self.imageFamily != ''Windows2025''
+                || !has(self.fipsMode) || self.fipsMode == ''FIPS'''
+            - message: TrustedLaunch is not supported for Windows2022
+              rule: '!has(self.imageFamily) || self.imageFamily != ''Windows2022''
+                || !has(self.security) || !has(self.security.trustedLaunch) || !((has(self.security.trustedLaunch.vtpm)
+                && self.security.trustedLaunch.vtpm) || (has(self.security.trustedLaunch.secureBoot)
+                && self.security.trustedLaunch.secureBoot))'
+            - message: linuxOSConfig is not supported for Windows image families
+              rule: '!has(self.linuxOSConfig) || !has(self.imageFamily) || !(self.imageFamily
+                in [''Windows2022'',''Windows2025''])'
+            - message: artifactStreaming is not supported for Windows image families
+              rule: '!has(self.artifactStreaming) || !has(self.artifactStreaming.enabled)
+                || self.artifactStreaming.enabled == false || !has(self.imageFamily)
+                || !(self.imageFamily in [''Windows2022'',''Windows2025''])'
+            - message: localDNS is not supported for Windows image families
+              rule: '!has(self.localDNS) || self.localDNS.mode == ''Disabled'' ||
+                !has(self.imageFamily) || !(self.imageFamily in [''Windows2022'',''Windows2025''])'
             - message: kubelet.failSwapOn must be set to false when linuxOSConfig.swapFileSize
                 is specified
               rule: '!has(self.linuxOSConfig) || !has(self.linuxOSConfig.swapFileSize)
--- a/pkg/apis/crds/karpenter.azure.com_aksnodeclasses.yaml
+++ b/pkg/apis/crds/karpenter.azure.com_aksnodeclasses.yaml
@@ -1005,6 +1005,8 @@ spec:
                 - Ubuntu2204
                 - Ubuntu2404
                 - AzureLinux
+                - Windows2022
+                - Windows2025
                 type: string
               kubelet:
                 description: |-
@@ -1680,6 +1682,12 @@ spec:
               rule: 'has(self.fipsMode) && self.fipsMode == ''FIPS'' ? (has(self.imageFamily)
                 && self.imageFamily != ''Ubuntu2204'' && self.imageFamily != ''Ubuntu2404'')
                 : true'
+            - message: FIPS is not supported for Windows image families
+              rule: '!has(self.fipsMode) || self.fipsMode != ''FIPS'' || !has(self.imageFamily)
+                || !(self.imageFamily in [''Windows2022'',''Windows2025''])'
+            - message: linuxOSConfig is not supported for Windows image families
+              rule: '!has(self.linuxOSConfig) || !has(self.imageFamily) || !(self.imageFamily
+                in [''Windows2022'',''Windows2025''])'
             - message: kubelet.failSwapOn must be set to false when linuxOSConfig.swapFileSize
                 is specified
               rule: '!has(self.linuxOSConfig) || !has(self.linuxOSConfig.swapFileSize)
--- a/pkg/apis/crds/karpenter.azure.com_aksnodeclasses.yaml
+++ b/pkg/apis/crds/karpenter.azure.com_aksnodeclasses.yaml
@@ -1031,6 +1031,8 @@ spec:
                 - Ubuntu2204
                 - Ubuntu2404
                 - AzureLinux
+                - Windows2022
+                - Windows2025
                 type: string
               kubelet:
                 description: |-
@@ -1724,14 +1726,35 @@ spec:
                 && ((has(self.security.trustedLaunch.vtpm) && self.security.trustedLaunch.vtpm)
                 || (has(self.security.trustedLaunch.secureBoot) && self.security.trustedLaunch.secureBoot)))))
                 : true'
-            - message: TrustedLaunch with FIPSMode FIPS is only supported for Ubuntu
-                and Ubuntu2204
+            - message: TrustedLaunch with FIPSMode FIPS is only supported for Ubuntu,
+                Ubuntu2204, and Windows2025
               rule: 'has(self.fipsMode) && self.fipsMode == ''FIPS'' && has(self.security)
                 && has(self.security.trustedLaunch) && ((has(self.security.trustedLaunch.vtpm)
                 && self.security.trustedLaunch.vtpm) || (has(self.security.trustedLaunch.secureBoot)
                 && self.security.trustedLaunch.secureBoot)) ? (!has(self.imageFamily)
-                || self.imageFamily == ''Ubuntu'' || self.imageFamily == ''Ubuntu2204'')
-                : true'
+                || self.imageFamily == ''Ubuntu'' || self.imageFamily == ''Ubuntu2204''
+                || self.imageFamily == ''Windows2025'') : true'
+            - message: FIPS is not supported for Windows2022
+              rule: '!has(self.fipsMode) || self.fipsMode != ''FIPS'' || !has(self.imageFamily)
+                || self.imageFamily != ''Windows2022'''
+            - message: fipsMode must be FIPS or omitted for Windows2025
+              rule: '!has(self.imageFamily) || self.imageFamily != ''Windows2025''
+                || !has(self.fipsMode) || self.fipsMode == ''FIPS'''
+            - message: TrustedLaunch is not supported for Windows2022
+              rule: '!has(self.imageFamily) || self.imageFamily != ''Windows2022''
+                || !has(self.security) || !has(self.security.trustedLaunch) || !((has(self.security.trustedLaunch.vtpm)
+                && self.security.trustedLaunch.vtpm) || (has(self.security.trustedLaunch.secureBoot)
+                && self.security.trustedLaunch.secureBoot))'
+            - message: linuxOSConfig is not supported for Windows image families
+              rule: '!has(self.linuxOSConfig) || !has(self.imageFamily) || !(self.imageFamily
+                in [''Windows2022'',''Windows2025''])'
+            - message: artifactStreaming is not supported for Windows image families
+              rule: '!has(self.artifactStreaming) || !has(self.artifactStreaming.enabled)
+                || self.artifactStreaming.enabled == false || !has(self.imageFamily)
+                || !(self.imageFamily in [''Windows2022'',''Windows2025''])'
+            - message: localDNS is not supported for Windows image families
+              rule: '!has(self.localDNS) || self.localDNS.mode == ''Disabled'' ||
+                !has(self.imageFamily) || !(self.imageFamily in [''Windows2022'',''Windows2025''])'
             - message: kubelet.failSwapOn must be set to false when linuxOSConfig.swapFileSize
                 is specified
               rule: '!has(self.linuxOSConfig) || !has(self.linuxOSConfig.swapFileSize)
--- a/pkg/apis/v1beta1/aksnodeclass.go
+++ b/pkg/apis/v1beta1/aksnodeclass.go
@@ -61,6 +61,8 @@ func (a *ArtifactStreaming) IsEnabled(arch string) bool {
 // AKSNodeClassSpec is the top level specification for the AKS Karpenter Provider.
 // This will contain configuration necessary to launch instances in AKS.
 // +kubebuilder:validation:XValidation:message="FIPS is not yet supported for Ubuntu2204 or Ubuntu2404",rule="has(self.fipsMode) && self.fipsMode == 'FIPS' ? (has(self.imageFamily) && self.imageFamily != 'Ubuntu2204' && self.imageFamily != 'Ubuntu2404') : true"
+// +kubebuilder:validation:XValidation:message="FIPS is not supported for Windows image families",rule="!has(self.fipsMode) || self.fipsMode != 'FIPS' || !has(self.imageFamily) || !(self.imageFamily in ['Windows2022','Windows2025'])"
+// +kubebuilder:validation:XValidation:message="linuxOSConfig is not supported for Windows image families",rule="!has(self.linuxOSConfig) || !has(self.imageFamily) || !(self.imageFamily in ['Windows2022','Windows2025'])"
 // +kubebuilder:validation:XValidation:message="kubelet.failSwapOn must be set to false when linuxOSConfig.swapFileSize is specified",rule="!has(self.linuxOSConfig) || !has(self.linuxOSConfig.swapFileSize) || (has(self.kubelet) && has(self.kubelet.failSwapOn) && self.kubelet.failSwapOn == false)"
 type AKSNodeClassSpec struct {
 	// vnetSubnetID is the subnet used by nics provisioned with this nodeclass.
@@ -79,7 +81,7 @@ type AKSNodeClassSpec struct {
 	ImageID *string `json:"-"`
 	// imageFamily is the image family that instances use.
 	// +default="Ubuntu"
-	// +kubebuilder:validation:Enum:={Ubuntu,Ubuntu2204,Ubuntu2404,AzureLinux}
+	// +kubebuilder:validation:Enum:={Ubuntu,Ubuntu2204,Ubuntu2404,AzureLinux,Windows2022,Windows2025}
 	// +optional
 	ImageFamily *string `json:"imageFamily,omitempty"`
 	// fipsMode controls FIPS compliance for the provisioned nodes
--- a/pkg/apis/v1beta1/aksnodeclass.go
+++ b/pkg/apis/v1beta1/aksnodeclass.go
@@ -62,7 +62,13 @@ func (a *ArtifactStreaming) IsEnabled(arch string) bool {
 // This will contain configuration necessary to launch instances in AKS.
 // +kubebuilder:validation:XValidation:message="FIPS is not yet supported for Ubuntu2404",rule="has(self.fipsMode) && self.fipsMode == 'FIPS' ? (has(self.imageFamily) && self.imageFamily != 'Ubuntu2404') : true"
 // +kubebuilder:validation:XValidation:message="TrustedLaunch is required for FIPS support with Ubuntu2204",rule="has(self.fipsMode) && self.fipsMode == 'FIPS' ? (has(self.imageFamily) && (self.imageFamily != 'Ubuntu2204' || (has(self.security) && has(self.security.trustedLaunch) && ((has(self.security.trustedLaunch.vtpm) && self.security.trustedLaunch.vtpm) || (has(self.security.trustedLaunch.secureBoot) && self.security.trustedLaunch.secureBoot))))) : true"
-// +kubebuilder:validation:XValidation:message="TrustedLaunch with FIPSMode FIPS is only supported for Ubuntu and Ubuntu2204",rule="has(self.fipsMode) && self.fipsMode == 'FIPS' && has(self.security) && has(self.security.trustedLaunch) && ((has(self.security.trustedLaunch.vtpm) && self.security.trustedLaunch.vtpm) || (has(self.security.trustedLaunch.secureBoot) && self.security.trustedLaunch.secureBoot)) ? (!has(self.imageFamily) || self.imageFamily == 'Ubuntu' || self.imageFamily == 'Ubuntu2204') : true"
+// +kubebuilder:validation:XValidation:message="TrustedLaunch with FIPSMode FIPS is only supported for Ubuntu, Ubuntu2204, and Windows2025",rule="has(self.fipsMode) && self.fipsMode == 'FIPS' && has(self.security) && has(self.security.trustedLaunch) && ((has(self.security.trustedLaunch.vtpm) && self.security.trustedLaunch.vtpm) || (has(self.security.trustedLaunch.secureBoot) && self.security.trustedLaunch.secureBoot)) ? (!has(self.imageFamily) || self.imageFamily == 'Ubuntu' || self.imageFamily == 'Ubuntu2204' || self.imageFamily == 'Windows2025') : true"
+// +kubebuilder:validation:XValidation:message="FIPS is not supported for Windows2022",rule="!has(self.fipsMode) || self.fipsMode != 'FIPS' || !has(self.imageFamily) || self.imageFamily != 'Windows2022'"
+// +kubebuilder:validation:XValidation:message="fipsMode must be FIPS or omitted for Windows2025",rule="!has(self.imageFamily) || self.imageFamily != 'Windows2025' || !has(self.fipsMode) || self.fipsMode == 'FIPS'"
+// +kubebuilder:validation:XValidation:message="TrustedLaunch is not supported for Windows2022",rule="!has(self.imageFamily) || self.imageFamily != 'Windows2022' || !has(self.security) || !has(self.security.trustedLaunch) || !((has(self.security.trustedLaunch.vtpm) && self.security.trustedLaunch.vtpm) || (has(self.security.trustedLaunch.secureBoot) && self.security.trustedLaunch.secureBoot))"
+// +kubebuilder:validation:XValidation:message="linuxOSConfig is not supported for Windows image families",rule="!has(self.linuxOSConfig) || !has(self.imageFamily) || !(self.imageFamily in ['Windows2022','Windows2025'])"
+// +kubebuilder:validation:XValidation:message="artifactStreaming is not supported for Windows image families",rule="!has(self.artifactStreaming) || !has(self.artifactStreaming.enabled) || self.artifactStreaming.enabled == false || !has(self.imageFamily) || !(self.imageFamily in ['Windows2022','Windows2025'])"
+// +kubebuilder:validation:XValidation:message="localDNS is not supported for Windows image families",rule="!has(self.localDNS) || self.localDNS.mode == 'Disabled' || !has(self.imageFamily) || !(self.imageFamily in ['Windows2022','Windows2025'])"
 // +kubebuilder:validation:XValidation:message="kubelet.failSwapOn must be set to false when linuxOSConfig.swapFileSize is specified",rule="!has(self.linuxOSConfig) || !has(self.linuxOSConfig.swapFileSize) || (has(self.kubelet) && has(self.kubelet.failSwapOn) && self.kubelet.failSwapOn == false)"
 type AKSNodeClassSpec struct {
 	// vnetSubnetID is the subnet used by nics provisioned with this nodeclass.
@@ -81,7 +87,7 @@ type AKSNodeClassSpec struct {
 	ImageID *string `json:"-"`
 	// imageFamily is the image family that instances use.
 	// +default="Ubuntu"
-	// +kubebuilder:validation:Enum:={Ubuntu,Ubuntu2204,Ubuntu2404,AzureLinux}
+	// +kubebuilder:validation:Enum:={Ubuntu,Ubuntu2204,Ubuntu2404,AzureLinux,Windows2022,Windows2025}
 	// +optional
 	ImageFamily *string `json:"imageFamily,omitempty"`
 	// fipsMode controls FIPS compliance for the provisioned nodes
@@ -751,6 +757,13 @@ func (in *AKSNodeClass) IsSecureBootEnabled() bool {
 // IsTrustedLaunchEnabled returns whether any Trusted Launch-backed setting is enabled.
 func (in *AKSNodeClass) IsTrustedLaunchEnabled() bool {
 	return in.IsVTPMEnabled() || in.IsSecureBootEnabled()
+}
+
+// IsFIPSEnabled returns the effective FIPS setting. Windows2025 requires FIPS in AKS,
+// so it is enabled even when fipsMode is omitted from the NodeClass.
+func (in *AKSNodeClass) IsFIPSEnabled() bool {
+	return lo.FromPtr(in.Spec.ImageFamily) == Windows2025ImageFamily ||
+		lo.FromPtr(in.Spec.FIPSMode) == FIPSModeFIPS
 }
 
 // IsArtifactStreamingEnabled returns whether artifact streaming should be enabled for this node class.
--- a/pkg/apis/v1beta1/labels.go
+++ b/pkg/apis/v1beta1/labels.go
@@ -42,7 +42,7 @@ func init() {
 	karpv1.WellKnownValuesForRequirements[AKSLabelMode] = sets.New(ModeSystem, ModeUser)
 	karpv1.WellKnownValuesForRequirements[AKSLabelScaleSetPriority] = sets.New(ScaleSetPriorityRegular, ScaleSetPrioritySpot)
 	karpv1.WellKnownValuesForRequirements[AKSLabelPriority] = sets.New(PriorityRegular, PrioritySpot)
-	karpv1.WellKnownValuesForRequirements[AKSLabelOSSKU] = sets.New(OSSKUUbuntu, OSSKUAzureLinux)
+	karpv1.WellKnownValuesForRequirements[AKSLabelOSSKU] = sets.New(OSSKUUbuntu, OSSKUAzureLinux, OSSKUWindows2022, OSSKUWindows2025)
 	karpv1.WellKnownValuesForRequirements[AKSLabelFIPSEnabled] = sets.New("true")
 }
 
@@ -190,11 +190,26 @@ const (
 	Ubuntu2204ImageFamily = "Ubuntu2204"
 	Ubuntu2404ImageFamily = "Ubuntu2404"
 	AzureLinuxImageFamily = "AzureLinux"
+
+	// Windows image families. Each maps 1:1 to an AKS Windows OSSKU.
+	Windows2022ImageFamily = "Windows2022"
+	Windows2025ImageFamily = "Windows2025"
 )
 
 const (
 	OSSKUUbuntu     = "Ubuntu"
 	OSSKUAzureLinux = "AzureLinux"
+
+	// Windows os-sku label values. These match the kubernetes.azure.com/os-sku
+	// values AKS applies to Windows nodes (version-specific, unlike Ubuntu which collapses).
+	OSSKUWindows2022 = "Windows2022"
+	OSSKUWindows2025 = "Windows2025"
+)
+
+// OS label values for kubernetes.io/os (and the AKS Machine OSType mapping).
+const (
+	OSLinux   = "linux"
+	OSWindows = "windows"
 )
 
 const (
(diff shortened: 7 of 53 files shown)


charts/karpenter-crd/templates/karpenter.azure.com_aksnodeclasses.yaml
```diff
@@ -1005,6 +1005,10 @@ spec:
                 - Ubuntu2204
                 - Ubuntu2404
                 - AzureLinux
+                - Windows2019
+                - Windows2022
+                - Windows2025
+                - WindowsAnnual
```
theunrepentantgeek: Maybe don't include `WindowsAnnual` as an option - according to [this announcement](https://learn.microsoft.com/en-us/azure/aks/windows-annual-channel?tabs=azure-cli#use-windows-server-annual-channel-for-containers-on-aks) the channel is deprecated.

pkg/apis/crds/karpenter.azure.com_aksnodeclasses.yaml
```diff
@@ -1005,6 +1005,10 @@ spec:
                 - Ubuntu2204
                 - Ubuntu2404
                 - AzureLinux
+                - Windows2019
+                - Windows2022
+                - Windows2025
+                - WindowsAnnual
```
theunrepentantgeek: Same here - maybe don't include `WindowsAnnual`.

pkg/apis/v1beta1/labels.go
```diff
@@ -238,3 +263,13 @@ func GetOSSKUFromImageFamily(imageFamily string) string {
...
 }
+
+// GetOSForImageFamily returns the kubernetes.io/os value ("linux" or "windows") for the
+// given imageFamily. Windows families return "windows"; everything else (including empty,
+// which defaults to Ubuntu) returns "linux".
+func GetOSForImageFamily(imageFamily string) string {
+	if WindowsFamilies.Has(imageFamily) {
+		return "windows"
+	}
+	return "linux"
```
theunrepentantgeek: Assuming this is a predicate used to trigger different functionality for each case, these strings should be defined as constants to eliminate the possibility of a typo.

pkg/apis/v1beta1/labels.go
```diff
@@ -238,3 +263,13 @@ func GetOSSKUFromImageFamily(imageFamily string) string {
 func IsAKSLabel(label string) bool {
 	return strings.HasPrefix(label, AKSLabelDomain+"/") || aksLegacyLabels.Has(label)
 }
+
+// GetOSForImageFamily returns the kubernetes.io/os value ("linux" or "windows") for the
+// given imageFamily. Windows families return "windows"; everything else (including empty,
+// which defaults to Ubuntu) returns "linux".
+func GetOSForImageFamily(imageFamily string) string {
+	if WindowsFamilies.Has(imageFamily) {
```
theunrepentantgeek: Is `imageFamily` case sensitive or insensitive? The current implementation appears to be case-sensitive, which means `windows2019` would return `linux` ... 

Either Fix this to be case-insensitive OR We need edge-validation to ensure case-incorrect values never get here.

charts/karpenter-crd/templates/karpenter.azure.com_aksnodeclasses.yaml
theunrepentantgeek: I checked in with the AKS Windows (@timmy-wright), and both `WindowsAnnual` and `Windows2019` are going, or gone. No need for this PR to support either.

pkg/providers/imagefamily/windows.go
```diff
@@ -0,0 +1,165 @@
...
+
+	WindowsAnnual23H2Gen2ImageDefinition = "windows-23H2-gen2"
+	WindowsAnnual23H2ImageDefinition     = "windows-23H2"
+)
+
+// Windows is the image family for AKS Windows nodes. It is parameterized by family
+// (one of the v1beta1 Windows*ImageFamily values), each of which maps to a Windows OSSKU.
+//
+// Windows nodes are only supported via the AKS Machine API provision mode, which always
+// uses the AKS-managed shared image gallery (SIG). Accordingly DefaultImages returns images
```
theunrepentantgeek:
```suggestion
// uses the AKS-managed shared image gallery (SIG). Accordingly, DefaultImages() returns images
```

pkg/providers/instance/aksmachineinstancehelpers.go
```diff
@@ -63,13 +63,9 @@ func (p *DefaultAKSMachineProvider) buildAKSMachineTemplate(ctx context.Context,
...
 	// E.g., "AKSUbuntu-2204gen2containerd-2023.11.15"
-	vmImageID, err := p.imageResolver.ResolveNodeImageFromNodeClass(nodeClass, instanceType)
-	if err != nil {
-		return nil, fmt.Errorf("failed to resolve VM image ID: %w", err)
-	}
-	nodeImageVersion, err := utils.GetAKSMachineNodeImageVersionFromImageID(vmImageID)
+	nodeImageVersionPtr, err := p.resolveNodeImageVersion(nodeClass, instanceType)
 	if err != nil {
-		return nil, fmt.Errorf("failed to convert VM image ID to NodeImageVersion: %w", err)
+		return nil, err
```
theunrepentantgeek: Why remove the wrapping of `err` - it seems to be adding useful context.

pkg/providers/instance/aksmachineinstancehelpers.go
```diff
@@ -284,38 +275,99 @@ func configurePriority(capacityType string) *armcontainerservice.ScaleSetPriorit
...
+	}
+}
+
+// configureOSType returns the AKS Machine OSType (Linux or Windows) for the NodeClass's
+// image family. Windows families map to OSTypeWindows; everything else to OSTypeLinux.
+func configureOSType(nodeClass *v1beta1.AKSNodeClass) *armcontainerservice.OSType {
+	if v1beta1.WindowsFamilies.Has(lo.FromPtr(nodeClass.Spec.ImageFamily)) {
+		return lo.ToPtr(armcontainerservice.OSTypeWindows)
+	}
+	return lo.ToPtr(armcontainerservice.OSTypeLinux)
```
theunrepentantgeek, line 328: Starting with Go 1.26, the `new()` function can take an expression, and it target-types the result.

So, this is equivilent:
```suggestion
	return new(armcontainerservice.OSTypeLinux)
```

More as a hint for next time, no pressing need to update the PR.

pkg/providers/instance/aksmachineinstanceutils.go
```diff
@@ -181,11 +181,24 @@ func FindNodePoolFromAKSMachine(ctx context.Context, aksMachine *armcontainerser
 // ASSUMPTION: NodeClaim name is in the format of <NodePool name>-<hash suffix>
 // If total length exceeds AKS machine name limit, the exceeded part will be replaced with another deterministic hash.
 // E.g., "thisisalongnodepoolname-a1b2c" --> "thisisalongnoz9y8x7-a1b2c"
-func GetAKSMachineNameFromNodeClaimName(nodeClaimName string) (string, error) {
+func GetAKSMachineNameFromNodeClaimName(nodeClaimName string, isWindows bool) (string, error) {
```
theunrepentantgeek: Boolean parameters like this become a maintenance problem because there's no affordance _at the call site_ about what it means:

``` go
machineName, err := GetAKSMachineNameFromNodeClaimName(nodeClaimName, false)
```

(Fun fact: Steve McConnell calls this out in his book [Code Complete](https://en.wikipedia.org/wiki/Code_Complete), published in 1993!)

Better to define two methods, one for Linux, and one for Windows, and do the test to decide which to call at the call site. Clearer, easier to maintain, and less likely for a future maintainer to break.

pkg/providers/instance/aksmachineinstanceutils.go
```diff
@@ -181,11 +181,24 @@ func FindNodePoolFromAKSMachine(ctx context.Context, aksMachine *armcontainerser
...
 	const prefixHashLength = 6         // The length of the hashed part replacing the exceeded part of the prefix.
 	// If 6, given alphanumeric hash, there will be a total of 36^6 = 2,176,782,336 combinations.
 
+	// Windows machines have a much shorter name limit in the AKS Machine API (12 chars in the
+	// reserved Karpenter agent pool, due to the Windows NetBIOS computer-name limit). The
+	// NodeClaim name is far longer, so derive a short, deterministic, collision-resistant name.
+	// Format: "w" + 11-char alphanumeric hash (starts with a letter; no hyphens), length 12.
+	if isWindows {
+		const maxWindowsAKSMachineNameLength = 12
+		hash, err := utils.GetAlphanumericHash(nodeClaimName, maxWindowsAKSMachineNameLength-1)
```
theunrepentantgeek: How much entropy (randomness) does this name have? 

I'm trying to work out how collision-resilient this is. How unique does the NetBIOS name need to be? Unique within the cluster? Virtual Network? Region? 

If `nodeClaimName` has 64 bits of entropy (which is what the hash function uses), then we're virtually guaranteed to be collision free (a 50% probability of collision requires 5 billion names), but if it has (say), just 16 bits of entropy, then a set of just 50 nodes has nearly a 2% chance of collision.

I guess that raises the other question - what happens if a naming collision occurs?

theunrepentantgeek (changes requested):
First pass review, needs another look from someone more familiar with the codebase.

Both `Windows2019` and `WindowsAnnual` are on their way out (going or already gone), so we can simplify this code by omitting those. 

Plus, I have a question about the uniqueness of Windows NetBIOS names - I suspect there's not sufficient entropy to make naming collisions unlikely.

pkg/providers/instance/aksmachineinstanceutils.go
theunrepentantgeek: I've dug into this further - the node claim includes a guaranteed unique "random" suffix generated by the Kubernetes API server. That's ~28 bits of entropy. 

With a perfect hashing function, we wouldn't get a collision until we exceeded ~60M nodes, far beyond any practical cluster size.

Conversely, if the hashing function was no better than a random number generator, any individual 50-node cluster would have a 1 in 219,000 chance of having a collision. At Azure scale, this becomes quite likely.

So the question becomes - how good is this hashing function? Can we rely on it generating unique names?

charts/karpenter-crd/templates/karpenter.azure.com_aksnodeclasses.yaml
MarcPow: Done. Per your follow-up (and the AKS Windows team), both `Windows2019` and `WindowsAnnual` were removed entirely — the imageFamily enum/CEL now only allow `Windows2022` and `Windows2025`, and this generated CRD was regenerated to match.

pkg/apis/crds/karpenter.azure.com_aksnodeclasses.yaml
MarcPow: Done — `WindowsAnnual` (and `Windows2019`) removed here too; both generated CRD copies were regenerated and now only list `Windows2022`/`Windows2025`.

pkg/apis/v1beta1/labels.go
MarcPow: Done. The image families are exported string constants, and the OS values are now constants too (`OSLinux`/`OSWindows`); `GetOSForImageFamily` returns those instead of bare strings, so a typo can't slip through.

pkg/apis/v1beta1/labels.go
MarcPow: Fixed to be case-insensitive: added `IsWindowsImageFamily` (using `strings.EqualFold`) and routed all family checks through it, so casing variants no longer fall through to `linux`. (Windows2019 was also removed, so that specific example is moot, but the casing is now handled generally.)

pkg/providers/imagefamily/windows.go
MarcPow: Applied your suggested wording verbatim.

pkg/providers/instance/aksmachineinstancehelpers.go
MarcPow: Restored the wrapping — it's back to `fmt.Errorf("failed to resolve node image version: %w", err)`.

pkg/providers/instance/aksmachineinstancehelpers.go
MarcPow: Thanks for the tip on Go 1.26 `new(expr)`. Kept `lo.ToPtr(...)` here for now to stay consistent with the rest of the codebase's pointer convention (no functional difference); happy to switch the codebase over once we're on 1.26.

pkg/providers/instance/aksmachineinstanceutils.go
MarcPow: Done — replaced the boolean parameter with two call-site-explicit methods, `GetLinuxAKSMachineName(name)` and `GetWindowsAKSMachineName(name, maxLen)`, dispatched by the caller. No more opaque bool at the call site.

pkg/providers/instance/aksmachineinstanceutils.go
MarcPow: Good question — here's the analysis. The AKS machine name must be unique within the agent pool (the RP composes the VM/NetBIOS name from `<pool>` + `<machine>`). `GetAlphanumericHash` hashes the NodeClaim name with `hashstructure.Hash` (FNV-1a 64-bit, well-distributed) and base36-extracts it.

The name is now **pool-aware**:
- **Reserved NAP pool (`aksmanagedap`) — the production path**: `"w"` + 11 base36 chars => 36^11 ≈ 2^57 of output space. As you noted, the NodeClaim carries the API-server random suffix (~28 bits of input entropy), and 2^57 ≫ 2^28, so even with a non-cryptographic hash there's effectively no collision risk at any realistic cluster size.
- **Custom/self-hosted pool**: only `"w"` + 4 chars => 36^4 ≈ 1.7M. This is intentionally for small self-hosted/test clusters only — the e2e suite gates on a pool name `<= 6` chars — not a production path.

So: production (NAP) is collision-free for all practical purposes; the short custom-pool budget is test-scoped by design.

MarcPow:
@theunrepentantgeek thanks for the thorough review — all of your comments are addressed and the conversations are resolved. Summary:

- **`Windows2019` + `WindowsAnnual` removed** entirely (enum, CEL, image-family defs, and both regenerated CRDs) — only `Windows2022`/`Windows2025` remain.
- **Case-insensitivity**: added `IsWindowsImageFamily` (`strings.EqualFold`); all family checks route through it.
- **Magic strings → constants**: `OSLinux`/`OSWindows`; `GetOSForImageFamily` returns those.
- **Boolean param → two methods**: `GetLinuxAKSMachineName` / `GetWindowsAKSMachineName(maxLen)`, chosen at the call site.
- **Restored `%w` error wrapping** on the node-image-version resolve.
- **Doc**: applied your `DefaultImages()` wording.
- **Naming entropy**: answered inline — the reserved NAP (production) pool gets an 11-char base36 hash (~2^57), far exceeding the NodeClaim's ~28-bit input entropy, so it's collision-free at any realistic scale; the short custom-pool budget is test-scoped (gated to pool names ≤ 6 chars).
- `new(expr)` Go 1.26 tip noted; kept `lo.ToPtr` for codebase consistency for now.

**Follow-up beyond the review** (worth a look):
- Found that for `Windows2022` the RP defaults to a **Gen1** image and only picks Gen2 when the caller sends `UseWindowsGen2VM`. Since NAP often selects Gen2-only SKUs, Karpenter now sends that header when the selected SKU supports Gen2 — so Windows provisions on **any** SKU generation. Validated end-to-end self-hosted on a Gen2-only SKU (`Standard_D2als_v7`) that previously failed.
- Added a separate commit wiring the Windows e2e suite into the CI matrix on a dedicated (non-Cilium) `az-mkaks-windows` cluster in machine-API mode.

Re-requesting review when you have a moment. 🙏

pkg/providers/azclient/aksmachinesheaderbatch/batchkey.go
```diff
@@ -38,7 +38,10 @@ func determineBatchKey(item *aksMachineCreatePayload) (string, error) {
 	}
 
 	hash := sha256.Sum256(jsonBytes)
-	prefix := item.resourceGroupName + "/" + item.resourceName + "/" + item.agentPoolName + "/"
+	// useWindowsGen2VM is part of the key so machines that disagree on the requested Windows image
+	// generation never coalesce into one batch (they also differ in vmSize/OSSKU, so this is
+	// defensive). The executor reads the value from the first request when composing batch headers.
+	prefix := fmt.Sprintf("%s/%s/%s/gen2=%t/", item.resourceGroupName, item.resourceName, item.agentPoolName, item.useWindowsGen2VM)
```
comtalyst: Suffix should be more appropriate, if we want to preserve prefix for the resource location.

pkg/providers/azclient/aksmachinesheaderbatch/executor.go
```diff
@@ -35,6 +36,10 @@ type aksMachineCreatePayload struct {
 	agentPoolName     string
 	machineName       string
 	machineBody       *armcontainerservice.Machine
+	// useWindowsGen2VM requests a Generation 2 Windows image from the RP for this machine. All
+	// requests in a batch share the same value (it is part of the batch key), so it is read from
+	// the first request when composing the batch's HTTP headers.
+	useWindowsGen2VM bool
```
comtalyst: Let's make it clear that a parameter like this could have been a part of/on the same dimension as `machineBody`, but not in practice due to specific reasons. In this case, it is that we want to customize this internal value that Machine API defaulted it incorrectly(?), and luckily the header allowed us to do so(?). This is so we don't add more dimensions to this interface and others.

Ideally, I could imagine passing another struct `additionalParameters` (one member for now) instead of individual parameters, so we don't expose lower-level details in this layer. Though _if_ we are sure this will be temporary, making the comments clear, here and elsewhere, should be enough.

pkg/providers/imagefamily/consts.go
```diff
@@ -22,7 +22,11 @@ const (
 
 	AKSUbuntuResourceGroup     = "AKS-Ubuntu"
 	AKSAzureLinuxResourceGroup = "AKS-AzureLinux"
+	// AKSWindowsResourceGroup is the resource group hosting the AKS Windows shared image gallery.
```
comtalyst: No need for comments, for consistency. (it is hard to watch over AI on something like this...)

pkg/providers/instance/aksmachineinstanceutils.go
```diff
@@ -181,9 +181,57 @@ func FindNodePoolFromAKSMachine(ctx context.Context, aksMachine *armcontainerser
...
+// name for Windows, bounded to maxNameLength (see WindowsMachineNameMaxLength). Windows machine names are
+// constrained by the NetBIOS computer-name limit, so the (much longer) NodeClaim name is hashed rather
+// than truncated. Format: "w" + (maxNameLength-1)-char alphanumeric hash (starts with a letter; no
+// hyphens). Collision-resistance scales with maxNameLength: the reserved NAP budget of 12 gives an
+// 11-char (~36^11 ≈ 2^57) hash space, far exceeding the NodeClaim's input entropy at any cluster scale.
+func GetWindowsAKSMachineName(nodeClaimName string, maxNameLength int) (string, error) {
+	if maxNameLength < 2 {
+		return "", fmt.Errorf("windows AKS machine name max length must be >= 2, got %d", maxNameLength)
+	}
+	hash, err := utils.GetAlphanumericHash(nodeClaimName, maxNameLength-1)
```
comtalyst: (discussing offline) Let's go with the options below unless you have other opinions.

<blockquote>

- 12 chars (NAP only):
  - **(1) Hash NodePool name, keep NodeClaim suffix hash, similar to Linux path**
    - Should probably hash the entire NodePool name rather than just the overflowing suffix. Keeping 1-2 chars un-hashed is awkward?
    - E.g., **<u>`thisisanodepoolname`</u>**`-a1b2c` --> **<u>`z9y8x7`</u>**`-a1b2c`
    - Node<u>Pool</u> name cardinality: 36^6 = 2176782336
      - Best case collision rate for 100 = **~2.2739e-6**
    - (+) Similarity between NodeClaim name and machine name
- 5 chars:
  - **(4) Use existing NodeClaim suffix**
    - E.g., `thisisanodepoolname-`**<u>`a1b2c`</u>** --> **<u>`a1b2c`</u>**
    - Node<u>Claim</u> name cardinality: 36^5 = 60466176
      - Best case collision rate for 1k, 5k, 100k = **~0.00823**, **~0.18673**, **~1**
    - (-) Insufficient hash cardinality for large clusters
    - (+) Similarity between NodeClaim name and machine name
    - (+) Simple
    - (-) Cross-module assumption - rely on Karpenter core hash to be fair

</blockquote>

pkg/providers/instance/aksmachineinstance.go
```diff
@@ -448,16 +459,38 @@ func (p *DefaultAKSMachineProvider) beginCreateMachine(
...
+// rejects the create if the chosen VM size does not support Gen2. Karpenter selects the cheapest
+// compatible SKU, which is frequently Gen2-only, so we request Gen2 exactly when the selected SKU
+// supports it: Gen2-only and dual-generation SKUs get Gen2 (preferred), while Gen1-only SKUs fall
+// back to the RP's Gen1 default. This mirrors the gen2-then-gen1 preference of the in-provider
+// (VM-direct) Windows image family and lets Windows NodePools provision on any SKU generation.
+func shouldUseWindowsGen2VM(nodeClass *v1beta1.AKSNodeClass, instanceType *corecloudprovider.InstanceType) bool {
+	if !v1beta1.IsWindowsImageFamily(lo.FromPtr(nodeClass.Spec.ImageFamily)) {
+		return false
+	}
+	return instanceType.Requirements.Get(v1beta1.LabelSKUHyperVGeneration).Has(v1beta1.HyperVGenerationV2)
```
comtalyst, line 494: Karpenter need to align with AKS, which means it should not provision instances with the configuration that AKS cannot. If it is _really_ temporary and high impact, that might be exempted.

In this case, sounds like we should get more alignment on those factors? The latest discussion is saying that current AKS/Machine API behavior is intentional, and won't be changed?

pkg/providers/instance/aksmachineinstanceutils.go
```diff
@@ -181,9 +181,57 @@ func FindNodePoolFromAKSMachine(ctx context.Context, aksMachine *armcontainerser
...
+	return windowsMachineNameMaxLenCustom
+}
+
+// GetWindowsAKSMachineName derives a short, deterministic, hyphen-free AKS machine name from a NodeClaim
+// name for Windows, bounded to maxNameLength (see WindowsMachineNameMaxLength). Windows machine names are
+// constrained by the NetBIOS computer-name limit, so the (much longer) NodeClaim name is hashed rather
+// than truncated. Format: "w" + (maxNameLength-1)-char alphanumeric hash (starts with a letter; no
+// hyphens). Collision-resistance scales with maxNameLength: the reserved NAP budget of 12 gives an
+// 11-char (~36^11 ≈ 2^57) hash space, far exceeding the NodeClaim's input entropy at any cluster scale.
+func GetWindowsAKSMachineName(nodeClaimName string, maxNameLength int) (string, error) {
```
comtalyst: I don't think this function interface should expose name length, which is lower level details. Either a boolean of whether the machine pool is managed machine pool or a string of machine pool name is more appropriate?

pkg/providers/instance/aksmachineinstancehelpers.go
```diff
@@ -284,38 +275,97 @@ func configurePriority(capacityType string) *armcontainerservice.ScaleSetPriorit
...
-	case v1beta1.UbuntuImageFamily:
-		fallthrough
-	default:
-		if enableFIPS {
-			ossku = armcontainerservice.OSSKUUbuntu
-		} else if imagefamily.UseUbuntu2404(orchestratorVersion) {
-			ossku = armcontainerservice.OSSKUUbuntu2404
-		} else {
-			ossku = armcontainerservice.OSSKUUbuntu2204
-		}
```
comtalyst, line 314: For this case, I think we should keep this clear, flat structure compared to the logic being distributed to `imageFamilyToARMOSSKU`, here, and `defaultUbuntuOSSKU()` without reuses.

pkg/providers/instance/aksmachineinstancehelpers.go
```diff
@@ -284,38 +275,97 @@ func configurePriority(capacityType string) *armcontainerservice.ScaleSetPriorit
...
+}
+
+// resolveNodeImageVersion resolves the AKS Machine API NodeImageVersion for the NodeClass.
+//
+// For Windows it returns nil: the AKS Machine API's node image version parser splits on "-"
+// and expects exactly "{gallery}-{name}-{version}", but Windows image names themselves
+// contain hyphens (e.g. "AKSWindows-2022-containerd-gen2-<ver>"), so passing an explicit
+// version is rejected as invalid. Leaving it nil lets the AKS RP resolve the latest image
+// from the OSSKU (it already owns Windows image/credential resolution server-side).
+func (p *DefaultAKSMachineProvider) resolveNodeImageVersion(nodeClass *v1beta1.AKSNodeClass, instanceType *corecloudprovider.InstanceType) (*string, error) {
```
comtalyst: We should rename this to `configureNodeImageVersion` to be consistent?

pkg/providers/instance/aksmachineinstancehelpers.go
```diff
@@ -284,38 +275,97 @@ func configurePriority(capacityType string) *armcontainerservice.ScaleSetPriorit
...
+	return lo.ToPtr(armcontainerservice.OSTypeLinux)
+}
+
+// resolveNodeImageVersion resolves the AKS Machine API NodeImageVersion for the NodeClass.
+//
+// For Windows it returns nil: the AKS Machine API's node image version parser splits on "-"
+// and expects exactly "{gallery}-{name}-{version}", but Windows image names themselves
+// contain hyphens (e.g. "AKSWindows-2022-containerd-gen2-<ver>"), so passing an explicit
+// version is rejected as invalid. Leaving it nil lets the AKS RP resolve the latest image
+// from the OSSKU (it already owns Windows image/credential resolution server-side).
```
comtalyst: I think this is a significant change and worth a wider discussion. Questions/concerns could be:
- So far, Karpenter always control node image version, and some assumptions (e.g., drift node image pinning, AKSNodeClass status fields) may have been made on that. Simply changing this may not be safe.
- Even if we can make it safe, is it really worth having two patterns to maintain?
- Of if we only omit the image ID here but not elsewhere, then there could be value inconsistency issue between what user expects and what it actually is.
- Why would Machine API reject it? Shouldn't the change be on that side?

cc @tallaxes @matthchr @rakechill

README.md
```diff
@@ -51,7 +51,7 @@ Karpenter provider for AKS can be used in two modes:
 ## Known limitations
 
 The following AKS features are not supported:
-* Windows nodes.
+* Windows nodes in self-hosted mode. Windows node provisioning is supported only with the AKS Machine API provision mode (used by Node Auto Provisioning); the self-hosted (scriptless) provision mode does not support Windows.
```
comtalyst: Note that Machine API works for self-hosted, just that the mode is not "announced" yet. Though there will be a lot of revisits from Machine API integration.

pkg/utils/image.go
```diff
@@ -38,6 +42,10 @@ func GetAKSMachineNodeImageVersionFromImageID(imageID string) (string, error) {
 
 // Convert from "/subscriptions/10945678-1234-1234-1234-123456789012/resourceGroups/AKS-Ubuntu/providers/Microsoft.Compute/galleries/AKSUbuntu/images/2204gen2containerd/versions/2022.10.03"
 // to "AKSUbuntu-2204gen2containerd-2022.10.03".
+//
+// For Windows the image definition is named like "windows-2022-containerd-gen2"; the
+// redundant "windows-" prefix is dropped so the result matches the AKS Windows
+// NodeImageVersion form, e.g. "AKSWindows-2022-containerd-gen2-20348.4529.251212".
```
comtalyst: Like Linux, could you explicitly surface how it looks like in ARM ID form?

pkg/apis/crds/karpenter.azure.com_aksnodeclasses.yaml
```diff
@@ -1680,6 +1682,12 @@ spec:
               rule: 'has(self.fipsMode) && self.fipsMode == ''FIPS'' ? (has(self.imageFamily)
                 && self.imageFamily != ''Ubuntu2204'' && self.imageFamily != ''Ubuntu2404'')
                 : true'
+            - message: FIPS is not supported for Windows image families
+              rule: '!has(self.fipsMode) || self.fipsMode != ''FIPS'' || !has(self.imageFamily)
+                || !(self.imageFamily in [''Windows2022'',''Windows2025''])'
+            - message: linuxOSConfig is not supported for Windows image families
+              rule: '!has(self.linuxOSConfig) || !has(self.imageFamily) || !(self.imageFamily
+                in [''Windows2022'',''Windows2025''])'
```
comtalyst: Good catch on these. 
If not already, let's dig deeper in the same category to ensure no more validation mismatches with AKS that could result in Karpenter trying to provision the config that AKS won't allow. E.g., K8s version (there is another ongoing issue that Karpenter might allow some bad combinations of Ubuntu version x K8s version), etc.

pkg/apis/v1beta1/labels.go
```diff
@@ -190,11 +190,26 @@ const (
 	Ubuntu2204ImageFamily = "Ubuntu2204"
 	Ubuntu2404ImageFamily = "Ubuntu2404"
 	AzureLinuxImageFamily = "AzureLinux"
+
+	// Windows image families. Each maps 1:1 to an AKS Windows OSSKU.
```
comtalyst: No need for comments for consistency

pkg/providers/imagefamily/windows.go
```diff
@@ -0,0 +1,151 @@
...
+// Windows is the image family for AKS Windows nodes. It is parameterized by family
+// (one of the v1beta1 Windows*ImageFamily values), each of which maps to a Windows OSSKU.
+//
+// Windows nodes are only supported via the AKS Machine API provision mode, which always
+// uses the AKS-managed shared image gallery (SIG). Accordingly, DefaultImages() returns images
+// only when useSIG is true; the community image gallery (CIG) is not supported for Windows.
+type Windows struct {
+	Options *parameters.StaticParameters
+	Family  string
+}
```
comtalyst: It would be more consistent if we have one family per struct, like Ubuntu and AzureLinux, and not having to pass in `Family`.

theunrepentantgeek (dismissed):

pkg/providers/azclient/aksmachinesheaderbatch/batchkey.go
MarcPow: Done. The batch key now preserves the resource path as the prefix (`<resource-group>/<cluster>/<pool>/`) and appends the template hash plus `/gen2=<bool>` as suffix data. Tests verify requests with different header options cannot coalesce.

pkg/providers/azclient/aksmachinesheaderbatch/executor.go
MarcPow: Done. Replaced the raw boolean plumbing with `CreateOptions`/`machineCreateOptions`. This makes it explicit that `UseWindowsGen2VM` is a create setting represented by an RP HTTP header rather than Machine body state, and keeps future header-only settings from adding more positional parameters.

pkg/providers/imagefamily/consts.go
MarcPow: Done. Removed the added comments from these constants to match the surrounding file.

pkg/providers/instance/aksmachineinstanceutils.go
MarcPow: Implemented the agreed split. Managed NAP (`aksmanagedap`) uses a 6-character hash of the full authoritative NodePool label plus `-` and the existing 5-character NodeClaim suffix (12 total). Custom/self-hosted pools use the existing 5-character suffix. This preserves traceability while giving managed NAP the larger collision space, and also handles maximum-length NodePool names safely.

pkg/providers/instance/aksmachineinstance.go
MarcPow: Kept the SKU-driven header after validating the current AKS behavior. `UseWindowsGen2VM` is an intentional, supported RP opt-in: Windows2022 otherwise defaults to Gen1, so Gen2-only SKUs fail without it. Karpenter sends it only for a Windows NodeClass when the selected SKU advertises Gen2; Gen1-only SKUs omit it. Full Azure Windows2022 and Windows2025 E2Es passed with unconstrained SKU selection.

pkg/providers/instance/aksmachineinstanceutils.go
MarcPow: Done. The helper now accepts the machines pool name rather than exposing the length limit. It internally selects the managed-NAP or custom-pool naming strategy.

pkg/providers/instance/aksmachineinstancehelpers.go
MarcPow: Done. Restored a single flat switch in `configureOSSKUAndFIPs`, including the Windows2022 and Windows2025 cases, instead of distributing the logic across a map and helper.

pkg/providers/instance/aksmachineinstancehelpers.go
MarcPow: Done. Renamed it to `configureNodeImageVersion`.

pkg/providers/instance/aksmachineinstancehelpers.go
MarcPow: Agreed that this requires deployment sequencing rather than becoming a permanent second pattern. The provider temporarily omits Windows `NodeImageVersion` because deployed Machine API parsers reject valid hyphenated AKSWindows values; it still records the actual version returned by the created Machine. A separate RP fix now parses all current AgentBaker Windows definitions while preserving non-Windows behavior. After that RP fix is deployed, a provider follow-up will restore explicit create-time pinning.

README.md
MarcPow: Corrected. The README now says Windows requires the Machine API provision mode, which is used by managed NAP and can also be used self-hosted; only the scriptless and bootstrapping-client modes are unsupported.

pkg/utils/image.go
MarcPow: Done. The comment now shows the complete Windows SIG ARM ID input and the resulting AKS Machine `NodeImageVersion`, including why the redundant `windows-` definition prefix is omitted from the public form.

pkg/apis/crds/karpenter.azure.com_aksnodeclasses.yaml
MarcPow: Audited the new Windows surface against the current RP and AgentBaker contracts. The PR now enforces Windows2022 FIPS-off; Windows2025 effective FIPS-on and Kubernetes 1.32+; Trusted Launch only for Windows2025 with its TL image; and rejects Windows artifact streaming, LocalDNS, and linuxOSConfig before provisioning. CEL/unit coverage and full Azure Windows2022/Windows2025 E2Es validate the aligned behavior. Unrelated existing Ubuntu/version combinations remain outside this PR.

pkg/apis/v1beta1/labels.go
MarcPow: Done. Removed the added comments here to match the surrounding constants.

pkg/providers/imagefamily/windows.go
MarcPow: Done. Split the parameterized type into separate `Windows2022` and `Windows2025` image-family structs, sharing only private helpers. Windows2025 also selects its dedicated Trusted Launch image when requested.

pkg/providers/instance/aksmachineinstanceutils.go
```diff
@@ -186,9 +186,44 @@ func FindNodePoolFromAKSMachine(ctx context.Context, aksMachine *armcontainerser
...
+	if len(nodeClaimName) < windowsNodeClaimSuffixLength {
+		return "", fmt.Errorf("invalid NodeClaim name %q, expected at least %d characters", nodeClaimName, windowsNodeClaimSuffixLength)
+	}
+	nodeClaimSuffix := nodeClaimName[len(nodeClaimName)-windowsNodeClaimSuffixLength:]
+
+	if machinesPoolName != reservedNAPMachinesPoolName {
+		return nodeClaimSuffix, nil
+	}
+	if nodePoolName == "" {
+		return "", fmt.Errorf("NodeClaim %q is missing the %q label", nodeClaimName, karpv1.NodePoolLabelKey)
```
comtalyst: This is irrelevant?

pkg/providers/instance/aksmachineinstanceutils.go
```diff
@@ -186,9 +186,44 @@ func FindNodePoolFromAKSMachine(ctx context.Context, aksMachine *armcontainerser
...
+		return "", fmt.Errorf("failed to hash NodePool name %q for Windows AKS machine name: %w", nodePoolName, err)
+	}
+	return nodePoolHash + "-" + nodeClaimSuffix, nil
+}
+
+// GetLinuxAKSMachineName derives a valid AKS machine name (<= 34 chars) from a NodeClaim name for Linux.
+// Names within the limit are used verbatim; longer names keep the legible prefix and trailing suffix and
+// hash the overflowing middle so the result stays deterministic and collision-resistant.
+func GetLinuxAKSMachineName(nodeClaimName string) (string, error) {
+	const prefixHashLength = 6 // The length of the hashed part replacing the exceeded part of the prefix.
```
comtalyst: I don't see that we have to make this different than Windows case (`windowsNodePoolHashLength`) in terms of placement. Either make both a global constant, or in-function.

pkg/providers/instance/aksmachineinstanceutils.go
```diff
@@ -186,9 +186,44 @@ func FindNodePoolFromAKSMachine(ctx context.Context, aksMachine *armcontainerser
...
+	reservedNAPMachinesPoolName  = "aksmanagedap"
+	windowsNodePoolHashLength    = 6
+	windowsNodeClaimSuffixLength = 5
+)
+
+// GetWindowsAKSMachineName derives a Windows AKS machine name from the NodeClaim and NodePool names.
+// Managed NAP pools have a 12-character budget and use a six-character hash of the full
+// NodePool name plus the existing five-character NodeClaim suffix. Custom pools have only
+// a five-character budget and use the existing NodeClaim suffix directly.
+func GetWindowsAKSMachineName(nodeClaimName, nodePoolName, machinesPoolName string) (string, error) {
```
comtalyst: We should be consistent with `GetLinuxAKSMachineName` as much as possible.

For example, let's abstract "nodepool name" as `prefix` and derive it from `nodeClaimName` like `GetLinuxAKSMachineName`.

pkg/providers/instance/aksmachineinstancehelpers.go
```diff
@@ -195,6 +186,9 @@ func configureGPUProfile(instanceType *corecloudprovider.InstanceType, nodeClass
 }
 
 func configureArtifactStreamingProfile(nodeClass *v1beta1.AKSNodeClass, instanceType *corecloudprovider.InstanceType) *armcontainerservice.AgentPoolArtifactStreamingProfile {
+	if v1beta1.IsWindowsImageFamily(lo.FromPtr(nodeClass.Spec.ImageFamily)) {
+		return nil
+	}
```
comtalyst: I was thinking that having only one layer of validation/block (on the API: NodeClass CRD) is enough. Duplicating that block here is harder to maintain. Thoughts?

comtalyst: (also apply to others of this category)

pkg/providers/instance/aksmachineinstancehelpers.go
comtalyst: RP fix - this is being investigated with RP team. Let's get more clarity on that before moving forward.

pkg/providers/instance/aksmachineinstance.go
comtalyst: Looks like the generation is reflected in node image ID, so we don't even need RP default if Karpenter can send node image ID to the API? And if so, we don't need the header in the first place. 
Let's wait for https://github.com/Azure/karpenter-provider-azure/pull/1726#discussion_r3509680818. I'm expecting that the issue there is easier to resolve than dealing with the fallout.

pkg/providers/instancetype/instancetypes.go
comtalyst, line 329: Need to filter out ARM if Windows only support AMD64?
```
func isArchitectureSupportedByImageFamily(architecture, imageFamily string) bool {
    if v1beta1.IsWindowsImageFamily(imageFamily) {
        return getArchitecture(architecture) == karpv1.ArchitectureAmd64
    }
    return true
}
```

test/pkg/environment/azure/environment.go
```diff
@@ -186,7 +187,9 @@ func NewEnvironment(t *testing.T) *Environment {
 	// Default to reserved managed machine agentpool name for NAP
 	azureEnv.MachineAgentPoolName = "aksmanagedap"
 	if azureEnv.InClusterController {
-		azureEnv.MachineAgentPoolName = "testmpool"
+		// Self-hosted machines pool name; matches AKS_MACHINES_POOL_NAME used at deploy time.
+		// Note: Windows machines require an agent pool name <= 6 characters, so keep this short.
+		azureEnv.MachineAgentPoolName = lo.Ternary(os.Getenv("AKS_MACHINES_POOL_NAME") == "", "testmpool", os.Getenv("AKS_MACHINES_POOL_NAME"))
```
comtalyst: I think we can change this "testmpool" everywhere to "mpool" for simplicity.

test/suites/windows/suite_test.go
```diff
@@ -0,0 +1,152 @@
+/*
```
comtalyst, line 1: Windows is more of a "dimension" to add to the matrix, and should be able to be tested using at least some of the existing suites in some way. 

The current structure might be enough to start, and more coverage is appreciated still. E.g., testing drift/nodes not getting drifted/consolidated outside of expectations.

comtalyst, line 1: And could you show the results of the added E2E passing? Ideally from GitHub actions.

comtalyst, line 1: Consider adding acceptance tests (i.e., unit end-to-end tests) in cloudprovider/suite_aksmachines_features_test.go in addition to this pure E2E.

comtalyst (commented):
The main thing left is https://github.com/Azure/karpenter-provider-azure/pull/1726#discussion_r3509680818 in which we are working with the team on it.

Also, could you update the PR description to be up-to-date (e.g., dropping `Windows2019`) and to match the format of other PRs?

pkg/providers/instance/aksmachineinstanceutils.go
MarcPow: Removed. The Windows naming helper no longer depends on the NodePool label or checks for its presence; it derives the prefix and generated suffix directly from `nodeClaimName`, matching the Linux helper's model.

pkg/providers/instance/aksmachineinstanceutils.go
MarcPow: Done. The Linux and Windows hash lengths now use package-level naming constants consistently.

pkg/providers/instance/aksmachineinstanceutils.go
MarcPow: Done. Windows naming now splits `nodeClaimName` at the final dash into the NodePool prefix and five-character generated suffix, like the Linux path. Managed NAP hashes that prefix; custom pools use the suffix directly.

pkg/providers/instance/aksmachineinstancehelpers.go
MarcPow: Agreed for the validation-only cases. Removed the redundant Windows guards from artifact-streaming and LocalDNS Machine-body configuration; AKSNodeClass CEL remains the single rejection layer. OS-specific serialization/defaulting such as omitting `LinuxProfile` and enforcing Windows2025 FIPS remains because those are output mappings rather than duplicate validation.

pkg/providers/instancetype/instancetypes.go
MarcPow: Fixed. Instance discovery now rejects ARM64 whenever the image family is Windows2022 or Windows2025, before offerings are advertised. Added coverage proving Linux retains ARM64 while both Windows families expose amd64 only.

test/pkg/environment/azure/environment.go
MarcPow: Done. Changed the default self-hosted Machine pool from `testmpool` to `mpool` across Makefiles, deployment configuration, workflow/action defaults, environment setup, and option tests.

test/suites/windows/suite_test.go
MarcPow: Kept the dedicated Windows live smoke suite as the starting point rather than expanding every E2E suite into an OS matrix in this PR. Added Machine API acceptance coverage for Windows2022/2025 creation and a Windows2022 no-drift case, covering the requested provider behaviors without multiplying live cluster jobs. Broader consolidation/matrix reuse remains follow-up scope.

MarcPow: Updated the PR description with the exact passing Azure results: Windows2022 passed in 356.40s and Windows2025 passed in 335.33s with the FIPS label asserted. Both used the committed Windows suite, self-hosted Machine API mode, Azure CNI Overlay, and unconstrained SKU selection. The suite is wired into GitHub Actions; an upstream-hosted run requires maintainer secrets because the PR source is a fork.

MarcPow: Done. Added Machine API cloud-provider acceptance tests for Windows2022 and Windows2025 (OSType, OSSKU, FIPS, Linux-only profile absence, os/arch/OSSKU labels) plus a Windows2022 no-drift acceptance case. Updated the fake RP image resolution to return realistic Windows versions so drift behavior is exercised correctly.

pkg/providers/instance/aksmachineinstance.go
MarcPow: Won't fix in this PR. The Gen2 header behavior is intentionally retained and decoupled from the initial Windows support change. We will revisit whether it can be removed only after explicit Windows image pinning is restored and the RP validation/defaulting order is verified end to end.

pkg/providers/instance/aksmachineinstancehelpers.go
MarcPow: Won't fix in this PR. Explicit Windows image pinning is deliberately decoupled until the companion aks-rp parser change is deployed. This PR records the RP-resolved image version on the NodeClaim; a follow-up will restore create-time pinning and validate drift semantics after that dependency is available.

MarcPow: Reversed — this is now implemented. Windows pins `NodeImageVersion` at create time exactly like Linux, so there is no longer a second pattern to maintain and the drift/status assumptions hold for Windows too. A Windows node now drifts on a new node image, which was impossible while the RP chose the image; there is acceptance coverage for both drift and no-drift.

One sequencing caveat worth flagging, because I verified the rollout rather than assuming it. I issued `PutMachine` with the exact string this PR now sends:

| Region | `AKSWindows-2022-containerd-gen2-<version>` |
|---|---|
| `westcentralus` | 200 OK, Machine created |
| `westus2` | 400 `invalid node image version` |

A malformed control (`AKSWindows-totally-bogus-xx`) was rejected in both, so the `westcentralus` success is the fixed parser accepting a valid name, not validation being skipped. The RP fix is merged (`abfc4ec2240`, Aug 20) but still rolling out. In a pre-fix region the pinned image 400s and there is no header fallback, so this should merge only once the rollout covers the regions where Windows NAP will run. Happy to run the live Windows E2E in `westcentralus`, which works today, or in the standalone E2E region.

pkg/providers/instance/aksmachineinstance.go
MarcPow: Reversed — the header is now gone entirely, along with its `CreateOptions` plumbing and batch-key participation; `pkg/providers/azclient/aksmachinesheaderbatch` is back to its pre-PR shape.

You were right that the generation is implied by the image. `setCustomizedImageForMachine` builds the SIG ID verbatim from the parsed `InputNodeImageVersion` and `continue`s past distro resolution, so `UseWindowsGen2VM` was never consulted on the Machine path at all — it only exists in `getDistroForWindows` on the AgentPool path. Since `windows-2022-containerd-gen2` and `windows-2022-containerd` are distinct definitions, pinning the image selects the generation, and Karpenter already filters images by `karpenter.azure.com/sku-hyperv-generation` so each SKU gets a bootable image. Removing the header also removes a failure mode, since that function only errored when the header was set on an incompatible SKU.

.github/workflows/e2e.yaml
```diff
@@ -217,9 +218,14 @@ jobs:
...
           k8s_version: ${{ inputs.k8s_version }}
-          provision_mode: ${{ inputs.provision_mode }}
+          # Windows is only provisionable via the AKS Machine API, so force that mode for the Windows
+          # suite regardless of the matrix's provision_mode (it would otherwise skip under aksscriptless).
+          provision_mode: ${{ inputs.suite == 'Windows' && 'aksmachineapi' || inputs.provision_mode }}
           identity_type: ${{ inputs.suite == 'Machine' && 'UserAssigned' || 'SystemAssigned' }}
           azure_vm_size: ${{ inputs.azure_vm_size }}
+          # Windows needs a dedicated (non-Cilium) cluster and a machines pool name <= 6 chars.
+          windows: ${{ inputs.suite == 'Windows' }}
+          aks_machines_pool_name: ${{ inputs.suite == 'Windows' && 'winmp' || 'mpool' }}
```
comtalyst, line 228: non-blocking: Can just go with `mpool` for both cases as they are both 5 chars. Although it is already defaulted in action.yaml, so this line may not be needed entirely.

.github/workflows/e2e.yaml
```diff
@@ -217,9 +218,14 @@ jobs:
           git_ref: ${{ inputs.git_ref }}
           location: ${{ inputs.location }}
           k8s_version: ${{ inputs.k8s_version }}
-          provision_mode: ${{ inputs.provision_mode }}
+          # Windows is only provisionable via the AKS Machine API, so force that mode for the Windows
+          # suite regardless of the matrix's provision_mode (it would otherwise skip under aksscriptless).
+          provision_mode: ${{ inputs.suite == 'Windows' && 'aksmachineapi' || inputs.provision_mode }}
```
comtalyst, line 223: minor: it would be more clear to explicitly skip it instead of hideously overriding the provided mode just on this test.

.github/workflows/README.md
```diff
@@ -7,6 +7,14 @@
 1. Create your new E2E testing suite `<test-suite-name>` within the `test/suites/` package. See: [test/README.md](../../test/README.md)
 2. Update the `workflows/e2e-matrix.yaml` workflow to include your E2E test case: `suite: [Utilization, GPU, ...]` - add in the name of your folder within the `test/suites/` package to the comma separated list. Casing does not matter.
 
+> **Note — suites that need a non-default cluster:** most suites run on the shared CI cluster
+> (`ci-mkcluster-all`, Azure CNI overlay + Cilium, in the matrix's `provision_mode`). The `Windows`
+> suite is special-cased in `workflows/e2e.yaml`: it always runs in `aksmachineapi` mode (Windows is
+> only provisionable via the AKS Machine API) on a dedicated cluster (`ci-mkcluster-all-windows`,
+> `az-mkaks-windows`) because Windows does not support the Cilium dataplane, and it uses a machines
+> pool name `<= 6` chars (`winmp`) to satisfy the Windows machine-name limit. Follow that pattern if a
```
comtalyst, line 15: (see https://github.com/Azure/karpenter-provider-azure/pull/1726/changes#r4042934296)

pkg/providers/instance/aksmachineinstancehelpers.go
```diff
@@ -301,32 +292,85 @@ func configureOSSKUAndFIPs(nodeClass *v1beta1.AKSNodeClass, orchestratorVersion
...
 		ossku = armcontainerservice.OSSKUUbuntu2404
 	case v1beta1.AzureLinuxImageFamily:
 		ossku = armcontainerservice.OSSKUAzureLinux
+	case v1beta1.Windows2022ImageFamily:
+		ossku = armcontainerservice.OSSKUWindows2022
+		enableFIPS = false
+	case v1beta1.Windows2025ImageFamily:
+		if !imagefamily.SupportsWindows2025(orchestratorVersion) {
+			return nil, nil, fmt.Errorf("Windows2025 requires Kubernetes 1.32.0 or newer, got %q", orchestratorVersion)
+		}
```
comtalyst, line 313: nit: this may not add much (and will add more point to maintain) given the filter already exist when retrieving node image, and the validation already exist within AKS Machine API? There will also be cleaner approaches in #1895.

pkg/providers/instance/aksmachineinstancehelpers.go
```diff
@@ -301,32 +292,85 @@ func configureOSSKUAndFIPs(nodeClass *v1beta1.AKSNodeClass, orchestratorVersion
...
+	switch family {
 	case v1beta1.Ubuntu2204ImageFamily:
 		ossku = armcontainerservice.OSSKUUbuntu2204
 	case v1beta1.Ubuntu2404ImageFamily:
 		ossku = armcontainerservice.OSSKUUbuntu2404
 	case v1beta1.AzureLinuxImageFamily:
 		ossku = armcontainerservice.OSSKUAzureLinux
+	case v1beta1.Windows2022ImageFamily:
+		ossku = armcontainerservice.OSSKUWindows2022
+		enableFIPS = false
```
comtalyst, line 309: nit: it would be cleaner to change from "force" to "default" here. I.e., set only when unspecified in NodeClass, as we want to avoid hideously overriding user inputs. It will also keep the logic of "which FIPS mode are supported in this image family" mainly on CRD layer, while leaving the logic of defaulting FIPS mode here (which could be improved further by moving it to a NodeClass helper to be reused elsewhere).
Although, that case of overriding user inputs is not yet possible given the current CRD validation, but good to establish a pattern nonetheless.

pkg/providers/instance/aksmachineinstancehelpers.go
```diff
@@ -301,32 +292,85 @@ func configureOSSKUAndFIPs(nodeClass *v1beta1.AKSNodeClass, orchestratorVersion
...
+	case v1beta1.Windows2025ImageFamily:
+		if !imagefamily.SupportsWindows2025(orchestratorVersion) {
+			return nil, nil, fmt.Errorf("Windows2025 requires Kubernetes 1.32.0 or newer, got %q", orchestratorVersion)
+		}
+		ossku = armcontainerservice.OSSKUWindows2025
+		enableFIPS = true
```
comtalyst, line 315: (same with https://github.com/Azure/karpenter-provider-azure/pull/1726/changes#r4049417308)

pkg/operator/options/suite_test.go
```diff
@@ -133,7 +133,7 @@ var _ = Describe("Options", func() {
 			os.Setenv("KUBELET_IDENTITY_CLIENT_ID", "12345678-1234-1234-1234-123456789012")
 			os.Setenv("LINUX_ADMIN_USERNAME", "customadminusername")
 			os.Setenv("ADDITIONAL_TAGS", "test-tag=test-value")
-			os.Setenv("AKS_MACHINES_POOL_NAME", "testmpool")
+			os.Setenv("AKS_MACHINES_POOL_NAME", "mpool")
```
comtalyst, line 136: non-blocking: we should update the option description of `AKS_MACHINES_POOL_NAME` to reflect the fact of length limitation if the pool will be used with Windows. Validation change is less of an option given linux doens't have the same restriction, and OS is NodePool-level

test/pkg/environment/azure/environment.go
```diff
@@ -253,6 +259,70 @@ func (env *Environment) AZLinuxNodeClass() *v1beta1.AKSNodeClass {
...
+	// Windows nodes are only provisionable via the AKS Machine API provision mode.
+	if !env.IsAKSMachineAPIMode() {
+		Skip("Windows node provisioning is only supported in AKS Machine API provision mode")
+	}
+	// Windows machine names are bounded by the Windows NetBIOS computer-name limit once the AKS RP
+	// composes the VM name from the pool and machine names. The reserved NAP pool ("aksmanagedap")
+	// uses a 12-character pool-hash plus NodeClaim-suffix name. Custom pools use only the
+	// five-character NodeClaim suffix and additionally require the pool name to be <= 6 chars.
+	if env.MachineAgentPoolName != "aksmanagedap" && len(env.MachineAgentPoolName) > 6 {
+		Skip(fmt.Sprintf("Windows machines require the reserved aksmanagedap pool or a custom machines pool name <= 6 chars; got %q (%d chars)",
```
comtalyst, line 298: non-blocking: I think we should allow it to fail in this case to not hide this requirement

MarcPow:
Ran the live Windows E2E in **westcentralus** — the region I confirmed has the RP parser fix — against this branch (`95af923e`). **Both families pass with the image pinned and no Gen2 header.**

| Suite | Result |
|---|---|
| Windows2022 | **PASS** — 1/1 specs, 368.81s |
| Windows2025 | **PASS** — 1/1 specs, 364.69s |

Self-hosted, `PROVISION_MODE=aksmachineapi`, `USE_SIG=true`, Azure CNI Overlay + windowsProfile, k8s 1.35.8, custom `winmp` machines pool.

### The pinned image reaches the RP and selects the generation

Parsed from the controller's `creating AKS machine` body:

```
Windows2022                                     Windows2025
  vmSize            Standard_D2als_v6             Standard_D2als_v6
  sku-hyperv-generation "2"  <-- GEN2-ONLY        "2"  <-- GEN2-ONLY
  osType/osSKU      Windows / Windows2022         Windows / Windows2025
  enableFIPS        false                         true
  linuxProfile      <nil>                         <nil>
  nodeImageVersion  AKSWindows-2022-containerd-gen2-20348.5499.260812
                                                  AKSWindows-2025-gen2-26100.33296.260812
```

`grep -ci usewindowsgen2vm` over both captured controller logs: **0 hits**.

Three things this establishes:

1. **Pinning works against a real RP.** `AKSWindows-2022-containerd-gen2-20348.5499.260812` is exactly the 5-part hyphenated shape that returns `400 invalid node image version` in a pre-fix region. Here the RP accepted it and built the node.

2. **Generation follows the image, not a header** — which is the point @comtalyst made. Karpenter freely selected a **Gen2-only** SKU (`Standard_D2als_v6`, `sku-hyperv-generation=2`) and paired it with the `-gen2` image definition, and the node booted. For contrast, before the Gen2 header existed an unconstrained Karpenter picked Gen2-only `Standard_D2als_v7` and the RP failed the LRO with `The selected VM size 'Standard_D2als_v7' cannot boot Hypervisor Generation '1'`. Pinning removes that whole class of failure with no header at all.

3. The SKU was genuinely unconstrained — the suite pins only the regional (zoneless) offering, never a VM size — so this was a free selection rather than a rigged size.

The suite's own assertions also passed for both: `kubernetes.io/os=windows`, the matching `kubernetes.azure.com/os-sku`, the NodePool label, and `kubernetes.azure.com/fips_enabled=true` on Windows2025.

### Merge sequencing unchanged

This validates the design, not the rollout. westus2 was still on the pre-fix parser as of 2026-09-16, so the gate in the PR description still stands: merge once the RP fix has reached the regions where Windows NAP will run. Happy to re-run this in any region you want checked, or in the standalone AKS E2E region.

Throwaway cluster and resource group deleted after the run.

.github/workflows/e2e.yaml
comtalyst, line 188: I ran self-hosted E2Es via GitHub Actions and this new suite failed because "windows" is apparently a reserved keyword for an ACR name... 

I don't think the suite deserves a rename and the test infra can be improved. For now, something like this mitigates it:

```suggestion
          ACR_PREFIX=$(echo "$SUITE" | awk '{print tolower($0)}')
          if [[ "$ACR_PREFIX" == *windows* ]]; then
            ACR_PREFIX="${ACR_PREFIX//windows/win}"
          fi
          ACR_NAME="${ACR_PREFIX}acr${{ github.run_id }}"
```

.github/workflows/e2e.yaml
```diff
@@ -217,9 +218,14 @@ jobs:
...
           identity_type: ${{ inputs.suite == 'Machine' && 'UserAssigned' || 'SystemAssigned' }}
           azure_vm_size: ${{ inputs.azure_vm_size }}
+          # Windows needs a dedicated (non-Cilium) cluster and a machines pool name <= 6 chars.
+          windows: ${{ inputs.suite == 'Windows' }}
```
comtalyst, line 227: Given we also add Windows test in other suites:

```suggestion
          # Windows contexts in Drift and Consolidation need the dedicated non-Cilium cluster.
          windows: ${{ inputs.suite == 'Windows' || inputs.suite == 'Drift' || inputs.suite == 'Consolidation' }}
```

Currently, the test apparently passes on our subs given there is an experimental flag that allows Windows + Cilium. But it wouldn't work/be supported anywhere else.

comtalyst, line 227: non-blocking: the phrasing here can imply "windows" to be something cluster-wide OS preferences, while it is in fact the existence of `windowsProfile` and the cluster using compatible network configurations. Using more explicit definitions like `networkPlugin` + `needWindowsProfile` is more clear. At least renaming it to `windowsReady` is already an improvement.

pkg/apis/v1beta1/aksnodeclass.go
```diff
@@ -103,7 +109,7 @@ type AKSNodeClassSpec struct {
 	ImageID *string `json:"-"`
 	// imageFamily is the image family that instances use.
 	// +default="Ubuntu"
-	// +kubebuilder:validation:Enum:={Ubuntu,Ubuntu2204,Ubuntu2404,AzureLinux}
+	// +kubebuilder:validation:Enum:={Ubuntu,Ubuntu2204,Ubuntu2404,AzureLinux,Windows2022,Windows2025}
 	// +optional
 	ImageFamily *string `json:"imageFamily,omitempty"`
```
comtalyst, line 114: non-blocking: given Windows + Cilium is not supported, it is better UX-wise to fail NodeClass validation (via status) like https://github.com/Azure/karpenter-provider-azure/pull/1926/changes.

comtalyst, line 114: non-blocking: there is a risk of Machine API rollout racing with this rollout, which may result in provisioning failures if Machine API rollout isn't done. For that, it is better UX-wise to fail NodeClass validation (via status) like https://github.com/Azure/karpenter-provider-azure/pull/1927/changes.

comtalyst (approved):
There are some non-blocking issues, can be follow-ups.

PR #1728, opened 2026-06-12
fix!: replace CreateInstanceFailed Launched reason with classified reasons (check dependent automations)

### LINKED ISSUE #1727: Surface specific failure reason on the NodeClaim Launched condition (capacity vs quota vs allocation)
### Summary

When instance creation fails, we already know *why* — capacity, quota, allocation, or config — but we throw that away before it reaches the user. Every failure shows up on the `Launched` condition with the same generic `reason: CreateInstanceFailed`.

The condition *message* spells out the cause in plain English, so a human reading one NodeClaim can usually figure it out. But the `reason` field — the part that's *meant* to be machine-readable — is identical every time. That leaves dashboards, alerts, and automation with nothing to branch on. Parsing the message instead is a trap: it's free-form prose that can get reworded between releases.

This is also a **parity** gap. `karpenter-provider-aws` already does this — it maps each launch error to a specific reason (`VCPULimitExceeded`, `Unauthorized`, `AMINotFound`, …) via [`errors.ToReasonMessage`](https://github.com/aws/karpenter-provider-aws/blob/fe623d537f85b0e6a69fc3eebc3692f80efda891/pkg/errors/errors.go?plain=1#L221).

### Current behavior

The error handlers in `pkg/providers/instance/offerings/commonerrorhandlers.go` already compute a precise reason — they use it to set the unavailable-offerings cache TTLs:

```go
SubscriptionQuotaReachedReason              = "SubscriptionQuotaReached"
AllocationFailureReason                     = "AllocationFailure"
ZonalAllocationFailureReason                = "ZonalAllocationFailure"
OverconstrainedZonalAllocationFailureReason = "OverconstrainedZonalAllocationFailure"
OverconstrainedAllocationFailureReason      = "OverconstrainedAllocationFailure"
```

But each handler returns only a friendly string — no `%w`, no code, no reason carried out:

```go
return fmt.Errorf("the requested SKU is unavailable for instance type %s in zone %s with capacity type %s, ...", ...)
```

That error flows up to `pkg/cloudprovider/cloudprovider.go`, where **every** create failure gets wrapped with the same hardcoded reason (`:187` VM, `:221` AKS machine, `:269` standalone):

```go
return nil, cloudprovider.NewCreateError(
    fmt.Errorf("creating instance failed, %w", err),
    CreateInstanceFailedReason,                 // ← generic, identical for every cause
    truncateMessage(err.Error()),               // ← code-stripped prose
)
```

So the condition always looks like this, no matter the cause:

```yaml
- type: Launched
  status: "Unknown"
  reason: CreateInstanceFailed     # capacity, quota, allocation — all identical
  message: "the requested SKU is unavailable for instance type Standard_D4s_v5 in zone 2 with capacity type spot, ..."
```

### Proposed behavior

Thread the reason the handler already computed out of `ResponseErrorHandler.Handle` / `ErrorDetailHandler.Handle` and pass it to `NewCreateError(...)` instead of the hardcoded `CreateInstanceFailedReason`:

```yaml
- type: Launched
  status: "Unknown"
  reason: SKUNotAvailable          # ← actionable, stable, machine-readable
  message: "the requested SKU is unavailable for instance type Standard_D4s_v5 in zone 2 with capacity type spot, ..."
```

One reason per handler:

| Handler | Reason on `Launched` |
|---|---|
| `handleSKUNotAvailableError` | `SKUNotAvailable` |
| `handleZonalAllocationFailureError` | `ZonalAllocationFailure` |
| `handleAllocationFailureError` | `AllocationFailure` |
| `handleOverconstrainedZonalAllocationFailureError` | `OverconstrainedZonalAllocationFailure` |
| `handleOverconstrainedAllocationFailureError` | `OverconstrainedAllocationFailure` |
| `handleSKUFamilyQuotaError` | `SubscriptionQuotaReached` |
| `handleLowPriorityQuotaError` | `SubscriptionQuotaReached` (or a new `LowPriorityQuotaReached`) |

ravishen:
Fixes #1727

### What this does

When instance creation fails, the offerings error handlers already know precisely *why* — capacity, quota, or allocation — and each maps the failure to a specific reason constant (`SKUNotAvailable`, `ZonalAllocationFailure`, …). But before this change that reason was only used as a **log label** when marking the offering unavailable in the cache; it was dropped before reaching the user. Create failures surfaced on the `Launched` condition with the same generic `reason: CreateInstanceFailed`, with the actual cause only in the free-form `message` — leaving dashboards/alerts/automation with nothing stable and machine-readable to branch on.

This threads the reason the handler already knows out to the `Launched` condition, matching the parity behavior `karpenter-provider-aws` 

### Reason mapping

| Handler | `Launched` reason |
|---|---|
| `handleSKUNotAvailableError` / `handleSKUNotAvailableForSubscriptionError` | `SKUNotAvailable` |
| `handleZonalAllocationFailureError` | `ZonalAllocationFailure` |
| `handleAllocationFailureError` | `AllocationFailure` |
| `handleOverconstrainedZonalAllocationFailureError` | `OverconstrainedZonalAllocationFailure` |
| `handleOverconstrainedAllocationFailureError` | `OverconstrainedAllocationFailure` |
| `handleSKUFamilyQuotaError` | `SubscriptionQuotaReached` |
| `handleLowPriorityQuotaError` | `SubscriptionQuotaReached` |
| `handleRegionalQuotaError` | *(unchanged — still `InsufficientCapacityError`)* |

### Before / after

```yaml
# before
- type: Launched
  status: "Unknown"
  reason: CreateInstanceFailed     # identical for capacity, quota, allocation
  message: "the requested SKU is unavailable for instance type Standard_D4s_v5 ..."

# after
- type: Launched
  status: "Unknown"
  reason: SKUNotAvailable          # actionable, stable, machine-readable
  message: "the requested SKU is unavailable for instance type Standard_D4s_v5 ..."
```
- [ ] Yes, issue opened: # 
- [x] No


ravishen:

pkg/cloudprovider/cloudprovider.go
```diff
@@ -706,6 +706,21 @@ func truncateMessage(msg string) string {
...
 }
 
+// toCreateError wraps an instance-create failure as a *cloudprovider.CreateError with the
+// wrapMsg context prefix. When err carries a classified CreateError (set by the offerings
+// error handlers, e.g. SKUNotAvailable or ZonalAllocationFailure), its specific Launched
+// condition reason and message are surfaced so consumers can branch on the cause; otherwise
+// it falls back to the generic CreateInstanceFailed reason and error text, preserving prior
+// behavior.
+func toCreateError(err error, wrapMsg string) error {
+	reason, message := CreateInstanceFailedReason, err.Error()
```
theunrepentantgeek: In this project, we don't generally use compound assignments like this unless capturing multiple return values from a method/function call. #minor #readability

```suggestion
	reason := CreateInstanceFailedReason
	, message := err.Error()
```

pkg/cloudprovider/cloudprovider.go
```diff
@@ -706,6 +706,21 @@ func truncateMessage(msg string) string {
...
+	var classified *cloudprovider.CreateError
+	if stderrors.As(err, &classified) {
```
theunrepentantgeek: The new `errors.AsType` method in the stdlib can make this cleaner: #minor #moderncode

```suggestion
	if cerr, ok := stderrors.AsType[*cloudprovider.CreateError](err, &classified); ok {
```

pkg/providers/instance/offerings/resperrorhandlers_test.go
```diff
@@ -96,10 +102,32 @@ type responseErrorTestCase struct {
...
 	expectedUnavailableOfferingsInformation []offeringToCheck
 	expectedAvailableOfferingsInformation   []offeringToCheck
 }
 
+// assertHandledError verifies the error returned by a handler dispatch. When a
+// reason is expected, the error must be a CreateError carrying that machine-readable
+// reason on the Launched condition, with the friendly message preserved. When no
+// reason is expected (e.g. InsufficientCapacityError passthrough), the error must
+// not carry a Launched reason and its message must match.
+func assertHandledError(g Gomega, actual, expected error, expectedReason string) {
```
theunrepentantgeek: Prefer explicit argument types #minor #readability


```suggestion
func assertHandledError(g Gomega, actual error, expected error, expectedReason string) {
```

theunrepentantgeek (commented):
Looks good, a few minor quibbles but nothing serious in the code.

Does this qualify as a breaking change? We're changing the `Reason` published for the error, I wonder if it's likely people have built automations/monitoring based on the existing values, and whether this change will need to be called out in the release notes.

pkg/cloudprovider/cloudprovider.go
ravishen: Done, thanks!

pkg/cloudprovider/cloudprovider.go
ravishen: updated.

pkg/providers/instance/offerings/resperrorhandlers_test.go
ravishen: done.

ravishen:
> Looks good, a few minor quibbles but nothing serious in the code.
> 
> Does this qualify as a breaking change? We're changing the `Reason` published for the error, I wonder if it's likely people have built automations/monitoring based on the existing values, and whether this change will need to be called out in the release notes.

I'm not sure if its breaking in strict sense since I don't know if these reasons were ever documented as a stable contract; But agreed that anyone exact-matching `reason: CreateInstanceFailed` would silently stop matching for the classified cases, so a release-notes callout is cheap insurance. WDYT ?

theunrepentantgeek:
> I'm not sure if its breaking in strict sense since I don't know if these reasons were ever documented as a stable contract; But agreed that anyone exact-matching `reason: CreateInstanceFailed` would silently stop matching for the classified cases, so a release-notes callout is cheap insurance. WDYT ?

Including this in the release notes is a good idea - at the very least it means someone can discover why things broke (if they did).

pkg/providers/instance/offerings/commonerrorhandlers.go
```diff
@@ -180,7 +181,8 @@ func handleSKUFamilyQuotaError(
 			unavailableOfferings.MarkUnavailableWithTTL(ctx, SubscriptionQuotaReachedReason, sku, getOfferingZone(offering), capacityType, LowQuotaTTL)
 		}
 	}
-	return fmt.Errorf("subscription level %s vCPU quota for %s has been reached (may try provision an alternative instance type)", capacityType, instanceType.Name)
+	err := fmt.Errorf("subscription level %s vCPU quota for %s has been reached (may try provision an alternative instance type)", capacityType, instanceType.Name)
+	return corecloudprovider.NewCreateError(err, SubscriptionQuotaReachedReason, err.Error())
```
matthchr, line 185: Can you move these reason const definitions from this package to `pkg/consts/condition_reasons.go` (or similarly named), along with the ones at the top of `pkg/cloudprovider/cloudprovider.go`, so we have all the condition reasons together?

If you'd rather we do that as a follow-up PR, I'm fine with that too, to not bloat this PR.

pkg/providers/instance/offerings/resperrorhandlers_test.go
```diff
@@ -96,10 +102,32 @@ type responseErrorTestCase struct {
...
 	expectedUnavailableOfferingsInformation []offeringToCheck
 	expectedAvailableOfferingsInformation   []offeringToCheck
 }
 
+// assertHandledError verifies the error returned by a handler dispatch. When a
+// reason is expected, the error must be a CreateError carrying that machine-readable
+// reason on the Launched condition, with the friendly message preserved. When no
+// reason is expected (e.g. InsufficientCapacityError passthrough), the error must
+// not carry a Launched reason and its message must match.
+func assertHandledError(g Gomega, actual error, expected error, expectedReason string) {
```
matthchr: minor: Move to `pkg/test/expectations.go` and export it as `ExpectHandledError`?

matthchr: Also mark as `GingkoHelper()`?

matthchr (dismissed):
Generally LGTM, had a few small comments. PTAL and let me know how you want to address them.

pkg/providers/instance/offerings/commonerrorhandlers.go
ravishen: I'll do that as a follow-up PR to keep this one focused.

pkg/providers/instance/offerings/resperrorhandlers_test.go
ravishen: > minor: Move to `pkg/test/expectations.go` and export it as `ExpectHandledError`?

Moving it to pkg/test/expectations creates an import cycle
`go list -deps ./pkg/test/expectations | grep -x \
          github.com/Azure/karpenter-provider-azure/pkg/providers/instance/offerings =>
github.com/Azure/karpenter-provider-azure/pkg/providers/instance/offerings` .

> Also mark as `GingkoHelper()`?

These are stdlib table tests (t.Run + NewWithT), not Ginkgo specs, so GinkgoHelper() wouldn't take effect here; applied the stdlib equivalent instead; the helper now takes *testing.T and calls t.Helper(), matching assertOfferingsState.

theunrepentantgeek (approved):

theunrepentantgeek:
Looks good, all workflow checks pass. We should be able to merge this as soon as it's up to date with the base branch, `main`.

theunrepentantgeek:
Merged! Thanks for the contribution.

PR #1736, opened 2026-06-19
feat: use skus api not price to determine vm availability

matthchr:
Previously we were intersecting price and skus to determine availability. This resulted in situations where for one reason or another the pricing API didn't have data, the VM was actually allocatable, but Karpenter couldn't allocate it.

Moving to rely solely on the SKUs API frees us from having to maintain a skus_missing_price list, and ensures that things work better in non-public clouds (where the pricing API is not available).

* Unit tests
* E2E tests: https://github.com/Azure/karpenter-provider-azure/actions/runs/27843517078

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


pkg/providers/instancetype/instancetypes.go
```diff
@@ -228,7 +228,6 @@ func (p *DefaultProvider) instanceTypeZones(sku *skewer.SKU) sets.Set[string] {
 	return sets.New(zones.Regional)
 }
 
-// TODO: review; switch to controller-driven updates
```
tallaxes, line 231: Just cleaning up the comment, or did you take a look and decide we won't need to change this?

tallaxes (dismissed):
LGTM. The change in logic is already well-covered by existing tests?

pkg/providers/instancetype/instancetypes.go
matthchr: I think we are controller-driven now aren't we? I changed it here: https://github.com/Azure/karpenter-provider-azure/pull/1536

Or is this talking about a different controller?

matthchr: Oooh AWS has a controller that actually writes the offerings? Maybe I should leave this comment.

matthchr: I'll revive this comment

matthchr:
> LGTM. The change in logic is already well-covered by existing tests?

I think we have good regression coverage but this is a good point, I should probably add a test that ensures that if something is missing in pricing but not in SKUs it now works (it didn't before). I don't think that's an existing test we had since it didn't work that way before.

pkg/providers/instancetype/instancetypes.go
tallaxes: I actually forgot we already did instance types controller, but yes would be good to take a look at one for offerings (and see if we need it)

tallaxes (dismissed):

tallaxes (approved):

PR #1419, opened 2026-02-09
feat: Add VMSizeProperties support to disable SMT/hyperthreading

fabioaraujopt:
Fixes #1541

**Description**

This PR adds support for Azure VM vCore Customization in `AKSNodeClass` by introducing `vmSizeProperties.vCPUsPerCore`, so users can configure SMT/hyperthreading behavior at VM creation time.

Changes included:
- Add `VMSizeProperties` to `AKSNodeClassSpec` with `vCPUsPerCore` (validated range: 1-2)
- Apply `vmSizeProperties` to VM `hardwareProfile` during VM object creation
- Add/extend unit tests for API validation and VM conversion logic
- Regenerate CRDs to surface the new field

Motivation/use cases:
- Close the SMT control gap between AWS and Azure providers
- Improve predictability for CAE/perf-sensitive workloads
- Enable security hardening scenarios that require SMT disabled

Example:
```yaml
apiVersion: karpenter.azure.com/v1beta1
kind: AKSNodeClass
metadata:
  name: smt-disabled
spec:
  imageFamily: AzureLinux
  vmSizeProperties:
    vCPUsPerCore: 1
```

**How was this change tested?**

- Added unit tests for `vCPUsPerCore` validation boundaries
- Added unit tests for VM conversion / hardware profile mapping
- Regenerated CRDs and validated field rendering
- Confirmed existing relevant tests pass locally

**Does this change impact docs?**
- [ ] Yes, PR includes docs updates
- [x] Yes, issue opened: #1541
- [ ] No

**Release Note**

```release-note
feat: add `vmSizeProperties.vCPUsPerCore` to AKSNodeClass to configure VM SMT/hyperthreading at creation time.
```

PR #1746, opened 2026-06-22
feat: Add VMSizeProperties support to disable SMT/hyperthreading (rebase of #1419)

### LINKED ISSUE #1541: Feature: Add SMT/hyperthreading control for AKSNodeClass via VMSizeProperties
## Summary

Azure Karpenter currently does not expose a first-class API to control SMT/hyperthreading behavior for nodes provisioned from `AKSNodeClass`.

AWS users can influence SMT behavior in their provider workflows, but Azure `AKSNodeClass` has no equivalent provider-native setting, creating a cross-provider capability gap.

## Problem

For workloads where SMT policy matters, the current Azure provider behavior is too rigid:
- No explicit way to request 1 vCPU per physical core (SMT disabled)
- No provider-level API field for users to express this intent in `AKSNodeClass`
- Teams must accept defaults or rely on non-ideal workarounds

## Impact

Lack of SMT controls affects users that need:
- **Performance consistency** for CAE/HPC-like workloads
- **Security hardening** scenarios where disabling SMT is part of mitigation policy
- **Provider parity** between AWS and Azure Karpenter experiences

## Proposed Solution

Add support for Azure VM vCore customization in `AKSNodeClass`:
- Add `vmSizeProperties` to `AKSNodeClassSpec`
- Support `vmSizeProperties.vCPUsPerCore` with validation range `1-2`
- Apply the value to VM `hardwareProfile` at VM creation time
- Regenerate CRDs and include tests for validation + conversion/mapping logic

## Constraints / Notes

- Azure requires support at VM size/image/API level
- Configuration is applied at VM creation time (not mutable on running VMs)
- Expected values:
  - `1` = SMT disabled
  - `2` = SMT enabled/default

## References

- PR implementing this feature: https://github.com/Azure/karpenter-provider-azure/pull/1419
- Azure docs (vCore customization): https://learn.microsoft.com/en-us/azure/virtual-machines/vm-customization
- Azure announcement (disable multithreading): https://techcommunity.microsoft.com/blog/azurecompute/announcing-preview-of-vcore-customization-disable-multithreading--configurable-c/4462417

## Outcome

Users can declaratively control SMT policy in Azure through `AKSNodeClass`, improving performance/security posture and reducing provider capability gaps.

fabioaraujopt:
Fixes #1541
Supersedes #1419

## Summary
- Rebased the SMT/hyperthreading control change onto latest `main` and aligned it with current provider structure.
- Add `vmSizeProperties.vCPUsPerCore` to both `v1beta1` and `v1alpha2` `AKSNodeClassSpec`, with validation and regenerated CRD/deepcopy artifacts.
- Apply `VMSizeProperties` only in VM API (`aksscriptless`) path; explicitly ignore it for AKS Machine API modes (`aksmachineapi`, `aksmachineapiheaderbatch`) with test coverage.

## Test plan
- [x] `go test ./pkg/providers/instance -run TestConvertVMSizeProperties -count=1`
- [x] `go test ./pkg/apis/v1beta1 -run TestAPIs -count=1 -args -ginkgo.focus=VMSizeProperties` (with envtest assets)
- [x] `go test ./pkg/apis/v1alpha2 -run TestAPIs -count=1 -args -ginkgo.focus=VMSizeProperties` (with envtest assets)
- [x] `go test ./pkg/apis/v1alpha2 ./pkg/apis/v1beta1 -run '^$' -count=1`

## Notes
- This preserves the original intent of #1419 while ensuring compatibility with the latest codebase and provision-mode behavior.


fabioaraujopt:
@tallaxes @matthchr @comtalyst @MaximilianoUribe could we get a scope decision on this PR?

Happy to make follow-up changes quickly, but I need clarity on one of these paths:
1) merge path (with requested changes), or
2) not roadmap-fit for now.

Quick notes to reduce review load:
- Default behavior is unchanged unless `vmSizeProperties.vCPUsPerCore` is set.
- The field is applied only in VM API mode (`aksscriptless`) and ignored in AKS Machine API modes (`aksmachineapi`, `aksmachineapiheaderbatch`) with tests.
- API validation + conversion tests are included.

If useful, I can split this into a smaller PR.

comtalyst:
@fabioaraujopt Karpenter AKS provider need to align with AKS on provisioning configurations. 
Using Machine API guarantees that, and this VM API path addition may move away from that AKS alignment, unless Machine API (and AKS entirely) does it behind the scenes, in which no change is needed to get the same outcome. 
Machine API mode will going to be the default mode moving forward, so I would say it is not a roadmap fit for now.

Could you elaborate more on what is the motivation for this change?

fabioaraujopt:
> Fábio Araújo (@fabioaraujopt) Karpenter AKS provider need to align with AKS on provisioning configurations. Using Machine API guarantees that, and this VM API path addition may move away from that AKS alignment, unless Machine API (and AKS entirely) does it behind the scenes, in which no change is needed to get the same outcome. Machine API mode will going to be the default mode moving forward, so I would say it is not a roadmap fit for now.
> 
> Could you elaborate more on what is the motivation for this change?

The motivation is our ML workloads, which require one thread per physical core for predictable performance. We currently run self-hosted Karpenter using aksscriptless, where the Azure Compute API supports hardwareProfile.vmSizeProperties.vCPUsPerCore.

I checked the AKS Machine API surface, and MachineHardwareProfile does not currently expose an equivalent setting. Therefore AKS cannot provide this configuration through Machine API today.

I agree this must not silently behave differently across provisioning modes. I can scope the feature explicitly to aksscriptless and make AKSNodeClass validation fail when it is configured with either Machine API mode.

If the project only accepts features available through Machine API, we can maintain this as a downstream extension. Is there an existing Machine API mechanism or roadmap item for configuring vCPUsPerCore that I have missed?

PR #1749, opened 2026-06-22
fix: distinguish env names of provider batch vs core batch

comtalyst:
**Description**

Rename Machine API instance-create batch settings to the `PROVIDER_BATCH_*` option family. The original names were confusable with [core batching parameters](https://github.com/kubernetes-sigs/karpenter/blob/61dff3a2603b8d1bd0715572089919b87b60d9ef/pkg/operator/options/options.go#L129-L130) of similar name. Also changed the format/value type to align with the core.

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note
```


pkg/operator/options/options.go
```diff
@@ -128,9 +129,9 @@ func (o *Options) AddFlags(fs *coreoptions.FlagSet) {
...
-	fs.IntVar(&o.BatchMaxTimeoutMS, "batch-max-timeout-ms", env.WithDefaultInt("BATCH_MAX_TIMEOUT_MS", 5000), "Maximum timeout in milliseconds for batch accumulation. Only used on provision mode aksmachineapiheaderbatch.")
-	fs.IntVar(&o.MaxBatchSize, "max-batch-size", env.WithDefaultInt("MAX_BATCH_SIZE", consts.AKSMachineAPIHeaderBatchMaxSize), fmt.Sprintf("Maximum number of machines per batch (AKS API limit is %d). Only used on provision mode aksmachineapiheaderbatch.", consts.AKSMachineAPIHeaderBatchMaxSize))
+	fs.DurationVar(&o.ProviderBatchIdleDuration, "provider-batch-idle-duration", env.WithDefaultDuration("PROVIDER_BATCH_IDLE_DURATION", time.Second), "Idle duration for provider batch accumulation. Only used on provision mode aksmachineapiheaderbatch.")
+	fs.DurationVar(&o.ProviderBatchMaxDuration, "provider-batch-max-duration", env.WithDefaultDuration("PROVIDER_BATCH_MAX_DURATION", 5*time.Second), "Maximum duration for provider batch accumulation. Only used on provision mode aksmachineapiheaderbatch.")
```
matthchr: Mention the "use go duration format such as `1s` here?"

pkg/operator/options/options.go
```diff
@@ -128,9 +129,9 @@ func (o *Options) AddFlags(fs *coreoptions.FlagSet) {
...
+	fs.IntVar(&o.ProviderBatchMaxSize, "provider-batch-max-size", env.WithDefaultInt("PROVIDER_BATCH_MAX_SIZE", consts.AKSMachineAPIHeaderBatchMaxSize), fmt.Sprintf("Maximum number of machines per provider batch (AKS API limit is %d). Only used on provision mode aksmachineapiheaderbatch.", consts.AKSMachineAPIHeaderBatchMaxSize))
```
tallaxes, line 134: I assume these are not used yet, so safe to rename? 

The intent is that in theory these could be used for any provisioning mode, so not Machine API specific? (Hmm, maybe not quite generic, at least not the current validation, e.g. `consts.AKSMachineAPIHeaderBatchMaxSize`)

Are these fully independent of core pod BATCH_MAX/IDLE_DURATION? Otherwise might be good to have some guidance on how these can / should be related (and possibly even validation if invalid pairing exists)

tallaxes (dismissed):
LGTM

pkg/operator/options/suite_test.go
```diff
@@ -162,6 +165,31 @@ var _ = Describe("Options", func() {
 		})
 	})
 	Context("Validation", func() {
+		It("should validate provider batch duration env vars", func() {
+			os.Setenv("PROVIDER_BATCH_IDLE_DURATION", "6s")
```
matthchr: minor: instead of testing these like this, maybe put the env var testing in:
`should correctly fallback to env vars when CLI flags aren't set`

and then just do plain flags testing here? I think that's more similar to what other tests are doing and ensures you have coverage for both the opts pattern and the env variable pattern

matthchr (dismissed):
LGTM, few small comments

pkg/operator/options/options.go
comtalyst: > I assume these are not used yet, so safe to rename?

That's right. AKS have been using defaults. Self-hosted doesn't support this (and provision mode `aksmachineapiheaderbatch` in general) .

> The intent is that in theory these could be used for any provisioning mode, so not Machine API specific?

That's right, though I don't currently have a strong opinion. The validation is currently specific to Machine API, though I would imagined that to be conditioned on what mode it uses.

On the same topic, the "batch" prefix in these env var names is quite generic even in the core, while their descriptions state that they are naturally for pending pods detection. I chose to align ours with it.

> Are these fully independent of core pod BATCH_MAX/IDLE_DURATION? Otherwise might be good to have some guidance on how these can / should be related (and possibly even validation if invalid pairing exists)

I would call it independent, as they are on a different layer and are running in parallel. 
Core batches their considerations of pending pods to issue a provider `Create` (the larger the batch, the less times `Create` will be called). We group those `Create` requests into batches.
There could be a more advanced algorithm to unify/optimize them though(?).

matthchr (dismissed):

pkg/operator/options/options.go
```diff
@@ -94,10 +95,10 @@ type Options struct {
...
-	BatchIdleTimeoutMS  int    `json:"batchIdleTimeoutMS,omitempty"`  // Idle timeout in milliseconds for batch accumulation (default 1000ms). Only used on provision mode aksmachineapiheaderbatch.
-	BatchMaxTimeoutMS   int    `json:"batchMaxTimeoutMS,omitempty"`   // Maximum timeout in milliseconds for batch accumulation (default 5000ms). Only used on provision mode aksmachineapiheaderbatch.
-	MaxBatchSize        int    `json:"maxBatchSize,omitempty"`        // Maximum number of machines per batch (default 50, AKS API limit). Only used on provision mode aksmachineapiheaderbatch.
+	AKSMachinesPoolName       string        `json:"aksMachinesPoolName,omitempty"` // The name of the agent pool for the AKS machine API, assuming that all machines belong to the same agent pool. Only used on AKS machine API provision modes.
+	ProviderBatchIdleDuration time.Duration `json:"providerBatchIdleDuration,omitempty"`
+	ProviderBatchMaxDuration  time.Duration `json:"providerBatchMaxDuration,omitempty"`
+	ProviderBatchMaxSize      int           `json:"providerBatchMaxSize,omitempty"`
```
theunrepentantgeek: Should the comment information from above be retained?

The change to use `time.Duration` is a good one though.

pkg/operator/options/suite_test.go
```diff
@@ -157,11 +164,29 @@ var _ = Describe("Options", func() {
 				ClusterDNSServiceIP:            lo.ToPtr("10.244.0.1"),
 				ManageExistingAKSMachines:      lo.ToPtr(true),
 				AKSMachinesPoolName:            lo.ToPtr("testmpool"),
+				ProviderBatchIdleDuration:      lo.ToPtr(1500 * time.Millisecond),
+				ProviderBatchMaxDuration:       lo.ToPtr(6 * time.Second),
+				ProviderBatchMaxSize:           lo.ToPtr(42),
```
theunrepentantgeek, line 169: Does this compile? 

The declarations are embedded:

```  go
	ProviderBatchIdleDuration time.Duration `json:"providerBatchIdleDuration,omitempty"`
	ProviderBatchMaxDuration  time.Duration `json:"providerBatchMaxDuration,omitempty"`
	ProviderBatchMaxSize      int           `json:"providerBatchMaxSize,omitempty"`
```

So you shouldn't need `lo.ToPtr()` here.

Aside: From Go 1.26, you can write this using `new(expr)` syntax:

``` go
				ProviderBatchIdleDuration:      new(1500 * time.Millisecond),
				ProviderBatchMaxDuration:       new(6 * time.Second),
				ProviderBatchMaxSize:           new(42),
```

comtalyst: Not sure I follow your first suggestion. For `new(expr)`, if we want to change, let's do so separately, and wider to not unnecessarily having two patterns in one place.

pkg/operator/options/options.go
comtalyst: (slightly awkward that some comments exist in both CLI description and here, but could be a follow-up)

matthchr (approved):

PR #1763, opened 2026-07-03
feat: Add design for quota/fungibility reactivity improvements

matthchr:
- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


designs/0012-quota-fungibility-reactivity-improvements.md
```diff
@@ -0,0 +1,714 @@
...
+
+## Decisions
+
+### Decision 1: How can we improve our reactivity to errors from CRP?
+
+Right now we manage our available set of VMs using the [unavailableofferings cache](https://github.com/Azure/karpenter-provider-azure/blob/main/pkg/cache/unavailableofferings.go). This cache is reactive and populated when we get errors from CRP.
+
+Problems to consider here:
+
+1. Hitting a capacity error for D64 doesn't guarantee that D32 isn't available, but on the other hand it _strongly suggests_ it. Right now, we have two options we either exclude D32 or we don't. If we exclude D32 as well (basically marking that whole family as unavailable), that's great if they're actually out of capacity, but bad if they're just low on capacity but it's the only size in the allocation list.
```
tallaxes: Currently we would exclude D64 and anything _larger_, right?

designs/0012-quota-fungibility-reactivity-improvements.md
```diff
@@ -0,0 +1,714 @@
...
+
+Phase 1: Call the API (+ cache), but do not actually use its results to provision anything. Instead just log what we _would_ provision. We can use this to build confidence: Is the API picking the same thing we would have? We can build reporting/monitoring on this. Note that Karpenter will still have to filter the large set of possible VM sizes down to a smaller set before calling the API.
+
+Phase 2: Start using the API as the main source of size picking.
+
+### Decision 3: How can we improve our allocation success rate in the near term?
+
+#### Conclusion: New quota provider (short term), to be replaced by SKU Split provider in the medium term
+
+Create a `quota` provider located at `pkg/providers/quota/quota.go`. This provider should call the [Azure Quota SDK](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/quota/armquota/v2) periodically (every 10m by default).
```
tallaxes, line 146: Will quota be a hard filter (presumably) or just a ranking signal? May be worth stating explicitly ...

designs/0012-quota-fungibility-reactivity-improvements.md
```diff
@@ -0,0 +1,714 @@
...
+
+- [#1323](https://github.com/Azure/karpenter-provider-azure/issues/1323) — Improve handling on quota failures (or avoid them altogether)
+- [#1733](https://github.com/Azure/karpenter-provider-azure/issues/1733) — Prefer availability over lowest-price for on-demand instance selection
+- [#1727 (traceability)](https://github.com/Azure/karpenter-provider-azure/issues/1727) — Surface specific failure reason on the NodeClaim Launched condition (capacity vs quota vs allocation)
+- [#943 (traceability)](https://github.com/Azure/karpenter-provider-azure/issues/943) — AMD64 v6 VM's never provisioned
+
+**The current behavior is:**
+
+- Karpenter is entirely reactive to errors that it gets from the compute API (quota, capacity, etc), and only adjusts its provisioning based on errors it has seen (with some cache expiry so errors don't prevent it from ever trying a given size again).
+- Karpenter always tries the cheapest sizes first. If those sizes are under heavy regional createVMInstanceontention or the user has low quota for them, they are tried anyway unless they are excluded at the `NodePool` level. Only advanced users even know to do that.
```
tallaxes: "under heavy regional createVMInstanceontention" - "under heavy regional contention"?

designs/0012-quota-fungibility-reactivity-improvements.md
```diff
@@ -0,0 +1,714 @@
...
+- Karpenter allocates the same way all the time (price-aware only). There is no way for Azure to perform capacity shaping among "similar" sizes. Note that capacity shaping does not exist anywhere today, and when it exists will likely only be enabled with explicit user opt-in, but the desire to eventually support the capability is there.
+
+> [!NOTE]
+> Some APIs referenced by this document are Microsoft internal-only. We've kept what we can public, but there are some links in this document that will not work for the general public.
+
+### Current statistics
+
+See [methodology](https://github.com/Azure/karpenter-poc/issues/1529) for more details about how these numbers were obtained.
+
+- Global success rate, measured as "successes allocating a VM" / ("successes allocating a VM" + "failures allocating a VM") is **~27%** over the last 30d. Many of those errors are intermittent (throttling, network blips, etc). Excluding transient errors and errors unrelated to quota/capacity the success rate is **~58%**.
```
tallaxes: This is about (success of) individual allocation attempts, not (success of) eventual allocation, right?

designs/0012-quota-fungibility-reactivity-improvements.md
```diff
@@ -0,0 +1,714 @@
...
+- **Improve handling of quota issues** by choosing the size we attempt to allocate more intelligently.
+- **Improve handling of capacity issues** by choosing the size we attempt to allocate more intelligently.
+- **Integrate with Azure Capacity Recommendation Service (CRS)** to enable capacity shaping.
+- **Improve `unavailableofferings` cache** to react more intelligently to errors.
+- **Improve traceability** to help users answer the question why a particular size was chosen.
+
+### Non-Goals
+
+- Changing how we provision VMs (PUT virtualMachine or PUT machine) - we've got ongoing investment in the Machine path so doing something other than that doesn't meet our short-term goals.
+- Improving "eventual allocation" success rate. Karpenter already tries every possible size allowed by the users configuration until it hits one that works. We're aiming to improve reaction _speed_ and _responsiveness_, Karpenter already picks a working size (eventually) today.
```
tallaxes: So maybe this should also be in the Goals list?
- Improve "eventual allocation" latency (This + shaping + traceability are likely the high-level goals?)
- Reduce probability of eviction

It would also be great to have something on how we could measure the improvement (maybe just for selected high-value goals) - especially since measuring some of this is tricky

designs/0012-quota-fungibility-reactivity-improvements.md
```diff
@@ -0,0 +1,714 @@
...
+
+This allows us to de-prioritize VMs whose families have hit capacity errors without excluding them entirely until we know for sure they won't work.
+
+For item 2: We already filter at allocation time as well, during `resolveInstanceTypes()` we consider if a VM is unavailable and exclude from attempting it when provisioning the node.
+
+### Decision 2: Which API should we use to get capacity recommendations from Azure?
+
+See [API comparison](#api-comparison) for a detailed breakdown of APIs considered and what they do for us.
+
+At this stage, the ideal API would:
```
tallaxes, line 89: Do we need to say anything about the potential need to expose allocation strategy choice to customization? (This is a more general question, asking here just in case it affects API choice)

designs/0012-quota-fungibility-reactivity-improvements.md
```diff
@@ -0,0 +1,714 @@
...
+| **5. One-stop-shop (quota+capacity+CRS)?** | ❌ Quota only |
+| **6. Public API?**                         | ✅            |
+| **7. Cacheable / high throttle limit?**    | ✅            |
+| **8. Supports 1-N VMs + partial success?** | N/A           |
+
+#### API reference
+
+**URL (usages):** `GET /subscriptions/{subscriptionId}/providers/Microsoft.Compute/locations/{location}/usages?api-version=2025-09-01`
+
+**URL (quota):** `GET /subscriptions/{subscriptionId}/providers/Microsoft.Compute/locations/{location}/quotas?api-version=2025-09-01`
```
tallaxes: Is it `Microsoft.Compute` or `Microsoft.Quota`?

designs/0012-quota-fungibility-reactivity-improvements.md
```diff
@@ -0,0 +1,714 @@
...
+Right now we manage our available set of VMs using the [unavailableofferings cache](https://github.com/Azure/karpenter-provider-azure/blob/main/pkg/cache/unavailableofferings.go). This cache is reactive and populated when we get errors from CRP.
+
+Problems to consider here:
+
+1. Hitting a capacity error for D64 doesn't guarantee that D32 isn't available, but on the other hand it _strongly suggests_ it. Right now, we have two options we either exclude D32 or we don't. If we exclude D32 as well (basically marking that whole family as unavailable), that's great if they're actually out of capacity, but bad if they're just low on capacity but it's the only size in the allocation list.
+2. Size filtering only at NodeClaim creation time isn't enough, because if multiple NodeClaims are created at the same time for the same SKUs, one hitting capacity issues means the others will too.
+
+#### Conclusion: Integrate unavailableofferings into allocation strategy alongside price
+
+This allows us to de-prioritize VMs whose families have hit capacity errors without excluding them entirely until we know for sure they won't work.
```
tallaxes, line 78: This would require shifting from binary available/unavailable to some kind of penalty/confidence score?

tallaxes (dismissed):

designs/0012-quota-fungibility-reactivity-improvements.md
matthchr: Correct. Updated the doc to clarify what we do today a bit better.

designs/0012-quota-fungibility-reactivity-improvements.md
matthchr: Great callout. I think that there are pros/cons to both. I've added a little section discussing this and concluded hard filter (which is what we were doing with unavailableofferings as well, just after-the-fact)

designs/0012-quota-fungibility-reactivity-improvements.md
matthchr: Fixed

designs/0012-quota-fungibility-reactivity-improvements.md
matthchr: Yes. Clarified in the doc and in the linked methodology issue.

Said:
> Note that these metrics are primarily focused on individual allocation attempts and NOT overall NodePool/NodeClaim allocation success. It's possible that an individual attempt for a given size fails, while the NodeClaim as a whole eventually allocates successfully. That initial failure results in increased latency for the user as well as increased Azure API throttling load, which is why we want to drive failed attempts down.

designs/0012-quota-fungibility-reactivity-improvements.md
matthchr: Refactored the goals list a bit to be clearer and call out the 3 top level goals you mentioned, as they are my top 3 goals and some of the others are really sub-goals in service of "Improve "eventual allocation" latency ".

My current thinking is that we will use the queries I shared in [methodology](https://github.com/Azure/karpenter-poc/issues/1529) (ln 31) to measure success of the "Improve reaction speed and responsiveness". I added a note to that effect above. I also looked to see if we had any better metrics exported and we may, so am digging into that a bit more as well -- some of the prom metrics from core seem like they might be useful.

designs/0012-quota-fungibility-reactivity-improvements.md
matthchr: I don't _think_ it impacts API choice that much.

It might be nice if the API exposed an allocation strategy for us to use, which we would then just proxy. On the other hand, if it _didn't_ we could always name our own and expose them. Even when we do just proxy the allocation strategy of an API, we can't really purely defer to it because for example user asking us for `LowestPrice` + NodeOverlay's might actually mean we ask the API to use `Prioritized`, because they are not aware of our price overlays.

So, it's an interesting consideration but not one that I think really actually impacts API choice all that much. WDYT?

designs/0012-quota-fungibility-reactivity-improvements.md
matthchr: Doh, it is `Microsoft.Quota`. Fixed

designs/0012-quota-fungibility-reactivity-improvements.md
matthchr: Yes, and after re-reading and re-thinking this, maybe we defer this until after we just take quota into account at all via quota provider. This might be an optimization we don't need.

matthchr: I've updated the doc to mention we defer this for now in favor of quota integration.

designs/0012-quota-fungibility-reactivity-improvements.md
tallaxes: Ack. Just want to make sure we consider the ask.

tallaxes (approved):

designs/0012-quota-fungibility-reactivity-improvements.md
matthchr: Yeah ok some of these core metrics are useful. Check the linked issue for them. I will see if I can bake them into a dashboard somewhere.

PR #1771, opened 2026-07-07
Raise LocalDNS Preferred Kubernetes threshold

saewoni:
Updates the LocalDNS Preferred auto-enable Kubernetes version threshold to 1.99.0 and adjusts the LocalDNS threshold test fixtures accordingly.


tallaxes (approved):
LGTM. The caveat is that this won't disable LocalDNS for NodeClasses already resolved as enabled - worth calling out (and checking for in existing deployments).

PR #1777, opened 2026-07-08
feat: add trusted launch support

### LINKED ISSUE #1634: Trusted Launch support in Karpenter/NAP
Description: [Trusted Launch](https://learn.microsoft.com/azure/aks/use-trusted-launch) enables administrators to deploy AKS nodes, which contain the underlying virtual machines, with verified and signed bootloaders, OS kernels, and drivers.

Request: 
This workstream would allow support for Trusted Launch, which is made of two experiences:
- vTPM
- Secure Boot

From a user standpoint I should be able to enter a field in the AKSNodeClass CRD that enables either of the two Trusted Launch Features:

```yaml
spec:
  security:
    trustedLaunch:
      vTPM: true
      secureBoot: true
```

PabloTriv:
Add support for TL via nodeclass. SIG images can be found in agentbaker https://github.com/Azure/AgentBaker/blob/main/pkg/agent/datamodel/sig_config.go.

Trusted Launch docs: https://learn.microsoft.com/en-us/azure/virtual-machines/trusted-launch

When a user enables vTPM or Secure Boot, this PR should make it so that selected Offerings support TL, the vm/machine is wired with TL enabled, and the node image resolves to the TL-enabled version.

Fixes #1634 

**Description**

E2E: 
[machineapi](https://github.com/Azure/karpenter-provider-azure/actions/runs/31191976060)
[scriptless](https://github.com/Azure/karpenter-provider-azure/actions/runs/31185471518)
*

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note

```


charts/karpenter-crd/templates/karpenter.azure.com_aksnodeclasses.yaml
```diff
@@ -776,9 +789,20 @@ spec:
                 type: string
             type: object
             x-kubernetes-validations:
-            - message: FIPS is not yet supported for Ubuntu2204 or Ubuntu2404
+            - message: FIPS is not yet supported for Ubuntu2204 without TrustedLaunch
+                or Ubuntu2404
```
theunrepentantgeek: The double-negative here makes this hard to parse, and I'm unsure whether Ubuntu2404 is supported or not.

Given that the user will have selected either `Ubuntu2404` or `Ubuntu2204`, could we have two separate rules (one for each OS) so they receive validation messages that are specific to their choice?

One could say

``` yaml
- message: FIPS is not yet supported for Ubuntu2404
```

and the other

``` yaml
- message: TrustedLaunch is required for FIPS support with Ubuntu2204
```

pkg/fake/nodeimageversionsapi.go
```diff
@@ -82,6 +82,12 @@ var (
 			SKU:      lo.ToPtr("2204gen2TLcontainerd"),
 			Version:  lo.ToPtr("202512.18.0"),
 		},
+		{
+			FullName: lo.ToPtr("AKSUbuntu-2204gen2fipsTLcontainerd-202512.18.0"),
+			OS:       lo.ToPtr("AKSUbuntu"),
+			SKU:      lo.ToPtr("2204gen2fipsTLcontainerd"),
+			Version:  lo.ToPtr("202512.18.0"),
+		},
```
theunrepentantgeek, line 90: As of Go 1.26, the **new** operator accepts an expression and returns the result, with target typing - no library needed for pointer creation:


```suggestion
		{
			FullName: new("AKSUbuntu-2204gen2fipsTLcontainerd-202512.18.0"),
			OS:       new("AKSUbuntu"),
			SKU:      new("2204gen2fipsTLcontainerd"),
			Version:  new("202512.18.0"),
		},
```

pkg/providers/imagefamily/ubuntu_2004_resolver_test.go
```diff
@@ -27,27 +27,31 @@ import (
 
 func TestResolvesToUbuntu2004(t *testing.T) {
 	cases := []struct {
```
theunrepentantgeek: I've been switching table tests to use `map[string]struct{...}` with the name of the test as the key - primarily because it promotes the name of the test for higher visibility. It's also useful because the non-deterministic ordering of map iteration might reveal a test weakness.

pkg/providers/imagefamily/ubuntu_2004_resolver_test.go
```diff
@@ -27,27 +27,31 @@ import (
...
-		{"empty family + FIPS", lo.ToPtr(""), lo.ToPtr(v1beta1.FIPSModeFIPS), true},
-		{"Ubuntu legacy + FIPS", lo.ToPtr(v1beta1.UbuntuImageFamily), lo.ToPtr(v1beta1.FIPSModeFIPS), true},
-		{"Ubuntu legacy + nil fips", lo.ToPtr(v1beta1.UbuntuImageFamily), nil, false},
-		{"Ubuntu legacy + Disabled", lo.ToPtr(v1beta1.UbuntuImageFamily), lo.ToPtr(v1beta1.FIPSModeDisabled), false},
-		{"Ubuntu2204 + FIPS", lo.ToPtr(v1beta1.Ubuntu2204ImageFamily), lo.ToPtr(v1beta1.FIPSModeFIPS), false},
-		{"Ubuntu2404 + FIPS", lo.ToPtr(v1beta1.Ubuntu2404ImageFamily), lo.ToPtr(v1beta1.FIPSModeFIPS), false},
-		{"AzureLinux + FIPS", lo.ToPtr(v1beta1.AzureLinuxImageFamily), lo.ToPtr(v1beta1.FIPSModeFIPS), false},
+		{"nil family + nil fips", nil, nil, false, false},
+		{"nil family + FIPS", nil, lo.ToPtr(v1beta1.FIPSModeFIPS), false, true},
+		{"nil family + FIPS + TrustedLaunch", nil, lo.ToPtr(v1beta1.FIPSModeFIPS), true, false},
```
theunrepentantgeek: Demo of map style:

```suggestion
		"nil family + nil fips": {nil, nil, false, false},
		"nil family + FIPS": {, nil, lo.ToPtr(v1beta1.FIPSModeFIPS), false, true},
		"nil family + FIPS + TrustedLaunch": {, nil, lo.ToPtr(v1beta1.FIPSModeFIPS), true, false},
```

works even better if the structs are formatted one field per line. 😁

pkg/providers/imagefamily/resolver.go
```diff
@@ -253,14 +262,21 @@ func GetImageFamily(familyName *string, fipsMode *v1beta1.FIPSMode, kubernetesVe
...
 	}
 }
 
-func defaultUbuntu(fipsMode *v1beta1.FIPSMode, kubernetesVersion string, parameters *template.StaticParameters) ImageFamily {
+func defaultUbuntu(fipsMode *v1beta1.FIPSMode, trustedLaunch bool, kubernetesVersion string, parameters *template.StaticParameters) ImageFamily {
 	if lo.FromPtr(fipsMode) == v1beta1.FIPSModeFIPS {
+		if trustedLaunch {
+			return &Ubuntu2204{Options: parameters}
+		}
 		return &Ubuntu2004{Options: parameters}
```
theunrepentantgeek, line 274: Should we still be defaulting anything to Ubuntu 2004? Isn't it well out of support?

theunrepentantgeek (dismissed):
Some suggestions and questions, but nothing to block.

pkg/fake/nodeimageversionsapi.go
PabloTriv: Neat!

pkg/providers/imagefamily/ubuntu_2004_resolver_test.go
PabloTriv: Applied suggestion

pkg/providers/imagefamily/resolver.go
PabloTriv: The original function returns 2004. I agree that that sounds like it needs to be updated, but I'm not sure if changing the default return is within the scope of this PR since it would affect more than just trusted launch?

pkg/providers/imagefamily/ubuntu_2004_resolver_test.go
PabloTriv: Makes sense, I updated it to use a map

charts/karpenter-crd/templates/karpenter.azure.com_aksnodeclasses.yaml
theunrepentantgeek: New rules look good.

pkg/providers/imagefamily/resolver.go
theunrepentantgeek: Fair.

theunrepentantgeek (approved):

pkg/providers/imagefamily/ubuntu_2204.go
```diff
@@ -45,15 +47,45 @@ func (u Ubuntu2204) Name() string {
...
+					ImageDefinition:      Ubuntu2204Gen2FIPSTLImageDefinition,
+					Requirements: scheduling.NewRequirements(
+						scheduling.NewRequirement(v1.LabelArchStable, v1.NodeSelectorOpIn, karpv1.ArchitectureAmd64),
+						scheduling.NewRequirement(v1beta1.LabelSKUHyperVGeneration, v1.NodeSelectorOpIn, v1beta1.HyperVGenerationV2),
+					),
+					Distro: "aks-ubuntu-fips-containerd-22.04-tl-gen2",
+				},
+			}
+		}
 		//TODO: Fill out when Ubuntu 22.04 with FIPS becomes available
```
matthchr, line 71: You removed the FIPS + 2204 block in the CRD but we don't actually return any FIPS images here. That seems incorrect to me?

matthchr: Or is the -tl` image the only FIPS image? Maybe the comment here is just somewhat out of date and we do support FIPS On 2204 but only if TL is enabled, in which case maybe we should update this?

charts/karpenter-crd/templates/karpenter.azure.com_aksnodeclasses.yaml
```diff
@@ -1676,9 +1715,22 @@ spec:
                 type: string
             type: object
             x-kubernetes-validations:
-            - message: FIPS is not yet supported for Ubuntu2204 or Ubuntu2404
+            - message: FIPS is not yet supported for Ubuntu2404
+              rule: 'has(self.fipsMode) && self.fipsMode == ''FIPS'' ? (has(self.imageFamily)
+                && self.imageFamily != ''Ubuntu2404'') : true'
+            - message: TrustedLaunch is required for FIPS support with Ubuntu2204
```
matthchr, line 1721: We dont actually support FIPS+ 2204 though?

pkg/providers/imagefamily/azlinux.go
```diff
@@ -79,6 +80,21 @@ func (u AzureLinux) DefaultImages(useSIG bool, fipsMode *v1beta1.FIPSMode) []typ
...
 	}
+	if trustedLaunch {
+		return []types.DefaultImageOutput{
+			{
+				PublicGalleryURL:     AKSAzureLinuxPublicGalleryURL,
+				GalleryResourceGroup: AKSAzureLinuxResourceGroup,
+				GalleryName:          AKSAzureLinuxGalleryName,
+				ImageDefinition:      AzureLinux2Gen2TrustedLaunchImageDefinition,
+				Requirements: scheduling.NewRequirements(
+					scheduling.NewRequirement(v1.LabelArchStable, v1.NodeSelectorOpIn, karpv1.ArchitectureAmd64),
```
matthchr, line 91: Is there no TrustedLaunch on ARM?

matthchr: Applies to the other AzureLinux families too

PabloTriv: No, based on this config for SIG which I used as a source of truth. https://github.com/Azure/AgentBaker/blob/main/pkg/agent/datamodel/sig_config.go

charts/karpenter-crd/templates/karpenter.azure.com_aksnodeclasses.yaml
PabloTriv: This PR introduces the FIPS + 2204 + TL image. So the cel is trying to say if you want 2204 and fips, it needs to also be TL

PR #1783, opened 2026-07-10
chore: bump sigs.k8s.io/karpenter to v1.14.0

tallaxes:
**Description**

Bump `sigs.k8s.io/karpenter` to `v1.14.0`. Update provider tests for the new provisioner constructor and ProvisioningResult API. Keeps DRA and CapacityBuffer unsupported.

* `make presubmit`
* https://github.com/Azure/karpenter-provider-azure/actions/runs/29468893323

- [x] Yes, issue opened: TBD
- [ ] No

**Release Note**

```release-note

```


charts/karpenter-crd/templates/autoscaling.x-k8s.io_capacitybuffers.yaml
```diff
@@ -0,0 +1,263 @@
+---
+apiVersion: apiextensions.k8s.io/v1
```
matthchr: Should we even be installing the CRD if we're not supporting it/disabling it?

charts/karpenter/values.yaml
```diff
@@ -201,8 +201,8 @@ settings:
   # -- How the Karpenter scheduler treats min values. Options include 'Strict' (fails scheduling when min values can't be met) and 'BestEffort' (relaxes min values when they can't be met).
   minValuesPolicy: ""
   # -- When set, Karpenter ignores pods' Dynamic Resource Allocation requests during scheduling simulations.
-  # This flag will be removed once formal DRA support is GA in Karpenter.
-  ignoreDRARequests: false
+  # DRA is not supported by the Azure provider, so this must remain enabled until support, RBAC, and validation are implemented.
+  ignoreDRARequests: true
```
matthchr, line 205: Does this value need to be changed for NAP too?

matthchr (dismissed):

charts/karpenter-crd/templates/autoscaling.x-k8s.io_capacitybuffers.yaml
tallaxes: Was thinking no harm & we will fast follow with support - but you are right, just skipping is better for now. https://github.com/Azure/karpenter-provider-azure/pull/1783/commits/5c1ff334c39e19a4a3bf2fd3280c9ccbf7cdad3c

charts/karpenter/values.yaml
tallaxes: No, the deployment has different shape there (still ends up disabled by default)

matthchr (approved):

PR #1784, opened 2026-07-10
feat: add and use quota provider

### LINKED ISSUE #1323: Improve handling on quota failures (or avoid them altogether)

### LINKED ISSUE #1733: Prefer availability over lowest-price for on-demand instance selection — Azure parity with aws/karpenter-provider-aws#6721

Add an opt-in way for NAP/Karpenter to **prefer capacity availability over lowest price** when selecting on-demand instance types. Any one of these would solve it:

- a capacity-optimized allocation strategy for on-demand (order offerings by observed availability, not only by price);
- a configurable / longer unavailable-offerings cache TTL, or a negative cache that persists across NodeClaims;
- a NodePool-level "prefer-stable / deprioritize newest-gen" knob.

This is the Azure-provider analog of aws/karpenter-provider-aws#6721 ("Ability to prefer generation over price"). I could not find an existing issue tracking this for the Azure provider.

**Root cause — Karpenter's on-demand selection logic.** For on-demand capacity, Karpenter takes the instance types that match a NodePool's `requirements`, **orders them by price, and attempts the cheapest first** (lowest-price allocation). The newest-generation AMD SKUs are usually the cheapest, so they are attempted first — but those are exactly the SKUs that a capacity-constrained region is most likely to be out of on-demand capacity for. Karpenter then walks down the price-sorted list, failing on each unavailable SKU, until it reaches one that actually has capacity.

There is currently **no way to tell Karpenter not to keep trying the cheap-but-unavailable SKUs first** — the lowest-price ordering is not configurable for on-demand.

**Observed impact.** On-demand D-family CI-runner NodePool (4–16 vCPU, >8Gi) on a dev cluster in **centralus**, 7 days:

| NodeClaim event | Count (7d) |
|---|---|
| created nodeclaim | 922 |
| **failed launching nodeclaim** | **784 (~85%)** |
| launched nodeclaim | 144 |
| registered / initialized | 138 |

- ~99% of failures are `insufficient capacity, all requested instance types were unavailable during launch`.
- Newest-gen AMD SKUs (`Standard_D4alds_v7`, `D4als_v6`, `D4als_v7`, `D4ls_v5`) are attempted first; Karpenter falls back to `Standard_D4as_v5`, which has reliable capacity.
- It is **self-healing**: per-hour over 3 days, every hour had both failures *and* successful launches (never an hour with failures>0 ∧ launches==0), so there is no provisioning outage. But every successful node is preceded by several failed launches.

**Why this is worth solving:** the retry churn (1) adds provisioning latency to CI jobs waiting for a runner node, and (2) produces high control-plane log noise (~784 launch-failure events/week from a single pool), which buries genuine provisioning issues.

Yes — by hand-curating the NodePool `requirements` to pin known-good SKUs (e.g. `karpenter.azure.com/sku-name In [...]`, or `karpenter.azure.com/sku-version NotIn ["6","7"]`) so the price-sorted candidate set only contains SKUs with reliable on-demand capacity in the region.

It works, but it is unsatisfying: it requires per-region capacity knowledge that drifts over time, it partially defeats the purpose of NAP choosing optimal SKUs, and it has to be re-derived and maintained by every user independently.

### Additional Context

- The unavailable-offerings cache (TTL set per-error in `pkg/providers/instance/offerings/commonerrorhandlers.go`, cf. #1727) is what eventually produces the fallback to an available SKU, but its TTL is short and not tunable under managed NAP, so the churn resumes after each expiry.
- Related issues:
  - aws/karpenter-provider-aws#6721 — prefer generation over price (the canonical upstream request).
  - #1727 — surface specific failure reason (capacity vs quota); observability counterpart.
  - #1604 — `allAzureVMSkus` handling; already mentions deprioritizing non-vetted SKUs, which overlaps with option 3 above.

### Attachments

_No response_

### Community Note

matthchr:
* Adds a new quota provider and controller.
* Refreshes quota every 10m (logs quota when it does).
* Considers quota as a hard blocker to instance allocation, similar to when a size is in unavailableofferings.

Fixes #1323
Fixes #1733

There is more we could do here (see [the design](https://github.com/Azure/karpenter-provider-azure/blob/main/designs/0012-quota-fungibility-reactivity-improvements.md)), but this is a stopgap in the meantime that should significantly improve behavior.

* Unit tests
* E2E tests (regression: Hard to have a dedicated quota e2e)
  * Machine: https://github.com/Azure/karpenter-provider-azure/actions/runs/29128641557
  * Standard selfhosted OSS: https://github.com/Azure/karpenter-provider-azure/actions/runs/29128627538

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/providers/azclient/azclient.go
+++ b/pkg/providers/azclient/azclient.go
@@ -211,6 +211,12 @@ func NewAZClient(ctx context.Context, cfg *auth.Config, env *auth.Environment, c
 		return nil, err
 	}
 
+	// Note that this is the Microsoft.Compute/locations/usages API,
+	// which is different than the Microsoft.Quota API. We use it here because:
+	//   * It is what the portal uses.
+	//   * It doesn't require additional RBAC over and above VM contributor which we already require.
+	//   * It is functionally identical to the Microsoft.Quota API.
+	// See designs/0012-quota-fungibility-reactivity-improvements.md for more details.
 	usageClient, err := armcompute.NewUsageClient(cfg.SubscriptionID, cred, opts)
 	if err != nil {
 		return nil, err
--- a/pkg/providers/azclient/azclient.go
+++ b/pkg/providers/azclient/azclient.go
@@ -37,6 +37,7 @@ import (
 	"github.com/Azure/karpenter-provider-azure/pkg/providers/instance/skuclient"
 	"github.com/Azure/karpenter-provider-azure/pkg/providers/loadbalancer"
 	"github.com/Azure/karpenter-provider-azure/pkg/providers/networksecuritygroup"
+	"github.com/Azure/karpenter-provider-azure/pkg/providers/quota"
 	"github.com/Azure/karpenter-provider-azure/pkg/providers/zone"
 	"github.com/Azure/karpenter-provider-azure/pkg/utils/batcher"
 	"github.com/Azure/skewer"
@@ -63,6 +64,7 @@ type AZClient struct {
 	LoadBalancersClient         loadbalancer.LoadBalancersAPI
 	NetworkSecurityGroupsClient networksecuritygroup.API
 	SubscriptionsClient         zone.SubscriptionsAPI
+	UsageClient                 quota.UsageAPI
 }
 
 func (c *AZClient) SubnetsClient() azapi.SubnetsAPI {
@@ -118,6 +120,7 @@ func NewAZClientFromAPI(
 	nodeBootstrappingClient imagefamilytypes.NodeBootstrappingAPI,
 	skuClient skewer.ResourceClient,
 	subscriptionsClient zone.SubscriptionsAPI,
+	usageClient quota.UsageAPI,
 ) *AZClient {
 	return &AZClient{
 		virtualMachinesClient:          virtualMachinesClient,
@@ -136,6 +139,7 @@ func NewAZClientFromAPI(
 		LoadBalancersClient:            loadBalancersClient,
 		NetworkSecurityGroupsClient:    networkSecurityGroupsClient,
 		SubscriptionsClient:            subscriptionsClient,
+		UsageClient:                    usageClient,
 	}
 }
 
@@ -203,6 +207,11 @@ func NewAZClient(ctx context.Context, cfg *auth.Config, env *auth.Environment, c
 	}
 
 	diskEncryptionSetsClient, err := armcompute.NewDiskEncryptionSetsClient(cfg.SubscriptionID, cred, opts)
+	if err != nil {
+		return nil, err
+	}
+
+	usageClient, err := armcompute.NewUsageClient(cfg.SubscriptionID, cred, opts)
 	if err != nil {
 		return nil, err
 	}
@@ -292,5 +301,6 @@ func NewAZClient(ctx context.Context, cfg *auth.Config, env *auth.Environment, c
 		nodeBootstrappingClient,
 		skuClient,
 		subscriptionsClient,
+		usageClient,
 	), nil
 }
--- a/pkg/providers/instancetype/suite_test.go
+++ b/pkg/providers/instancetype/suite_test.go
@@ -3231,8 +3231,6 @@ var _ = Describe("InstanceType Provider", func() {
 		})
 
 		It("should fail open when quota data is unavailable", func() {
-			// No quota data set up — provider has no data
-
 			instanceTypes, err := azureEnv.InstanceTypesProvider.List(ctx, nodeClass)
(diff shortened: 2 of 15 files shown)


pkg/providers/quota/quota.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+}
+
+func (p *DefaultProvider) Update(ctx context.Context) error {
+	freshUsages := map[string]*armcompute.Usage{}
+
+	pager := p.usageClient.NewListPager(p.location, nil)
+	for pager.More() {
+		page, err := pager.NextPage(ctx)
+		if err != nil {
```
tallaxes, line 88: If quota API becomes unavailable, should we remove the existing data (or fail open in some other way)?

pkg/providers/azclient/azclient.go
```diff
@@ -207,6 +211,11 @@ func NewAZClient(ctx context.Context, cfg *auth.Config, env *auth.Environment, c
 		return nil, err
 	}
 
+	usageClient, err := armcompute.NewUsageClient(cfg.SubscriptionID, cred, opts)
```
tallaxes, line 220: Do we need to document that Karpenter now needs quota read?

pkg/providers/quota/quota.go
```diff
@@ -0,0 +1,165 @@
...
+				freshUsages[*usage.Name.Value] = usage
+			}
+		}
+	}
+
+	p.mu.Lock()
+	defer p.mu.Unlock()
+	p.usages = freshUsages
+	atomic.AddUint64(&p.seqNum, 1)
+	if p.cm.HasChanged("quota-usages", freshUsages) {
```
tallaxes: Should probably only update seqNum if something hasChanged?

pkg/providers/instancetype/suite_test.go
```diff
@@ -3098,6 +3098,213 @@ var _ = Describe("InstanceType Provider", func() {
...
+				},
+			)
+			lo.Must0(azureEnv.QuotaProvider.Update(ctx))
+
+			instanceTypes, err := azureEnv.InstanceTypesProvider.List(ctx, nodeClass)
+			Expect(err).To(BeNil())
+
+			foundFamily := false
+			for _, it := range instanceTypes {
+				sku := fake.MakeSKU(it.Name)
```
tallaxes, line 3121: AI found, explains CI failure:
> This is order-dependent and breaks the full suite. The quota specs reconstruct every returned type using fake.MakeSKU(it.Name) at suite_test.go:3120 and several similar loops. Earlier specs can leave Standard_D2_v2_Promo in the provider snapshot, but MakeSKU panics when that name is absent from its static catalog. The full package run reproduced five panics; the PR’s ci-test (1.34.x) job also fails in this package. Resolve each SKU through InstanceTypesProvider.Get(ctx, it.Name), or fully rehydrate the provider during reset.

pkg/providers/quota/quota.go
matthchr: Updated to fail open after 5m of inability to call the quota API. Given controller-runtime exponential backoff that should be something like 4-5 consecutive failures. (10m requeue + 4-5 attempts at least to get us to 15m of staleness).

pkg/providers/quota/quota.go
matthchr: Fixed

pkg/providers/instancetype/suite_test.go
matthchr: Fixed by correctly cleaning up instanceTypesProvider in reset.

tallaxes (dismissed):

pkg/providers/azclient/azclient.go
matthchr: I don't see any RBAC documented that actually gives only Quota/read (the only built-in role that has it is quota request creator), and our UMI for E2E seems to be able to call this API without it, which makes me think we don't have to document it.

Do we actually have documentation for what roles we require anywhere?

matthchr: Let me check on a bare UMI with no permissions

matthchr: Added a new commit updating the docs and adding a comment explaining why we use the API we do and why it doesn't require perms

tallaxes (approved):

PR #1787, opened 2026-07-13
fix: install aks-gpu-cuda-lts (R580.159) driver for modern CUDA GPU SKUs

MaximilianoUribe:
Fixes # 

**Description**

Modern CUDA compute SKUs (T4, V100, A100, H100, H200, ...) must install the R580 **LTS** NVIDIA driver published as the `aks-gpu-cuda-lts` image. That LTS branch is what the AKS GPU VHD driver prebake is built against. Karpenter's scriptless bootstrap was still routing these SKUs to the older `aks-gpu-cuda` image (R580.126), so it installed and loaded a kernel module whose version did not match the userspace NVIDIA libraries already baked into the VHD.

The result on the node:

- `nvidia-smi` fails with `Failed to initialize NVML: Driver/library version mismatch` (loaded kernel module `580.126.09` vs `libnvidia-ml.so.1 -> 580.159.04`).
- The NVIDIA container runtime cannot generate its CDI spec, so the device-plugin pod `CrashLoopBackOff`s.
- The node never advertises `nvidia.com/gpu`, so GPU workloads stay `Pending` until the test times out.

This showed up as *flaky* GPU e2e because it tracks the image family, not luck: the Ubuntu 24.04 entry fails while the AzureLinux entry passes (its VHD/driver-install path stays self-consistent).

This change mirrors the upstream node-bootstrap `GetGPUDriverType` / `GetGPUDriverVersion` mapping:

- Route modern CUDA SKUs to driver type `cuda-lts` at `580.159.04` (image suffix `20260629214430`).
- Keep the legacy `cuda` / R470 path for NCv1 (K80).
- Use the LTS image suffix for the CUDA image SHA.

Only the scriptless bootstrap path is affected — it is the only path that pins an explicit driver version/type; the other provisioning paths request driver installation without a version and let the platform pick the version that matches the VHD.

* `go test ./pkg/utils/` (GPU driver type/version/SHA mapping)
* `go test ./pkg/providers/instancetype/ -args -ginkgo.focus="GPU"` — verifies a T4 node's bootstrap now emits `GPU_DRIVER_TYPE="cuda-lts"`, `GPU_DRIVER_VERSION="580.159.04"`, and the LTS image SHA.

- [ ] Yes, issue opened: #
- [x] No

**Release Note**

```release-note
Modern CUDA GPU SKUs (T4, V100, A100, H100, H200, ...) now install the R580 LTS NVIDIA driver (aks-gpu-cuda-lts), matching the driver prebaked into the node image and fixing an NVML "Driver/library version mismatch" that could leave GPU nodes without an advertised nvidia.com/gpu resource.
```


MaximilianoUribe:
e2e run: https://github.com/Azure/karpenter-provider-azure/actions/runs/29272628445

Bryce-Soghigian (approved):

matthchr (approved):

PR #1791, opened 2026-07-15
fix: bug where loadbalancer deletion could block node provisioning

matthchr:
* When CloudProvider deletes a managed loadbalancer it can take Karpenter up to 2 hours to pick up on that deletion. In the meantime, NIC creation fails due to InvalidResourceReference. This change fixes this behavior so that Karpenter detects the deletion immedaitely and reacts to unblock VM/NIC creation.

* Unit tests
* E2E: https://github.com/Azure/karpenter-provider-azure/actions/runs/29440679302

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```

### DIFF
--- a/pkg/providers/instance/lberrors.go
+++ b/pkg/providers/instance/lberrors.go
@@ -1,0 +1,53 @@
+/*
+Portions Copyright (c) Microsoft Corporation.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package instance
+
+import (
+	"strings"
+
+	sdkerrors "github.com/Azure/azure-sdk-for-go-extensions/pkg/errors"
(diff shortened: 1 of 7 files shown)


pkg/providers/instance/lberrors.go
```diff
@@ -0,0 +1,53 @@
...
+// isMissingSubmittedBackendPool returns true if err is an Azure InvalidResourceReference
+// whose message identifies one of the backend-pool IDs that were submitted in the NIC request.
+// This avoids retrying errors about missing subnets, NSGs, or other unrelated resources.
+func isMissingSubmittedBackendPool(err error, pools *loadbalancer.BackendAddressPools) bool {
+	azErr := sdkerrors.IsResponseError(err)
+	if azErr == nil {
+		return false
+	}
+	if !strings.EqualFold(azErr.ErrorCode, invalidResourceReferenceCode) {
+		return false
```
tallaxes, line 37: Is empty poolID possible here? In which case it should probably be skipped as well ... FWIW, checking with AI: 
> Technically yes, but very unlikely from a normal Azure LIST response. The consequence is minor: strings.Contains(errorMessage, "") is always true, so an unrelated InvalidResourceReference could trigger one unnecessary refresh and retry. It remains bounded and would not create a loop. I would downgrade this from a meaningful PR concern to defensive hardening. The best fix is upstream, preventing empty IDs from entering BackendAddressPools.

tallaxes (dismissed):

pkg/providers/instance/lberrors.go
matthchr: > The best fix is upstream, preventing empty IDs from entering BackendAddressPools.

Fixed

tallaxes (approved):

PR #1794, opened 2026-07-16
fix: use aks-gpu-cuda-lts for compute GPU SKUs to match AgentBaker VHD prebake

ganeshkumarashok:
## Problem

karpenter installs `aks-gpu-cuda` (580.126.09) on CUDA compute SKUs (NCv2/NCv3/T4/A100/H100 — everything except NCv1 K80 and the GRID NV-series). AgentBaker moved these SKUs to `aks-gpu-cuda-lts` (580.159.04) in Azure/AgentBaker#8811 and #8822, and the shared Ubuntu VHDs now prebake cuda-lts.

The driver karpenter installs at bootstrap therefore no longer matches the userspace `libnvidia-ml` prebaked in the VHD. NVML fails to initialize (`Driver/library version mismatch`), the device plugin never registers, and `nvidia.com/gpu` is not advertised — GPU pods stay Pending.

Grid, grid-v20, and NCv1 driver selection are already correct; only the default CUDA path is stale.

## Change

Align `pkg/utils/gpu.go` with AgentBaker's `GetGPUDriverType` / `GetGPUDriverVersion`:

- Add `NvidiaCudaLTSDriverVersion` / `AKSGPUCudaLTSVersionSuffix` and return them from `GetGPUDriverVersion` and `GetAKSGPUImageSHA` for the default CUDA path.
- `GetGPUDriverType` returns `cuda-lts` for the CUDA path; NCv1 (K80) still returns `cuda` (R470).
- Refresh the retained `aks-gpu-cuda` suffix to AgentBaker's current value for parity (it is no longer selected at runtime).

Values are taken verbatim from AgentBaker `parts/common/components.json`.

## Scope / safety

Only the self-contained CSE bootstrap path (`Options` → `cse_cmd.sh.gtpl`) consumes these functions; it now emits `GPU_DRIVER_TYPE="cuda-lts"`, which AgentBaker's CSE resolves to `mcr.microsoft.com/aks/aks-gpu-cuda-lts` with the matching version/SHA. The RP-API bootstrap path (`ProvisionClientBootstrap`, `GPUProfile.DriverType`) does not use these functions and is unchanged — it already resolves the driver server-side. Both paths stay consistent.

No change to which SKUs are GPU-enabled, to grid/grid-v20/NCv1 selection, or to `InstallGPUDriver` behavior.

## Risk

Medium. This changes the driver installed on every CUDA compute GPU node karpenter provisions — but that is the fix: it aligns the installed driver with the VHD prebake, replacing a mismatch that already breaks GPU advertisement. The blast radius is bounded to the CUDA compute path; grid/grid-v20/NCv1 and the RP-API path are untouched, and the values are copied verbatim from AgentBaker. The main residual risk is future drift: these constants are still manually mirrored (see the existing `// TODO: Get these from agentbaker`), so the next AgentBaker driver bump reintroduces the same skew until that sync is automated.

## Testing

- `pkg/utils` unit tests updated and passing (compute SKUs → cuda-lts type + LTS version/SHA; NCv1 → cuda/R470).
- Updated the T4 (`Standard_NC16as_T4_v3`) bootstrap-payload assertion in `pkg/providers/instancetype/suite_test.go` to expect `GPU_DRIVER_TYPE="cuda-lts"` and the LTS version/SHA.
- `go build ./...` and `go vet ./pkg/utils/...` clean. The instancetype Ginkgo suite requires envtest binaries not present locally (panics at `test.NewEnvironment` on unmodified `main` too); it runs in CI.

Refs: Azure/AgentBaker#8811, Azure/AgentBaker#8822.

### DIFF
--- a/pkg/providers/instancetype/suite_test.go
+++ b/pkg/providers/instancetype/suite_test.go
@@ -1640,9 +1640,9 @@ var _ = Describe("InstanceType Provider", func() {
 					ContainSubstring("MIG_NODE=false"),
 					ContainSubstring("CONFIG_GPU_DRIVER_IF_NEEDED=true"),
 					ContainSubstring("ENABLE_GPU_DEVICE_PLUGIN_IF_NEEDED=false"),
-					ContainSubstring("GPU_DRIVER_TYPE=\"cuda\""),
-					ContainSubstring(fmt.Sprintf("GPU_DRIVER_VERSION=\"%s\"", utils.NvidiaCudaDriverVersion)),
-					ContainSubstring(fmt.Sprintf("GPU_IMAGE_SHA=\"%s\"", utils.AKSGPUCudaVersionSuffix)),
+					ContainSubstring("GPU_DRIVER_TYPE=\"cuda-lts\""),
+					ContainSubstring(fmt.Sprintf("GPU_DRIVER_VERSION=\"%s\"", utils.NvidiaCudaLTSDriverVersion)),
+					ContainSubstring(fmt.Sprintf("GPU_IMAGE_SHA=\"%s\"", utils.AKSGPUCudaLTSVersionSuffix)),
 					ContainSubstring("GPU_NEEDS_FABRIC_MANAGER=\"false\""),
 					ContainSubstring("GPU_INSTANCE_PROFILE=\"\""),
 				))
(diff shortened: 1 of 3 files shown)


tallaxes:
@ganeshkumarashok, @MaximilianoUribe - this is very similar to https://github.com/Azure/karpenter-provider-azure/pull/1787 (which has an extra E2E test, not sure if related)

MaximilianoUribe:
duplicate, closing

PR #1795, opened 2026-07-16
test(byok): remove runner label from CMK disk-encryption tests

MaximilianoUribe:
### Description

Removes the `runner` label from the two BYOK customer-managed-key (CMK) disk-encryption tests, because neither can pass under the `runner` suite in either provisioning mode:

- In **Machine API mode**, both tests `Skip()` immediately — the `NODE_OSDISK_DISKENCRYPTIONSET_ID` override is a no-op there.
- **Otherwise**, they only create the DiskEncryptionSet when running with an in-cluster controller. Without that, no DES is configured, so the Phase 7 assertion that the VM has a `DiskEncryptionSet` fails.

Since there's no configuration under which these tests pass in the `runner` suite, they shouldn't carry the `runner` label. They continue to run as part of the normal `byok` suite, so coverage is unchanged.

### How was this change tested?

- `go build ./...` and `golangci-lint` (via pre-commit) pass.
- The two tests still execute in the standard `byok` suite; only their inclusion in the label-filtered `runner` set is removed.

### Does this change impact docs?

- [x] No


tallaxes (approved):
LGTM, though maybe @rakechill should take a look too

PR #1796, opened 2026-07-16
feat: do not hard block VM sizes that are not known

matthchr:
Instead, deprioritize them by setting price to 999 -- user can override with NodeOverlays.

This unblocks using preview SKUs that are not yet in known SKUs while still avoiding the outage risk that a new cheap SKU shows up that doesn't work and everybody gets by default.

* Unit test
* E2E: https://github.com/Azure/karpenter-provider-azure/actions/runs/29542497397

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


pkg/providers/instancetype/suite_test.go
```diff
@@ -2574,6 +2574,78 @@ var _ = Describe("InstanceType Provider", func() {
 				Expect(onDemandAvailable[0].Price).To(BeNumerically("==", pricing.MissingPrice))
 				Expect(spotAvailable[0].Price).To(BeNumerically("==", pricing.MissingPrice))
 			})
+
+			It("should have available offerings with MissingPrice for a SKU not in known_skus.yaml", func() {
+				// Add a SKU to the fake SKU API that is NOT in known_skus.yaml.
+				// This simulates a newly added or private VM size that hasn't been added to the known list yet.
+				// The SKU should still be provisionable but deprioritized with MissingPrice.
+				unknownSKUName := "Standard_D2ts_v6"
```
MaximilianoUribe: assuming this sku is never going to be added? or is it fake?

pkg/providers/instancetype/instancetypes.go
```diff
@@ -436,6 +446,17 @@ func (p *DefaultProvider) UpdateInstanceTypes(ctx context.Context) error {
...
 
+	// Log unknown SKUs (not in known_skus.yaml) that passed all filters, grouped by family.
+	unknownFamilies := sets.New[string]()
+	for name, sku := range instanceTypes {
+		if !IsKnownSKU(name) {
+			unknownFamilies.Insert(sku.GetFamilyName())
+		}
+	}
+	if unknownFamilies.Len() > 0 {
+		log.FromContext(ctx).Info("discovered VM SKU families not in known_skus.yaml (deprioritized with MissingPrice)", "families", unknownFamilies.UnsortedList())
```
MaximilianoUribe: so we are putting the skus that we are missing in known_skus _with_ pricing and the skus that we are missing in known_skus _without_ pricing in the same bucket? not sure if this is relevant but worth taking into account because of the differences in both apis

MaximilianoUribe (dismissed):

pkg/providers/instancetype/instancetypes.go
matthchr: Yes basically if it is unknown (not in known_skus), we force price to 999.
If it doesn't have a price, we (elsewhere) force price to 999.

The net result is that things we don't know about (price or sku) get deprioritized, but the user can override that with NodeOverlay.

pkg/providers/instancetype/suite_test.go
matthchr: It's fake. I will update comment/name to make this clearer.

tallaxes (dismissed):

pkg/providers/instancetype/suite_test.go
matthchr: Changed to Standard_FakeSKU_v6 (more obviously fake)

MaximilianoUribe (approved):

tallaxes (approved):

PR #1818, opened 2026-07-27
feat: align NAP/Karpenter node reservations with AKS node hardening

SriHarsha001:
Fixes # 

**Description**

Adds opt-in node-hardening reservation parity to the Azure provider, gated behind a
new operator option `--enable-node-hardening` (env `ENABLE_NODE_HARDENING`, default
`false`).

When effectively enabled for `aksscriptless` or AKS Machine API modes, `InstanceType.Overhead` and the scriptless (`aksscriptless`) bootstrap
mirror the AKS RP node-hardening formulas:
- **kube-reserved (memory):** `min(35*maxPods + max(250, 2% of RAM), 25% of RAM)` MiB
  (CPU unchanged — same tax brackets as RP).
- **system-reserved:** `200 + 100*floor(memGiB/32)` MiB (+100 for Azure CNI),
  `cpu=100m`, `ephemeral-storage=1Gi`.
- **eviction-hard (memory.available):** VM-size ladder 250 / 375 / 512 MiB.
- **--enforce-node-allocatable:** `pods,kube-reserved,system-reserved` (was `pods`).

All selfhosted tests have succeeded and are green. 
[E2EMatrixTrigger · Azure/karpenter-provider-azure@5c52f3d](https://github.com/Azure/karpenter-provider-azure/actions/runs/32454322204)
NAP E2E test have succeeded as well. 

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


pkg/providers/imagefamily/resolver.go
```diff
@@ -229,7 +232,38 @@ func prepareKubeletConfiguration(ctx context.Context, instanceType *cloudprovide
 	// TODO: revisit computeResources implementation
 	kubeletConfig.KubeReserved = utils.StringMap(instanceType.Overhead.KubeReserved)
 	kubeletConfig.SystemReserved = utils.StringMap(instanceType.Overhead.SystemReserved)
-	kubeletConfig.EvictionHard = map[string]string{instancetype.MemoryAvailable: instanceType.Overhead.EvictionThreshold.Memory().String()}
+	// resource.Quantity canonicalizes 1000 as "1k"; retain the RP's literal
+	// --kube-reserved component for exact bootstrap parity.
+	kubeletConfig.KubeReserved["pid"] = instancetype.KubeReservedPIDs
+	kubeletConfig.EvictionHard = map[string]string{
```
mxj220, line 245: this is not gated on node hardening, so this will change the default behavior of hard eviction in NAP to match what we do in aks-rp. That makes sense, I just want to confirm that NAP team agrees with this

mxj220:
AI-assisted suggestion:

opts.EnableNodeHardening and opts.NetworkPlugin should be added to `type instanceTypeParameters struct` in `/pkg/providers/instancetype/instancetypes.go` because otherwise "`paramsHash` won't change if the flag flips. Cache will keep serving stale Overhead until the 23h TTL rolls or `instanceTypesSeqNum` bumps." Add both to the struct and the List, then reference them from the list instead of from `opts`

example:

      type instanceTypeParameters struct {
	      ImageFamily              string
	      OSDiskSizeGB             int32
	      MaxPods                  int32
	      EncryptionAtHost         bool
	      GPUMode                  v1beta1.GPUMode
	      ArtifactStreamingEnabled bool
	      LocalDNSEnabled          bool
	      EnableNodeHardening      bool  // NEW
	      IsAzureCNI               bool  // NEW
      }

...

	opts := options.FromContext(ctx)  // NEW
	instanceTypeParams := &instanceTypeParameters{
		ImageFamily:              lo.FromPtr(nodeClass.Spec.ImageFamily),
		OSDiskSizeGB:             lo.FromPtr(nodeClass.Spec.OSDiskSizeGB),
		MaxPods:                  utils.GetMaxPods(nodeClass, opts.NetworkPlugin, opts.NetworkPluginMode),
		EncryptionAtHost:         nodeClass.GetEncryptionAtHost(),
		GPUMode:                  nodeClass.GetGPUMode(),
		ArtifactStreamingEnabled: nodeClass.IsArtifactStreamingExplicitlyEnabled(),
		FIPSMode:                 lo.FromPtr(nodeClass.Spec.FIPSMode),
		LocalDNSEnabled:          nodeClass.IsLocalDNSEnabled(),
		EnableNodeHardening:      opts.EnableNodeHardening, // NEW
		IsAzureCNI:               opts.NetworkPlugin == consts.NetworkPluginAzure, // NEW
	}


then reference this list instead of opts in .../instancetype.go:
    
  KubeReserved:      KubeReservedResources(lo.Must(sku.VCPU()), memoryGiB, params.MaxPods, params.EnableNodeHardening),
  SystemReserved:    SystemReservedResources(memoryGiB, params.IsAzureCNI, params.EnableNodeHardening),
  EvictionThreshold: EvictionThreshold(memoryGiB, params.EnableNodeHardening),

pkg/providers/imagefamily/resolver.go
```diff
@@ -229,7 +232,38 @@ func prepareKubeletConfiguration(ctx context.Context, instanceType *cloudprovide
 	// TODO: revisit computeResources implementation
 	kubeletConfig.KubeReserved = utils.StringMap(instanceType.Overhead.KubeReserved)
 	kubeletConfig.SystemReserved = utils.StringMap(instanceType.Overhead.SystemReserved)
-	kubeletConfig.EvictionHard = map[string]string{instancetype.MemoryAvailable: instanceType.Overhead.EvictionThreshold.Memory().String()}
+	// resource.Quantity canonicalizes 1000 as "1k"; retain the RP's literal
```
mxj220: can we just store these as "1000"

mxj220:
Also, for the NAP/Karpenter team, how do we mirror these changes for `aksmachineapi`/`aksmachineapiheaderbatch`?

pkg/providers/imagefamily/resolver.go
SriHarsha001: Addressed this comment

SriHarsha001:
> AI-assisted suggestion:
> 
> opts.EnableNodeHardening and opts.NetworkPlugin should be added to `type instanceTypeParameters struct` in `/pkg/providers/instancetype/instancetypes.go` because otherwise "`paramsHash` won't change if the flag flips. Cache will keep serving stale Overhead until the 23h TTL rolls or `instanceTypesSeqNum` bumps." Add both to the struct and the List, then reference them from the list instead of from `opts`
> 
> example:
> 
> ```
>   type instanceTypeParameters struct {
>       ImageFamily              string
>       OSDiskSizeGB             int32
>       MaxPods                  int32
>       EncryptionAtHost         bool
>       GPUMode                  v1beta1.GPUMode
>       ArtifactStreamingEnabled bool
>       FIPSMode                 v1beta1.FIPSMode
>       LocalDNSEnabled          bool
>       EnableNodeHardening      bool  // NEW
>       IsAzureCNI               bool  // NEW
>   }
> ```
> 
> ...
> 
> ```
> opts := options.FromContext(ctx)  // NEW
> instanceTypeParams := &instanceTypeParameters{
> 	ImageFamily:              lo.FromPtr(nodeClass.Spec.ImageFamily),
> 	OSDiskSizeGB:             lo.FromPtr(nodeClass.Spec.OSDiskSizeGB),
> 	MaxPods:                  utils.GetMaxPods(nodeClass, opts.NetworkPlugin, opts.NetworkPluginMode),
> 	EncryptionAtHost:         nodeClass.GetEncryptionAtHost(),
> 	GPUMode:                  nodeClass.GetGPUMode(),
> 	ArtifactStreamingEnabled: nodeClass.IsArtifactStreamingExplicitlyEnabled(),
> 	FIPSMode:                 lo.FromPtr(nodeClass.Spec.FIPSMode),
> 	LocalDNSEnabled:          nodeClass.IsLocalDNSEnabled(),
> 	EnableNodeHardening:      opts.EnableNodeHardening, // NEW
> 	IsAzureCNI:               opts.NetworkPlugin == consts.NetworkPluginAzure, // NEW
> }
> ```
> 
> then reference this list instead of opts in .../instancetype.go:
> 
> KubeReserved: KubeReservedResources(lo.Must(sku.VCPU()), memoryGiB, params.MaxPods, params.EnableNodeHardening), SystemReserved: SystemReservedResources(memoryGiB, params.IsAzureCNI, params.EnableNodeHardening), EvictionThreshold: EvictionThreshold(memoryGiB, params.EnableNodeHardening),

[EnableNodeHardening]/[NetworkPlugin] are CLI-flag/env options parsed once at startup (injection.WithOptionsOrDie in [main.go]) and never change for the life of the process — no reload mechanism exists. Changing either requires a pod restart, which also rebuilds [instanceTypesCache] from scratch, so there's no window for stale [Overhead] to be served.

This also matches how we already treat [opts.NodeResourceGroup] in the same function — another static, process-global option read directly from [opts] rather than added to instanceTypeParameters, since it can't vary within a process's lifetime either.

So I'd rather not add this unless Options becomes hot-reloadable later.

pkg/operator/options/options.go
```diff
@@ -63,12 +63,17 @@ func (s *nodeIdentitiesValue) String() string { return strings.Join(*s, ",") }
...
-	KubeletClientTLSBootstrapToken string  `json:"-"` // => TLSBootstrapToken in bootstrap (may need to be per node/nodepool)
-	LinuxAdminUsername             string  `json:"-"`
-	SSHPublicKey                   string  `json:"-"` // ssh.publicKeys.keyData => VM SSH public key // TODO: move to v1beta1.AKSNodeClass?
+	ClusterName             string  `json:"clusterName,omitempty"`
+	ClusterEndpoint         string  `json:"clusterEndpoint,omitempty"` // => APIServerName in bootstrap, except needs to be w/o https/port
+	VMMemoryOverheadPercent float64 `json:"vmMemoryOverheadPercent,omitempty"`
+	// EnableNodeHardening mirrors the cluster-level enableNodeHardening setting
+	// in the AKS RP. Keep this option synchronized with the RP so Karpenter's
+	// scheduling simulation and rendered kubelet configuration match the node.
+	// Remove it when node hardening becomes unconditional in the RP.
```
comtalyst:
```suggestion
	// Remove it when node hardening becomes unconditional in AKS.
```

pkg/providers/imagefamily/resolver.go
comtalyst: Is this going to match what we do in AKS **as of today**?

comtalyst: See the [note above](https://github.com/Azure/karpenter-provider-azure/pull/1818#discussion_r3738961741) on provision mode clarification--this will not affect NAP.

In ideal case, the defaulting here (for scriptless mode), bootstrapping API, and machine API should be aligned. If this is getting us closer to that point, it is good.

pkg/providers/imagefamily/resolver.go
comtalyst, line 230: Note that your changes to this block is not effective on the nodes provisioned with `bootstrappingclient` mode, in which NAP is using it.

For that mode, Karpenter only sends what's in `pkg/provisionclients/models/custom_kubelet_config.go` through `pkg/providers/imagefamily/customscriptsbootstrap/provisionclientbootstrap.go` to the bootstrapping API to retrieve custom data/CSE. The unexposed kubelet configs, in which your changes are targeting, are instead being defaulted behind the bootstrapping API (that code we talked about). 
So, I suppose changes in `instancetype` (for Karpenter scheduling awareness) and changes behind the API (for actual node provisioning values) is sufficient. Machine API works similarly.

This, however, will affect `aksscriptless` mode, in which self-hosted Karpenter is using it. For that, I would recommend running self-hosted E2E as well, either through [GitHub Actions](https://github.com/Azure/karpenter-provider-azure/actions/workflows/e2e-matrix-trigger.yaml) or creating codespaces, then `make az-all` and `make e2etest`.

pkg/providers/instancetype/instancetype.go
```diff
@@ -126,15 +127,17 @@ func newInstanceType(
...
 		Offerings:    offerings,
 		Capacity:     computeCapacity(ctx, sku, params),
 		Overhead: &cloudprovider.InstanceTypeOverhead{
-			KubeReserved:      KubeReservedResources(lo.Must(sku.VCPU()), lo.Must(sku.Memory())),
-			SystemReserved:    SystemReservedResources(),
-			EvictionThreshold: EvictionThreshold(),
+			KubeReserved:      KubeReservedResources(lo.Must(sku.VCPU()), memoryGiB, params.MaxPods, opts.EnableNodeHardening),
+			SystemReserved:    SystemReservedResources(memoryGiB, opts.NetworkPlugin == consts.NetworkPluginAzure, opts.EnableNodeHardening),
+			EvictionThreshold: EvictionThreshold(memoryGiB, opts.EnableNodeHardening),
 		},
```
comtalyst, line 142: Just a note: in the future, once these are exposed through Machine API (just saw it on the latest preview API, but not sure if ready), we should consider sending these to Machine API rather than relying on "Machine API will set the same value" so we have the source of truth in one place.

comtalyst (commented):
(let's also wait for more NAP test results)

pkg/providers/imagefamily/resolver.go
comtalyst: For the modes, `bootstrappingclient` will be deprecated once machine API migration is done. `aksscriptless` will eventually stop being the default mode on self-hosted in favor of machine mode. The code will likely be retained, but _likely_ deprecated from our support, but that might take a bit longer.

pkg/operator/options/suite_test.go
```diff
@@ -170,6 +174,15 @@ var _ = Describe("Options", func() {
 			})
 			Expect(opts).To(BeComparableTo(expectedOpts, cmpopts.IgnoreUnexported(options.Options{})))
 		})
+
+		It("should expose the enable-node-hardening external contract", func() {
```
matthchr: minor: This test doesn't seem amazingly useful to me. I would just remove it.

pkg/operator/options/options.go
```diff
@@ -108,6 +113,7 @@ func (o *Options) AddFlags(fs *coreoptions.FlagSet) {
 	fs.StringVar(&o.ClusterName, "cluster-name", env.WithDefaultString("CLUSTER_NAME", ""), "[REQUIRED] The kubernetes cluster name for resource tags.")
 	fs.StringVar(&o.ClusterEndpoint, "cluster-endpoint", env.WithDefaultString("CLUSTER_ENDPOINT", ""), "[REQUIRED] The external kubernetes cluster endpoint for new nodes to connect with.")
 	fs.Float64Var(&o.VMMemoryOverheadPercent, "vm-memory-overhead-percent", utils.WithDefaultFloat64("VM_MEMORY_OVERHEAD_PERCENT", 0.075), "The VM memory overhead as a percent that will be subtracted from the total memory for all instance types.")
+	fs.BoolVar(&o.EnableNodeHardening, "enable-node-hardening", env.WithDefaultBool("ENABLE_NODE_HARDENING", false), "If set to true, Karpenter applies (kube-reserved, system-reserved, hard-eviction, soft-eviction) matching AKS node hardening, keeping scheduling simulation and rendered kubelet configuration aligned with hardened values.")
```
matthchr: Currently unclear to me if we even want this value... can you explain the desire/need for it? Toggle in AKS RP?

pkg/providers/imagefamily/bootstrap/aksbootstrap_test.go
```diff
@@ -205,4 +207,79 @@ func TestKubeletConfigMap(t *testing.T) {
 	for k, v := range expectedKubeletConfigs {
 		g.Expect(actualKubeletConfig[k]).To(Equal(v), fmt.Sprintf("parameter mismatch for %s", k))
 	}
+	g.Expect(strings.Split(actualKubeletConfig["--kube-reserved"], ",")).To(ConsistOf("cpu=180m", "memory=2250Mi", "pid=1000"))
```
matthchr: Just confirming: This is actually just bringing OSS Karpenter up to parity with what AKS RP is doing today is that right?

pkg/providers/imagefamily/bootstrap/aksbootstrap_test.go
```diff
@@ -205,4 +207,79 @@ func TestKubeletConfigMap(t *testing.T) {
...
+}
+
+func TestKubeletConfigMapEnforceNodeAllocatable(t *testing.T) {
+	g := NewWithT(t)
+
+	// Node hardening on: emit the AKS RP hardening signal; AgentBaker owns the
```
matthchr: minor: This seems like 2 tests to me? One for `kubeletConfigToMap` w/ empty kubeletConfiguration, and one with KubeReserved/SystemReserved set?

pkg/providers/imagefamily/bootstrap/bootstrap.go
```diff
@@ -42,6 +42,10 @@ type KubeletConfiguration struct {
 	// EvictionMaxPodGracePeriod is the maximum allowed grace period (in seconds) to use when terminating pods in
 	// response to soft eviction thresholds being met.
 	EvictionMaxPodGracePeriod *int32
+	// EnforceNodeAllocatable is the list of node-allocatable enforcement scopes passed to
+	// --enforce-node-allocatable (e.g. pods, kube-reserved, system-reserved). When empty the
+	// kubelet default ("pods") is left in place.
+	EnforceNodeAllocatable []string
```
matthchr, line 48: minor:
```suggestion
	// EnforceNodeAllocatable is the list of node-allocatable enforcement scopes passed to
	// --enforce-node-allocatable (e.g. pods, kube-reserved, system-reserved). When empty the
	// default is the kubelet default ("pods")
	EnforceNodeAllocatable []string
```

pkg/providers/imagefamily/resolver.go
```diff
@@ -229,7 +232,34 @@ func prepareKubeletConfiguration(ctx context.Context, instanceType *cloudprovide
 	// TODO: revisit computeResources implementation
 	kubeletConfig.KubeReserved = utils.StringMap(instanceType.Overhead.KubeReserved)
 	kubeletConfig.SystemReserved = utils.StringMap(instanceType.Overhead.SystemReserved)
-	kubeletConfig.EvictionHard = map[string]string{instancetype.MemoryAvailable: instanceType.Overhead.EvictionThreshold.Memory().String()}
+	kubeletConfig.KubeReserved["pid"] = instancetype.KubeReservedPIDs
```
matthchr, line 244: This is safe to override because `instanceType.Overhead.KubeReserved` never contains `pid`, is that right?

Can you leave a comment here to that effect?

matthchr: minor: stopped here

pkg/providers/instancetype/instancetype.go
```diff
@@ -126,15 +127,17 @@ func newInstanceType(
...
 		Name:         sku.GetName(),
-		Requirements: computeRequirements(options.FromContext(ctx), sku, vmsize, architecture, offerings, region, params),
+		Requirements: computeRequirements(opts, sku, vmsize, architecture, offerings, region, params),
 		Offerings:    offerings,
 		Capacity:     computeCapacity(ctx, sku, params),
 		Overhead: &cloudprovider.InstanceTypeOverhead{
-			KubeReserved:      KubeReservedResources(lo.Must(sku.VCPU()), lo.Must(sku.Memory())),
-			SystemReserved:    SystemReservedResources(),
-			EvictionThreshold: EvictionThreshold(),
+			KubeReserved:      KubeReservedResources(lo.Must(sku.VCPU()), memoryGiB, params.MaxPods, opts.EnableNodeHardening),
```
matthchr: need to make sure to force enableNodeHardening parameter to false if Bootstrap client mode is set in opts.

pkg/operator/options/options.go
```diff
@@ -63,12 +63,17 @@ func (s *nodeIdentitiesValue) String() string { return strings.Join(*s, ",") }
...
-	LinuxAdminUsername             string  `json:"-"`
-	SSHPublicKey                   string  `json:"-"` // ssh.publicKeys.keyData => VM SSH public key // TODO: move to v1beta1.AKSNodeClass?
+	ClusterName             string  `json:"clusterName,omitempty"`
+	ClusterEndpoint         string  `json:"clusterEndpoint,omitempty"` // => APIServerName in bootstrap, except needs to be w/o https/port
+	VMMemoryOverheadPercent float64 `json:"vmMemoryOverheadPercent,omitempty"`
+	// EnableNodeHardening mirrors the cluster-level enableNodeHardening setting
+	// in the AKS RP. Keep this option synchronized with the RP so Karpenter's
+	// scheduling simulation and rendered kubelet configuration match the node.
+	// Remove it when node hardening becomes unconditional in AKS.
+	EnableNodeHardening            bool   `json:"enableNodeHardening,omitempty"`
```
matthchr, line 75: Spoke in person, the idea is that the cluster level setting on ManagedCluster flows down here, just for preview.

Then eventually, this field will not GA with ManagedCluster and instead we will do some defaulting based on KubernetesVersion.

Can you maybe update this comment to make it more clear that this field will be removed when AKS starts performing this defaulting on new nodes (probably k8s 1.37?) and the matching ManagedCluster field is removed from the preview API

pkg/providers/imagefamily/resolver.go
matthchr, line 244: just confirming, the bits of this that are outside of the `options.FromContext(ctx).EnableNodeHardening` if statement are to make NAP/Karpenter match what AKS RP is already unconditionally doing is that right?

pkg/providers/imagefamily/resolver.go
```diff
@@ -229,7 +232,34 @@ func prepareKubeletConfiguration(ctx context.Context, instanceType *cloudprovide
...
+		kubeletConfig.SystemReserved["pid"] = instancetype.SystemReservedPIDs
+		totalMemoryMiB := lo.Must(strconv.ParseInt(instanceType.Requirements.Get(v1beta1.LabelSKUMemory).Any(), 10, 64))
+		softEvictionThreshold := instancetype.SoftEvictionThreshold(totalMemoryMiB)
+		kubeletConfig.EvictionSoft = map[string]string{
+			instancetype.MemoryAvailable:  softEvictionThreshold.Memory().String(),
+			instancetype.NodeFSAvailable:  instancetype.SoftEvictionNodeFSAvailable,
+			instancetype.NodeFSInodesFree: instancetype.SoftEvictionNodeFSInodesFree,
+		}
+		kubeletConfig.EvictionSoftGracePeriod = map[string]metav1.Duration{
+			instancetype.MemoryAvailable:  {Duration: 30 * time.Second},
```
matthchr: minor: should these durations be defined as consts in instancetype too, since everything else seems to be?

pkg/providers/imagefamily/resolver_unit_test.go
```diff
@@ -0,0 +1,128 @@
...
+		t.Fatalf("hard eviction thresholds = %#v, want %#v", configuration.EvictionHard, expectedHardThresholds)
+	}
+
+	expectedThresholds := map[string]string{
+		"memory.available":  "1Gi",
+		"nodefs.available":  "12%",
+		"nodefs.inodesFree": "7%",
+	}
+	if !reflect.DeepEqual(configuration.EvictionSoft, expectedThresholds) {
+		t.Fatalf("soft eviction thresholds = %#v, want %#v", configuration.EvictionSoft, expectedThresholds)
```
matthchr: minor, style: I think usually we prefer gomega asserts. Test doesn't/shouldn't be ginkgo, but using gomega for asserts to match general pattern elsewhere in the repo.

pkg/providers/instancetype/nodehardening.go
```diff
@@ -0,0 +1,112 @@
...
+// Karpenter via options.Options.EnableNodeHardening) the kube-reserved,
+// system-reserved and eviction values that Karpenter uses for scheduling
+// simulation and kubelet configuration must match what the AKS RP /
+// NodeProvisioner configures. Otherwise the bin-packing
+// simulation (allocatable = capacity - overhead) diverges from reality and
+// nodes are over- or under-provisioned.
+//
+// The AKS RP is the source of truth; keep the formulas below in sync with the
+// go.goms.io/aks/rp repository:
+//   - kube-reserved memory: resourceprovider/sharedlib/common/kubereserved/utils.go
```
matthchr: Let's not refer to AKS RP code that the public cannot access here in OSS.

If we really think this is valuable, you can put it in an issue in karpenter-poc and then reference the issue here rather than direct code links... but ideally we wouldn't even do that we would just say something like "keep in sync with AKS RP".

It's fine to put links to this code in the RP though.

pkg/providers/instancetype/instancetype.go
```diff
@@ -343,19 +346,28 @@ func pods(params *instanceTypeParameters) *resource.Quantity {
...
+		corev1.ResourceCPU:              *resource.NewScaledQuantity(systemReservedCPUMillicores, resource.Milli),
+		corev1.ResourceMemory:           *resource.NewQuantity(mibToBytes(systemReservedMemoryMiB(memoryGiB, isAzureCNI)), resource.BinarySI),
+		corev1.ResourceEphemeralStorage: resource.MustParse(systemReservedEphemeralStorage),
 	}
 }
 
-func KubeReservedResources(vcpus int64, memoryGib float64) corev1.ResourceList {
-	reservedMemoryMi := int64(1024 * reservedMemoryTaxGi.Calculate(memoryGib))
+func KubeReservedResources(vcpus int64, memoryGiB float64, maxPods int32, enableNodeHardening bool) corev1.ResourceList {
+	reservedMemoryMi := int64(1024 * reservedMemoryTaxGi.Calculate(memoryGiB))
```
matthchr: minor, style: Weird to have `Mi` and `GiB`... `GiB` * 1024 should be `MiB` right?

pkg/providers/instancetype/instancetype.go
```diff
@@ -343,19 +346,28 @@ func pods(params *instanceTypeParameters) *resource.Quantity {
 	return resource.NewQuantity(int64(params.MaxPods), resource.DecimalSI)
 }
 
-func SystemReservedResources() corev1.ResourceList {
-	// AKS does not set system-reserved values and only CPU and memory are considered
-	// https://learn.microsoft.com/en-us/azure/aks/concepts-clusters-workloads#resource-reservations
+func SystemReservedResources(memoryGiB float64, isAzureCNI, enableNodeHardening bool) corev1.ResourceList {
```
matthchr: minor, style maybe just pass the actual network plugin string and do the compare in here?

pkg/providers/instancetype/nodehardening.go
```diff
@@ -0,0 +1,112 @@
...
+	HardEvictionPIDAvailable             = "2000"
+	SoftEvictionNodeFSAvailable          = "12%"
+	SoftEvictionNodeFSInodesFree         = "7%"
+	SoftEvictionMaxPodGracePeriodSeconds = int32(60)
+	KubeReservedPIDs                     = "1000"
+	SystemReservedPIDs                   = "1000"
+)
+
+// mibToBytes converts a MiB quantity to bytes.
+func mibToBytes(mib int64) int64 {
```
matthchr: minor, style, feel free to ignore: Maybe could just be `const bytesPerMiB int64 = 1024 * 1024` and then multiple by that rather than calling a function?

matthchr: Also there are some places this isn't used like `*resource.NewQuantity(reservedMemoryMi*1024*1024, resource.BinarySI)`

pkg/providers/instancetype/instancetype.go
```diff
@@ -343,19 +346,28 @@ func pods(params *instanceTypeParameters) *resource.Quantity {
...
 }
 
-func KubeReservedResources(vcpus int64, memoryGib float64) corev1.ResourceList {
-	reservedMemoryMi := int64(1024 * reservedMemoryTaxGi.Calculate(memoryGib))
+func KubeReservedResources(vcpus int64, memoryGiB float64, maxPods int32, enableNodeHardening bool) corev1.ResourceList {
+	reservedMemoryMi := int64(1024 * reservedMemoryTaxGi.Calculate(memoryGiB))
 	reservedCPUMilli := int64(1000 * reservedCPUTaxVCPU.Calculate(float64(vcpus)))
 
+	if enableNodeHardening {
+		reservedMemoryMi = hardenedKubeReservedMemoryMiB(maxPods, int64(math.Floor(memoryGiB*1024)))
```
matthchr: Assuming memoryGiB cannot be negative, I don't think math.Floor is needed here

pkg/providers/instancetype/instancetype.go
```diff
@@ -364,8 +376,23 @@ func KubeReservedResources(vcpus int64, memoryGib float64) corev1.ResourceList {
 	return resources
 }
 
-func EvictionThreshold() corev1.ResourceList {
+func EvictionThreshold(memoryGiB float64, enableNodeHardening bool) corev1.ResourceList {
+	if enableNodeHardening {
+		_, hardMemoryMiB := evictionMemoryLadder(int64(math.Floor(memoryGiB * 1024)))
```
matthchr: Assuming memoryGiB cannot be negative, I don't think math.Floor is needed here

pkg/providers/instancetype/nodehardening.go
```diff
@@ -0,0 +1,112 @@
...
+	reservedMiB := 35*int64(maxPods) + max(totalMemoryMiB*2/100, 250)
+	return min(reservedMiB, capMiB)
+}
+
+// systemReservedMemoryMiB returns the hardened system-reserved memory in MiB:
+//
+//	200 + 100*floor(memGiB/32) (+100 when Azure CNI)
+//
+// Mirrors calculateSystemReservedMemoryMiB in the AKS RP.
+func systemReservedMemoryMiB(memoryGiB float64, isAzureCNI bool) int64 {
```
matthchr: Some of these helpers take MiB, some take GiB.

Feels like we should unify on a common unit, even though the callers sometimes have different ones? Then the caller can convert to the standard (MiB, maybe?)

pkg/providers/instancetype/instancetype.go
```diff
@@ -364,8 +376,23 @@ func KubeReservedResources(vcpus int64, memoryGib float64) corev1.ResourceList {
...
+		}
+	}
 	return corev1.ResourceList{
 		corev1.ResourceMemory: resource.MustParse(DefaultMemoryAvailable),
 	}
 }
+
+// SoftEvictionThreshold returns the hardened soft-eviction memory threshold
+// for the VM's total memory using the AKS RP memory ladder.
+func SoftEvictionThreshold(totalMemoryMiB int64) corev1.ResourceList {
```
matthchr, line 393: Same comment as below - some of these take memory MiB, some GiB. Feels like we should unify on what the helper functions take for consistency sake.

pkg/providers/instancetype/nodehardening_unit_test.go
```diff
@@ -0,0 +1,120 @@
...
+
+import (
+	"testing"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
+// These cases mirror calculateMemoryReservation(enableNodeHardening=true) in
+// resourceprovider/sharedlib/common/kubereserved/utils.go in the AKS RP.
```
matthchr: similar comment to elsewhere, avoid RP codepaths

pkg/providers/instancetype/nodehardening_unit_test.go
```diff
@@ -0,0 +1,120 @@
...
+		{name: "128 GiB with 250 pods", vcpus: 16, memoryGiB: 128, maxPods: 250, wantMemoryMiB: 11371},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			resources := KubeReservedResources(test.vcpus, test.memoryGiB, test.maxPods, true)
+			memory := resources[corev1.ResourceMemory]
+			wantMemory := *resource.NewQuantity(mibToBytes(test.wantMemoryMiB), resource.BinarySI)
+			if memory.Cmp(wantMemory) != 0 {
+				t.Fatalf("resources = memory=%s; want memory=%s", memory.String(), wantMemory.String())
```
matthchr: minor: Same comment as elsewhere, consider Gomega for assertions (standard Go test is fine, no Ginkgo)

pkg/providers/instancetype/suite_test.go
```diff
@@ -1212,6 +1212,42 @@ var _ = Describe("InstanceType Provider", func() {
...
+
+				expectedFlags := map[string]string{
+					"enforce-node-allocatable":      "pods,kube-reserved,system-reserved",
+					"eviction-max-pod-grace-period": "60",
+				}
+
+				ExpectKubeletFlags(azureEnv, customData, expectedFlags)
+				kubeletFlags := ExpectKubeletFlagsPassed(customData)
+				Expect(kubeletFlags).To(ContainSubstring("--eviction-soft="))
+				Expect(kubeletFlags).To(ContainSubstring("memory.available<"))
```
matthchr: missing an actual number? Or is the issue that it may vary?

pkg/providers/instancetype/instancetype_test.go
```diff
@@ -0,0 +1,120 @@
+/*
```
matthchr, line 1: minor: file should just be `nodehardening_test.go`

pkg/providers/instancetype/instancetype_test.go
```diff
@@ -0,0 +1,120 @@
...
+import (
+	"testing"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
+// These cases mirror calculateMemoryReservation(enableNodeHardening=true) in
+// resourceprovider/sharedlib/common/kubereserved/utils.go in the AKS RP.
+func TestKubeReservedResourcesHardeningParity(t *testing.T) {
```
matthchr, line 30: The functions these are testing are actually defined in instancetype.go, so feels like this should be `instancetype_test.go` rather than `nodehardening_unit_test.go`

pkg/providers/instancetype/suite_test.go
matthchr: Also is this not very similar to `ExpectHardEvictionThresholds`?

pkg/providers/instancetype/suite_test.go
```diff
@@ -3428,14 +3455,66 @@ var _ = Describe("Tax Calculator", func() {
...
+			Expect(cpu.String()).To(Equal("180m"))
+			Expect(mem.String()).To(Equal("4505Mi"))
+		})
+		It("KubeReservedResources is capped at 25% of memory", func() {
+			// Tiny node, very high pod density -> capped at 25% of 8Gi = 2048Mi (= 2Gi).
+			resources := instancetype.KubeReservedResources(int64(2), 8.0, int32(250), true)
+			mem := resources[v1.ResourceMemory]
+			Expect(mem.String()).To(Equal("2Gi"))
+		})
+		It("SystemReservedResources adds the Azure CNI bonus", func() {
```
matthchr: These seem unit unit tests of SystemReservedResources and KubeReservedResources, why not put them next to the tests in `nodehardening_unit_test.go` -- or since I asked you to rename it into `instancetype_test.go`, there instead?

pkg/providers/instancetype/instancetype.go
SriHarsha001: Made the suggested changes.

pkg/operator/options/suite_test.go
SriHarsha001: Made the suggested changes.

pkg/providers/imagefamily/bootstrap/aksbootstrap_test.go
SriHarsha001: Made the suggested changes. Added 2 test functions.

pkg/providers/imagefamily/bootstrap/bootstrap.go
SriHarsha001: Made the suggested changes.

pkg/providers/imagefamily/bootstrap/aksbootstrap_test.go
SriHarsha001: Yes. This change makes the OSS provider use the same values for Karpenter’s allocatable calculation and rendered kubelet flags, so scheduling simulation matches the resulting AKS node. However, this is an existing test case, so just removed the check.

pkg/operator/options/options.go
SriHarsha001: Updated the comments.

pkg/providers/imagefamily/resolver.go
SriHarsha001: Made the changes.

pkg/providers/imagefamily/resolver_unit_test.go
SriHarsha001: Updated the code.

pkg/providers/instancetype/nodehardening.go
SriHarsha001: Updated the comments.

pkg/providers/instancetype/instancetype.go
SriHarsha001: Updated the code.

pkg/providers/instancetype/instancetype.go
SriHarsha001: Updated the code.

pkg/providers/instancetype/nodehardening.go
SriHarsha001: Updated the code.

pkg/providers/instancetype/instancetype.go
SriHarsha001: Updated the code.

pkg/providers/instancetype/instancetype.go
SriHarsha001: Updated the code.

pkg/providers/instancetype/nodehardening.go
SriHarsha001: Updated the code.

pkg/providers/instancetype/instancetype.go
SriHarsha001: Updated the code.

pkg/providers/instancetype/nodehardening_unit_test.go
SriHarsha001: Updated the code.

pkg/providers/instancetype/nodehardening_unit_test.go
SriHarsha001: Updated the code.

pkg/providers/instancetype/instancetype_test.go
SriHarsha001: Updated the file name.

pkg/providers/instancetype/instancetype_test.go
SriHarsha001: Update the code.

pkg/providers/instancetype/suite_test.go
SriHarsha001: Updated the code.

pkg/providers/instancetype/suite_test.go
SriHarsha001: updated the code.

pkg/providers/imagefamily/resolver.go
```diff
@@ -238,7 +241,34 @@ func prepareKubeletConfiguration(ctx context.Context, instanceType *cloudprovide
...
+		instancetype.MemoryAvailable:  instanceType.Overhead.EvictionThreshold.Memory().String(),
+		instancetype.NodeFSAvailable:  instancetype.HardEvictionNodeFSAvailable,
+		instancetype.NodeFSInodesFree: instancetype.HardEvictionNodeFSInodesFree,
+		instancetype.PIDAvailable:     instancetype.HardEvictionPIDAvailable,
+	}
+
+	if options.FromContext(ctx).EnableNodeHardening {
+		kubeletConfig.SystemReserved["pid"] = instancetype.SystemReservedPIDs
+		totalMemoryMiB := lo.Must(strconv.ParseInt(instanceType.Requirements.Get(v1beta1.LabelSKUMemory).Any(), 10, 64))
+		softEvictionThreshold := instancetype.SoftEvictionThreshold(totalMemoryMiB)
```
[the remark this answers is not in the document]
SriHarsha001: Updated the code.

pkg/providers/imagefamily/resolver.go
SriHarsha001: We had a call and we decided that we will enable node hardening only for machineAPI mode and selfhosted/scripless mode and not for bootstrappAPI mode. aksscriptless: hardening enabled for scheduling and rendered kubelet configuration.
aksmachineapi and aksmachineapiheaderbatch: hardening enabled for scheduling; AKS Machine API supplies the actual node configuration.
bootstrappingclient: hardening disabled, even if the feature flag is set.

pkg/providers/imagefamily/resolver.go
SriHarsha001: Yes, that’s correct. KubeReservedResources only populates CPU and memory, so instanceType.Overhead.KubeReserved cannot already contain a PID reservation.

SriHarsha001: Yes, that is correct. We already have hardeviction set by default in kubelet (even without the nodehardening flag set).

pkg/providers/imagefamily/resolver.go
```diff
@@ -238,7 +240,36 @@ func prepareKubeletConfiguration(ctx context.Context, instanceType *cloudprovide
...
+			instancetype.NodeFSAvailable:  {Duration: instancetype.SoftEvictionNodeFSGracePeriod},
+			instancetype.NodeFSInodesFree: {Duration: instancetype.SoftEvictionNodeFSInodesGracePeriod},
+		}
+		kubeletConfig.EvictionMaxPodGracePeriod = lo.ToPtr(instancetype.SoftEvictionMaxPodGracePeriodSeconds)
+
+		// Signal node hardening to AgentBaker, which owns the reserved-cgroup paths.
+		// Mirrors nodeAllocatableEnforcementHardened in the AKS RP.
+		kubeletConfig.EnforceNodeAllocatable = []string{"pods", "kube-reserved", "system-reserved"}
+	}
 	return kubeletConfig
```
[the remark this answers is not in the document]
SriHarsha001: LabelSKUMemory is added internally for every instance type in computeRequirements(). It is derived from the validated SKU memory value, so it should always exist and contain a valid integer. It is not user-provided data.

pkg/providers/instancetype/instancetype.go
```diff
@@ -128,20 +128,31 @@ func newInstanceType(
...
+// ShouldUseNodeHardening returns whether node hardening applies to the configured provisioning mode.
+func ShouldUseNodeHardening(enableNodeHardening bool, provisionMode string) bool {
+	if !enableNodeHardening {
+		return false
+	}
+	return provisionMode == consts.ProvisionModeAKSScriptless ||
+		provisionMode == consts.ProvisionModeAKSMachineAPI ||
+		provisionMode == consts.ProvisionModeAKSMachineAPIHeaderBatch
+}
+
```
comtalyst: This might better be in `options` module?

pkg/operator/options/options.go
```diff
@@ -66,10 +66,9 @@ type Options struct {
...
 	ClusterEndpoint         string  `json:"clusterEndpoint,omitempty"` // => APIServerName in bootstrap, except needs to be w/o https/port
 	VMMemoryOverheadPercent float64 `json:"vmMemoryOverheadPercent,omitempty"`
-	// EnableNodeHardening mirrors the cluster-level enableNodeHardening setting
-	// in the AKS RP. Keep this option synchronized with the RP so Karpenter's
-	// scheduling simulation and rendered kubelet configuration match the node.
-	// Remove it when node hardening becomes unconditional in the RP.
+	// EnableNodeHardening is a temporary preview only flag on ManagedCluster.
+	// Will remove this option when AKS defaults node hardening for new nodes based on Kubernetes version
+	// (expected 1.39+) and removes the corresponding field from the preview API.
 	EnableNodeHardening            bool   `json:"enableNodeHardening,omitempty"`
```
comtalyst, line 75: Let's also document/leave a comment on our decision w/ `bootstrappingclient` provision mode. This is to prevent anyone from accidentally relying on this option blindly rather than `ShouldUseNodeHardening`.

(_Ideally_, the name `Enable` should be changed to reflect this nature. But that could be overkill once we drop `bootstrappingclient` provision mode)

pkg/providers/imagefamily/resolver.go
```diff
@@ -238,7 +240,37 @@ func prepareKubeletConfiguration(ctx context.Context, instanceType *cloudprovide
...
+		kubeletConfig.EvictionSoftGracePeriod = map[string]metav1.Duration{
+			instancetype.MemoryAvailable:  {Duration: instancetype.SoftEvictionMemoryGracePeriod},
+			instancetype.NodeFSAvailable:  {Duration: instancetype.SoftEvictionNodeFSGracePeriod},
+			instancetype.NodeFSInodesFree: {Duration: instancetype.SoftEvictionNodeFSInodesGracePeriod},
+		}
+		kubeletConfig.EvictionMaxPodGracePeriod = lo.ToPtr(instancetype.SoftEvictionMaxPodGracePeriodSeconds)
+
+		// Signal node hardening to AgentBaker, which owns the reserved-cgroup paths.
+		// Mirrors nodeAllocatableEnforcementHardened in the AKS RP.
+		kubeletConfig.EnforceNodeAllocatable = []string{"pods", "kube-reserved", "system-reserved"}
```
comtalyst, line 272: AI found the below. If not already, I would also recommend testing setting the option to `true` on self-hosted.

<blockquote>
When hardening is enabled, the PR emits:

```text
--enforce-node-allocatable=pods,kube-reserved,system-reserved
```

See [resolver.go lines 252–273](https://github.com/Azure/karpenter-provider-azure/blob/5c52f3d1f51514497bb092911b9feccad920b0c4/pkg/providers/imagefamily/resolver.go#L252-L273).

However, the scriptless renderer does not emit:

```text
--kube-reserved-cgroup=/kubereserved.slice
--system-reserved-cgroup=/system.slice
```

See [aksbootstrap.go lines 397–418](https://github.com/Azure/karpenter-provider-azure/blob/5c52f3d1f51514497bb092911b9feccad920b0c4/pkg/providers/imagefamily/bootstrap/aksbootstrap.go#L397-L418).

Worse, the new test explicitly requires both flags to remain absent, claiming AgentBaker owns them:

[aksbootstrap_test.go lines 212–226](https://github.com/Azure/karpenter-provider-azure/blob/5c52f3d1f51514497bb092911b9feccad920b0c4/pkg/providers/imagefamily/bootstrap/aksbootstrap_test.go#L212-L226).

That assumption is false for `aksscriptless`:

1. Scriptless directly renders `KUBELET_FLAGS` and calls `provision_start.sh`:  
   [cse_cmd.sh.gtpl lines 127–152](https://github.com/Azure/karpenter-provider-azure/blob/5c52f3d1f51514497bb092911b9feccad920b0c4/pkg/providers/imagefamily/bootstrap/cse_cmd.sh.gtpl#L127-L152)
2. It therefore bypasses AgentBaker’s Go-side cgroup derivation.
3. AgentBaker’s image shell only consumes explicitly supplied cgroup values; it does not infer them from `enforce-node-allocatable`.
4. Kubernetes rejects kube/system reservation enforcement without the corresponding cgroup fields:  
   [Kubernetes 1.30 validation](https://github.com/kubernetes/kubernetes/blob/e2e64ef58da04369882d6187fc6612d1d090d133/pkg/kubelet/apis/config/validation/validation.go#L208-L218) · [Kubernetes 1.35 validation](https://github.com/kubernetes/kubernetes/blob/6637d2842831a59ee736fbd5d0fefa6cfd63a85a/pkg/kubelet/apis/config/validation/validation.go#L258-L267)

**Consequence:** every `ENABLE_NODE_HARDENING=true` launch using `aksscriptless` receives a kubelet configuration that fails startup validation, so the node cannot register.

### Required correction

Either:

- emit the two AgentBaker-compatible reserved-cgroup flags for scriptless, then validate the image/version contract; or
- remove `aksscriptless` from `ShouldUseNodeHardening` until that bootstrap path can produce a complete configuration.

The test must be inverted to require the cgroups rather than their absence. An enabled live `aksscriptless` E2E should then verify node readiness and observed allocatable values.
</blockquote>

pkg/providers/instancetype/instancetype.go
SriHarsha001: ok, made the suggested change.

pkg/operator/options/options.go
SriHarsha001: ok, updated the comments.

pkg/operator/options/options.go
```diff
@@ -108,6 +115,7 @@ func (o *Options) AddFlags(fs *coreoptions.FlagSet) {
 	fs.StringVar(&o.ClusterName, "cluster-name", env.WithDefaultString("CLUSTER_NAME", ""), "[REQUIRED] The kubernetes cluster name for resource tags.")
 	fs.StringVar(&o.ClusterEndpoint, "cluster-endpoint", env.WithDefaultString("CLUSTER_ENDPOINT", ""), "[REQUIRED] The external kubernetes cluster endpoint for new nodes to connect with.")
 	fs.Float64Var(&o.VMMemoryOverheadPercent, "vm-memory-overhead-percent", utils.WithDefaultFloat64("VM_MEMORY_OVERHEAD_PERCENT", 0.075), "The VM memory overhead as a percent that will be subtracted from the total memory for all instance types.")
+	fs.BoolVar(&o.EnableNodeHardening, "enable-node-hardening", env.WithDefaultBool("ENABLE_NODE_HARDENING", false), "Requests AKS node hardening reservations and eviction settings where supported. This option is ignored in bootstrappingclient mode.")
 	fs.StringVar(&o.KubeletClientTLSBootstrapToken, "kubelet-bootstrap-token", env.WithDefaultString("KUBELET_BOOTSTRAP_TOKEN", ""), "[REQUIRED] The bootstrap token for new nodes to join the cluster.")
```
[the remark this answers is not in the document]
SriHarsha001: Thanks, the PR description retained names from an earlier implementation. I updated it to document the implemented --enable-node-hardening flag and ENABLE_NODE_HARDENING environment variable, including the provisioning-mode restriction.

pkg/providers/imagefamily/resolver.go
```diff
@@ -238,7 +240,37 @@ func prepareKubeletConfiguration(ctx context.Context, instanceType *cloudprovide
 	// TODO: revisit computeResources implementation
 	kubeletConfig.KubeReserved = utils.StringMap(instanceType.Overhead.KubeReserved)
 	kubeletConfig.SystemReserved = utils.StringMap(instanceType.Overhead.SystemReserved)
-	kubeletConfig.EvictionHard = map[string]string{instancetype.MemoryAvailable: instanceType.Overhead.EvictionThreshold.Memory().String()}
+	// KubeReservedResources only calculates CPU and memory, instanceType.Overhead.KubeReserved does not contain a PID reservation.
+	kubeletConfig.KubeReserved["pid"] = instancetype.KubeReservedPIDs
+	kubeletConfig.EvictionHard = map[string]string{
+		instancetype.MemoryAvailable:  instanceType.Overhead.EvictionThreshold.Memory().String(),
+		instancetype.NodeFSAvailable:  instancetype.HardEvictionNodeFSAvailable,
```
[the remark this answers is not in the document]
SriHarsha001: These hard-eviction thresholds are already present on existing non-hardened AKS nodes (memory.available<100Mi,nodefs.available<10%,nodefs.inodesFree<5%,pid.available<2000), so the nodefs and PID thresholds are baseline kubelet settings rather than hardening-only behavior. We intentionally keep them outside ShouldUseNodeHardening().

pkg/providers/imagefamily/resolver.go
SriHarsha001: Manually tested the code changes with enable node hardening set to true -  

k get nodes
NAME                                STATUS   ROLES    AGE   VERSION
aks-general-purpose-fvwfg           Ready    <none>   74s   v1.35.7
aks-nodepool1-36669291-vmss000000   Ready    <none>   29m   v1.35.7
aks-nodepool1-36669291-vmss000001   Ready    <none>   29m   v1.35.7
aks-nodepool1-36669291-vmss000002   Ready    <none>   29m   v1.35.7

root@aks-general-purpose-fvwfg:/# cat /etc/default/kubelet
KUBELET_FLAGS=--address=0.0.0.0 --anonymous-auth=false --authentication-token-webhook=true --authorization-mode=Webhook --cgroups-per-qos=true --client-ca-file=/etc/kubernetes/certs/ca.crt --cloud-provider=external --cluster-dns=10.0.0.10 --cluster-domain=cluster.local **--enforce-node-allocatable=pods,kube-reserved,system-reserved --event-qps=0 --eviction-hard=nodefs.available<10%,nodefs.inodesFree<5%,pid.available<2000,memory.available<250Mi --eviction-max-pod-grace-period=60 --eviction-soft=nodefs.available<12%,nodefs.inodesFree<7%,memory.available<500Mi --eviction-soft-grace-period=memory.available=30s,nodefs.available=2m0s,nodefs.inodesFree=2m0s** --image-credential-provider-bin-dir=/var/lib/kubelet/credential-provider --image-credential-provider-config=/var/lib/kubelet/credential-provider-config.yaml --image-gc-high-threshold=85 --image-gc-low-threshold=80 --kube-reserved=pid=1000,cpu=100m,memory=1Gi **--kube-reserved-cgroup=/kubereserved.slice** --kubeconfig=/var/lib/kubelet/kubeconfig --max-pods=250 --node-status-update-frequency=10s --pod-manifest-path=/etc/kubernetes/manifests --pod-max-pids=-1 --protect-kernel-defaults=true --read-only-port=0 --register-with-taints=node.cilium.io/agent-not-ready=true:NoExecute,karpenter.sh/unregistered:NoExecute --resolv-conf=/run/systemd/resolve/resolv.conf --rotate-certificates=true --streaming-connection-idle-timeout=4h --system-reserved=memory=300Mi,ephemeral-storage=1Gi,pid=1000,cpu=100m **--system-reserved-cgroup=/system.slice** --tls-cert-file=/etc/kubernetes/certs/kubeletserver.crt --tls-cipher-suites=TLS_ECDHE_ECDSA_WI

comtalyst (approved):

pkg/providers/imagefamily/bootstrap/aksbootstrap.go
```diff
@@ -405,6 +414,14 @@ func kubeletConfigToMap(kubeletConfig *KubeletConfiguration) map[string]string {
 		return v.Duration.String()
 	}), "=")
 
+	if len(kubeletConfig.EnforceNodeAllocatable) > 0 {
+		args["--enforce-node-allocatable"] = strings.Join(kubeletConfig.EnforceNodeAllocatable, ",")
+		if lo.Contains(kubeletConfig.EnforceNodeAllocatable, "kube-reserved") && lo.Contains(kubeletConfig.EnforceNodeAllocatable, "system-reserved") {
+			args["--kube-reserved-cgroup"] = nodeHardeningKubeReservedCgroup
+			args["--system-reserved-cgroup"] = nodeHardeningSystemReservedCgroup
```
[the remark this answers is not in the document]
SriHarsha001: Thanks for pointing this out. JoinParameterArgsToMap does retain the existing nondeterministic Go map iteration order, but that behavior predates this change. The rendered custom data is used only during VM creation and is neither part of the NodeClass drift hash nor compared during reconciliation, so ordering differences cannot cause drift or node churn. The new reserved-cgroup flags are scalar values and are already covered by the top-level flag-name sorting. I’d prefer to handle deterministic serialization of nested map values as a separate cleanup since it is not required for node hardening.

Issue #1439, opened 2026-02-18
Support SecurityPatch channel in managed NAP


matthchr:
This is not currently supported but should be.

Supporting it requires pre-requisite work to move to the machine API, which is ongoing as of Feb 2026.

matthchr:
Assigning this to myself primarily so I don't lose track of it

matthchr:
@comtalyst and I met with @yewmsft and determined the following:

**Step 1:** Move Karpenter to use the new SecurityPatchOnly header for ListNodeImageVersions:
```
az rest --method get \
    --headers SecurityPatchOnly=true \
    --uri /subscriptions/.../providers/Microsoft.ContainerService/locations/westus2/nodeimageversions?api-version=2026-03-02-preview
```

We must call this API when the cluster has opted into SecurityPatch for NodeImage updates. Note that in this mode, images should _always_ come from the SecurityPatchOnly channel. If the user was previously on standard channel and moves to SecurityPatch channel, their nodes will be rolled.

This should be relatively easy to do and will improve the CVE patch speed compared to standard images.

**Step 2:** Use Machine API + drift field (or new field, since semantics are slightly different than drift) to indicate "action required for live patching". Then Karpetner can PUT/POST to the Machine API with an updated image which will trigger live-patching. The signal in the Machine API will need to indicate which type of patching is required - it should not say live patching but then actually require a drain - if it requires a drain a different signal should be sent (Karpenter can just ignore this signal and do its normal thing of periodically calling the ListNodeImageVersions API and rolling nodes to get to the new one).

One other wart that came up for discussion with @rakechill is that the security patch VHDs come from a _different SIG_ than the standard VHDs. and the version strings themselves may look slightly different. We need to cater for this in the image pinning/decouple image update workflows, and answer the question what happens for pinning on securitypatch channel. Do we allow it? How does it work?

matthchr:
Plan is to just do Step 1 initially

PR #1821, opened 2026-07-29
feat: support the SecurityPatch node OS upgrade channel on the AKS Machine API path

MaximilianoUribe:
## Description

Adds support for the existing AKS `SecurityPatch` node OS upgrade channel in managed NAP (part of #1439).

When a cluster uses the `SecurityPatch` channel, Karpenter selects security-patched AKS node images instead of the standard node-image lineage, improving CVE patch speed.

## Design

This implementation mirrors the existing AKS cluster-level node OS upgrade policy:

```yaml
properties:
  autoUpgradeProfile:
    nodeOSUpgradeChannel: SecurityPatch
```

`SecurityPatch` remains a **cluster-level setting** rather than a per-`AKSNodeClass` image selector.

Managed NAP passes the configured channel to Karpenter through `NODE_OS_UPGRADE_CHANNEL`. Karpenter uses it to select the appropriate node-image lineage:

| Channel | Image lineage |
|---|---|
| `SecurityPatch` | Security-patched AKS node images |
| `NodeImage` | Standard AKS node images |

`AKSNodeClass` continues to select properties within that lineage, including image family, FIPS mode, architecture, and generation.

Future version-selection controls can select or pin a version within the active lineage. Supporting different upgrade channels per NodeClass would be a separate API and product decision.

## Behavior

When the cluster uses `SecurityPatch`, Karpenter:

1. Requests SecurityPatch node-image versions from AKS by setting `SecurityPatchOnly: true` and `CapturedImagesOnly: true` on `ListNodeImageVersions`. This requests physical images suitable for new nodes while leaving the full catalog available to other consumers.
2. Publishes the resolved versions into `AKSNodeClass.status.images`.
3. Uses those versions when creating new nodes through the AKS Machine API.
4. Rolls existing nodes through normal Karpenter image drift and disruption behavior when the effective image version changes.

In-place live patching of existing Karpenter nodes is not included in this change.

Changing between `NodeImage` and `SecurityPatch` refreshes the in-memory image cache and persisted `AKSNodeClass.status.images`, so versions from the previous lineage are not reused.

## Configuration validation

`SecurityPatch` is supported only for managed NAP configurations that use:

- AKS-managed image resolution (`UseSIG=true`).
- An AKS Machine API provision mode.

Unsupported combinations fail validation instead of silently using another image lineage.

Only the AKS-managed `SecurityPatch` and `NodeImage` values are accepted. `None` and `Unmanaged` are not surfaced to managed NAP image selection.

## Why the AKS Machine API is required

SecurityPatch versions are logical node-image versions, for example:

```text
AKSUbuntu-2404gen2containerd-202606.08.1-2026.06.13
```

With the AKS Machine API, Karpenter sends the logical version to AKS when creating a node. Direct-VM modes construct image references locally and do not support this SecurityPatch image-resolution flow.

## Version ordering

SecurityPatch versions use a composite format:

```text
<baseVersion>-<securityPatchDate>
```

They are ordered by security patch date first, with base version as a tiebreaker:

```text
202602.19.0-2026.03.02
    is newer than
202602.20.0-2026.03.01
```

The newer patch date carries newer security fixes even when its base image is older.

## Testing

- Option parsing and validation:
  - Valid SecurityPatch + managed image resolution + Machine API configuration.
  - Rejection without managed image resolution.
  - Rejection on direct-VM modes.
  - NodeImage and invalid channel behavior.
- Request policy:
  - Both `SecurityPatchOnly: true` and `CapturedImagesOnly: true` are sent by the SecurityPatch image-list policy; neither is sent by the default path.
- Image resolution:
  - Patch-date-first version ordering.
  - Correct latest SecurityPatch selection.
- Catalog transitions:
  - In-memory cache keys distinguish SecurityPatch and NodeImage.
  - Persisted NodeClass images refresh when the configured lineage changes.
- Full CI passes for Kubernetes 1.30 through 1.36.
- End-to-end validation confirmed that managed NAP nodes become Ready using the SecurityPatch image lineage.

(AI-generated) Validation: three consecutive manual managed-NAP runs passed. The opt-in returned only captured images; legacy SecurityPatch-only requests retained abstract entries. Captured-image nodes became Ready with Running workloads and the exact expected physical image; abstract-only NodeClasses reported ImagesNotFound with no NodeClaims during the observation window. Requires service-side support before controller rollout. Existing cached-image rollout and in-place live patching were not tested.

## Documentation

- [ ] Yes, issue opened: #
- [x] No

## Release Note

```release-note
Managed NAP honors the AKS SecurityPatch node OS upgrade channel when using the AKS Machine API.
```


pkg/operator/options/options.go
```diff
@@ -82,7 +82,17 @@ type Options struct {
 	SubnetID                string   `json:"subnetId,omitempty"`                // => VnetSubnetID to use (for nodes in Azure CNI Overlay and Azure CNI + pod subnet; for for nodes and pods in Azure CNI), unless overridden via AKSNodeClass
 	setFlags                map[string]bool
 
-	ProvisionMode              string            `json:"provisionMode,omitempty"`
+	ProvisionMode string `json:"provisionMode,omitempty"`
+	// NodeOSUpgradeChannel mirrors the managed cluster's autoUpgradeProfile.nodeOSUpgradeChannel.
```
matthchr, line 86: minor: this documentation is good but IMO surface it here or in pkg/consts/consts.go not both?

pkg/operator/options/options_validation.go
```diff
@@ -142,6 +143,17 @@ func (o *Options) validateProvisionMode() error {
 	return nil
 }
 
+func (o *Options) validateNodeOSUpgradeChannel() error {
```
matthchr, line 146: Can we not validate `IsAKSMachineAPIMode` and `UseSIG` here too, and just reject attempts to set security patch when those aren't configured?

pkg/operator/options/options.go
```diff
@@ -147,6 +158,26 @@ func (o *Options) IsAKSMachineAPIMode() bool {
...
+//   - an AKS Machine API provision mode.
+//
+// The provision mode requirement is not incidental. On the AKS Machine API modes Karpenter sends the
+// logical node image version (e.g. "AKSUbuntu-2404gen2containerd-202606.08.1-2026.06.13") and the
+// service resolves it to the physical security-patch image. On the direct-VM modes Karpenter builds
+// the image ID itself from the standard node image gallery, which cannot address the security-patch
+// gallery (different gallery, image definition and physical version encoding), so requesting
+// security-patch images there yields unprovisionable image references.
+func (o *Options) IsSecurityPatchChannel() bool {
+	return o.NodeOSUpgradeChannel == consts.NodeOSUpgradeChannelSecurityPatch &&
```
matthchr: If we do https://github.com/Azure/karpenter-provider-azure/pull/1821/changes#r3730714327, then in theory you can just do `o.NodeOSUpgradeChannel == consts.NodeOSUpgradeChannelSecurityPatch` because the other two conditions checked here would not be possible

pkg/operator/options/suite_test.go
```diff
@@ -447,6 +447,32 @@ var _ = Describe("Options", func() {
...
+				"--cluster-endpoint", "https://karpenter-000000000000.hcp.westus2.staging.azmk8s.io",
+				"--kubelet-bootstrap-token", "flag-bootstrap-token",
+				"--ssh-public-key", "flag-ssh-public-key",
+				"--vnet-subnet-id", "/subscriptions/12345678-1234-1234-1234-123456789012/resourceGroups/sillygeese/providers/Microsoft.Network/virtualNetworks/karpentervnet/subnets/karpentersub",
+				"--node-resource-group", "my-node-rg",
+				"--node-os-upgrade-channel", "NotAChannel",
+			)
+			Expect(err).To(MatchError(ContainSubstring("node-os-upgrade-channel is invalid")))
+		})
+		It("should fail validation when NodeOSUpgradeChannel is None (not surfaced to Karpenter)", func() {
```
matthchr: minor: There's in most practical ways no difference between these two tests. Maybe just have this one and remove the other one?

pkg/operator/options/suite_test.go
```diff
@@ -821,6 +847,82 @@ var _ = Describe("Options", func() {
...
+				"--aks-machines-pool-name", "testmpool",
+				"--use-sig",
+				"--sig-subscription-id", "92345678-1234-1234-1234-123456789012",
+				"--node-os-upgrade-channel", "SecurityPatch",
+			)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(opts.IsSecurityPatchChannel()).To(BeTrue())
+		})
+
+		It("should NOT enable security-patch images on a non AKS machine API mode", func() {
```
matthchr: minor: If you do https://github.com/Azure/karpenter-provider-azure/pull/1821/changes#r3730714327, I think you can remove some of these tests as they don't apply anymore

pkg/providers/azclient/securitypatchheader.go
```diff
@@ -0,0 +1,42 @@
...
+const securityPatchOnlyHeader = "SecurityPatchOnly"
+
+var _ policy.Policy = &securityPatchOnlyPolicy{}
+
+// securityPatchOnlyPolicy is an Azure SDK per-call policy that sets the SecurityPatchOnly header on
+// ListNodeImageVersions requests. When set, the service returns node image versions from the
+// security-patch lineage instead of the standard node images, so clusters on the SecurityPatch node
+// OS upgrade channel receive security-patched images.
+//
+// The policy is only attached when security-patch images should be used (see
```
matthchr: minor: Maybe remove this section about how it is used? User can inspect code and better to describe what something does on its comments versus how it is used (since usage is flexible)

pkg/providers/azclient/securitypatchheader_test.go
```diff
@@ -0,0 +1,74 @@
...
+
+package azclient
+
+import (
+	"net/http"
+	"testing"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
+	"github.com/stretchr/testify/assert"
```
matthchr: please don't use testify, use gomega for assertions

pkg/providers/azclient/azclient.go
```diff
@@ -186,7 +186,14 @@ func NewAZClient(ctx context.Context, cfg *auth.Config, env *auth.Environment, c
 		return nil, err
 	}
 
-	nodeImageVersionsClient, err := imagefamily.NewNodeImageVersionsClient(cfg.SubscriptionID, cred, opts)
+	// copy the options to avoid modifying the original, so the SecurityPatchOnly header policy is
+	// only applied to the node image versions client.
+	var nodeImageVersionsClientOptions = *opts
+	if o.IsSecurityPatchChannel() {
```
matthchr, line 195: **Bot-generated review comment:** Could we ensure that changing `NODE_OS_UPGRADE_CHANNEL` or `PROVISION_MODE` invalidates `AKSNodeClass.Status.Images`? This gate is evaluated when constructing the client, but the selected image IDs persist in status. If configuration rolls from Machine API/`SecurityPatch` back to a direct-VM mode, existing composite versions may be retained and then used as literal gallery references—the failure this gate is intended to prevent.

pkg/operator/options/suite_test.go
```diff
@@ -447,6 +447,32 @@ var _ = Describe("Options", func() {
 			)
 			Expect(err).To(MatchError(ContainSubstring("invalid")))
 		})
+		It("should fail validation when NodeOSUpgradeChannel is not valid", func() {
```
matthchr, line 451: **Bot-generated review comment:** Please add `NODE_OS_UPGRADE_CHANNEL` to the `environmentVariables` cleanup list above. These new option tests depend on it being unset; an ambient value can otherwise leak into parsing and make the negative cases dependent on the CI environment or test order.

pkg/providers/imagefamily/nodeimageversionsclient_test.go
```diff
@@ -47,3 +64,55 @@ func TestIsNewerVersion(t *testing.T) {
...
+		img("2404gen2containerd", "202606.08.1-2026.06.13"), // newest
+		img("2404gen2containerd", "202605.14.0-2026.05.24"),
+	}
+
+	filtered := FilteredNodeImages(input)
+	if len(filtered) != 1 {
+		t.Fatalf("expected 1 image after filtering, got %d", len(filtered))
+	}
+	if got := lo.FromPtr(filtered[0].Version); got != "202606.08.1-2026.06.13" {
+		t.Errorf("FilteredNodeImages selected version %q; want the newest %q", got, "202606.08.1-2026.06.13")
```
matthchr: minor: Use gomega for assertions

matthchr: and below

pkg/operator/options/options.go
MaximilianoUribe: (AI-generated) Done. Kept the detailed behavior and constraints on `Options.NodeOSUpgradeChannel`, where operators encounter the setting, and reduced the constants comment to one line.

pkg/operator/options/options_validation.go
MaximilianoUribe: (AI-generated) Done. `SecurityPatch` now fails option validation unless `UseSIG=true` and the provision mode is `aksmachineapi` or `aksmachineapiheaderbatch`. Unsupported combinations no longer silently act like standard images.

pkg/operator/options/options.go
MaximilianoUribe: (AI-generated) Done. Since validation now guarantees SIG + Machine API for `SecurityPatch`, `IsSecurityPatchChannel()` is simplified to the channel comparison only.

pkg/operator/options/suite_test.go
MaximilianoUribe: (AI-generated) Done. Removed the duplicate `None` validation case; the remaining invalid-channel case covers the same validation branch.

pkg/operator/options/suite_test.go
MaximilianoUribe: (AI-generated) Done. Removed the runtime “not enabled” cases for unsupported provision mode / missing SIG and replaced them with startup-validation failures. Kept the positive SecurityPatch case and the distinct NodeImage case.

pkg/providers/azclient/securitypatchheader.go
MaximilianoUribe: (AI-generated) Done. The policy comment now only describes what it does: sets `SecurityPatchOnly` so the service returns the security-patch lineage.

pkg/providers/azclient/securitypatchheader_test.go
MaximilianoUribe: (AI-generated) Done. Replaced Testify with Gomega in `securitypatchheader_test.go`.

pkg/providers/azclient/azclient.go
MaximilianoUribe: (AI-generated) Addressed both caches involved in a configuration transition. Startup validation rejects `SecurityPatch` with direct-VM or without SIG. For valid `SecurityPatch <-> NodeImage` changes, the in-memory image cache key now includes the active catalog, and the status reconciler detects when persisted `AKSNodeClass.Status.Images` use composite SecurityPatch versions that do not match the configured catalog (or vice versa), forcing a full refresh even outside a maintenance window. Added pure catalog-shape/cache-key tests and an envtest covering SecurityPatch status refreshing back to NodeImage.

pkg/operator/options/suite_test.go
MaximilianoUribe: (AI-generated) Done. Added `NODE_OS_UPGRADE_CHANNEL` to the environment cleanup list so ambient values cannot leak between option tests.

pkg/providers/imagefamily/nodeimageversionsclient_test.go
MaximilianoUribe: (AI-generated) Done. Converted the new node-image-version tests to Gomega, including `isNewerVersion` and filtered-image assertions.

MaximilianoUribe: (AI-generated) Done below as well; all new assertions in this file now use Gomega.

pkg/controllers/nodeclass/status/images_test.go
```diff
@@ -218,6 +218,36 @@ var _ = Describe("NodeClass NodeImage Status Controller", func() {
 			})
 		})
 
+		Context("image catalog transitions", func() {
+			It("refreshes persisted SecurityPatch images when switching to NodeImage", func() {
```
matthchr, line 223: Shouldn't you test the other direction too?

pkg/controllers/nodeclass/status/images_test.go
```diff
@@ -218,6 +218,36 @@ var _ = Describe("NodeClass NodeImage Status Controller", func() {
...
+				old, had := os.LookupEnv("SYSTEM_NAMESPACE")
+				os.Setenv("SYSTEM_NAMESPACE", "kube-system")
+				DeferCleanup(func() {
```
matthchr: Do you actually need to set SYSTEM_NAMESPACE here at all? I observe that many other tests in this folder (such as the FIPS image tests) do not.

As far as I can tell, the only thing that variable does is enable MW checks.

Since you are not writing/configuring MWs here, it doesn't matter and you can probably remove this SYSTEM_NAMESPACE stuff?

pkg/controllers/nodeclass/status/images.go
```diff
@@ -186,6 +186,53 @@ func imageVersionsUnready(nodeClass *v1beta1.AKSNodeClass) bool {
...
+		return false
+	}
+	base, patchDate, found := strings.Cut(parts[len(parts)-1], "-")
+	if !found {
+		return false
+	}
+	return isNumericDotVersion(base) && isNumericDotVersion(patchDate)
+}
+
+func isNumericDotVersion(version string) bool {
```
matthchr, line 221: minor: maybe `// comment` add to this method and isSecurityPatchImageID?

pkg/controllers/nodeclass/status/images.go
```diff
@@ -186,6 +186,53 @@ func imageVersionsUnready(nodeClass *v1beta1.AKSNodeClass) bool {
...
+// imageCatalogChanged reports whether persisted images came from a different node image catalog
+// than the one configured for this controller process. AKSNodeClass status survives a controller
+// restart, so a valid SecurityPatch <-> NodeImage channel transition must force a full refresh even
+// when the node OS maintenance window is closed. Security-patch versions are composite
+// "<baseVersion>-<securityPatchDate>" values; standard node image versions have no patch-date suffix.
+func imageCatalogChanged(ctx context.Context, images []v1beta1.NodeImage) bool {
+	if len(images) == 0 {
+		return false
+	}
+	wantSecurityPatch := options.FromContext(ctx).IsSecurityPatchChannel()
```
matthchr: minor, style: The only thing you use ctx for here is to pass `IsSecurityPatchChannel`. I think it would be clearer if you just made that the parameter and extracted it at the callsite.

matthchr: I think it would make your tests slightly clearer as well.

pkg/controllers/nodeclass/status/images_catalog_test.go
```diff
@@ -0,0 +1,63 @@
...
+
+import (
+	"testing"
+
+	"github.com/Azure/karpenter-provider-azure/pkg/apis/v1beta1"
+	"github.com/Azure/karpenter-provider-azure/pkg/operator/options"
+	. "github.com/onsi/gomega"
+)
+
+func TestImageCatalogChanged(t *testing.T) {
```
matthchr: minor, style: Just put this in images_test.go.

I think _generally_ we try to have our test files match our impl files, at least for unit tests.

matthchr: IMO it's OK to do these as just standard Go tests at the end of that file.

pkg/providers/azclient/azclient.go
```diff
@@ -186,7 +186,16 @@ func NewAZClient(ctx context.Context, cfg *auth.Config, env *auth.Environment, c
 		return nil, err
 	}
 
-	nodeImageVersionsClient, err := imagefamily.NewNodeImageVersionsClient(cfg.SubscriptionID, cred, opts)
+	// copy the options to avoid modifying the original, so the SecurityPatchOnly header policy is
+	// only applied to the node image versions client.
+	var nodeImageVersionsClientOptions = *opts
+	// Clone the slice to ensure appends here don't mutate opts.PerCallPolicies via a shared backing array.
+	nodeImageVersionsClientOptions.PerCallPolicies = append(nodeImageVersionsClientOptions.PerCallPolicies[:0:0], nodeImageVersionsClientOptions.PerCallPolicies...)
```
matthchr: You could just use `slices.Clone(original)` here, I think?

Or `append([]T(nil), original...)`?

Both are probably cleaner than this?

matthchr: but yeah probably 
```suggestion
	nodeImageVersionsClientOptions.PerCallPolicies = slices.Clone(nodeImageVersionsClientOptions.PerCallPolicies)
```

is clearest

pkg/providers/azclient/azclient.go
```diff
@@ -186,7 +186,16 @@ func NewAZClient(ctx context.Context, cfg *auth.Config, env *auth.Environment, c
...
+	if o.IsSecurityPatchChannel() {
+		log.FromContext(ctx).Info("cluster is on the SecurityPatch node OS upgrade channel; requesting security-patch node images")
```
matthchr:
```suggestion
		log.FromContext(ctx).Info("cluster is on the SecurityPatch node OS upgrade channel; configuring client to request security-patch node images")
```

pkg/providers/imagefamily/nodeimage_cache_test.go
```diff
@@ -0,0 +1,35 @@
...
+
+package imagefamily
+
+import (
+	"testing"
+
+	. "github.com/onsi/gomega"
+)
+
+func TestNodeImageCacheKeyIncludesSecurityPatchCatalog(t *testing.T) {
```
matthchr: minor: in nodeimage_test instead? (At the bottom, non-ginko is fine?)

pkg/test/options.go
```diff
@@ -84,6 +85,7 @@ func Options(overrides ...OptionsFields) *azoptions.Options {
 		SubnetID:                       lo.FromPtrOr(options.SubnetID, "/subscriptions/12345678-1234-1234-1234-123456789012/resourceGroups/test-resourceGroup/providers/Microsoft.Network/virtualNetworks/aks-vnet-12345678/subnets/aks-subnet"),
 		NodeResourceGroup:              lo.FromPtrOr(options.NodeResourceGroup, "test-resourceGroup"),
 		ProvisionMode:                  lo.FromPtrOr(options.ProvisionMode, "aksscriptless"),
+		NodeOSUpgradeChannel:           lo.FromPtrOr(options.NodeOSUpgradeChannel, ""),
```
matthchr, line 88: Is the default in RP going to be empty string or `NodeImage`?

pkg/controllers/nodeclass/status/images_test.go
MaximilianoUribe: i think this was one of the copilot comments that did not make sense, i think i commited it by mistake :/

pkg/controllers/nodeclass/status/images_test.go
MaximilianoUribe: (AI-generated) Done. Both directions are now covered through the public reconciler behavior: SecurityPatch -> NodeImage verifies composite versions are removed, and NodeImage -> SecurityPatch supplies composite fake API results and verifies those versions become effective.

pkg/controllers/nodeclass/status/images_test.go
MaximilianoUribe: (AI-generated) Done. Removed SYSTEM_NAMESPACE and the closed maintenance-window ConfigMap setup as requested. The catalog transition tests now exercise the refresh directly without involving MW behavior.

pkg/controllers/nodeclass/status/images.go
MaximilianoUribe: (AI-generated) Done. Added concise doc comments to `isSecurityPatchImageID` and `isNumericDotVersion`; `imageCatalogChanged` already had one.

pkg/controllers/nodeclass/status/images.go
MaximilianoUribe: (AI-generated) Done. `imageCatalogChanged` now accepts `wantSecurityPatch bool`; the caller extracts it from options. This also makes the helper contract clearer.

pkg/controllers/nodeclass/status/images_catalog_test.go
MaximilianoUribe: (AI-generated) Done. Removed the separate private-helper test file and consolidated coverage into `images_test.go` through public reconciler behavior.

pkg/providers/azclient/azclient.go
MaximilianoUribe: (AI-generated) Done. Replaced the full-slice append idiom with `slices.Clone(nodeImageVersionsClientOptions.PerCallPolicies)`.

pkg/providers/azclient/azclient.go
MaximilianoUribe: (AI-generated) Done. Applied the suggested log wording: “configuring client to request security-patch node images”.

pkg/providers/imagefamily/nodeimage_cache_test.go
MaximilianoUribe: (AI-generated) Done. Removed the separate cache test file and moved coverage into `nodeimage_test.go`. The test now observes that changing from NodeImage to SecurityPatch bypasses the previous cached result and returns composite SecurityPatch versions.

pkg/test/options.go
MaximilianoUribe: (AI-generated) RP surfaces `NodeImage` when the managed cluster channel is explicitly/defaulted to NodeImage, `SecurityPatch` when configured, and empty for None/Unmanaged/unspecified. I kept the generic test helper default empty because that is the neutral/unset contract; tests requiring NodeImage or SecurityPatch now set it explicitly.

matthchr (dismissed):
LGTM. I am not sure we strictly speaking need an E2E test pass for this PR as most of what it changes is behind an option that is not configured by default (so E2E is just for regression which we also have UTs covering).

Please consider if you ought to run E2E before merging this - if you don't, please ensure you check E2E after you merge and that it passes.

matthchr (commented):
resetting review actually because we are waiting on Machine API. I am good to approve this exactly as-is but can't merge yet because blocked on Machine API it sounds like.

If we want this to work for Bootstrapping client too we'd need to at the very least pass in an alternative SIG on the RP side when the mode changes (either in a separate variable or the same one)

PR #1823, opened 2026-07-31
feat: Quality of life Makefile improvements

theunrepentantgeek:
**Description**

Adds breadcrumbs to all targets in `Makefile-az.mk` to make it obvious which target is executing when a failure occurs.

Also fixes some minor bugs:

* `NAME_SUFFIX` was incorrectly formed if the git email address configured for the user contained a period (`.`)
* Cluster namespace creation would fail if the namespace already existed.
* Helm installs would fail if the chart was already installed.
* Grafana v1.5 dashboards are no longer available (404 errors)

Local execution.

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
NONE
```


Makefile-az.mk
```diff
@@ -58,6 +59,7 @@ az-all-cniv1:        az-login az-create-workload-msi az-mkaks-cniv1
 
 az-all-cni-overlay:  az-login az-create-workload-msi az-mkaks-overlay             az-create-federated-cred $(AZ_ALL_PERMS) az-configure-values az-build az-run az-run-sample ## Provision the infra (ACR,AKS); build and deploy Karpenter; deploy sample Provisioner and workload
 az-all-perftest:     az-login az-create-workload-msi az-mkaks-perftest            az-create-federated-cred $(AZ_ALL_PERMS) az-configure-values
+	@echo "----- az-all-perftest -----"
```
tallaxes: Not sure I like this everywhere (and not in the other Makefile), and feels like something like `make --debug` would do the job when needed, but if we add it could probably be

```suggestion
	@echo "----- $@ -----"
```

There is probably a way to inject the echo automatically ...

tallaxes: AI came up with this (probably too convoluted):

```Makefile
PHONY_TARGETS := az-login az-build # list of targets to announce
.PHONY: $(PHONY_TARGETS) FORCE
$(PHONY_TARGETS): %: | .announce-%
.announce-%: FORCE
	@printf '----- %s -----\n' '$*'
FORCE:
```

For az-build, the static pattern rule effectively adds:

```Makefile
az-build: | .announce-az-build
```

Then:

* `|` makes `.announce-az-build` an order-only prerequisite.
* `.announce-%` matches it, with `$*` equal to `az-build`.
* `FORCE` is phony, ensuring the announcement always runs.
* The announcement prints before `az-build`’s existing recipe.
* The rule adds a prerequisite without replacing the existing recipe.

`$@` inside the announcement would be `.announce-az-build`, so `$*` is used to print the original target name.

theunrepentantgeek: I took that idea and came up with a simple(ish) wayh to automagically show breadcrumbs for all targets.

cmd/controller/kodata/HEAD
theunrepentantgeek, line 1: This link (and the following one) are unused, don't work at all in a git worktree, and pull in a lot of noise (such as details of all known branches, etc).

hack/monitoring/grafana-values.yaml
```diff
@@ -34,9 +34,9 @@ dashboardProviders:
 dashboards:
   default:
     capacity-dashboard:
-      url: https://karpenter.sh/v1.5/getting-started/getting-started-with-karpenter/karpenter-capacity-dashboard.json
```
theunrepentantgeek, line 37: The v1.5 dashboards are no longer available; v1.14 is the latest version.

Makefile
```diff
@@ -161,3 +161,20 @@ define newline
 
 
 endef
+
+# Announce all explicit targets as they run to allow easier tracing of issues
+# NOTE: This block must be at the very end of the entrypoint Makefile so it's parsed last
+
+ANNOUNCED_TARGETS := $(filter-out FORCE .PHONY,$(shell awk -F: \
+    '/^[a-zA-Z0-9_.\/-]+([[:space:]]+[a-zA-Z0-9_.\/-]+)*[[:space:]]*:/ \
+    { print $$1 }' $(MAKEFILE_LIST)))
```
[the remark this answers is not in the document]
theunrepentantgeek: The regex covers all the targets we're interested in.

Makefile-az.mk
```diff
@@ -11,7 +11,8 @@ ifeq ($(CODESPACES),true)
   AZURE_RESOURCE_GROUP ?= $(CODESPACE_NAME)
   AZURE_ACR_NAME ?= $(subst -,,$(CODESPACE_NAME))
 else
-  NAME_SUFFIX ?= $(shell git config user.email | cut -d'@' -f1 | tr -d '+')
+  # generate a unique suffix based on the user's email
+  NAME_SUFFIX ?= $(shell git config user.email | cut -d'@' -f1 | tr -cd '[:alnum:]')
```
theunrepentantgeek, line 15: The old transformation assumed email addresses never have an embedded period (`.`) - the new one works by rejecting everything except alphanumerics.

tallaxes (approved):

Makefile-az.mk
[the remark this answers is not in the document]
theunrepentantgeek: Not an issue with the current scale of the team.

PR #1824, opened 2026-07-31
feat: allow Node Hardening kubelet config overrides via CRD

mxj220:
Fixes #<issue if any>

## Description

Follow-up to #1818. Adds five customer-configurable overrides to
`AKSNodeClass.spec.kubelet` so users with node hardening enabled can tune kubelet reservations and eviction thresholds like they can using a standard AKS cluster:

- `kubeReserved.cpuMillicores` and `kubeReserved.memoryMB`
- `evictionHard.memoryAvailable`, `evictionHard.nodeFsAvailable`, and `evictionHard.nodeFsInodesFree`
- `evictionSoft.memoryAvailable`, `evictionSoft.nodeFsAvailable`, and `evictionSoft.nodeFsInodesFree`
- `evictionSoftGracePeriod.memoryAvailable`, `evictionSoftGracePeriod.nodeFsAvailable`, and `evictionSoftGracePeriod.nodeFsInodesFree`
- `evictionMaxPodGracePeriod`

- All existing unit tests succeed
- Added unit tests
- Added 1 new E2E
  
- [ ] Yes, issue opened: # 
- [X] No


pkg/providers/instance/aksmachineinstancehelpers.go
```diff
@@ -431,7 +437,82 @@ func configureKubeletConfig(nodeClass *v1beta1.AKSNodeClass) *armcontainerservic
...
+	unsupportedFields := []string{}
+	if len(nodeClass.Spec.Kubelet.EvictionSoft) > 0 {
+		unsupportedFields = append(unsupportedFields, "evictionSoft")
+	}
+	if len(nodeClass.Spec.Kubelet.EvictionSoftGracePeriod) > 0 {
+		unsupportedFields = append(unsupportedFields, "evictionSoftGracePeriod")
+	}
+	if nodeClass.Spec.Kubelet.EvictionMaxPodGracePeriod != nil {
+		unsupportedFields = append(unsupportedFields, "evictionMaxPodGracePeriod")
+	}
```
comtalyst: What is the reasoning of making these ones available on the API, but not support it? And doesn't look like these are being used in `instancetype`?

pkg/providers/instance/aksmachineinstancehelpers.go
```diff
@@ -431,7 +437,82 @@ func configureKubeletConfig(nodeClass *v1beta1.AKSNodeClass) *armcontainerservic
...
+	if len(unsupportedFields) > 0 {
+		return nil, fmt.Errorf("kubelet overrides %s are not supported by the AKS Machine API; supported override fields are kubeReserved and evictionHard", strings.Join(unsupportedFields, ", "))
```
comtalyst: We should avoid using provisioning time error for expected failure case unless there are meaningful handling logic (e.g., quota unavailability, where Karpenter mark certain config as unavailable and retry differently). Otherwise it will keep retrying without progress. 
In this case, that is not possible given these are user-configured fields. It should failed during CRD-level validation instead (which kills the point of supporting them in the API in the first place?).

pkg/providers/imagefamily/resolver.go
```diff
@@ -271,9 +271,44 @@ func prepareKubeletConfiguration(ctx context.Context, instanceType *cloudprovide
...
+			kubeletConfig.EvictionSoftGracePeriod = map[string]metav1.Duration{}
+		}
+		for key, value := range overrides.EvictionSoftGracePeriod {
+			kubeletConfig.EvictionSoftGracePeriod[key] = value
+		}
+	}
+	if overrides.EvictionMaxPodGracePeriod != nil {
+		kubeletConfig.EvictionMaxPodGracePeriod = lo.ToPtr(*overrides.EvictionMaxPodGracePeriod)
+	}
+}
```
comtalyst, line 297: Karpenter is not going to supporting things in Scriptless mode that is unsupported in AKS/Machine path.
Either we work on Machine path to support those fields first, or they should be dropped.

pkg/providers/instance/aksmachineinstancehelpers.go
```diff
@@ -431,7 +437,82 @@ func configureKubeletConfig(nodeClass *v1beta1.AKSNodeClass) *armcontainerservic
...
+		case string(v1.ResourceMemory):
+			bytes := quantity.Value()
+			const bytesPerMiB = int64(1024 * 1024)
+			memoryMiB := bytes / bytesPerMiB
+			if bytes%bytesPerMiB != 0 || memoryMiB < math.MinInt32 || memoryMiB > math.MaxInt32 {
+				return nil, fmt.Errorf("kubeReserved[%q] value %q cannot be represented as whole MiB", key, value)
+			}
+			result.MemoryMB = lo.ToPtr(int32(memoryMiB))
+		default:
+			return nil, fmt.Errorf("kubeReserved[%q] is not supported by the AKS Machine API", key)
```
comtalyst, line 499: Similar concerns with https://github.com/Azure/karpenter-provider-azure/pull/1824/changes#r3908066639. For this and nearby. Note that some fallback errors here are still useful in addition to that.

comtalyst: Unit conversion can still happen here, if that conversion is necessary.

pkg/providers/instancetype/instancetype.go
```diff
@@ -372,11 +374,33 @@ func KubeReservedResources(vcpus, totalMemoryMiB int64, maxPods int32, enableNod
...
+		if value, ok := overrides[0][MemoryAvailable]; ok {
+			if strings.HasSuffix(value, "%") {
+				if percentage, err := strconv.ParseFloat(strings.TrimSuffix(value, "%"), 64); err == nil && !math.IsNaN(percentage) && percentage >= 0 && percentage <= 100 {
+					return corev1.ResourceList{
+						corev1.ResourceMemory: *resource.NewQuantity(int64(math.Ceil(float64(totalMemoryMiB)*percentage/100))*bytesPerMiB, resource.BinarySI),
+					}
+				}
+			} else if quantity, err := resource.ParseQuantity(value); err == nil {
+				return corev1.ResourceList{corev1.ResourceMemory: quantity}
+			}
```
comtalyst, line 402: Silently ignoring errors here may not be ideal though. May consider https://github.com/Azure/karpenter-provider-azure/pull/1824/changes#r3908066639 for some of these too?

pkg/providers/instance/aksmachineinstancehelpers.go
```diff
@@ -431,7 +437,82 @@ func configureKubeletConfig(nodeClass *v1beta1.AKSNodeClass) *armcontainerservic
...
+		case instancetype.NodeFSAvailable:
+			result.NodeFsAvailable = lo.ToPtr(value)
+		case instancetype.NodeFSInodesFree:
+			result.NodeFsInodesFree = lo.ToPtr(value)
+		default:
+			return nil, fmt.Errorf("evictionHard[%q] is not supported by the AKS Machine API", key)
+		}
+	}
+	return result, nil
 }
```
comtalyst, line 520: Consider adding acceptance unit tests in cloudprovider/suite_*_test.go

pkg/providers/instance/aksmachineinstancehelpers.go
```diff
@@ -431,7 +437,82 @@ func configureKubeletConfig(nodeClass *v1beta1.AKSNodeClass) *armcontainerservic
...
```
comtalyst, line 521: Consider adding E2E tests in test/... Especially on the newly exposed fields.

comtalyst (commented):
Please also update the PR description to match the repo-wide format (preferably to a human-written, so it is a self-review in away). 
And also include E2E (GitHub Actions) results.

pkg/providers/instance/aksmachineinstancehelpers.go
mxj220: This was a mistake, these fields are supported as of this PR: https://dev.azure.com/msazure/CloudNativeCompute/_git/aks-rp/pullrequest/16635303?path=/resourceprovider/server/microsoft.com/containerservice/datamodel/v20260702preview/machine_convert.go

I just need to bump the armcontainerservice version.

pkg/apis/v1beta1/aksnodeclass.go
```diff
@@ -446,8 +446,56 @@ type KubeletConfiguration struct {
...
+	EvictionHard map[string]EvictionHardValue `json:"evictionHard,omitempty"`
+	// EvictionSoft maps supported eviction signals (`memory.available`, `nodefs.available`, `nodefs.inodesFree`) to quantities and overrides Karpenter's soft eviction thresholds per key.
+	// It matches AKS `kubeletConfig.softEvictionThreshold`. Each soft threshold must be paired with a grace period in evictionSoftGracePeriod.
+	// +kubebuilder:validation:MaxProperties=3
+	// +kubebuilder:validation:XValidation:rule="self.all(key, key == 'memory.available' || key == 'nodefs.available' || key == 'nodefs.inodesFree')",message="evictionSoft supports only memory.available, nodefs.available, and nodefs.inodesFree"
+	// +kubebuilder:validation:XValidation:rule="self.all(key, key == 'nodefs.inodesFree' ? !self[key].endsWith('Ki') && !self[key].endsWith('Mi') && !self[key].endsWith('Gi') : self[key].endsWith('Ki') || self[key].endsWith('Mi') || self[key].endsWith('Gi') || self[key].endsWith('%'))",message="memory.available and nodefs.available must be integer Ki, Mi, or Gi quantities or percentages; nodefs.inodesFree must be an integer count or percentage"
```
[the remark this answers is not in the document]
mxj220: Enforcement is at the RP level in customnodeconfigvalidator

pkg/providers/instance/aksmachineinstancehelpers.go
mxj220: moved primary checks to CRD validation, kept some runtime checks in case.

pkg/providers/instance/aksmachineinstancehelpers.go
mxj220: moved primary checks to CRD validation, kept some runtime checks in case.

pkg/providers/instancetype/instancetype.go
mxj220: moved primary checks to CRD validation, kept some runtime checks in case.

pkg/apis/v1beta1/aksnodeclass.go
```diff
@@ -467,8 +467,56 @@ type KubeletConfiguration struct {
...
+	// KubeReserved maps supported resource names (`cpu`, `memory`) to reserved values and overrides Karpenter's computed defaults per key.
+	// It matches AKS `kubeletConfig.cpuReserved` and `memoryReserved`.
+	// +kubebuilder:validation:MaxProperties=2
+	// +kubebuilder:validation:XValidation:rule="self.all(key, key == 'cpu' || key == 'memory')",message="kubeReserved supports only cpu and memory"
+	// +kubebuilder:validation:XValidation:rule="self.all(key, key == 'cpu' ? !self[key].endsWith('Mi') && !self[key].endsWith('Gi') : self[key].endsWith('Mi') || self[key].endsWith('Gi'))",message="kubeReserved cpu must be positive whole cores or millicores and memory must be positive whole MiB or GiB"
+	// +optional
+	//nolint:kubeapilinter // ssatags: server-side apply is not used for kubelet reservation maps
+	KubeReserved map[string]KubeReservedValue `json:"kubeReserved,omitempty"`
+	// EvictionHard maps supported eviction signals (`memory.available`, `nodefs.available`, `nodefs.inodesFree`) to quantities and overrides Karpenter's defaults per key.
+	// It matches AKS `kubeletConfig.hardEvictionThreshold`.
```
comtalyst: (Applying to all new fields) The fields here is different from AKS, though I assume is intentional to match K8s?

Thoughts? @matthchr

pkg/providers/instance/aksmachineinstancehelpers.go
```diff
@@ -109,6 +112,10 @@ func (p *DefaultAKSMachineProvider) buildAKSMachineTemplate(ctx context.Context,
 	// see batch_field_registry.go — ClearPerMachineFields must cover any new per-machine
 	// MachineProperties field so batch grouping and header extraction stay correct.
 	tags := ConfigureAKSMachineTags(options.FromContext(ctx), nodeClass, nodeClaim)
+	kubeletConfig, err := configureKubeletConfig(nodeClass)
```
comtalyst:
```suggestion

	kubeletConfig, err := configureKubeletConfig(nodeClass)
```

pkg/providers/instance/aksmachineinstancehelpers.go
```diff
@@ -396,11 +403,11 @@ func ConfigureAKSMachineTags(opts *options.Options, nodeClass *v1beta1.AKSNodeCl
 }
 
 //nolint:gocyclo // borderline complexity violation, code is not hard to read
-func configureKubeletConfig(nodeClass *v1beta1.AKSNodeClass) *armcontainerservice.KubeletConfig {
+func configureKubeletConfig(nodeClass *v1beta1.AKSNodeClass) (*armcontainerservice.KubeletConfig, error) {
 	// Counterpart for ProvisionModeBootstrappingClient is in customscriptsbootstrap/provisionclientbootstrap.go and imagefamily/resolver.go
 
 	if nodeClass == nil || nodeClass.Spec.Kubelet == nil {
-		return nil
+		return nil, nil
```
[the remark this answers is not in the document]
comtalyst: The plan is to not supporting that mode, so this divergence is intentional (right? @mxj220)

Still, given this is an API change and Machine API integration rollout haven't finish, users still on `bootstrappingclient` mode will see unusable API. Will need to resolve this (wait, or toggle the fields, or just let go?) once we take this change to NAP.

pkg/providers/instancetype/instancetypes.go
```diff
@@ -182,6 +188,16 @@ func (p *DefaultProvider) List(
...
 
+func copySelectedValues[T ~string](values map[string]T, keys ...string) map[string]string {
+	selected := map[string]string{}
+	for _, key := range keys {
+		if value, ok := values[key]; ok {
+			selected[key] = string(value)
+		}
+	}
+	return selected
+}
```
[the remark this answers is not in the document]
comtalyst: @mxj220 thoughts? I think this is more correct.

pkg/providers/instance/aksmachineinstancehelpers.go
```diff
@@ -441,7 +448,119 @@ func configureKubeletConfig(nodeClass *v1beta1.AKSNodeClass) *armcontainerservic
 
 	kubeletConfig.FailSwapOn = nodeClass.Spec.Kubelet.FailSwapOn
 
-	return kubeletConfig
+	if len(nodeClass.Spec.Kubelet.KubeReserved) > 0 {
+		kubeReserved, err := configureAKSMachineKubeReserved(nodeClass.Spec.Kubelet.KubeReserved)
```
comtalyst: Per https://github.com/Azure/karpenter-provider-azure/pull/1818#discussion_r3739030120, we should wire `KubeReserved` and others from `instanceType` instead of directly from `nodeClass`. This will resolve the split source of truth issue for the case where there is no override. Otherwise, when `KubeReserved` unspecified in the `nodeClass` (and Machine API payload), what's defaulted behind AKS Machine API can continue to drift from `KubeReserved` in `instanceType`.

pkg/apis/v1beta1/aksnodeclass.go
```diff
@@ -467,8 +467,56 @@ type KubeletConfiguration struct {
...
+	// It matches AKS `kubeletConfig.softEvictionThreshold`. Each soft threshold must be paired with a grace period in evictionSoftGracePeriod.
+	// +kubebuilder:validation:MaxProperties=3
+	// +kubebuilder:validation:XValidation:rule="self.all(key, key == 'memory.available' || key == 'nodefs.available' || key == 'nodefs.inodesFree')",message="evictionSoft supports only memory.available, nodefs.available, and nodefs.inodesFree"
+	// +kubebuilder:validation:XValidation:rule="self.all(key, key == 'nodefs.inodesFree' ? !self[key].endsWith('Ki') && !self[key].endsWith('Mi') && !self[key].endsWith('Gi') : self[key].endsWith('Ki') || self[key].endsWith('Mi') || self[key].endsWith('Gi') || self[key].endsWith('%'))",message="memory.available and nodefs.available must be integer Ki, Mi, or Gi quantities or percentages; nodefs.inodesFree must be an integer count or percentage"
+	// +optional
+	//nolint:kubeapilinter // ssatags: server-side apply is not used for kubelet eviction maps
+	EvictionSoft map[string]EvictionSoftValue `json:"evictionSoft,omitempty"`
```
comtalyst: (AI-generated)

**High — enforce soft-threshold/grace pairing at admission**

`evictionSoft` and `evictionSoftGracePeriod` are currently validated independently. The CRD test even expects a NodeClass containing `evictionSoft` without a grace map to be admitted. In `aksscriptless`, the two maps are rendered independently, while kubelet rejects every configured soft signal that has no corresponding grace period. That can leave a created VM with a kubelet that cannot start and a node that never registers.

Please add `KubeletConfiguration`-level CEL requiring every explicit `evictionSoft` key to exist in `evictionSoftGracePeriod`, plus missing- and mismatched-key tests. One-way containment is preferable if grace-only overrides of existing hardening defaults are intentional.

pkg/apis/v1beta1/aksnodeclass.go
```diff
@@ -467,8 +467,56 @@ type KubeletConfiguration struct {
...
+	// EvictionSoftGracePeriod maps supported eviction signals (`memory.available`, `nodefs.available`, `nodefs.inodesFree`) to grace periods and overrides Karpenter's defaults per key.
+	// It matches AKS `kubeletConfig.softEvictionGracePeriod`.
+	// +kubebuilder:validation:MaxProperties=3
+	// +kubebuilder:validation:XValidation:rule="self.all(key, key == 'memory.available' || key == 'nodefs.available' || key == 'nodefs.inodesFree')",message="evictionSoftGracePeriod supports only memory.available, nodefs.available, and nodefs.inodesFree"
+	// +optional
+	//nolint:kubeapilinter // ssatags,nodurations: metav1.Duration matches upstream kubelet types
+	EvictionSoftGracePeriod map[string]metav1.Duration `json:"evictionSoftGracePeriod,omitempty"`
```
comtalyst: (AI-generated)

**High — validate duration values at the CRD boundary**

`map[string]metav1.Duration` constrains the Go type, but the generated CRD schema permits any string value here. I reproduced an unstructured create with `memory.available: not-a-duration`: admission succeeded, and a subsequent typed `AKSNodeClassList` failed with `time: invalid duration "not-a-duration"` while that object existed.

Please add value-level admission validation for duration syntax and the supported range, with an unstructured negative test. A typed fixture cannot exercise this failure because it rejects the malformed value before sending the request.

pkg/apis/v1beta1/aksnodeclass.go
```diff
@@ -467,8 +467,56 @@ type KubeletConfiguration struct {
...
+	// evictionMaxPodGracePeriod is the maximum grace period (in seconds) kubelet honors when terminating pods for soft eviction.
+	// It matches AKS `kubeletConfig.evictionMaxPodGracePeriodInSeconds`.
+	// +kubebuilder:validation:Minimum=0
+	// +optional
+	EvictionMaxPodGracePeriod *int32 `json:"evictionMaxPodGracePeriod,omitempty"`
 }
 
+// +kubebuilder:validation:MaxLength=20
+// +kubebuilder:validation:Pattern=`^([1-9][0-9]*m|[1-9][0-9]*(Mi|Gi)?)$`
+type KubeReservedValue string
```
comtalyst: (AI-generated)

**Medium — enforce downstream integer representation bounds at admission**

The current pattern checks syntax and length but has no numeric upper bound. I reproduced `kubeReserved.cpu: 2147483648m` passing CRD admission, after which the Machine converter rejects it because whole millicores must fit in `int32`. This is deterministic user input, so it should not first fail during NodeClaim provisioning.

Please add admission bounds for CPU millicores and memory MiB matching the downstream `int32` representation, retain the defensive runtime checks, and test the maximum representable value plus maximum + 1.

pkg/apis/v1beta1/aksnodeclass.go
```diff
@@ -467,8 +467,56 @@ type KubeletConfiguration struct {
...
+	// +kubebuilder:validation:MaxProperties=3
+	// +kubebuilder:validation:XValidation:rule="self.all(key, key == 'memory.available' || key == 'nodefs.available' || key == 'nodefs.inodesFree')",message="evictionSoftGracePeriod supports only memory.available, nodefs.available, and nodefs.inodesFree"
+	// +optional
+	//nolint:kubeapilinter // ssatags,nodurations: metav1.Duration matches upstream kubelet types
+	EvictionSoftGracePeriod map[string]metav1.Duration `json:"evictionSoftGracePeriod,omitempty"`
+	// evictionMaxPodGracePeriod is the maximum grace period (in seconds) kubelet honors when terminating pods for soft eviction.
+	// It matches AKS `kubeletConfig.evictionMaxPodGracePeriodInSeconds`.
+	// +kubebuilder:validation:Minimum=0
+	// +optional
+	EvictionMaxPodGracePeriod *int32 `json:"evictionMaxPodGracePeriod,omitempty"`
```
comtalyst, line 507: What is the plan on preview experience of these fields on NAP?
On AKS, my understanding is user need to register `CustomNodeConfigPreview` to be able to use AgentPool API. For NAP here, the same requirement exists for Machine API(?). So, it will fail at provision time instead. 

I think this is worth discussing w/ more audiences as it is a more generalized API problem.

pkg/providers/instance/aksmachineinstancehelpers.go
```diff
@@ -441,7 +448,119 @@ func configureKubeletConfig(nodeClass *v1beta1.AKSNodeClass) *armcontainerservic
 
 	kubeletConfig.FailSwapOn = nodeClass.Spec.Kubelet.FailSwapOn
 
-	return kubeletConfig
+	if len(nodeClass.Spec.Kubelet.KubeReserved) > 0 {
```
comtalyst: From https://github.com/Azure/karpenter-provider-azure/pull/1824/changes#r3973091302, if we are failing at provisioning time, let's add the comment here to prevent these fields from being used unintentionally, regressing everyone.

pkg/providers/instance/aksmachineinstancehelpers.go
comtalyst: Once the preview flag issue from https://github.com/Azure/karpenter-provider-azure/pull/1824/changes#r3973106487 is resolved, per https://github.com/Azure/karpenter-provider-azure/pull/1818#discussion_r3739030120, we should wire `KubeReserved` and others from `instanceType` instead of directly from `nodeClass`. This will resolve the split source of truth issue for the case where there is no override. Otherwise, when `KubeReserved` unspecified in the `nodeClass` (and Machine API payload), what's defaulted behind AKS Machine API can continue to drift from `KubeReserved` in `instanceType`. 

For now, if we are going to restrict this for the preview, let's at least leave a comment/TODO on this.

pkg/apis/v1beta1/aksnodeclass.go
comtalyst: If we ended up with no better choice than this, let's at least leave a comment in the CRD(?).

comtalyst (commented):
Let's include E2E run results in the PR description as well.

pkg/apis/v1beta1/aksnodeclass.go
matthchr: I think we should match AKS with a well-typed object here not an arbitrary map. We only allow certain keys so we should just say so on the object. i.e. this should be:

```
KubeReserved *KubeReserved
```

and `KubeReserved` struct would be:
```
MemoryAvailable *KubeReservedValue
NodeFsAvailable *KubeReservedValue
etc
```

Also I think we want `*KubeReservedValue` when we do it this way so the default isn't ""

pkg/apis/v1beta1/aksnodeclass.go
```diff
@@ -467,8 +467,56 @@ type KubeletConfiguration struct {
 	// Must be set to false to allow linuxOSConfig.swapFileSize to take effect.
 	// +optional
 	FailSwapOn *bool `json:"failSwapOn,omitempty"`
+	// KubeReserved maps supported resource names (`cpu`, `memory`) to reserved values and overrides Karpenter's computed defaults per key.
```
matthchr: Missing v1alpha2 copies of this

pkg/apis/v1beta1/aksnodeclass.go
```diff
@@ -467,8 +467,56 @@ type KubeletConfiguration struct {
...
+	// +optional
+	//nolint:kubeapilinter // ssatags: server-side apply is not used for kubelet reservation maps
+	KubeReserved map[string]KubeReservedValue `json:"kubeReserved,omitempty"`
+	// EvictionHard maps supported eviction signals (`memory.available`, `nodefs.available`, `nodefs.inodesFree`) to quantities and overrides Karpenter's defaults per key.
+	// It matches AKS `kubeletConfig.hardEvictionThreshold`.
+	// +kubebuilder:validation:MaxProperties=3
+	// +kubebuilder:validation:XValidation:rule="self.all(key, key == 'memory.available' || key == 'nodefs.available' || key == 'nodefs.inodesFree')",message="evictionHard supports only memory.available, nodefs.available, and nodefs.inodesFree"
+	// +kubebuilder:validation:XValidation:rule="self.all(key, key == 'nodefs.inodesFree' ? !self[key].endsWith('Ki') && !self[key].endsWith('Mi') && !self[key].endsWith('Gi') : self[key].endsWith('Ki') || self[key].endsWith('Mi') || self[key].endsWith('Gi') || self[key].endsWith('%'))",message="memory.available and nodefs.available must be integer Ki, Mi, or Gi quantities or percentages; nodefs.inodesFree must be an integer count or percentage"
```
matthchr: Can we have different validation rules for each field rather than lumping them together? (Applies to other fields as well). That way it is clearer for the user which field they got wrong and what the actual problem is, since the errors for the different fields are different.

pkg/apis/v1beta1/aksnodeclass.go
matthchr: 🤖 **High:** The API documentation requires every soft-eviction threshold to have a matching grace period, but the two maps are currently validated independently. An object containing `evictionSoft` without the corresponding `evictionSoftGracePeriod` key can therefore pass admission and produce invalid kubelet configuration or a deterministic Machine API provisioning failure. Could we add a structure-level CEL rule requiring every soft-threshold key to exist in the grace-period map, with missing and mismatched-key tests?

pkg/apis/v1beta1/aksnodeclass.go
matthchr: 🤖 **High:** The generated schema represents these `metav1.Duration` values as unconstrained strings. An unstructured client can persist a value such as `not-a-duration`, even though typed Go fixtures cannot construct it, which may later fail typed deserialization/list-watch processing. Could we validate duration syntax and the supported non-negative range at admission and add an `unstructured.Unstructured` test for malformed and negative values?

pkg/providers/instance/aksmachineinstancehelpers.go
```diff
@@ -441,7 +448,119 @@ func configureKubeletConfig(nodeClass *v1beta1.AKSNodeClass) *armcontainerservic
...
+	result := &armcontainerservice.KubeReserved{}
+	for key, value := range overrides {
+		quantity, err := resource.ParseQuantity(string(value))
+		if err != nil {
+			return nil, fmt.Errorf("invalid kubeReserved[%q] value %q: %w", key, value, err)
+		}
+		switch key {
+		case string(v1.ResourceCPU):
+			milliValue := quantity.MilliValue()
+			if milliValue < math.MinInt32 || milliValue > math.MaxInt32 || quantity.Cmp(*resource.NewMilliQuantity(milliValue, resource.DecimalSI)) != 0 {
```
matthchr: 🤖 **Medium:** These runtime bounds checks safely prevent truncation, but the CRD still admits quantities larger than the Machine API `int32` representation. Such input will fail only after a NodeClaim reaches provisioning and can retry deterministically. Could the admission rules reject values above `MaxInt32` millicores/MiB, while retaining these checks defensively?

pkg/providers/instance/aksmachineinstancehelpers.go
```diff
@@ -396,11 +403,11 @@ func ConfigureAKSMachineTags(opts *options.Options, nodeClass *v1beta1.AKSNodeCl
 }
 
 //nolint:gocyclo // borderline complexity violation, code is not hard to read
-func configureKubeletConfig(nodeClass *v1beta1.AKSNodeClass) *armcontainerservice.KubeletConfig {
+func configureKubeletConfig(nodeClass *v1beta1.AKSNodeClass) (*armcontainerservice.KubeletConfig, error) {
```
matthchr: 🤖 **Medium:** This public kubelet API now has multiple defaulting authorities: the bootstrap path overlays locally computed defaults, while the Machine API path sends only explicit fields and relies on downstream defaults for omitted values. Scheduling also computes its own effective values. How do we guarantee these remain equivalent across provisioning modes? A shared resolved internal configuration, or explicit parity coverage for every supported mode, would reduce drift risk.

pkg/providers/instancetype/instancetype.go
```diff
@@ -372,7 +375,7 @@ func SystemReservedResources(totalMemoryMiB int64, networkPlugin string, enableN
 	}
 }
 
-func KubeReservedResources(vcpus, totalMemoryMiB int64, maxPods int32, enableNodeHardening bool) corev1.ResourceList {
+func KubeReservedResources(vcpus, totalMemoryMiB int64, maxPods int32, enableNodeHardening bool, overrides ...map[string]string) corev1.ResourceList {
```
matthchr: 🤖 **Medium:** This variadic signature suggests that multiple override maps are supported, but the implementation reads only `overrides[0]`; `EvictionThreshold` follows the same pattern. A future caller passing more than one map would have later values silently ignored. Could these take one explicit map (or a named options structure) unless ordered multi-overlay behavior is intended?

pkg/providers/instancetype/instancetypes.go
```diff
@@ -182,6 +188,16 @@ func (p *DefaultProvider) List(
 	return append([]*cloudprovider.InstanceType{}, result...), nil
 }
 
+func copySelectedValues[T ~string](values map[string]T, keys ...string) map[string]string {
+	selected := map[string]string{}
```
matthchr: 🤖 **Low:** This eagerly returns a non-nil empty map when the kubelet block contains no scheduling-relevant keys. Because the result participates in `instanceTypeParameters` hashing, that can create a separate long-lived cache entry for configuration that produces identical instance types. Could this allocate lazily and return `nil` when no requested key is selected?

pkg/providers/imagefamily/resolver.go
```diff
@@ -285,9 +285,44 @@ func prepareKubeletConfiguration(ctx context.Context, instanceType *cloudprovide
...
+
+func mergeStringMap[T ~string](base map[string]string, overrides map[string]T) map[string]string {
```
matthchr: There's a `lo` helper for this which we use elsewhere, though I think this may just go away if we use actual object types at the API layer.

pkg/providers/imagefamily/resolver_unit_test.go
```diff
@@ -116,3 +117,30 @@ func TestPrepareKubeletConfigurationSoftEvictionDisabledForBootstrappingClient(t
 	g.Expect(configuration.EnforceNodeAllocatable).To(BeNil())
 	g.Expect(configuration.SystemReserved).ToNot(HaveKey("pid"))
 }
+
+func TestOverlayKubeletConfiguration(t *testing.T) {
```
matthchr: IMO test is at the wrong level. Test the result of a call to `prepareKubeletConfiguration` like the other tests here and keep the tests at that level rather than assuming about its implementation.

pkg/providers/instance/aksmachineinstancehelpers.go
matthchr: Stopped here

matthchr (commented):
Got through some of it but didn't finish. Will continue tomorrow

pkg/providers/instance/aksmachineinstancehelpers.go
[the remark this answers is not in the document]
mxj220: Correct, we are not supporting bootstrappingClient.

pkg/apis/v1beta1/aksnodeclass.go
mxj220: Correct, on second thought, it makes more sense to rename them to match AKS

pkg/apis/v1beta1/aksnodeclass.go
mxj220: Ok, I think I misunderstood what v1alpha2 is for then. I'll revisit.

mxj220: There was an AI comment much earlier that said it was deprecated so I removed the implementation there. Can you confirm that it should still be served?

pkg/providers/instance/aksmachineinstancehelpers.go
```diff
@@ -441,7 +448,119 @@ func configureKubeletConfig(nodeClass *v1beta1.AKSNodeClass) *armcontainerservic
...
+			}
+			result.MemoryMB = lo.ToPtr(int32(memoryMiB))
+		default:
+			return nil, fmt.Errorf("kubeReserved[%q] is not supported by the AKS Machine API", key)
+		}
+	}
+	return result, nil
+}
+
+func configureAKSMachineHardEviction(overrides map[string]v1beta1.EvictionHardValue) (*armcontainerservice.HardEvictionThreshold, error) {
```
matthchr: I think these methods get a bit easier/simpler (no need for default case because there would be no way to set an unsupported field) if you move to struct like I suggested earlier

pkg/providers/instancetype/instancetypes.go
```diff
@@ -70,6 +70,8 @@ type instanceTypeParameters struct {
 	ArtifactStreamingEnabled bool
 	FIPSMode                 v1beta1.FIPSMode
 	LocalDNSEnabled          bool
+	KubeReserved             map[string]string
```
matthchr, line 73: minor, possibly worth a note that these are needed in instance type parameters because even though they don't filter the list of allowed instance types like fields like FIPS do, they do impact the scheduling simulation since Karpenter considers the reserved + eviction levels to fit pods within nodes, so we need to ensure that one NodeClass doesn't get another NodeClasses instances.

matthchr (commented):
Finished rest of my review

pkg/apis/v1beta1/aksnodeclass.go
comtalyst: @mxj220 we discussed this in the standup today and think that we will get away with best-effort documentation.
For the scope of this PR, let's at least ensure this fact is well-included in field description.

pkg/providers/instance/aksmachineinstancehelpers.go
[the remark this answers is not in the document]
mxj220: BootstrappingClient remains intentionally unsupported for these fields. I did not add partial plumbing there. The field descriptions now explicitly say the overrides are supported only by AKSScriptless and AKSMachineAPI, and require CustomNodeConfigPreview registration.

pkg/apis/v1beta1/aksnodeclass.go
mxj220: Updated the API to use AKS-aligned typed objects instead of arbitrary maps

pkg/apis/v1beta1/aksnodeclass.go
mxj220: Added field-specific CEL rules requiring each soft eviction threshold and its corresponding grace period to be configured together. Admission tests cover valid and mismatched fields.

pkg/apis/v1beta1/aksnodeclass.go
mxj220: Added per-field duration validation requiring a valid duration of at least 30 seconds. The unstructured negative test confirms malformed duration strings are rejected at admission before typed decoding.

pkg/apis/v1beta1/aksnodeclass.go
mxj220: KubeReserved now uses int32 fields matching the Machine API, with explicit Maximum=2147483647 schema validation because format: int32 alone does not enforce the range. Tests cover the maximum and an unstructured maximum-plus-one rejection.

pkg/apis/v1beta1/aksnodeclass.go
mxj220: Added documentation to every new field. Specified that CustomNodeConfigPreview registration is required and that bootstrappingClient is not supported.

pkg/apis/v1beta1/aksnodeclass.go
mxj220: Added

pkg/providers/instance/aksmachineinstancehelpers.go
mxj220: Added the preview requirement to the API field descriptions and a warning at the Machine payload boundary

pkg/providers/instance/aksmachineinstancehelpers.go
mxj220: Kept an explicit TODO at the Machine configuration call and expanded it to cover a shared effective kubelet configuration across scheduling, AKSScriptless, and AKSMachineAPI. This can be completed when Node Hardening becomes GA

pkg/apis/v1beta1/aksnodeclass.go
mxj220: Done when switching from maps to typed objects

pkg/providers/instance/aksmachineinstancehelpers.go
mxj220: Set to int32 with explicit max

pkg/providers/instance/aksmachineinstancehelpers.go
mxj220: The original TODO was added in response to the earlier request to leave a follow-up marker for passing effective instance-type values. I expanded it to track one shared resolved configuration across scheduling, AKSScriptless, and AKSMachineAPI. Awaiting Node Hardening GA

pkg/apis/v1alpha2/aksnodeclass.go
```diff
@@ -450,6 +453,81 @@ type KubeletConfiguration struct {
 	// Must be set to false to allow linuxOSConfig.swapFileSize to take effect.
 	// +optional
 	FailSwapOn *bool `json:"failSwapOn,omitempty"`
+	// kubeReserved configures resources reserved for Kubernetes system daemons and overrides Karpenter's computed defaults per field.
+	// This field requires the CustomNodeConfigPreview feature registration and is supported only by the AKSScriptless and AKSMachineAPI provisioning modes.
+	// +optional
+	KubeReserved *KubeReserved `json:"kubeReserved,omitempty"`
```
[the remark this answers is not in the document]
mxj220: @matthchr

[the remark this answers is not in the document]
matthchr: It's still expected you can ignore this bot.

pkg/apis/v1beta1/aksnodeclass.go
```diff
@@ -479,6 +482,81 @@ type KubeletConfiguration struct {
 	// Must be set to false to allow linuxOSConfig.swapFileSize to take effect.
 	// +optional
 	FailSwapOn *bool `json:"failSwapOn,omitempty"`
+	// kubeReserved configures resources reserved for Kubernetes system daemons and overrides Karpenter's computed defaults per field.
+	// This field requires the CustomNodeConfigPreview feature registration and is supported only by the AKSScriptless and AKSMachineAPI provisioning modes.
```
matthchr: >  This field requires the CustomNodeConfigPreview feature registration 

Is that actually true? I don't think it is?

matthchr: > and is supported only by the AKSScriptless and AKSMachineAPI provisioning modes.

Not sure it's worth calling this out in the docs, as this is going to be "all supported modes" here very shortly. IMO just cut this here and elsewhere.

pkg/apis/v1beta1/aksnodeclass.go
```diff
@@ -479,6 +482,81 @@ type KubeletConfiguration struct {
...
+	// memoryMB is the memory reservation in MiB.
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=2147483647
+	// +optional
+	MemoryMB *int32 `json:"memoryMB,omitempty"`
+}
+
+// EvictionThreshold defines thresholds for supported kubelet eviction signals.
+type EvictionThreshold struct {
+	// memoryAvailable is the available-memory threshold as an integer Ki, Mi, or Gi quantity or percentage.
```
matthchr: minor: state when percentage what it is a percentage of (I think total node memory?)

matthchr: Applies to all of the below too.

pkg/apis/v1beta1/aksnodeclass.go
```diff
@@ -479,6 +482,81 @@ type KubeletConfiguration struct {
...
+	NodeFsInodesFree *string `json:"nodeFsInodesFree,omitempty"`
+}
+
+// EvictionSoftGracePeriod defines grace periods for supported soft eviction signals.
+type EvictionSoftGracePeriod struct {
+	// memoryAvailable is the grace period for the memoryAvailable signal.
+	// +kubebuilder:validation:XValidation:rule="duration(self) >= duration('30s')",message="memoryAvailable must be a valid duration of at least 30s"
+	// +optional
+	//nolint:kubeapilinter // nodurations: metav1.Duration matches upstream kubelet types
+	MemoryAvailable *metav1.Duration `json:"memoryAvailable,omitempty"`
```
matthchr: Should be using `karpv1.NillableDuration` here.

pkg/apis/v1beta1/aksnodeclass.go
```diff
@@ -479,6 +482,81 @@ type KubeletConfiguration struct {
...
+	// +optional
+	NodeFsInodesFree *string `json:"nodeFsInodesFree,omitempty"`
+}
+
+// EvictionSoftGracePeriod defines grace periods for supported soft eviction signals.
+type EvictionSoftGracePeriod struct {
+	// memoryAvailable is the grace period for the memoryAvailable signal.
+	// +kubebuilder:validation:XValidation:rule="duration(self) >= duration('30s')",message="memoryAvailable must be a valid duration of at least 30s"
+	// +optional
+	//nolint:kubeapilinter // nodurations: metav1.Duration matches upstream kubelet types
```
matthchr: Remove this. See LocalDNS for examples of what you should be doing.

pkg/cloudprovider/suite_aksmachineapi_features_test.go
```diff
@@ -557,6 +571,15 @@ var _ = Describe("CloudProvider", func() {
...
 				Expect(*aksMachine.Properties.Kubernetes.KubeletConfig.ImageGcLowThreshold).To(Equal(int32(80)))
 				Expect(lo.FromPtr(aksMachine.Properties.Kubernetes.KubeletConfig.FailSwapOn)).To(BeFalse())
+				Expect(*aksMachine.Properties.Kubernetes.KubeletConfig.KubeReserved.CPUMillicores).To(Equal(int32(250)))
+				Expect(*aksMachine.Properties.Kubernetes.KubeletConfig.KubeReserved.MemoryMB).To(Equal(int32(512)))
+				Expect(*aksMachine.Properties.Kubernetes.KubeletConfig.HardEvictionThreshold.MemoryAvailable).To(Equal("333Mi"))
+				Expect(*aksMachine.Properties.Kubernetes.KubeletConfig.HardEvictionThreshold.NodeFsAvailable).To(Equal("12%"))
+				Expect(*aksMachine.Properties.Kubernetes.KubeletConfig.HardEvictionThreshold.NodeFsInodesFree).To(Equal("7%"))
+
+				Expect(*aksMachine.Properties.Kubernetes.KubeletConfig.SoftEvictionThreshold.MemoryAvailable).To(Equal("500Mi"))
+				Expect(*aksMachine.Properties.Kubernetes.KubeletConfig.SoftEvictionGracePeriod.MemoryAvailable).To(Equal("1m30s"))
```
matthchr: This shouldn't normalize I don't think, if you use the `karpv1.NillableDuration`.

pkg/providers/instance/aksmachineinstancehelpers.go
```diff
@@ -441,9 +448,63 @@ func configureKubeletConfig(nodeClass *v1beta1.AKSNodeClass) *armcontainerservic
...
+	return &armcontainerservice.SoftEvictionThreshold{
+		MemoryAvailable:  config.MemoryAvailable,
+		NodeFsAvailable:  config.NodeFsAvailable,
+		NodeFsInodesFree: config.NodeFsInodesFree,
+	}
+}
+
+func configureAKSMachineSoftEvictionGracePeriod(config *v1beta1.EvictionSoftGracePeriod) *armcontainerservice.SoftEvictionGracePeriod {
+	return &armcontainerservice.SoftEvictionGracePeriod{
+		MemoryAvailable:  durationString(config.MemoryAvailable),
```
matthchr, line 492: May want to pass on the exact string the user specified using `.Raw` with `NillableDuration`? Not 100% clear to me it's required but might be useful so that Machine gets exactly what we got in our API?

pkg/providers/instancetype/instancetypes.go
```diff
@@ -158,6 +162,12 @@ func (p *DefaultProvider) List(
 		LocalDNSEnabled:          nodeClass.IsLocalDNSEnabled(),
 		KataEnabled:              nodeClass.IsKataEnabled(),
 	}
+	if nodeClass.Spec.Kubelet != nil && (opts.ProvisionMode == consts.ProvisionModeAKSScriptless || opts.IsAKSMachineAPIMode()) {
```
matthchr: Check: Are we rejecting configurations other than this? (i.e. user can't ask for this field if not in machine API mode, the status controller of the nodeclass rejects them?)

matthchr: It's not totally clear we need this guard here to me, we could just check `if nodeClass.Spec.Kubelet != nil` since it's not possible for the fields in here to be set if these modes anyway I think?

pkg/providers/instancetype/instancetypes.go
```diff
@@ -182,6 +192,40 @@ func (p *DefaultProvider) List(
 	return append([]*cloudprovider.InstanceType{}, result...), nil
 }
 
+func kubeReservedOverrides(config *v1beta1.KubeReserved) map[string]string {
```
matthchr: Is this duplicate?

mxj220: Yeah they can be consolidated

pkg/providers/instancetype/instancetypes.go
mxj220: I believe today it is possible to set these in bootstrappingClient. However, since this PR will not be merged until bootstrappingClient is retired, we should be okay to remove this check.

matthchr:
These two comments look possible accurate to me?

> Severity: Medium; Category: Scheduling — The percentage path rounds the threshold up to a whole MiB, while kubelet computes percentage eviction thresholds against the capacity in bytes and truncates the result (the byte-level behavior is documented in the adjacent ephemeral-storage implementation). For example, 5% of an 8-GiB node is about 409.6 MiB, but this models 410 MiB, so InstanceType.Allocatable can reject a request that kubelet would accept at the boundary. Compute the percentage directly in bytes using the same float32/truncation behavior as the node configuration.

> Severity: Medium; Category: Test Coverage — These new spec fields participate in AKSNodeClass.Hash(), which drives static drift, but the existing hash suite still changes only CPUManagerPolicy under Kubelet. A regression that omits one of these nested fields would leave existing NodeClaims on old kubelet settings without drift; add hash cases for the new override groups, including a duration, in aksnodeclass_hash_test.go.

pkg/apis/v1beta1/crd_validation_cel_test.go
```diff
@@ -92,6 +93,100 @@ var _ = Describe("CEL/Validation", func() {
...
+			}
+		},
+			Entry("valid supported signals", &v1beta1.EvictionThreshold{MemoryAvailable: lo.ToPtr("333Mi"), NodeFsAvailable: lo.ToPtr("12%"), NodeFsInodesFree: lo.ToPtr("100000")}, true),
+			Entry("valid percentage boundary", &v1beta1.EvictionThreshold{MemoryAvailable: lo.ToPtr("100%")}, true),
+			Entry("invalid memory quantity", &v1beta1.EvictionThreshold{MemoryAvailable: lo.ToPtr("invalid")}, false),
+			Entry("node filesystem percentage above 100", &v1beta1.EvictionThreshold{NodeFsAvailable: lo.ToPtr("101%")}, false),
+			Entry("inode byte quantity", &v1beta1.EvictionThreshold{NodeFsInodesFree: lo.ToPtr("100Mi")}, false),
+		)
+
+		DescribeTable("should validate evictionSoft", func(evictionSoft *v1beta1.EvictionThreshold, gracePeriods *v1beta1.EvictionSoftGracePeriod, expected bool) {
```
matthchr, line 148: More tests here than in v1alpha2 path? Shouldn't they have basically identical tests, since the structure is identical?

matthchr (approved):
LGTM. As discussed, need to hold off on merge for a while while Machine API finishes rollout

PR #1834, opened 2026-08-10
feat: add design for capacity reservations

tallaxes:
**Description**

Adds `designs/0014-capacity-reservations.md` — a proposal for consuming Azure Capacity Reservation Groups from Karpenter-managed clusters. Design only, no code changes.

An `AKSNodeClass` names one CRG, and every VM it launches targets that group. Nodes stay `on-demand` rather than `reserved`, because Azure overallocates instead of failing a launch and does not report which VM holds a reserved unit — so the `reserved` contract could not be verified or repaired. The increment covers association, readiness, offering projection, quota preflight relief, drift, and aggregate telemetry, plus an operator playbook for filling a reservation with existing NodePool controls. Reservation-first provisioning and reservation-aware consolidation are deliberately deferred to upstream core work, with the reasoning and phasing in the document.

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


designs/0014-capacity-reservations.md
```diff
@@ -0,0 +1,675 @@
+# Capacity reservations for Karpenter on Azure
+
+**Status:** proposal, for review.
+
```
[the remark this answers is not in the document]
tallaxes: What's wrong with line 431?

designs/0014-capacity-reservations.md
```diff
@@ -0,0 +1,675 @@
+# Capacity reservations for Karpenter on Azure
```
matthchr, line 1: Just be aware there are a few other 0014's as well in PR so we may need to deduplicate

designs/0014-capacity-reservations.md
```diff
@@ -0,0 +1,675 @@
...
+- [External dependencies](#external-dependencies)
+
+## Summary
+
+Let an `AKSNodeClass` name one Azure Capacity Reservation Group (CRG). Every VM
+that NodeClass launches targets that group, so a cluster managed by Karpenter
+can consume capacity the customer has already reserved and is already paying
+for.
+
+Nodes are labelled `karpenter.sh/capacity-type=on-demand`. Karpenter does not
```
matthchr, line 27: Is there another label they could/should be labelled? (maybe it will become clear as I get further through the doc).

designs/0014-capacity-reservations.md
```diff
@@ -0,0 +1,675 @@
...
+assert that a particular node occupies an SLA-backed reserved unit, because
+Azure does not report that per VM. Reservation state is published as aggregate
```
matthchr, line 29: let's raise this to CRP as an ask, if that would help?

designs/0014-capacity-reservations.md
```diff
@@ -0,0 +1,675 @@
...
+opt-in; there is no separate feature flag. The field is per-NodeClass rather
+than cluster-wide, so different NodeClasses can target different groups, or
+none. A group that cannot be read, is in
+the wrong region, has an unsupported type, or has no eligible member reservation
+matching any candidate VM size and placement makes the NodeClass unready rather
+than being silently ignored. A member reservation whose quantity is zero is
+valid: it is the documented way to associate VMs before raising the reserved
+quantity.
+
+**Member eligibility.** A member with an explicit provisioning state other than
```
matthchr, line 89: Are members nodes or nodeclaims?

This paragraph is somewhat confusing to me

designs/0014-capacity-reservations.md
```diff
@@ -0,0 +1,675 @@
...
+member reported as `Succeeded`, or with no provisioning state reported, is
+eligible. Treating a missing value as ineligible would turn an optional field
+into a provisioning outage, and the field is read-only and has been populated in
+every observed response; at worst, guessing wrong here can cost a launch attempt
+that ARM rejects, and it cannot produce an unassociated VM, because every
+request still targets the group. Readiness turns on whether any *eligible*
+member remains, not on the member list being non-empty — otherwise a group of
+entirely unusable members reports Ready and emits no offerings.
+
+Projection follows current state rather than holding a member at its last good
```
matthchr, line 101: This paragraph also confusign to me

designs/0014-capacity-reservations.md
```diff
@@ -0,0 +1,675 @@
...
+shape while it changes: no `Updating` window was observable in probing, a
+rejected quantity change leaves the member `Succeeded` and unchanged, and
+retaining state would make the reconciler stateful for a window nothing has
+demonstrated. Revisit if a successful quantity raise is seen to expose
+`Updating` for a material duration.
+
+**Permissions.** The verified minimum actions at the group's scope are:
+
+- `Microsoft.Compute/capacityReservationGroups/read`
+- `Microsoft.Compute/capacityReservationGroups/capacityReservations/read`
```
matthchr, line 111: Why do we need to read individual reservations?

designs/0014-capacity-reservations.md
```diff
@@ -0,0 +1,675 @@
...
+- `Microsoft.Compute/capacityReservationGroups/deploy/action`
+
+These names were checked against Azure's operation catalog, and a live probe
```
matthchr, line 114: This paragraph seems unnecessary. It is just saying "yeah I tested this" and confirmed this is all that is needed -- but IMO that's sorta implied by the fact that you're putting it in the design document as well.

matthchr: Maybe instead of this just clarify what the UX will be for this in NAP and in selfhosted Karp?

designs/0014-capacity-reservations.md
```diff
@@ -0,0 +1,675 @@
...
+That half surfaces only as a scoped launch failure, which is why the failure
+reason names the group.
+
+**Offerings.** For a configured NodeClass the provider emits only Regular
+offerings backed by a matching member reservation, and only in the group's
+placement: a zone-less group yields regional offerings, a zonal group yields
+exactly its zones. Spot and other configurations Azure does not support with
+capacity reservations are excluded.
+
+**Quota preflight.** Ordinary family-quota preflight can reject a launch that
```
matthchr, line 131: How does this work with quota provider? I guess we just can't include that signal at all if CRG is enabled?

matthchr: Though it sounds like SKU Mix Placement will take this into consideration eventually so when we get to the point where we just rely on that and drop the quota provider entirely this will be better again.

designs/0014-capacity-reservations.md
```diff
@@ -0,0 +1,675 @@
...
+Azure would have accepted, because capacity within a reservation does not
+consume incremental quota: that quota was taken when the reservation was
+created. For CRG-backed offerings that check is suppressed and Azure stays
+authoritative at launch. The cost of being wrong is bounded and local. If the
+reserved units are already taken, Azure does not reject the launch — it creates
+the VM as an overallocation, which does draw on ordinary quota. The launch fails
```
matthchr, line 137: Not sure I understand about how mis-pricing comes into the picture here...

designs/0014-capacity-reservations.md
```diff
@@ -0,0 +1,675 @@
...
+only when that quota or regional capacity is genuinely unavailable, and the
+failure is scoped to the group, VM size, and placement, costing a scheduling
+round trip rather than mispricing anything.
+
+**Launch.** The group ID goes in the VM's capacity reservation profile. Launch
```
matthchr, line 142: What if the CRG is overprovisioned... (so it's consuming normal quota) then if we don't mark ordinary capacity unavailable we could get stuck in a loop of provisioning failures?

Is this just a particular way of saying that CRG will have a separate entry per CRG in the unavailableofferings and we will track availability for it separately or something? That seems possibly unnecessary though... is the capacity from a CRG _guaranteed_ available? If so... then a failure to provision size X from CRG presumably means size X cannot be provisioned in standard allocations either?

It's the other way around that doesn't hold? Just because a size is unavailable in standard allocations doesn't mean it is unavailable in CRG?

designs/0014-capacity-reservations.md
```diff
@@ -0,0 +1,675 @@
...
+failures are scoped to that group, VM size, and placement, so a CRG-specific
+failure does not mark ordinary capacity unavailable.
+
+**Classification.** Nodes are `on-demand`. The configured group ID is recorded
```
matthchr, line 146: An annotation and not a label?

Why? Because label length restrictions?

matthchr: How do I as a user affinitize my pods to nodes from a CRG, is there any way?

designs/0014-capacity-reservations.md
```diff
@@ -0,0 +1,675 @@
...
+as an annotation, and the observed member reservation may be recorded as a
+second annotation marked as a point-in-time observation. Annotations do not
+participate in scheduling and can be corrected without replacing the node.
+
+**Telemetry.** Per member reservation: reserved quantity, allocated count,
```
matthchr, line 151: again somewhat confused by what a member is

designs/0014-capacity-reservations.md
```diff
@@ -0,0 +1,675 @@
...
+unused units, overallocated count, and observation age. This is what makes the
+size of the potential economic opportunity measurable rather than assumed — see
+[phases](#phases). These are deliberately aggregate signals: they do not claim
```
matthchr, line 154: minor: document ordering: Talking about phases but haven't actually discussed phases yet. I guess fine

designs/0014-capacity-reservations.md
matthchr: it might be helpful to have a section that walks through common failure cases for CRG and what exactly we do with them.

designs/0014-capacity-reservations.md
matthchr: Is it a CR in the CRG?

designs/0014-capacity-reservations.md
```diff
@@ -0,0 +1,675 @@
...
+associated with the group. For simpler setup, the built-in `Contributor` role
+at the group's scope is the known-good alternative, is what AKS asks users to
+grant today, and is what provider E2E tests use. A grant is usually required,
+because groups are commonly pre-created in a capacity team's own resource
+group. Note the asymmetry: readiness can verify read access but not deploy, so
+an identity with read-only access reports Ready and then fails every launch.
+That half surfaces only as a scoped launch failure, which is why the failure
+reason names the group.
+
+**Offerings.** For a configured NodeClass the provider emits only Regular
```
matthchr, line 125: Ahh so we are intersecting the NodePool's configured sizes with the CRGs configured sizes + zones to build our offering collection? That makes sense I guess.

designs/0014-capacity-reservations.md
```diff
@@ -0,0 +1,675 @@
...
+
+Keeping nodes `on-demand` is the conservative and truthful choice: it prices
+CRG-associated capacity at its worst-case marginal cost and makes no per-node
+guarantee.
+
+## What this does not do, and why
+
+Two capabilities are worth having and are not in this increment.
+
+**Reservation-first provisioning.** Ideally, when a group has unused reserved
```
matthchr, line 203: Does NodeOverlay allow for overlaying CRG-specific sizes? I assume probably not, the overlay applies to that size regardless?

designs/0014-capacity-reservations.md
```diff
@@ -0,0 +1,675 @@
...
+units, provisioning would prefer them over launching incremental capacity, even
+when the reserved VM size has a higher list price. Two levers could express that
+preference, and the same consolidation behavior bounds both.
+
+Core expresses preference through offering price, and price is also what
```
matthchr, line 208: It would be helpful (maybe there is a section below) where we discuss what we need from the compute API to make this preference possible.

When we generate offerings, if we "knew" that there was available space in the CRG and/or even if we didn't but could express on the individual VM allocation that we want a "guaranteed CRG VM" that failed provisioning if CRG for that size wasn't available, then we could generate 2x set of offerings, 1x for CRG sizes that cost 0 (already paid for), and 1x for overallocated sizes which cost more. Then Karp naturally prefers the cheaper options and everything is good.

This has the (possibly added) benefit of making CRG work very well on a single NodePool - we just try with the "must be CRG flag" first, and fall back to "overprovision OK" if that offering fails and users get a good experience in a single NodePool.

basically I want to understand what we are going to go to CRP with and ask for API changes/improvements.

matthchr: Though there may be additional warts with this pattern + VMs moving from on-demand to reserved dynamically, I am not sure.

Though... is there anything stopping us from updating the label dynamically on the node when that happens? We could in theory have a controller that did that. It would mean that the nodeClaim and the node differed though, but once the node exists I am not sure that matters?

designs/0014-capacity-reservations.md
```diff
@@ -0,0 +1,675 @@
...
+consolidation uses to value running nodes, so discounting an offering to attract
+provisioning under-prices every node in the pool, including any running beyond
+the reserved quantity. The provider has a second lever, since it re-derives the
+candidate set at launch and picks one, but as [what is left for the provider to
+do](#what-is-left-for-the-provider-to-do) explains, using that to prefer a more
+expensive reserved size runs into the same problem with no price change
+anywhere.
+
+**Reservation-aware consolidation.** Ideally, consolidation would move workloads
```
matthchr, line 217: Can you explain more about why it requires cost to be evaluated for the whole pool rather than per node? Assuming the labels on the nodes were correct (reserved vs on-demand), the fact that on-demand can become reserved later is somewhat immaterial to consolidation _now_, isn't it, especially considering the fact that we only consider offerings for a pool that are contained in the CRG?

i.e. if I have
VM A (CRG on-demand)
VM B (CRG reserved)
VM C (CRG on-demand)

and we correctly understand price as B = 0, A,C = $$, then consolidation would prefer to pack as much onto B as possible?

is the issue that without considering overall CRG usage (and in the absence of limits? Do we propagate CRG limits onto the NodePool explicitly? I guess no?) core would calculate the cost of a _new_ VM on CRG as 0 and so might consolidate A+C into D (from CRG) where D is actually not as cheap as core thinks it is?

Does AWS deal with this by attempting that consolidation but then allocation fails (CRG full) and so then it drops that from consolidation consideration and gives up since nothing is actually allocatable + cheaper? Would a similar CRG provider to quota provider help here, where we snapshot CRG usage (assuming that's possible) and exclude from offerings would make this more correct (except for snapshot latency?)

designs/0014-capacity-reservations.md
```diff
@@ -0,0 +1,675 @@
...
+whether or not a VM occupies it. Consolidation exists to stop paying for idle
+capacity, and against a reservation there is nothing to reclaim. That holds for
+the compute charge, which is the dominant term but not the only one — a running
+node still draws disks, networking, IP addresses, any per-node licensing and
+extensions, and operational attention. Standing capacity is close to free
+against a paid-for reservation, not free. The arrangements below follow from
+that observation.
+
+**First, one NodePool per member reservation.** A group is a container; the unit
+that carries a quantity is the member reservation, which is one VM size in one
```
matthchr, line 241: Ahh yes ok member is member of the CRG. Does that have an official ARM ID/name? is it really member? can we use that name here, or say "CRG member" everywhere just for clarity.

matthchr: Or have a diagram to start that explains exactly what we mean by member.

designs/0014-capacity-reservations.md
```diff
@@ -0,0 +1,675 @@
...
+Before reaching for anything reservation-specific: reserved capacity is paid for
+whether or not a VM occupies it. Consolidation exists to stop paying for idle
+capacity, and against a reservation there is nothing to reclaim. That holds for
+the compute charge, which is the dominant term but not the only one — a running
+node still draws disks, networking, IP addresses, any per-node licensing and
+extensions, and operational attention. Standing capacity is close to free
+against a paid-for reservation, not free. The arrangements below follow from
+that observation.
+
+**First, one NodePool per member reservation.** A group is a container; the unit
```
matthchr, line 240: What does "per member reservation" mean?

A member reservation is something like: D8v5 100 cores prepaid? Are you saying you have to have a single NodePool for D8v5, even if there is another member reservation for D8v6? That doesn't seem to make sense?

matthchr: Oh wait you spell this out here:
https://github.com/Azure/karpenter-provider-azure/pull/1834/changes#diff-a2151eda94cc1c75a975cd5cdba5842a0dc9d783f5acd5313086103896b6b108R253

But still not following... I can't have multiple sizes + zones in a single NodePool anymore?

matthchr: Oh I _can_ but this is an example of a different way to do it? really not clear what this section is talking about/why, TBH.

designs/0014-capacity-reservations.md
```diff
@@ -0,0 +1,675 @@
...
+  imageFamily: Ubuntu
+  capacityReservationGroupID: /subscriptions/<sub>/resourceGroups/<rg>/providers/Microsoft.Compute/capacityReservationGroups/prod-eastus
+```
+
+Each example below shows member 1. Members 2 and 3 differ only in name, zone,
+VM size, and quantity, except where an option says otherwise. For a regional
+group, pin `karpenter.azure.com/placement-scope: regional` rather than a zone;
+Azure reports regional nodes with `topology.kubernetes.io/zone: "0"`.
+
+**Stable baseline.** For a reservation bought to guarantee a steady baseline,
```
matthchr, line 277: Doesn't this approach assume that the CRG is entirely for this cluster?

Do we have any evidence that's the case? I would somewhat expect that users have fewer CRGs that may be shared among clusters...

AKS RP has data on this today in via AP Snapshots I think, could we gather data there and inform this approach? If 80% of users are sharing CRGs across APs or even across clusters I am not sure we want to push this as a recommended approach since it mismatches with what users actually do.

designs/0014-capacity-reservations.md
```diff
@@ -0,0 +1,675 @@
...
+partly idle.
+
+This is deliberately close to an AKS agent pool associated with the same group,
+which is fixed capacity pinned to a reservation and has been available all
+along. The difference is whose lifecycle it lives in: the static NodePool uses
+the same NodeClass, drift, and repair as the rest of the Karpenter-managed
+cluster, so the reserved baseline and the burst around it are configured and
+upgraded the same way.
+
+**Simple elastic.** When the baseline is not steady enough to pin a replica
```
matthchr, line 371: minor, style: there are markdown headings after `##` - consider using them here? =) This is a pretty big section to have under just a small bold heading here.

Applies a few other places too.

designs/0014-capacity-reservations.md
```diff
@@ -0,0 +1,675 @@
...
+  template:
+    spec:
+      nodeClassRef:
+        group: karpenter.azure.com
+        kind: AKSNodeClass
+        name: default     # no capacityReservationGroupID
+      requirements: []   # required field; empty admits anything
+```
+
+**Better elastic, with an overlay.** Weight decides which pool is tried first,
```
matthchr, line 434: Hmm... maybe you need a summary of these + links at the top of this section as well? Hard for me to follow what these sections are doing/showcasing and when a user might want to use one versus another.

designs/0014-capacity-reservations.md
```diff
@@ -0,0 +1,675 @@
...
+zero budget removes their nodes from consolidation candidacy altogether — their
+work does not move. If consolidating D8 work into D16 is also wanted, drop the
+budget from the D8 pools and accept that their reserved nodes may be vacated in
+the process.
+
+Note what the overlay is *not* doing here: the `Underutilized: 0` budget already
+prevents consolidation away from a reserved D16 node. The overlay is about
+moving *into* one.
+
+Agreement is not accuracy. The declared price is an operator's assertion that
```
matthchr, line 493: These sorts of things seem worth maybe a bulleted list somewhere rather than buried in what seems to be an examples section, since it calls out a fundamental limitation of the approach discussed rather than anything specific to the example?

designs/0014-capacity-reservations.md
```diff
@@ -0,0 +1,675 @@
...
+
+`spec.replicas` and `NodeOverlay` are both alpha. They are off by default in the
+Helm chart, and on by default in AKS Node Auto Provisioning, so on NAP clusters
+the arrangements above need no gate changes.
+
+## What is left for the provider to do
+
+Everything above is operator configuration. The one thing left that the provider
+could do on its own, with no core change, is let observed headroom influence
+which candidate it launches. That is bounded, for a reason worth stating once.
```
matthchr, line 527: "For a reason worth stating once" (and a bunch of other stuff in here) -- clearly AI-generated fluff. Not clear it adds value. Maybe do a pass and clear up/simplify language? Prefer short lists + bullet points with expository paragraphs only for things that require complex callouts?

Not 100% sure the best way to fix this without going over the document with a fine-tooth comb yourself but it feels longer/more verbose than necessary to me.

designs/0014-capacity-reservations.md
```diff
@@ -0,0 +1,675 @@
...
+
+The provider receives candidates core has already priced and ordered. Were it to
```
matthchr, line 529: This problem was discussed elsewhere in the document too. Maybe consolidate discussions about challenges/problems into a clear section? May be helpful to more closely follow the design template which has goals/non-goals section at the top (missing here I think?), decsions section where you discuss/consider alternatives (if appropriate -- it's also 100% fine to just state a single solution with no alternatives if there is no great alternative) and a FAQ section which maybe some of these "problems" could be repurposed into.

i.e. we decide we're going to just use on-demand for everything because of xyz reasons. Then FAQ outlines the various consequences of that succinctly in a Q+A style format?

https://github.com/Azure/karpenter-provider-azure/blob/main/designs/0013-ultra-ssd-support.md and https://github.com/Azure/karpenter-provider-azure/blob/main/designs/0009-regional-vm-provisioning.md both do this well, IMO

designs/0014-capacity-reservations.md
```diff
@@ -0,0 +1,675 @@
...
+> prices equally or more cheaply than the one it would otherwise have picked. It
+> must not use headroom to select a candidate that core prices higher.
+
+**What that leaves, honestly.** One benefit, conditional on one configuration
+choice. A pool pinned to one VM size and one placement offers a single
+candidate, so headroom cannot change the pick no matter what it observes — the
+operator has already chosen. Steering only has alternatives to choose among when
+a NodePool deliberately spans several member reservations.
+
+That is a trade, not a convenience. A multi-member pool gives up per-member
```
matthchr, line 548: More AI-generated text of meh value IMO

designs/0014-capacity-reservations.md
```diff
@@ -0,0 +1,675 @@
...
+| The protection belongs to one workload rather than to a whole pool. | `karpenter.sh/do-not-disrupt` on that workload's pods | Blocks consolidation and drift alike. If the NodePool sets `terminationGracePeriod`, drift may select the node immediately and the annotation only holds during draining, up to that deadline |
+
+`spec.replicas` and `NodeOverlay` are both alpha. They are off by default in the
+Helm chart, and on by default in AKS Node Auto Provisioning, so on NAP clusters
+the arrangements above need no gate changes.
+
+## What is left for the provider to do
+
+Everything above is operator configuration. The one thing left that the provider
+could do on its own, with no core change, is let observed headroom influence
```
matthchr, line 526: is "observed headroom" based on particular fields on the CRG API? Where? Can you be more clear about what exactly this is and where the data is coming from?

matthchr: This is the first occurrence of this phrase in the document and I have no idea what it means =)

designs/0014-capacity-reservations.md
```diff
@@ -0,0 +1,675 @@
...
+worth having then only if same-price stranding is actually costing something.
+
+Phase 1 telemetry narrows that question without settling it. Per-member counts
+can show member A sitting unused while member B is overallocated, which is
+evidence of candidate stranding — but not proof that steering would have helped,
+since topology spread, affinity, volume placement, or any other requirement may
+have excluded A from the launch in the first place. That aggregate pattern is
+enough to decide whether a closer measurement is worth making.
+
+If it is material, gate the feature with a bounded shadow study in the launch
```
matthchr, line 576: I have no idea what this paragraph is saying. What is a "bounded shadow study"? What is "equal-price-headroom-stranded"?

designs/0014-capacity-reservations.md
```diff
@@ -0,0 +1,675 @@
...
+   physical-zone translation. A same-subscription group may still be
+   nonexclusive, which Phase 1 already handles truthfully.
+5. **Upstream ownership.** If full elastic optimization is committed, may
+   Karpenter core evolve, and who owns the Phase 2 upstream work? The accounting
+   and cost-model changes may land separately, but they are one product
+   capability.
+
+## External dependencies
+
+| Dependency | Needed for | Effect while unavailable |
```
matthchr, line 670: I think I have another comment around this somewhere but missing section on what we want/need the platform to deliver to make this better.

designs/0014-capacity-reservations.md
matthchr: On second read maybe this is just saying: "If we want to consider headroom in the CRG during provisioning, we can track/log it and decide based on that logging if it would be valuable to consider for offering generation"?

PR #1835, opened 2026-08-11
design: choosing between NodeClass and Well-Known Labels

PabloTriv:
Fixes # 

**Description**
The main idea of this doc is to provide some guidance on when to makea new feature a well-known label and when to use NodeClass

*

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note

```


designs/0015-nodeclass-vs-labels.md
```diff
@@ -0,0 +1,90 @@
+# Labels vs. AKSNodeClass
```
matthchr, line 1: minor: heads up that other designs are 0014 as well so you might need to renumber based on those.

designs/0014-nodeclass-vs-labels
```diff
@@ -0,0 +1,90 @@
...
+
+**Last updated:** August 11, 2026
+
+**Status:** Guidance
+
+## Overview
+
+Karpenter for Azure has two main places where provider-specific behavior can be exposed to users:
+
+- Well-known labels can be used as scheduling requirements on workloads and NodePools.
```
matthchr: minor, style: could link to the code locations for both of these?

For example:

```suggestion
- [Well-known labels](linky-link) can be used as scheduling requirements on workloads and NodePools.
```

designs/0014-nodeclass-vs-labels
```diff
@@ -0,0 +1,90 @@
...
+- Strongly typed fields on `AKSNodeClass`, which describe how Azure capacity should be provisioned.
+
+Both surfaces are useful, but they solve different problems. A label is best when the user is making a scheduling decision: they need to express that a workload can run only on nodes with a particular capability, offering, operating mode, or provider-computed property. A NodeClass field is best when the user is configuring a feature: they are choosing how nodes should be created, often with multiple parameters, validation rules, defaults, or drift behavior.
+
+This document is guidance for new features being added to Karpenter for Azure. It defines the rule feature authors should use when deciding whether a new Azure concept belongs in a label or in `AKSNodeClass`.
```
matthchr: very minor/nit:
```suggestion
This document is guidance for new features being added to Karpenter for Azure. It defines the rules feature authors should use when deciding whether a new Azure concept belongs in a label or in `AKSNodeClass`.
```

designs/0014-nodeclass-vs-labels
```diff
@@ -0,0 +1,90 @@
...
+
+> Which nodes are compatible with this workload?
+
+This usually means the value participates in Karpenter requirements and changes which instance types, zones, offerings, or nodes are eligible for a `NodeClaim`.
+
+Choose a label when most of these are true:
+
+- The concept is a node capability, offering property, or provider-computed scheduling attribute.
+- Workloads reasonably need to select for it using `nodeSelector`, node affinity, or `NodePool` requirements.
+- The provider can safely expose a default value for all relevant offerings, such as `true` or `false`.
```
matthchr: Is this strictly required/necessary? We also support the exists/doesnotexist dichotomy too don't we?

designs/0015-nodeclass-vs-labels.md
```diff
@@ -0,0 +1,90 @@
...
+
+Use a well-known label when the primary user question is:
+
+> Which nodes are compatible with this workload?
+
+This usually means the value participates in Karpenter requirements and changes which instance types, zones, offerings, or nodes are eligible for a `NodeClaim`.
+
+Choose a label when most of these are true:
+
+- The concept is a node capability, offering property, or provider-computed scheduling attribute.
```
matthchr: Not sure that "node capability" is actually that clear of a line, because there are node capabilities that maybe make sense to schedule against and others that maybe dont (or am I misunderstanding what you mean by node capability?) - plus probably everything that is in the label or AKSNodeClass is in some way a "node capability".

The scheduling aspect here is covered by the next bullet point as well so could maybe be dropped from this one?

designs/0015-nodeclass-vs-labels.md
```diff
@@ -0,0 +1,90 @@
...
+- Workloads reasonably need to select for it using `nodeSelector`, node affinity, or `NodePool` requirements.
+- The provider can safely expose a default value for all relevant offerings, such as `true` or `false`.
+- Omitting the requirement should preserve normal behavior.
+- The value is simple enough to model as a label value without losing important structure.
```
matthchr, line 47: In addition to simple enough (which I 100% agree is a key requirement), I think there is something about "complexity of validation" too - i.e. if the field is very simple but it can only be enabled when managedCluster is in state X, and other AKSNodeClass fields are in state Y, Z, then it is better as a AKSNodeClass property because we can surface more complex validation there. So I think we should say something about: "The value may restrict what other VM sizes/features are used, but those restrictions are simple or easy to understand. Overly complex or interrelated validations should be avoided in labels as the failure mode (no matching instance types) is harder to understand versus clear AKSNodeClass status error"

designs/0014-nodeclass-vs-labels
```diff
@@ -0,0 +1,90 @@
...
+Use a well-known label when the primary user question is:
+
+> Which nodes are compatible with this workload?
+
+This usually means the value participates in Karpenter requirements and changes which instance types, zones, offerings, or nodes are eligible for a `NodeClaim`.
+
+Choose a label when most of these are true:
+
+- The concept is a node capability, offering property, or provider-computed scheduling attribute.
+- Workloads reasonably need to select for it using `nodeSelector`, node affinity, or `NodePool` requirements.
```
matthchr: Consider adding topology spreading into this list as well.

designs/0014-nodeclass-vs-labels
```diff
@@ -0,0 +1,90 @@
...
+- The provider can safely expose a default value for all relevant offerings, such as `true` or `false`.
+- Omitting the requirement should preserve normal behavior.
```
matthchr: Isn't this sorta true for both AKSNodeClass and labels though?

designs/0014-nodeclass-vs-labels
```diff
@@ -0,0 +1,90 @@
...
+- The value is simple enough to model as a label value without losing important structure.
+
+For these cases, labels are the right API because they are part of Kubernetes and Karpenter scheduling semantics. They let users express compatibility without creating extra NodeClasses or encoding workload placement in provisioning configuration.
+
+Examples:
+
+- `topology.kubernetes.io/zone`, because zone is a placement constraint that directly changes which offerings can satisfy a workload.
```
matthchr: and pods need to schedule against it/set topology constraints against it

designs/0015-nodeclass-vs-labels.md
```diff
@@ -0,0 +1,90 @@
...
+- The value is not useful as a pod-level scheduling constraint.
+
+For these cases, NodeClass is the right API because labels are stringly typed and intentionally limited. A field on `AKSNodeClass` gives us a typed schema, API validation, clearer defaults, and room to evolve the feature without creating many loosely related labels.
+
+Examples:
+
+- Disk encryption configuration, because it involves resource IDs, precedence between cluster defaults and NodeClass overrides, mutability, and drift behavior.
+- Networking configuration, because it can include subscriptions, resource groups, subnet names, dataplane choices, and multi-tenancy settings that should not be encoded into string labels.
+- Trusted Launch configuration, because it has related settings such as vTPM and Secure Boot that need typed configuration and validation. It is primarily concerned with how Azure creates the VM and does not describe a pod-level scheduling compatibility requirement.
+
```
matthchr, line 82: I think missing a section on "choose both", which we also sometimes do? I'm pretty sure we have some examples of that.

For "choose both" I think usually AKSNodeClass drives, but there is a label whose value is 100% derived from the nodeclass that is then considered for scheduling.

matthchr: and we should probably point out that the "do both" approach is not the preferred approach and should only be used when one of the above 2 doesn't work by itself.

designs/0015-nodeclass-vs-labels.md
matthchr: Oh maybe "node capability" means "intrinsic property of the VM size/instance type" and as such is tightly coupled/related to VM size?

examples: gpu support, disk capabilities?

designs/0014-nodeclass-vs-labels
```diff
@@ -0,0 +1,90 @@
...
+
+> How should Azure create or configure this node?
+
+This usually means the option is part of the node provisioning contract rather than a workload placement constraint.
+
+Choose `AKSNodeClass` when most of these are true:
+
+- The concept enables or configures a feature rather than only describing compatibility.
+- The feature has multiple fields, modes, or possible values.
+- The setting needs strong typing, defaulting, validation, or clear API documentation.
```
matthchr: minor: expand a bit as the opposite of https://github.com/Azure/karpenter-provider-azure/pull/1835/changes#r3771515144, complex validation or complex defaulting?

designs/0014-nodeclass-vs-labels
```diff
@@ -0,0 +1,90 @@
...
+Use `AKSNodeClass` when the primary user question is:
+
+> How should Azure create or configure this node?
+
+This usually means the option is part of the node provisioning contract rather than a workload placement constraint.
+
+Choose `AKSNodeClass` when most of these are true:
+
+- The concept enables or configures a feature rather than only describing compatibility.
+- The feature has multiple fields, modes, or possible values.
```
matthchr:
```suggestion
- The feature has multiple interrelated fields or modes
```

multiple fields/modes/values could go for either, couldn't it? Since some labels such as disk size definitely have many possible values? Even multiple interrelated fields isn't necessarily a huge problem (see for example GPU labels, there are a bunch of interrelated ones IIRC).

I wonder if this is just saying the same as the next item around: "the more complex the feature, with more complicated conditional defaulting and/or validation, the more likely it should be on the AKSNodeClass" -- example LocalDNS.

designs/0014-nodeclass-vs-labels
matthchr: Consider slipstreaming in something about:
It requires complex structured data — Selectors with tags, nested objects like block device mappings, or multi-field configurations don't fit the simple key-value model of labels.

matthchr: It's a reference to an external cloud resource — Subnet selectors, security group selectors, AMI selectors, placement group selectors — these resolve external resources and appear in the NodeClass status.

(Similar to above, don't need to take all of these just trying to give examples/ideas that can be included here).

designs/0014-nodeclass-vs-labels
```diff
@@ -0,0 +1,90 @@
...
+
+## Rule of Thumb
+
+If it is a scheduling decision, and it is something that can be on by default or represented as a simple capability, make it a well-known label.
+
+If it is a feature, especially a complex feature with many possible values or related settings, make it an `AKSNodeClass` field because the hard-typed API is better.
+
+## Final Note: AKS Precedent
+
+Feature design should generally try to mimic how the same capability works in AKS. AKS behavior is useful precedent because it gives users a familiar model and helps keep provider behavior aligned with the platform. However, it should not be treated as the end-all-be-all; Karpenter may expose a capability differently when scheduling semantics, API ergonomics, or provider constraints justify it.
```
matthchr: minor: add EoL at EoF?

designs/0015-nodeclass-vs-labels.md
PabloTriv: Yes to the second comment. The example I was thinking of was `sku-storage-premium-capable`, but those other examples you listed work as well. The idea was to pick a label if the feature to passively say that the VM size is capable of it. (Although I do agree the wording is misleading here as basically everything is a node capability)

designs/0014-nodeclass-vs-labels
PabloTriv: Yeah that's the intent, rewording for clarity

designs/0014-nodeclass-vs-labels
PabloTriv: Added more clarity and included LocalDNS as an example of complexity.

designs/0015-nodeclass-vs-labels.md
matthchr: Maybe rephrase to talk about "property of the instance type" and give some examples (things like memory, sku-storage-premium-capable, etc)?

designs/0015-nodeclass-vs-labels.md
```diff
@@ -0,0 +1,96 @@
...
+- The value is not useful as a pod-level scheduling constraint.
+
+For these cases, NodeClass is the right API because labels are stringly typed and intentionally limited. A field on `AKSNodeClass` gives us a typed schema, API validation, clearer defaults, and room to evolve the feature without creating many loosely related labels.
+
+Examples:
+
+- Disk encryption configuration, because it involves resource IDs, precedence between cluster defaults and NodeClass overrides, mutability, and drift behavior.
+- Networking configuration, because it can include subscriptions, resource groups, subnet names, dataplane choices, and multi-tenancy settings that should not be encoded into string labels.
+- Trusted Launch configuration, because it has related settings such as vTPM and Secure Boot that need typed configuration and validation. It is primarily concerned with how Azure creates the VM and does not describe a pod-level scheduling compatibility requirement.
+- LocalDNS configuration, because it includes modes, nested DNS override objects, forwarding policy, cache settings, and validation that depends on the whole configuration.
```
matthchr:
```suggestion
- LocalDNS configuration, because it is complex and includes modes, nested DNS override objects, forwarding policy, cache settings, and validation that depends on the whole configuration.
```

matthchr (dismissed):

matthchr (approved):

PR #1841, opened 2026-08-13
feat: support Azure Container Linux image family

MaximilianoUribe:
Add `AzureContainerLinux` as an `AKSNodeClass` image family for AKS Machine API provision modes only.

**Design flow**

1. **NodeClass selection**
   - Users select `spec.imageFamily: AzureContainerLinux`.
   - Admission validation enforces the 60 GB minimum OS disk and prevents disabling Secure Boot or vTPM.
   - Runtime validation accepts `aksmachineapi` and `aksmachineapiheaderbatch`; other provision modes are rejected.

2. **Image resolution**
   - Karpenter selects the Azure Container Linux image definition from the existing Azure Linux gallery.
   - Architecture and FIPS determine one of four variants: AMD64, ARM64, AMD64 FIPS, or ARM64 FIPS.
   - All variants require Gen2 and Trusted Launch-capable instance types.

3. **Instance-type filtering**
   - Trusted Launch-incompatible SKUs are removed from consideration.
   - GPU nodes use the Azure Container Linux compatibility list; supported NVIDIA SKUs remain available while unsupported NVIDIA and AMD SKUs are excluded.

4. **Machine creation**
   - AKS Machine API receives `osSKU=AzureContainerLinux` and the resolved node image version.
   - Secure Boot and vTPM are enabled automatically.
   - Direct-VM bootstrap modes are out of scope because ACL requires Ignition generated by the AKS Machine API path.

- Affected unit suites pass after rebasing onto current `main`.
- Three fresh Machine API clusters reached NodeClass/NodeClaim/node Ready with workloads, CNS, and Cilium healthy.
- Trusted Launch, Secure Boot, vTPM, and AgentBaker Phase-2 completion were verified.

- [ ] Yes, issue opened: #
- [x] No

**Release Note**

```release-note
Add Azure Container Linux support to AKSNodeClass when using AKS Machine API provision modes.
```

### DIFF
--- a/charts/karpenter-crd/templates/karpenter.azure.com_aksnodeclasses.yaml
+++ b/charts/karpenter-crd/templates/karpenter.azure.com_aksnodeclasses.yaml
@@ -105,7 +105,6 @@ spec:
                 - Ubuntu2204
                 - Ubuntu2404
                 - AzureLinux
-                - AzureContainerLinux
                 type: string
               kubelet:
                 description: |-
@@ -799,22 +798,14 @@ spec:
                 && ((has(self.security.trustedLaunch.vtpm) && self.security.trustedLaunch.vtpm)
                 || (has(self.security.trustedLaunch.secureBoot) && self.security.trustedLaunch.secureBoot)))))
                 : true'
-            - message: TrustedLaunch with FIPSMode FIPS is only supported for Ubuntu,
-                Ubuntu2204, and AzureContainerLinux
+            - message: TrustedLaunch with FIPSMode FIPS is only supported for Ubuntu
+                and Ubuntu2204
               rule: 'has(self.fipsMode) && self.fipsMode == ''FIPS'' && has(self.security)
                 && has(self.security.trustedLaunch) && ((has(self.security.trustedLaunch.vtpm)
                 && self.security.trustedLaunch.vtpm) || (has(self.security.trustedLaunch.secureBoot)
                 && self.security.trustedLaunch.secureBoot)) ? (!has(self.imageFamily)
-                || self.imageFamily == ''Ubuntu'' || self.imageFamily == ''Ubuntu2204''
-                || self.imageFamily == ''AzureContainerLinux'') : true'
-            - message: AzureContainerLinux requires Secure Boot and vTPM
-              rule: '!has(self.imageFamily) || self.imageFamily != ''AzureContainerLinux''
-                || !has(self.security) || !has(self.security.trustedLaunch) || ((!has(self.security.trustedLaunch.vtpm)
-                || self.security.trustedLaunch.vtpm) && (!has(self.security.trustedLaunch.secureBoot)
-                || self.security.trustedLaunch.secureBoot))'
-            - message: AzureContainerLinux requires an OS disk of at least 60 GB
-              rule: '!has(self.imageFamily) || self.imageFamily != ''AzureContainerLinux''
-                || !has(self.osDiskSizeGB) || self.osDiskSizeGB >= 60'
+                || self.imageFamily == ''Ubuntu'' || self.imageFamily == ''Ubuntu2204'')
+                : true'
             - message: kubelet.failSwapOn must be set to false when linuxOSConfig.swapFileSize
                 is specified
               rule: '!has(self.linuxOSConfig) || !has(self.linuxOSConfig.swapFileSize)
--- a/charts/karpenter-crd/templates/karpenter.azure.com_aksnodeclasses.yaml
+++ b/charts/karpenter-crd/templates/karpenter.azure.com_aksnodeclasses.yaml
@@ -105,6 +105,7 @@ spec:
                 - Ubuntu2204
                 - Ubuntu2404
                 - AzureLinux
+                - AzureContainerLinux
                 type: string
               kubelet:
                 description: |-
@@ -798,14 +799,22 @@ spec:
                 && ((has(self.security.trustedLaunch.vtpm) && self.security.trustedLaunch.vtpm)
                 || (has(self.security.trustedLaunch.secureBoot) && self.security.trustedLaunch.secureBoot)))))
                 : true'
-            - message: TrustedLaunch with FIPSMode FIPS is only supported for Ubuntu
-                and Ubuntu2204
+            - message: TrustedLaunch with FIPSMode FIPS is only supported for Ubuntu,
+                Ubuntu2204, and AzureContainerLinux
               rule: 'has(self.fipsMode) && self.fipsMode == ''FIPS'' && has(self.security)
                 && has(self.security.trustedLaunch) && ((has(self.security.trustedLaunch.vtpm)
                 && self.security.trustedLaunch.vtpm) || (has(self.security.trustedLaunch.secureBoot)
                 && self.security.trustedLaunch.secureBoot)) ? (!has(self.imageFamily)
-                || self.imageFamily == ''Ubuntu'' || self.imageFamily == ''Ubuntu2204'')
-                : true'
+                || self.imageFamily == ''Ubuntu'' || self.imageFamily == ''Ubuntu2204''
+                || self.imageFamily == ''AzureContainerLinux'') : true'
+            - message: AzureContainerLinux requires Secure Boot and vTPM
+              rule: '!has(self.imageFamily) || self.imageFamily != ''AzureContainerLinux''
+                || !has(self.security) || !has(self.security.trustedLaunch) || ((!has(self.security.trustedLaunch.vtpm)
+                || self.security.trustedLaunch.vtpm) && (!has(self.security.trustedLaunch.secureBoot)
+                || self.security.trustedLaunch.secureBoot))'
+            - message: AzureContainerLinux requires an OS disk of at least 60 GB
+              rule: '!has(self.imageFamily) || self.imageFamily != ''AzureContainerLinux''
+                || !has(self.osDiskSizeGB) || self.osDiskSizeGB >= 60'
             - message: kubelet.failSwapOn must be set to false when linuxOSConfig.swapFileSize
                 is specified
               rule: '!has(self.linuxOSConfig) || !has(self.linuxOSConfig.swapFileSize)
@@ -1031,6 +1040,7 @@ spec:
                 - Ubuntu2204
                 - Ubuntu2404
                 - AzureLinux
+                - AzureContainerLinux
                 type: string
               kubelet:
                 description: |-
@@ -1724,14 +1734,22 @@ spec:
                 && ((has(self.security.trustedLaunch.vtpm) && self.security.trustedLaunch.vtpm)
                 || (has(self.security.trustedLaunch.secureBoot) && self.security.trustedLaunch.secureBoot)))))
                 : true'
-            - message: TrustedLaunch with FIPSMode FIPS is only supported for Ubuntu
-                and Ubuntu2204
+            - message: TrustedLaunch with FIPSMode FIPS is only supported for Ubuntu,
+                Ubuntu2204, and AzureContainerLinux
               rule: 'has(self.fipsMode) && self.fipsMode == ''FIPS'' && has(self.security)
                 && has(self.security.trustedLaunch) && ((has(self.security.trustedLaunch.vtpm)
                 && self.security.trustedLaunch.vtpm) || (has(self.security.trustedLaunch.secureBoot)
                 && self.security.trustedLaunch.secureBoot)) ? (!has(self.imageFamily)
-                || self.imageFamily == ''Ubuntu'' || self.imageFamily == ''Ubuntu2204'')
-                : true'
+                || self.imageFamily == ''Ubuntu'' || self.imageFamily == ''Ubuntu2204''
+                || self.imageFamily == ''AzureContainerLinux'') : true'
+            - message: AzureContainerLinux requires Secure Boot and vTPM
+              rule: '!has(self.imageFamily) || self.imageFamily != ''AzureContainerLinux''
+                || !has(self.security) || !has(self.security.trustedLaunch) || ((!has(self.security.trustedLaunch.vtpm)
+                || self.security.trustedLaunch.vtpm) && (!has(self.security.trustedLaunch.secureBoot)
+                || self.security.trustedLaunch.secureBoot))'
+            - message: AzureContainerLinux requires an OS disk of at least 60 GB
+              rule: '!has(self.imageFamily) || self.imageFamily != ''AzureContainerLinux''
+                || !has(self.osDiskSizeGB) || self.osDiskSizeGB >= 60'
             - message: kubelet.failSwapOn must be set to false when linuxOSConfig.swapFileSize
                 is specified
               rule: '!has(self.linuxOSConfig) || !has(self.linuxOSConfig.swapFileSize)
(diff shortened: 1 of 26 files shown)


MaximilianoUribe:
(AI-generated)

RP-side `bootstrappingclient` E2E exposed and fixed nested ACL Ignition token hydration in Karpenter. Node Provisioner returns the NBC script inside a gzip+Base64 Ignition file; hydration now replaces bootstrap-token placeholders inside those files while preserving the full Ignition document.

Validated end to end with Karpenter direct VM provisioning: NodeBootstrapping HTTP 200, Compute VM + CSE created directly, no AKS Machine annotation, NodeClaim/node Ready, workload Running, CNS/Cilium healthy, Trusted Launch ACL image, node-controller/kubelet active, and CNI writable.

KyleGospo (changes requested):
Changes look good to me, no complaints about code quality and this both builds and tests well locally.

@MaximilianoUribe my only concern is the same issue that Copilot found. Are the dropped SKUs intentional?

pkg/utils/supported-gpus.yaml
```diff
@@ -118,27 +112,21 @@ standard_nc6s_v2:
...
 standard_nd96amsr_a100_v4:
   gpu: nvidia
-  os: ["ubuntu", "windows", "azurelinux3"]
-standard_nd112amsr_a100_v4:
-  gpu: nvidia
-  os: ["ubuntu", "windows", "azurelinux3"]
-standard_nd120amsr_a100_v4:
-  gpu: nvidia
-  os: ["ubuntu", "windows", "azurelinux3"]
+  os: ["ubuntu", "windows", "azurelinux3", "azurecontainerlinux"]
```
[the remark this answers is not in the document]
MaximilianoUribe: The entries were dropped while synchronizing this file with the current AKS RP GPU map, where these SKUs were removed as unusable. That cleanup is unrelated to ACL and would change existing Karpenter behavior, so I restored both entries with their previous OS list and intentionally did not add `azurecontainerlinux`. I also added regression coverage. We will track RP parity/removal separately.

pkg/utils/supported-gpus.yaml
```diff
@@ -1,77 +1,71 @@
...
 standard_nd96asr_v4:
   gpu: nvidia
-  os: ["ubuntu", "windows", "azurelinux3"]
-standard_nd112asr_a100_v4:
-  gpu: nvidia
-  os: ["ubuntu", "windows", "azurelinux3"]
-standard_nd120asr_a100_v4:
-  gpu: nvidia
-  os: ["ubuntu", "windows", "azurelinux3"]
+  os: ["ubuntu", "windows", "azurelinux3", "azurecontainerlinux"]
```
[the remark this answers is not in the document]
MaximilianoUribe: The entries were dropped while synchronizing this file with the current AKS RP GPU map, where these SKUs were removed as unusable. That cleanup is unrelated to ACL and would change existing Karpenter behavior, so I restored both entries with their previous OS list and intentionally did not add `azurecontainerlinux`. I also added regression coverage. We will track RP parity/removal separately.

KyleGospo (approved):
Changes look good to me, thanks for moving the servers to a separate PR.

charts/karpenter-crd/templates/karpenter.azure.com_aksnodeclasses.yaml
```diff
@@ -1724,14 +1725,22 @@ spec:
...
-                || self.imageFamily == ''Ubuntu'' || self.imageFamily == ''Ubuntu2204'')
-                : true'
+                || self.imageFamily == ''Ubuntu'' || self.imageFamily == ''Ubuntu2204''
+                || self.imageFamily == ''AzureContainerLinux'') : true'
+            - message: AzureContainerLinux requires Secure Boot and vTPM
+              rule: '!has(self.imageFamily) || self.imageFamily != ''AzureContainerLinux''
+                || !has(self.security) || !has(self.security.trustedLaunch) || ((!has(self.security.trustedLaunch.vtpm)
+                || self.security.trustedLaunch.vtpm) && (!has(self.security.trustedLaunch.secureBoot)
+                || self.security.trustedLaunch.secureBoot))'
+            - message: AzureContainerLinux requires an OS disk of at least 60 GB
```
matthchr, line 1794: Should confirm with @comtalyst if this impacts any of the disk size work he is doing. I assume this impacts our VM size filtering elsewhere in this PR too, as we have to filter out sizes that don't support OS Disk of 60GB?

pkg/apis/v1beta1/aksnodeclass.go
```diff
@@ -735,13 +737,19 @@ func (in *AKSNodeClass) GetEncryptionAtHost() bool {
 }
 
 func (in *AKSNodeClass) IsVTPMEnabled() bool {
+	if lo.FromPtr(in.Spec.ImageFamily) == AzureContainerLinuxImageFamily {
```
matthchr: @tallaxes thoughts on this? Do we require users to explicitly opt-in to the vTPM fields, or do we do this and dynamically default based on OS type if they weren't set explicitly?

pkg/apis/v1beta1/aksnodeclass.go
```diff
@@ -62,7 +62,9 @@ func (a *ArtifactStreaming) IsEnabled(arch string) bool {
...
+// +kubebuilder:validation:XValidation:message="AzureContainerLinux requires Secure Boot and vTPM",rule="!has(self.imageFamily) || self.imageFamily != 'AzureContainerLinux' || !has(self.security) || !has(self.security.trustedLaunch) || ((!has(self.security.trustedLaunch.vtpm) || self.security.trustedLaunch.vtpm) && (!has(self.security.trustedLaunch.secureBoot) || self.security.trustedLaunch.secureBoot))"
+// +kubebuilder:validation:XValidation:message="AzureContainerLinux requires an OS disk of at least 60 GB",rule="!has(self.imageFamily) || self.imageFamily != 'AzureContainerLinux' || !has(self.osDiskSizeGB) || self.osDiskSizeGB >= 60"
 // +kubebuilder:validation:XValidation:message="kubelet.failSwapOn must be set to false when linuxOSConfig.swapFileSize is specified",rule="!has(self.linuxOSConfig) || !has(self.linuxOSConfig.swapFileSize) || (has(self.kubelet) && has(self.kubelet.failSwapOn) && self.kubelet.failSwapOn == false)"
 type AKSNodeClassSpec struct {
```
matthchr, line 85: Missing changes for v1alpha2.

pkg/controllers/nodeclass/status/validation_test.go
```diff
@@ -64,7 +66,7 @@ var _ = Describe("Validation Reconciler", func() {
 	var emptyDiskEncryptionSetID *arm.ResourceID
 
 	BeforeEach(func() {
-		ctx = context.Background()
+		ctx = options.ToContext(context.Background(), &options.Options{ProvisionMode: consts.ProvisionModeAKSMachineAPI})
```
matthchr: Seems incorrect to do (for now?)

If we require this just do it for the test that requires it rather than all of them?

pkg/fake/nodeimageversionsapi.go
matthchr, line 39: Did you actually do this or manually include the below images? Did you at least confirm that this data you put here matches what we're getting from the API used by `make az-codegen-nodeimageversions`?

pkg/providers/imagefamily/azurecontainerlinux.go
```diff
@@ -0,0 +1,117 @@
...
+type AzureContainerLinux struct {
+	Options *parameters.StaticParameters
+}
+
+func (a AzureContainerLinux) Name() string {
+	return v1beta1.AzureContainerLinuxImageFamily
+}
+
+func (a AzureContainerLinux) DefaultImages(_ bool, fipsMode *v1beta1.FIPSMode, _ bool) []types.DefaultImageOutput {
+	if lo.FromPtr(fipsMode) == v1beta1.FIPSModeFIPS {
```
matthchr, line 57: should have guards here for trustedLaunch false that explicitly returns no images. Same for Kata?
What about SIG?

As-is this code doesn't seem right to me because it will return TL images even when the cluster is not configured for TL (though I know that's not technically possilbe because we guard against it in CEL). Other things like "not on SIG" though probably are technically possible

pkg/providers/imagefamily/nodeimage_test.go
matthchr, line 111: Do we support CIG for ACL? I don't think we do actually so I dont think the test here is right?

pkg/providers/imagefamily/resolver.go
```diff
@@ -156,6 +156,10 @@ func (r *defaultResolver) Resolve(
 		vtpmEnabled = nodeClass.Spec.Security.TrustedLaunch.VTPM
 		secureBootEnabled = nodeClass.Spec.Security.TrustedLaunch.SecureBoot
 	}
+	if lo.FromPtr(nodeClass.Spec.ImageFamily) == v1beta1.AzureContainerLinuxImageFamily {
```
matthchr: As mentioned above in the CRD types, I am not sure if we want to do this or just make users set these explicitly. I see pros/cons both ways... Curious @tallaxes opinion.

pkg/providers/instancetype/instancetypes.go
```diff
@@ -361,6 +361,8 @@ func (p *DefaultProvider) isInstanceTypeSupportedByImageFamily(skuName, imageFam
 	case imageFamily == v1beta1.AzureLinuxImageFamily:
 		return utils.IsGPUSKUSupportedOnOS(skuName, "azurelinux") ||
 			utils.IsGPUSKUSupportedOnOS(skuName, "azurelinux3")
+	case imageFamily == v1beta1.AzureContainerLinuxImageFamily:
```
matthchr: Do we need to update our support file to add the fact that some GPUs are supported on ACL, or are none supported?

pkg/utils/gpu_test.go
```diff
@@ -176,6 +176,33 @@ func TestIsAMDEnabledSKU(t *testing.T) {
...
+	g.Expect(IsGPUSKUSupportedOnOS("standard_nc24ads_a100_v4", "azurecontainerlinux")).To(BeTrue())
+	g.Expect(IsGPUSKUSupportedOnOS("standard_nd112asr_a100_v4", "azurecontainerlinux")).To(BeFalse())
+	g.Expect(IsGPUSKUSupportedOnOS("standard_nd120asr_a100_v4", "azurecontainerlinux")).To(BeFalse())
+	g.Expect(IsGPUSKUSupportedOnOS("standard_nd112amsr_a100_v4", "azurecontainerlinux")).To(BeFalse())
+	g.Expect(IsGPUSKUSupportedOnOS("standard_nd120amsr_a100_v4", "azurecontainerlinux")).To(BeFalse())
+	g.Expect(IsGPUSKUSupportedOnOS("standard_nc144ds_xl_rtxpro6000bse_v6", "azurecontainerlinux")).To(BeFalse())
+	g.Expect(IsGPUSKUSupportedOnOS("standard_nv4ads_v710_v5", "azurecontainerlinux")).To(BeFalse())
+}
+
+func TestLegacyA100GPUSupport(t *testing.T) {
```
matthchr: Is this test actually related?

pkg/utils/supported-gpus.yaml
```diff
@@ -1,51 +1,51 @@
 #  StandardNCADSA10v4Family:
 standard_nc16ads_a10_v4:
   gpu: nvidia
-  os: ["ubuntu", "windows", "azurelinux3"]
+  os: ["ubuntu", "windows", "azurelinux3", "azurecontainerlinux"]
```
matthchr, line 4: Just confirming, we know that this actually is supported is that right? (Same for all the below?)

pkg/apis/v1beta1/aksnodeclass.go
MaximilianoUribe: I am confused with this, thought we were going in the direction of not updating this anymore?

pkg/controllers/nodeclass/status/validation_test.go
MaximilianoUribe: yes you are right, maybe not in scope for this pr, will just update the testing for this pr

pkg/providers/imagefamily/azurecontainerlinux.go
MaximilianoUribe: yeah, as you said it is technically not possible, but I do think we should be doing some guard an not listing images that are not valid, will update

pkg/providers/imagefamily/resolver.go
MaximilianoUribe: I am more inclined on doing this? seems more straight forward to maintain and we could be avoiding some edge cases... but interested in what @tallaxes thinks

pkg/utils/gpu_test.go
MaximilianoUribe: forgot to delete when not updating the gpu list :/

pkg/utils/supported-gpus.yaml
MaximilianoUribe: yes, this is out current provider updated list

charts/karpenter-crd/templates/karpenter.azure.com_aksnodeclasses.yaml
comtalyst: Currently we either default OS disk size to 128 (survives this) or let the user pins it (this validation catches it once they enable this). I don't think there will be immediate issue, and no need for filtering?

After #1723 there could be an issue though, as OS disk size default will varies (in most cases >= 128, but not guaranteed IIRC). Shouldn't be too hard(?) for the solution. E.g., just not default below 60.

Have anyone looked at how AKS API defaults OS disk size in this ACL case?

pkg/apis/v1beta1/aksnodeclass.go
matthchr: We should update v1alpha2 until it's actually removed (code gone). Better safe than sorry (I dont think forgetting it will hurt but I am not 100% sure that is true in all clusters/all cases)

MaximilianoUribe: ok, will update

pkg/apis/v1beta1/aksnodeclass.go
MaximilianoUribe: i am in favor of doing this dynamically default based on os type as you mentioned, implementation for that is in the latest iteration

pkg/fake/nodeimageversionsapi.go
MaximilianoUribe: removed

pkg/controllers/nodeclass/status/validation.go
```diff
@@ -70,6 +71,14 @@ func NewValidationReconciler(
 
 func (r *ValidationReconciler) Reconcile(ctx context.Context, nodeClass *v1beta1.AKSNodeClass) (reconcile.Result, error) {
 	logger := log.FromContext(ctx)
+	if incompatibleACLProvisionMode(ctx, nodeClass) {
+		nodeClass.StatusConditions().SetFalse(
+			v1beta1.ConditionTypeValidationSucceeded,
+			IncompatibleProvisionMode,
+			"AzureContainerLinux requires an AKS Machine API provision mode",
```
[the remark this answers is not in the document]
MaximilianoUribe: good find, will fix

PR #1842, opened 2026-08-13
feat: enforce the AKS cluster-level FIPS setting

MaximilianoUribe:
## Description

Adds support for enforcing the AKS cluster-level FIPS setting in managed NAP.

When FIPS is enabled for the cluster, Karpenter accepts only `AKSNodeClass` resources that explicitly set `spec.fipsMode: FIPS`. NodeClasses with an unset or `Disabled` FIPS mode become not ready and cannot provision new nodes.

## Design

This implementation mirrors the existing AKS cluster-level setting:

```yaml
properties:
  enableFIPS: true
```

Managed NAP passes the setting to Karpenter through `ENABLE_FIPS`. The option is a boolean to preserve the semantics of the cluster API and defaults to `false`, leaving self-hosted installations and clusters without FIPS enforcement unchanged.

When `ENABLE_FIPS=true`, Karpenter requires:

```yaml
apiVersion: karpenter.azure.com/v1beta1
kind: AKSNodeClass
spec:
  fipsMode: FIPS
```

Both an unset `fipsMode` and an explicit `fipsMode: Disabled` are rejected. Karpenter does not dynamically default an unset NodeClass field from the cluster setting because that would change the effective NodeClass configuration, and potentially its image selection, without a NodeClass spec update.

## Runtime validation

The check uses the existing `ValidationSucceeded` condition rather than admission-time validation:

- Cluster configuration is not available to CRD CEL validation.
- Runtime validation is continuously reevaluated and applies to existing as well as newly created NodeClasses.
- An incompatible NodeClass receives `ValidationSucceeded=False` with reason `FIPSRequired`, which makes the aggregate `Ready` condition false and blocks provisioning.

The cloud-provider create path repeats the check before provisioning. This preserves enforcement during startup if an existing NodeClass still has a stale `Ready=True` status before its first status reconciliation.

The setting does not rewrite NodeClass specs and does not automatically drift existing nodes. Cluster-level update validation and lifecycle behavior remain owned by AKS.

## Testing

- Option parsing from `ENABLE_FIPS`.
- Unset `fipsMode` is rejected when cluster FIPS is enabled.
- Explicit `fipsMode: Disabled` is rejected when cluster FIPS is enabled.
- Explicit `fipsMode: FIPS` succeeds.
- Provisioning-boundary validation rejects an incompatible NodeClass even if its stored status is still `Ready=True`.
- `make verify` passes.
- `make presubmit` passes all 39 suites with 68.5% composite coverage.

## Documentation

- [ ] Yes, issue opened: #
- [x] No

## Release Note

```release-note
Managed NAP blocks non-FIPS AKSNodeClasses when FIPS is enabled at the AKS cluster level.
```

### DIFF
--- a/pkg/operator/options/options.go
+++ b/pkg/operator/options/options.go
@@ -91,6 +91,7 @@ type Options struct {
 	AdditionalTags             map[string]string `json:"additionalTags,omitempty"`
 	EnableAzureSDKLogging      bool              `json:"enableAzureSDKLogging,omitempty"` // Controls whether Azure SDK middleware logging is enabled
 	DiskEncryptionSetID        string            `json:"diskEncryptionSetId,omitempty"`
+	EnableFIPS                 bool              `json:"enableFIPS,omitempty"`
 
 	// If set to true, existing AKS machines created with an AKS Machine API provision mode will be managed even with other provision modes. This option does not have any effect if PROVISION_MODE is already an AKS Machine API mode, as it will behave as if this option is set to true.
 	ManageExistingAKSMachines bool `json:"manageExistingAKSMachines,omitempty"`
@@ -127,6 +128,7 @@ func (o *Options) AddFlags(fs *coreoptions.FlagSet) {
 	fs.StringVar(&o.SIGAccessTokenServerURL, "sig-access-token-server-url", env.WithDefaultString("SIG_ACCESS_TOKEN_SERVER_URL", ""), "The URL for the SIG access token server. Only used for AKS managed karpenter. UseSIG must be set tot true for this to take effect.")
 	fs.StringVar(&o.SIGSubscriptionID, "sig-subscription-id", env.WithDefaultString("SIG_SUBSCRIPTION_ID", ""), "The subscription ID of the shared image gallery.")
 	fs.StringVar(&o.DiskEncryptionSetID, "node-osdisk-diskencryptionset-id", env.WithDefaultString("NODE_OSDISK_DISKENCRYPTIONSET_ID", ""), "The ARM resource ID of the disk encryption set to use for customer-managed key (BYOK) encryption.")
+	fs.BoolVar(&o.EnableFIPS, "enable-fips", env.WithDefaultBool("ENABLE_FIPS", false), "If set to true, AKSNodeClasses must explicitly enable FIPS mode. This setting mirrors the AKS cluster-level FIPS configuration.")
 	fs.BoolVar(&o.ManageExistingAKSMachines, "manage-existing-aks-machines", env.WithDefaultBool("MANAGE_EXISTING_AKS_MACHINES", false), "If set to true, existing AKS machines created with an AKS Machine API provision mode will be managed even with other provision modes. This option does not have any effect when already on an AKS Machine API mode.")
 	fs.StringVar(&o.AKSMachinesPoolName, "aks-machines-pool-name", env.WithDefaultString("AKS_MACHINES_POOL_NAME", ""), "The name of the agent pool that the AKS machines are/will be in with AKS machine API provision modes. Existing AKS machines outside of this pool will be ignored. Required when PROVISION_MODE is an AKS machine API mode.")
 	fs.DurationVar(&o.ProviderBatchIdleDuration, "provider-batch-idle-duration", env.WithDefaultDuration("PROVIDER_BATCH_IDLE_DURATION", time.Second), "Idle duration for provider batch accumulation. Use Go duration format such as `1s`. Only used on provision mode aksmachineapiheaderbatch.")
--- a/pkg/cloudprovider/cloudprovider.go
+++ b/pkg/cloudprovider/cloudprovider.go
@@ -114,7 +114,10 @@ func (c *CloudProvider) WaitForInstancePromises() {
 	c.instancePromiseWg.Wait()
 }
 
-func (c *CloudProvider) validateNodeClass(nodeClass *v1beta1.AKSNodeClass) error {
+func (c *CloudProvider) validateNodeClass(ctx context.Context, nodeClass *v1beta1.AKSNodeClass) error {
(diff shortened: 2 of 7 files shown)


MaximilianoUribe:
(AI-generated)

Rebased onto current main and completed repeated managed-NAP validation on both `bootstrappingclient` and `aksmachineapiheaderbatch`.

- Two runs per backend of greenfield FIPS enforcement/recovery, enabling cluster FIPS on an existing NAP cluster, and enabling NAP on an existing FIPS cluster: **12 successful cases**.
- Verified omitted/Disabled NodeClasses are blocked, explicit FIPS recovers provisioning, default and system-surge nodes become Ready, and new Ubuntu22.04 nodes report kernel FIPS enabled.
- Validation exposed a required VM capability: Ubuntu22.04 FIPS images need `additionalCapabilities.enableFips1403Encryption` for guest-agent certificates and protected extension settings. Added propagation from the resolved image distro plus payload tests; fresh nodes passed after the fix.
- `make presubmit`: **40 suites passed**, **70.2% composite coverage**. Image scans passed.

Current head: `6a8c9fa1c8878cefa9089f5005867dfd756baa03`. Live validation covered Ubuntu22.04; the startup stale-readiness guard remains covered by its targeted unit test.

pkg/providers/imagefamily/resolver.go
```diff
@@ -227,6 +228,18 @@ func (r *defaultResolver) getStorageProfile(ctx context.Context, instanceType *c
 	return consts.StorageProfileManagedDisks, nil, nil
 }
 
+// Ubuntu 22.04 FIPS images require Compute's FIPS 140-3 encryption for guest-agent
+// certificates and protected extension settings. Use the resolved distro so generic
+// Ubuntu and explicit Ubuntu2204 NodeClasses receive the same capability.
+func requiresFIPS1403Encryption(distro string) bool {
+	switch distro {
+	case "aks-ubuntu-fips-containerd-22.04", "aks-ubuntu-fips-containerd-22.04-gen2", "aks-ubuntu-fips-containerd-22.04-tl-gen2":
```
PabloTriv, line 241: Is this something that will apply to other FIPS images in the future, such as 2404?

MaximilianoUribe: Yes, when we get the functionality we should update this as well, not sure if this is the best way of doing it since we also have to exclude some cases lmkwyt :)

PabloTriv: When you say excluding some cases do you mean excluding distros in the same image family, or excluding whole image families in general? If it's a per image family thing it might be worth it to change the function to something like this AI suggestion:

```
func requiresFIPS1403Encryption(
    imageFamily ImageFamily,
    fipsMode *v1beta1.FIPSMode,
) bool {
    if lo.FromPtr(fipsMode) != v1beta1.FIPSModeFIPS {
        return false
    }

    switch imageFamily.(type) {
    case *Ubuntu2204:
        return true
    default:
        return false
    }
}
```

If it's a per-distro thing where some distors in the same image family might be invalid then it's probably fine to leave as is.

In any case we should probably leave a comment about it needing to be updated in the future so AI can at least pick it up as something to consider when updating the image families.

MaximilianoUribe: (AI-generated)

Added the maintenance comment in 4602c15c. All currently supported Ubuntu22.04 FIPS variants (Gen1, Gen2, and Trusted Launch Gen2) require this capability; there is no intentional exclusion among those variants. Kept the resolved-distro mapping and documented that it must be revisited when adding image families or FIPS variants, since FIPS mode alone does not imply this Compute capability is required. Generic Ubuntu and explicit Ubuntu2204 still use the same resolved-distro behavior.

MaximilianoUribe: good catch, I fixed with a more concise comment for future applications :)

pkg/providers/imagefamily/resolver.go
```diff
@@ -206,9 +206,10 @@ func (r *defaultResolver) Resolve(
 
 		// TODO: We could potentially use the instance type to do defaulting like
 		// traditional AKS, so putting this here along with the other settings
-		StorageProfileSizeGB: lo.FromPtr(nodeClass.Spec.OSDiskSizeGB),
-		ImageID:              imageID,
-		IsWindows:            false, // TODO(Windows)
+		StorageProfileSizeGB:     lo.FromPtr(nodeClass.Spec.OSDiskSizeGB),
+		ImageID:                  imageID,
+		EnableFIPS1403Encryption: requiresFIPS1403Encryption(imageDistro),
```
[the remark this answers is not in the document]
MaximilianoUribe: good catch

pkg/operator/options/options.go
```diff
@@ -137,6 +138,7 @@ func (o *Options) AddFlags(fs *coreoptions.FlagSet) {
 	fs.StringVar(&o.SIGAccessTokenServerURL, "sig-access-token-server-url", env.WithDefaultString("SIG_ACCESS_TOKEN_SERVER_URL", ""), "The URL for the SIG access token server. Only used for AKS managed karpenter. UseSIG must be set tot true for this to take effect.")
 	fs.StringVar(&o.SIGSubscriptionID, "sig-subscription-id", env.WithDefaultString("SIG_SUBSCRIPTION_ID", ""), "The subscription ID of the shared image gallery.")
 	fs.StringVar(&o.DiskEncryptionSetID, "node-osdisk-diskencryptionset-id", env.WithDefaultString("NODE_OSDISK_DISKENCRYPTIONSET_ID", ""), "The ARM resource ID of the disk encryption set to use for customer-managed key (BYOK) encryption.")
+	fs.BoolVar(&o.EnableFIPS, "enable-fips", env.WithDefaultBool("ENABLE_FIPS", false), "If set to true, AKSNodeClasses must explicitly enable FIPS mode. This setting mirrors the AKS cluster-level FIPS configuration.")
```
tallaxes: nit: something like "enforce FIPS" would be a closer reflection of the functionality

tallaxes (approved):

PR #1843, opened 2026-08-13
feat: add FIPS support for ubuntu 2204

PabloTriv:
Fixes # 

**Description**
This change adds support for FIPS in Ubuntu 2204.

* unit testing

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
support for FIPS in Ubuntu 2204
```


pkg/providers/imagefamily/ubuntu_2204.go
```diff
@@ -68,8 +70,31 @@ func (u Ubuntu2204) DefaultImages(useSIG bool, fipsMode *v1beta1.FIPSMode, trust
...
 			}
 		}
-		//TODO: Fill out when Ubuntu 22.04 with FIPS becomes available
-		return []types.DefaultImageOutput{}
+
+		return []types.DefaultImageOutput{
+			{
+				PublicGalleryURL:     AKSUbuntuPublicGalleryURL,
+				GalleryResourceGroup: AKSUbuntuResourceGroup,
+				GalleryName:          AKSUbuntuGalleryName,
```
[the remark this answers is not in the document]
PabloTriv: Added test

pkg/providers/imagefamily/ubuntu_2204.go
```diff
@@ -68,8 +70,31 @@ func (u Ubuntu2204) DefaultImages(useSIG bool, fipsMode *v1beta1.FIPSMode, trust
...
+				),
+				Distro: "aks-ubuntu-fips-containerd-22.04-gen2",
+			},
+			{
+				PublicGalleryURL:     AKSUbuntuPublicGalleryURL,
+				GalleryResourceGroup: AKSUbuntuResourceGroup,
+				GalleryName:          AKSUbuntuGalleryName,
+				ImageDefinition:      Ubuntu2204Gen1FIPSImageDefinition,
+				Requirements: scheduling.NewRequirements(
+					scheduling.NewRequirement(v1.LabelArchStable, v1.NodeSelectorOpIn, karpv1.ArchitectureAmd64),
```
matthchr, line 92: I think I asked this question elsewhere as well... but no ARM FIPS image?

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/Azure/karpenter-provider-azure
 
-go 1.26.5
+go 1.26.6
```
matthchr, line 3: minor: unrelated go.mod changes? Revert?

pkg/providers/imagefamily/nodeimage_test.go
```diff
@@ -253,13 +253,13 @@ var _ = Describe("NodeImageProvider tests", func() {
...
 			})
 
-			//TODO: Modify when Ubuntu 22.04 with FIPS becomes available
 			It("should match expected images for FIPS Ubuntu2204", func() {
 				nodeClass.Spec.ImageFamily = lo.ToPtr(v1beta1.Ubuntu2204ImageFamily)
 
 				foundImages, err := nodeImageProvider.List(ctx, nodeClass)
 				Expect(err).ToNot(HaveOccurred())
-				Expect(foundImages).To(BeEmpty())
+				expectedImages := renderExpectedSIGNodeImages(&imagefamily.Ubuntu2204{}, nodeClass.Spec.FIPSMode, nodeClass.IsTrustedLaunchEnabled())
```
matthchr: 🤖 `renderExpectedSIGNodeImages` uses one hardcoded `sigImageVersion`, and this PR adjusts the generated snapshot entries for the newly exercised SKUs to match it. Since the snapshot legitimately contains per-SKU versions, could we derive the expected version from the fake API row instead? That would keep generated fixture data generator-owned and avoid failures after the next regeneration.

pkg/providers/imagefamily/ubuntu_2204_test.go
```diff
@@ -127,3 +127,52 @@ func TestUbuntu2204_Name(t *testing.T) {
...
+		g.Expect(images).To(BeEmpty())
+	})
+
+	t.Run("should return images for FIPS mode with SIG", func(t *testing.T) {
+		g := NewWithT(t)
+		fipsMode := v1beta1.FIPSModeFIPS
+		images := ubuntu.DefaultImages(true, &fipsMode, false)
+		g.Expect(images).To(HaveLen(2))
+
+		g.Expect(images[0].ImageDefinition).To(Equal(imagefamily.Ubuntu2204Gen2FIPSImageDefinition))
```
matthchr, line 172: 🤖 Could these tests independently pin the new selection contract rather than comparing against the same exported constants? In particular, please assert the literal Gen2/Gen1 SKU values, amd64 and Hyper-V generation requirements, ordering, and the `useSIG=true, FIPS, trustedLaunch=true` case. A wrong constant or swapped generation requirement can currently be copied into both actual and expected results.

pkg/providers/imagefamily/ubuntu_2204.go
```diff
@@ -68,8 +70,31 @@ func (u Ubuntu2204) DefaultImages(useSIG bool, fipsMode *v1beta1.FIPSMode, trust
...
-		return []types.DefaultImageOutput{}
+
+		return []types.DefaultImageOutput{
+			{
+				PublicGalleryURL:     AKSUbuntuPublicGalleryURL,
+				GalleryResourceGroup: AKSUbuntuResourceGroup,
+				GalleryName:          AKSUbuntuGalleryName,
+				ImageDefinition:      Ubuntu2204Gen2FIPSImageDefinition,
+				Requirements: scheduling.NewRequirements(
+					scheduling.NewRequirement(v1.LabelArchStable, v1.NodeSelectorOpIn, karpv1.ArchitectureAmd64),
```
matthchr: 🤖 Both Ubuntu2204 FIPS images are amd64-only. An arm64-only NodePool can therefore see its NodeClass become `ImagesReady=True`, then repeatedly fail image resolution because no compatible image exists. Since Azure does not publish an arm64 Ubuntu2204 FIPS SKU, could we surface a targeted architecture/FIPS diagnostic rather than failing later per NodeClaim?

pkg/providers/imagefamily/ubuntu_2204.go
```diff
@@ -68,8 +70,31 @@ func (u Ubuntu2204) DefaultImages(useSIG bool, fipsMode *v1beta1.FIPSMode, trust
...
+					scheduling.NewRequirement(v1beta1.LabelSKUHyperVGeneration, v1.NodeSelectorOpIn, v1beta1.HyperVGenerationV2),
+				),
+				Distro: "aks-ubuntu-fips-containerd-22.04-gen2",
```
matthchr, line 84: 🤖 Do we have a successful provisioning result for these new image definitions and distro strings without Trusted Launch? Unit tests validate local selection, but they cannot establish that the SIG images are available in the intended regions or that AgentBaker accepts these distro values. Please record a manual or end-to-end result in the PR.

go.mod
matthchr, line 3: 🤖 This toolchain and `golang.org/x/*` dependency update has a separate risk and rollback profile from Ubuntu2204 FIPS support. If it is not required for this feature, could it move to a separate PR? Otherwise, please document why the CVE update needs to ship atomically here.

pkg/providers/imagefamily/ubuntu_2204.go
```diff
@@ -68,8 +70,31 @@ func (u Ubuntu2204) DefaultImages(useSIG bool, fipsMode *v1beta1.FIPSMode, trust
 				},
 			}
 		}
-		//TODO: Fill out when Ubuntu 22.04 with FIPS becomes available
-		return []types.DefaultImageOutput{}
+
+		return []types.DefaultImageOutput{
```
matthchr, line 74: 🤖 Should generic Ubuntu + FIPS now resolve to Ubuntu2204?

matthchr: and if not now... when? At a k8s version boundary?

pkg/providers/imagefamily/ubuntu_2204.go
matthchr: Did you test manually? Can you share what testing you did?

pkg/fake/nodeimageversionsapi.go
```diff
@@ -167,10 +167,10 @@ var (
 			Version:  lo.ToPtr("202512.18.0"),
 		},
 		{
-			FullName: lo.ToPtr("AKSUbuntu-2204fipscontainerd-202404.09.0"),
+			FullName: lo.ToPtr("AKSUbuntu-2204fipscontainerd-202512.18.0"),
```
matthchr: Was this data not generated with `az-codegen-nodeimageversions`? (See comment up above)? Did it get changed manually?

I guess this is fine?

pkg/providers/imagefamily/ubuntu_2204.go
MaximilianoUribe: Yes. The answer should follow AKS’s existing Kubernetes-version boundary:                                                                                                                                                                                               
                                                                                                                                                                                                                                                                             
     - Kubernetes < 1.35: generic Ubuntu + FIPS resolves to Ubuntu 20.04 FIPS.                                                                                                                                                                                               
     - Kubernetes >= 1.35: it resolves to Ubuntu 22.04 FIPS.                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                             
RP defines max-k8s-version-for-2004-fips-enabled-vhd as 1.35.0 and switches to 22.04 at that boundary.

pkg/providers/imagefamily/ubuntu_2204.go
MaximilianoUribe: There is currently no ubuntu 2204 arm fips available

pkg/providers/imagefamily/ubuntu_2204.go
MaximilianoUribe: let me know if you want to resolve this in this pr @PabloTriv or you want me to include it as a part of: https://github.com/Azure/karpenter-provider-azure/pull/1842 , since we need the k8s specific resolution for this

pkg/fake/nodeimageversionsapi.go
PabloTriv: Was debating how to handle that. I went ahead and regenerated it with `make az-codegen-nodeimageversions` and copied the results into the file.

pkg/providers/imagefamily/ubuntu_2204.go
PabloTriv: Incorporated it into this PR

go.mod
PabloTriv: Removed

pkg/providers/imagefamily/ubuntu_2204.go
PabloTriv: I built the image and swapped it in standalone. Then made a nodepool/class/workload with TL, 2204 gen1 and gen2. (Will follow up with images). I also tested generic Ubuntu.

As for a more permanent test, as far as I can tell our E2Es in this repo can't use SIG (which is where all the FIPS images are.)? I was going to make an E2E in RP

pkg/providers/imagefamily/ubuntu_2204_test.go
PabloTriv: I made the tests verify against literals (I think that's what this comment is asking for if I understood correctly)

pkg/providers/imagefamily/ubuntu_2204.go
PabloTriv: Standalone Tests:

This is k8s version 1.34, so the generic fips evaluates to 20.04. It also has a gen1 example, and a gen2 example
image

This is k8s version 1.35, so generic ubuntu evaluates to 22.04
image

pkg/providers/imagefamily/utils.go
```diff
@@ -60,8 +72,8 @@ func UseUbuntu2404(kubernetesVersion string) bool {
 // intentionally not used from defaultUbuntu itself, to keep that function's
 // existing logic flow untouched. If the rule in defaultUbuntu ever changes,
 // update this helper to match.
-func ResolvesToUbuntu2004(familyName *string, fipsMode *v1beta1.FIPSMode, trustedLaunch bool) bool {
+func ResolvesToUbuntu2004(familyName *string, fipsMode *v1beta1.FIPSMode, trustedLaunch bool, kubernetesVersion string) bool {
 	family := lo.FromPtr(familyName)
 	isUbuntuLegacyOrUnset := family == "" || family == v1beta1.UbuntuImageFamily
-	return isUbuntuLegacyOrUnset && lo.FromPtr(fipsMode) == v1beta1.FIPSModeFIPS && !trustedLaunch
+	return isUbuntuLegacyOrUnset && lo.FromPtr(fipsMode) == v1beta1.FIPSModeFIPS && !UseUbuntu2204FIPS(kubernetesVersion, trustedLaunch)
```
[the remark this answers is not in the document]
PabloTriv: Updated the comment

pkg/providers/imagefamily/nodeimage_test.go
```diff
@@ -44,7 +44,8 @@ const (
 	cigImageVersion      = "202505.27.0"
 	laterCIGImageVersion = "202605.27.0"
 
-	sigImageVersion = "202512.18.0" // Updated to match fake data versions
+	sigImageVersion             = "202608.26.0"
+	azureLinuxV2SIGImageVersion = "202512.06.0"
```
matthchr, line 48: minor: worth a comment about why this is different? I assume because it's not published anymore...?

pkg/providers/imagefamily/ubuntu_2004_resolver_test.go
```diff
@@ -26,31 +26,35 @@ import (
...
+		"nil family + FIPS + TrustedLaunch":          {nil, lo.ToPtr(v1beta1.FIPSModeFIPS), true, oldVersion, false},
+		"nil family + FIPS Disabled":                 {nil, lo.ToPtr(v1beta1.FIPSModeDisabled), false, oldVersion, false},
+		"empty family + FIPS":                        {lo.ToPtr(""), lo.ToPtr(v1beta1.FIPSModeFIPS), false, oldVersion, true},
+		"empty family + FIPS + TrustedLaunch":        {lo.ToPtr(""), lo.ToPtr(v1beta1.FIPSModeFIPS), true, oldVersion, false},
+		"Ubuntu legacy + FIPS":                       {lo.ToPtr(v1beta1.UbuntuImageFamily), lo.ToPtr(v1beta1.FIPSModeFIPS), false, oldVersion, true},
+		"Ubuntu legacy + FIPS + TrustedLaunch":       {lo.ToPtr(v1beta1.UbuntuImageFamily), lo.ToPtr(v1beta1.FIPSModeFIPS), true, oldVersion, false},
+		"Ubuntu legacy + nil fips":                   {lo.ToPtr(v1beta1.UbuntuImageFamily), nil, false, oldVersion, false},
+		"Ubuntu legacy + Disabled":                   {lo.ToPtr(v1beta1.UbuntuImageFamily), lo.ToPtr(v1beta1.FIPSModeDisabled), false, oldVersion, false},
+		"Ubuntu legacy + FIPS + K8S version >= 1.35": {lo.ToPtr(v1beta1.UbuntuImageFamily), lo.ToPtr(v1beta1.FIPSModeFIPS), false, newVersion, false},
```
matthchr, line 48: Is this `Ubuntu legacy`? I thought it's just standard Ubuntu, that's not legacy right? Though I guess I see that's what these were called before as well...

Does this really mean "Ubuntu (unspecified)" or "Ubuntu (no specific version)"? Might be more clear... since AFAIK it is not legacy?

pkg/controllers/nodeclass/status/localdns_test.go
```diff
@@ -124,15 +124,15 @@ func TestPreferred_BYOCNI_Disabled(t *testing.T) {
 	expectState(t, nc, v1beta1.LocalDNSStateDisabled)
 }
 
-func TestPreferred_Ubuntu2004_Disabled(t *testing.T) {
+func TestPreferred_UbuntuFIPS_Enabled(t *testing.T) {
```
matthchr, line 127: Just confirming: The reason that this was disabled previously is because of Ubuntu 20.04, and NOT because of FIPS, is that right?

Does AKS actually run LocalDNS in preferred mode on 2204 FIPS?

pkg/providers/imagefamily/ubuntu_2004_resolver_test.go
matthchr: Though I do see we were setting this already and we do refer to it in some other comments that way too... so feel free to ignore me if it IS in fact legacy...

matthchr (approved):

pkg/providers/imagefamily/ubuntu_2204.go
matthchr, line 81: 🤖 These new FIPS images are AMD64-only, but `GetInstanceTypes()` does not filter ARM64 SKUs against the resolved image requirements. An ARM64-constrained NodePool—or an unconstrained workload for which Karpenter selects ARM64—can therefore receive a valid, Ready NodeClass and fail later with `no compatible images found`. The same gap affects a Hyper-V generation whose SIG definition is temporarily absent. Please filter candidate instance types against the requirements of the images actually present in `status.images`, or add equivalent FIPS architecture/generation filtering, with negative tests.

pkg/providers/imagefamily/utils.go
```diff
@@ -60,8 +73,8 @@ func UseUbuntu2404(kubernetesVersion string) bool {
...
```
matthchr, line 79: 🤖 This comment confirms there are now two sources of truth for Ubuntu resolution: `defaultUbuntu` and `ResolvesToUbuntu2004`. A future threshold/default change can make LocalDNS eligibility disagree with the image selected for provisioning. Can the resolver expose a shared resolved-family/version trait so both paths consume the same decision?

pkg/providers/imagefamily/nodeimage_test.go
```diff
@@ -67,10 +68,14 @@ func renderExpectedSIGNodeImages(
 	fips *v1beta1.FIPSMode,
 	trustedLaunch bool,
 ) []imagefamily.NodeImage {
+	version := sigImageVersion
+	if _, ok := fam.(*imagefamily.AzureLinux); ok {
+		version = azureLinuxV2SIGImageVersion
```
matthchr, line 73: 🤖 This expected value is built by calling `DefaultImages()`, which is also the implementation being tested. An incorrect architecture or Hyper-V requirement can therefore appear in both actual and expected results. Please add independent assertions for the Ubuntu 22.04 FIPS AMD64 and Gen1/Gen2 requirements.

matthchr (changes requested):
🤖 Automated code review

Requesting changes for the ARM64/image-compatibility issue below. The resolver-duplication and test-oracle comments are non-blocking follow-ups.

The AKS Machine path should also have a focused test proving that the resolved Ubuntu 22.04 FIPS `NodeImageVersion`, `OSSKU`, and `EnableFIPS` remain consistent across the Kubernetes 1.35 boundary.

pkg/providers/imagefamily/ubuntu_2204.go
matthchr: Arch choice is per-nodepool so I don't know that we have a better experience for this now than `no compatible images found` since we can't block it at the NodePool level.

pkg/providers/imagefamily/utils.go
matthchr: This seems like a fair comment, to me.

matthchr (commented):
clearing AI's overzealous "waiting"

pkg/providers/imagefamily/utils.go
PabloTriv: Called GetImageFamily instead, which goes through defaultUbuntu

pkg/providers/imagefamily/ubuntu_2004_resolver_test.go
PabloTriv: This test was calling it Legacy before I made edits (I didn't change the test names). I think it's calling it "legacy" because generic Ubuntu + FIPS is the only way to wind up with 20.04. In any case, I switched it to "generic Ubuntu" which seems more accurate.

pkg/providers/imagefamily/nodeimage_test.go
PabloTriv: Added a comment. Yes, the image version is frozen at 202512.06.0 https://learn.microsoft.com/en-us/azure/aks/node-images

pkg/controllers/nodeclass/status/localdns_test.go
PabloTriv: Yes, the restriction was 20.04. The docs for local dns specifically say it is only supported for 2204+. The gates in the code also already stop Ubuntu2004, which is why this test expected disabled before. (Note that Preferred requires a higher k8s version than what would resolve to 2004)

matthchr (approved):

PR #1847, opened 2026-08-16
fix: bound instance type cache retention

### LINKED ISSUE #1844: Potential memory leak in v1.14
### Version

__Karpenter Version:__ v1.14.0
__Kubernetes Version:__ v1.34.6

### Expected Behavior

Stable memory usage when provisioning and destroying nodes. 

### Actual Behavior

Using self-hosted Karpenter we are seeing increasing memory consumption until an eventual OOMkill (based on our low limit) after upgrading to v1.14.0. Downgrading to v1.13.1 returns to stable memory consumption. 

Our usage is as follows: 
- karpenter used to provision new nodes for temporary cronjobs to run hourly
- this results in constant scale out/scale in from 0-1 nodes hourly
- each scaling event increases memory usage that is never cleaned up until OOM
- mem req/limit of 280Mi/1Gi

We run self hosted for Azure and AWS. We do not see the same behavior with the AWS provider. 

Run self hosted karpenter v1.14.0 and trigger repeated node creation/deletion events on a cadence and observe memory stats. 

### Resource Specs and Logs

Our nodepool configuration is: 
```
    spec:
      nodeClassRef:
        group: karpenter.azure.com
        kind: AKSNodeClass
        name: default
      requirements:
      - key: kubernetes.io/arch
        operator: In
        values:
        - amd64
        operator: In
        values:
        - on-demand
      - key: "karpenter.azure.com/sku-cpu"
        operator: In
        values:
        - "8"
      - key: karpenter.azure.com/sku-family
        operator: In
        values:
        - D
      - key: karpenter.azure.com/sku-storage-premium-capable
        operator: In
        values:
        - "true"
      - key: karpenter.azure.com/sku-storage-ephemeralos-maxsize
        operator: Exists
      expireAfter: Never
  disruption:
    consolidationPolicy: WhenEmpty
    consolidateAfter: 1m
```

We do not see any meaningful log events in the controller. 

Image

### Community Note

tallaxes:
Fixes #1844

**Description**

The fully initialized instance type cache currently includes monotonically increasing SKU, unavailable-offering, and quota sequence numbers in each cache key. Every availability change therefore retains another complete instance type graph until the 23-hour TTL expires.

This change:

- keys derived instance types only by the NodeClass parameters that can legitimately coexist;
- flushes derived entries when quota, unavailable-offering, or Skewer SKU data changes;
- serializes cache rebuilding so concurrent misses share one result;
- coalesces duplicate unavailable-offering marks so TTL refreshes do not trigger redundant instance type rebuilds;
- invalidates cached quota availability across fail-open reset and recovery;
- uses typed atomic counters for quota and unavailable-offering generations;
- returns an independent copy of built instance types on a cache miss, matching the cache-hit path, so a caller reordering its returned slice cannot mutate the cached entry.

The existing TTL remains responsible for removing unused NodeClass parameter variants rather than obsolete source-data generations.

* Added cache-cardinality coverage for quota, unavailable-offering, and Skewer invalidation.
* Added quota reset and recovery coverage.
* Added deterministic coverage for quota availability changing during instance type construction.
* Added duplicate, stricter, global spot, TTL refresh, and concurrent unavailable-offering mark coverage.
* Added coverage that reorders an instance-type slice returned from a cache miss and asserts a later cache hit for the same key is unaffected.
* `go test ./pkg/cache ./pkg/providers/quota ./pkg/providers/instancetype -count=1`
* `go test -race ./pkg/cache ./pkg/providers/quota -count=1`
* `make verify`
* ✅ [E2EMatrixTrigger · Azure/karpenter-provider-azure@3d55e8c](https://github.com/Azure/karpenter-provider-azure/actions/runs/31962583272)
* ✅ [E2EMatrixTrigger · Azure/karpenter-provider-azure@d763392](https://github.com/Azure/karpenter-provider-azure/actions/runs/31977443520)
* ✅ [E2EMatrixTrigger · Azure/karpenter-provider-azure@8478d70](https://github.com/Azure/karpenter-provider-azure/actions/runs/32084795044)

- [ ] Yes, issue opened: #
- [x] No

**Release Note**

```release-note
Fix excessive controller memory retention when Azure quota or offering availability changes.
```

### DIFF
--- a/pkg/cache/unavailableofferings.go
+++ b/pkg/cache/unavailableofferings.go
@@ -54,7 +54,8 @@ type UnavailableOfferings struct {
 	singleOfferingCache *cache.Cache
 	// key: <skuFamilyName>:<zone>:<capacityType> (lowercase), value: int64 (CPU count at or above which we block, or wholeVMFamilyBlockedSentinel if entire family is blocked)
 	vmFamilyCache *cache.Cache
-	seqNum        atomic.Uint64
+	// seqNum is updated on any material changes to unavailable offerrings cache (not updated on TTL only changes)
+	seqNum atomic.Uint64
 }
 
 func NewUnavailableOfferingsWithCache(singleOfferingCache, vmFamilyCache *cache.Cache) *UnavailableOfferings {
--- a/pkg/cache/unavailableofferings.go
+++ b/pkg/cache/unavailableofferings.go
@@ -54,7 +54,7 @@ type UnavailableOfferings struct {
 	singleOfferingCache *cache.Cache
 	// key: <skuFamilyName>:<zone>:<capacityType> (lowercase), value: int64 (CPU count at or above which we block, or wholeVMFamilyBlockedSentinel if entire family is blocked)
 	vmFamilyCache *cache.Cache
-	// seqNum is updated on any material changes to unavailable offerrings cache (not updated on TTL only changes)
+	// seqNum is updated on any material changes to unavailable offerings cache (not updated on TTL only changes)
 	seqNum atomic.Uint64
 }
 
--- a/pkg/cache/unavailableofferings.go
+++ b/pkg/cache/unavailableofferings.go
@@ -52,20 +52,19 @@ type UnavailableOfferings struct {
 	singleOfferingCache *cache.Cache
 	// key: <skuFamilyName>:<zone>:<capacityType> (lowercase), value: int64 (CPU count at or above which we block, or wholeVMFamilyBlockedSentinel if entire family is blocked)
 	vmFamilyCache *cache.Cache
-	SeqNum        uint64
+	seqNum        atomic.Uint64
 }
 
 func NewUnavailableOfferingsWithCache(singleOfferingCache, vmFamilyCache *cache.Cache) *UnavailableOfferings {
 	uo := &UnavailableOfferings{
 		singleOfferingCache: singleOfferingCache,
 		vmFamilyCache:       vmFamilyCache,
-		SeqNum:              0,
 	}
 	uo.singleOfferingCache.OnEvicted(func(_ string, _ any) {
-		atomic.AddUint64(&uo.SeqNum, 1)
+		uo.seqNum.Add(1)
 	})
 	uo.vmFamilyCache.OnEvicted(func(_ string, _ any) {
-		atomic.AddUint64(&uo.SeqNum, 1)
+		uo.seqNum.Add(1)
 	})
 	return uo
 }
@@ -75,6 +74,10 @@ func NewUnavailableOfferings() *UnavailableOfferings {
 		cache.New(UnavailableOfferingsTTL, UnavailableOfferingsCleanupInterval),
 		cache.New(UnavailableOfferingsTTL, UnavailableOfferingsCleanupInterval),
 	)
+}
+
+func (u *UnavailableOfferings) SeqNum() uint64 {
+	return u.seqNum.Load()
 }
 
 // IsUnavailable returns true if the offering appears in the cache
@@ -165,7 +168,7 @@ func (u *UnavailableOfferings) markFamilyUnavailableAtCPUCountImpl(ctx context.C
 
 	// call Set to update the cache entry, even if it already exists, to extend its TTL
 	u.vmFamilyCache.Set(key, cpuCount, ttl)
-	atomic.AddUint64(&u.SeqNum, 1)
+	u.seqNum.Add(1)
 }
 
 // MarkSpotUnavailable communicates recently observed temporary capacity shortages for spot
@@ -177,7 +180,7 @@ func (u *UnavailableOfferings) MarkSpotUnavailableWithTTL(ctx context.Context, t
 		"capacity-type", capacityType,
 		"ttl", ttl)
 	u.singleOfferingCache.Set(singleInstanceKey("", "", capacityType), struct{}{}, ttl)
-	atomic.AddUint64(&u.SeqNum, 1)
+	u.seqNum.Add(1)
 }
 
 // MarkUnavailableWithTTL allows us to mark an offering unavailable with a custom TTL.
@@ -193,7 +196,7 @@ func (u *UnavailableOfferings) MarkUnavailableWithTTL(ctx context.Context, unava
 		"capacity-type", capacityType,
 		"ttl", ttl)
 	u.singleOfferingCache.Set(singleInstanceKey(instanceType, zone, capacityType), struct{}{}, ttl)
-	atomic.AddUint64(&u.SeqNum, 1)
+	u.seqNum.Add(1)
 
 	// Also mark the VM family unavailable at this SKU's vCPU count, so larger sizes of the same family are blocked too
 	u.markFamilyUnavailableAtCPUCount(ctx, sku, zone, capacityType, ttl)
@@ -207,7 +210,7 @@ func (u *UnavailableOfferings) MarkUnavailable(ctx context.Context, unavailableR
 func (u *UnavailableOfferings) Flush() {
 	u.singleOfferingCache.Flush()
 	u.vmFamilyCache.Flush()
-	atomic.AddUint64(&u.SeqNum, 1)
+	u.seqNum.Add(1)
 }
 
 // singleInstanceKey returns the cache singleInstanceKey for all offerings in the cache
--- a/pkg/cache/unavailableofferings.go
+++ b/pkg/cache/unavailableofferings.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"strings"
+	"sync"
 	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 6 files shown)


pkg/cache/unavailableofferings.go
```diff
@@ -117,24 +122,28 @@ func (u *UnavailableOfferings) isFamilyUnavailable(sku *skewer.SKU, zone, capaci
...
 	}
-	u.markFamilyUnavailableAtCPUCountImpl(ctx, sku, zone, capacityType, cpuCount, ttl)
+	return u.markFamilyUnavailableAtCPUCountImpl(ctx, sku, zone, capacityType, cpuCount, ttl)
 }
 
 // MarkFamilyUnavailable marks the entire VM family as unavailable in a specific zone for a specific capacity type with custom TTL.
 // Family name is derived from the provided skewer.SKU.
 func (u *UnavailableOfferings) MarkFamilyUnavailable(ctx context.Context, sku *skewer.SKU, zone, capacityType string, ttl time.Duration) {
-	u.markFamilyUnavailableAtCPUCountImpl(ctx, sku, zone, capacityType, wholeVMFamilyBlockedSentinel, ttl)
+	u.mu.Lock()
```
matthchr, line 138: is the lock only required to get accurate change-counting? because otherwise two separate entries could enter and both think they changed something when in reality only 1 really did?

pkg/cache/unavailableofferings.go
```diff
@@ -45,27 +46,27 @@ var (
...
+	mu sync.Mutex
 	// TODO: I think this singleOfferingCache could basically be removed in favor of the family cache at this point, as we are now marking family unavailable at CPU count for all error cases.
 	// I didn't do that purely because of the defensive fallback we have in markFamilyUnavailableAtCPUCountImpl, but in practice I am not sure if we'll ever have a nil family (I don't
 	// see any evidence it can happen in logs)
 	// key: <capacityType>:<instanceType>:<zone>, value: struct{}{}
 	singleOfferingCache *cache.Cache
 	// key: <skuFamilyName>:<zone>:<capacityType> (lowercase), value: int64 (CPU count at or above which we block, or wholeVMFamilyBlockedSentinel if entire family is blocked)
 	vmFamilyCache *cache.Cache
-	SeqNum        uint64
+	seqNum        atomic.Uint64
```
matthchr: minor: maybe explain in a comment here or on the new `SeqNum` function (probably better here) what exactly the goal/objective of seqNum is.

I think one of the key changes you're making here is that changes that _only_ update a TTL, but dont fundamentally edit the items in the cache, don't rev the seqNum. I think that makes sense based on where this is used (instanceType cache), but I think it would be good to call out the contract on this guy explicitly.

pkg/providers/instancetype/instancetypes.go
```diff
@@ -95,15 +99,14 @@ type DefaultProvider struct {
...
 	quotaProvider        quota.Provider
 
-	// Values cached *before* considering insufficient capacity errors from the unavailableOfferings cache.
-	// Fully initialized Instance Types are also cached based on the set of all instance types,
-	// unavailableOfferings cache, AWSNodeClass, and kubelet configuration from the NodePool
-	instanceTypesCache *cache.Cache
+	// Fully initialized instance types are cached by the parameters that affect their construction.
+	// Source-data generations invalidate the cache as a whole instead of creating unreachable keys.
+	instanceTypesCache           *cache.Cache
+	muInstanceTypesCache         sync.Mutex
```
matthchr: `*cache.Cache` is already threadsafe AFAIK so `muInstanceTypesCache` at least at first blush is confusing. That can't be _all_ it's protecting because if it were it wouldn't be needed.

pkg/providers/instancetype/instancetypes.go
```diff
@@ -95,15 +99,14 @@ type DefaultProvider struct {
 	unavailableOfferings *kcache.UnavailableOfferings
 	quotaProvider        quota.Provider
 
-	// Values cached *before* considering insufficient capacity errors from the unavailableOfferings cache.
-	// Fully initialized Instance Types are also cached based on the set of all instance types,
-	// unavailableOfferings cache, AWSNodeClass, and kubelet configuration from the NodePool
-	instanceTypesCache *cache.Cache
+	// Fully initialized instance types are cached by the parameters that affect their construction.
+	// Source-data generations invalidate the cache as a whole instead of creating unreachable keys.
```
matthchr: minor: Consider rephrasing slightly?

```suggestion
	// Changes in the source data generation invalidate the cache as a whole instead of creating unreachable keys.
```

pkg/providers/instancetype/instancetypes.go
```diff
@@ -181,20 +203,23 @@ func (p *DefaultProvider) List(
...
 		}
-
 		result = append(result, instanceType)
 	}
+	return result
+}
 
-	p.instanceTypesCache.SetDefault(key, result)
-	return result, nil
+func (p *DefaultProvider) currentInstanceTypesCacheGeneration() instanceTypesCacheGeneration {
```
matthchr: minor, naming: is this the `currentInstanceTypesCacheGeneration` or the `currentInstanceTypesGeneration` (no cache?) or `currentInstanceTypesSourceDataGeneration` or something?

The stored local variable `instanceTypesCacheGeneration instanceTypesCacheGeneration` feels like the "cache generation" to me, and differentiating that with the "current" cache generation is a bit confusing (because the current cache generation would presumably be the one in the cache?)

pkg/providers/instancetype/instancetypes.go
```diff
@@ -154,20 +156,40 @@ func (p *DefaultProvider) List(
...
 	if item, ok := p.instanceTypesCache.Get(key); ok {
 		// Ensure what's returned from this function is a shallow-copy of the slice (not a deep-copy of the data itself)
 		// so that modifications to the ordering of the data don't affect the original
 		return append([]*cloudprovider.InstanceType{}, item.([]*cloudprovider.InstanceType)...), nil
 	}
 
-	// Get Viable offerings
-	// Azure has zones availability directly from SKU info
+	result := p.buildInstanceTypes(ctx, instanceTypeParams)
+	if latest := p.currentInstanceTypesCacheGeneration(); generation != latest {
```
matthchr: Is this actually necessary? We're holding the RLock and the muInstanceTypesCache, what's the difference between just saving this data in the cache (it is stale) and flushing the cache and setting the latest generation.

In both cases, we just rebuild instance types on the next list? Why is this better?

pkg/providers/instancetype/suite_test.go
```diff
@@ -3187,6 +3221,45 @@ var _ = Describe("InstanceType Provider", func() {
...
+			Expect(instanceTypes).ToNot(BeEmpty())
+			Expect(instanceTypeCache.ItemCount()).To(BeZero())
+
+			instanceTypes, err = provider.List(ctx, nodeClass)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(instanceTypes).ToNot(BeEmpty())
+			Expect(instanceTypeCache.ItemCount()).To(Equal(1))
+		})
+
+		It("should not let a caller reordering the returned slice affect a later cache hit", func() {
```
matthchr: minor, test location: Is this behavior actually "quota filtering" specific?

pkg/providers/quota/quota.go
```diff
@@ -143,13 +145,18 @@ func (p *DefaultProvider) HasQuotaFor(ctx context.Context, sku *skewer.SKU) bool
...
 }
 
 func (p *DefaultProvider) Reset() {
 	p.mu.Lock()
 	defer p.mu.Unlock()
+	if len(p.usages) == 0 {
+		return
+	}
 	p.usages = map[string]*armcompute.Usage{}
+	p.reset = true
```
matthchr: Does this exist for the case where we have
Data set A
reset
Data set A again

Which would then _not_ increment the seqNum because `cm` thinks it hasn't changed?

Assuming yes - maybe cleaner to just reconstruct `cm` here too, that way `cm` just doesn't have the old data in it?

pkg/cache/unavailableofferings.go
tallaxes: It is both to avoid double generation bumps and also to make compare-and-set atomic (which should ensure the correct/strictest result is stored when multiple threads try to update it concurrently)

pkg/cache/unavailableofferings.go
tallaxes: Added a brief comment

pkg/providers/instancetype/instancetypes.go
tallaxes: Agreed. Renamed both the function and the struct for clarity, used `currentInstanceTypesSourceDataGeneration`

pkg/providers/quota/quota.go
tallaxes: I don't think there is a bug since it checks for HasChanged || p.reset, but recreating CM is cleaner; done.

pkg/providers/instancetype/suite_test.go
tallaxes: Nope, not quota specific; moved

pkg/providers/instancetype/instancetypes.go
tallaxes: Was just trying to avoid storing the known outdated capture - even if briefly - but agree it is not worth it; simplified

pkg/providers/quota/quota.go
matthchr: yeah agreed - definitely was not a bug I was just trying to understand the reasoning for why we have it. Agree reconstructing `cm` is better (though only slightly).

matthchr (dismissed):

matthchr (approved):

PR #1858, opened 2026-08-21
feat: add capacityrecommendation provider

matthchr:

* Unit tests
* E2E test: (TODO: share link)
  * https://github.com/Azure/karpenter-provider-azure/actions/runs/32533004094/job/96928690467
  * https://github.com/Azure/karpenter-provider-azure/actions/runs/34555251377
- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


pkg/providers/capacityrecommendation/capacity_recommendation.go
```diff
@@ -0,0 +1,458 @@
...
+	}
+	return armrecommender.SKUMixPlacementRequest{
+		Zones: zones,
+		CapacityProfile: &armrecommender.SKUMixPlacementCapacityProfile{
+			Capacity:           to.Ptr(input.Count),
+			CapacityType:       to.Ptr(armrecommender.SKUMixPlacementCapacityTypeVM),
+			Priority:           to.Ptr(priority),
+			AllocationStrategy: to.Ptr(armrecommender.SKUMixPlacementAllocationStrategyPrioritized),
+			OSType:             to.Ptr(osType),
+		},
```
MaximilianoUribe, line 228: (AI-generated)

The API contract describes `spotPriorityProfile` as required when `priority` is `Spot`, but Spot requests currently omit it. Has omission been validated against the live API? If not, could we explicitly send a profile with `maxPricePerVm: -1` to preserve the documented no-price-limit default?

pkg/providers/allocationstrategy/stages/sku_mix_rank.go
```diff
@@ -0,0 +1,374 @@
...
+		)
+		if err != nil {
+			log.FromContext(ctx).Error(err, "failed to get SKU Mix Placement recommendation",
+				"capacityType", group.key.capacityType,
+				"placementScope", group.key.placementScope,
+				"zones", group.zones,
+			)
+			if s.mode == consts.ComputeRecommendationModeEnabled {
+				// If our API failed, we want to fail open so add all items in the group to recommended
+				addLocalCandidates(group, recommendedOfferings)
```
MaximilianoUribe, line 97: (AI-generated)

On an API error, the code says it will fail open and add all items in the group, but the group retains only the first 10 locally ranked VM sizes. If limiting fallback to the same top-10 candidate window is intentional, could this comment make that explicit? Otherwise, the group would need to retain the complete local candidate set separately so an API outage does not remove size 11 and later.

MaximilianoUribe (commented):
(AI-generated)

I left three questions about how the SKU Mix API contract maps into the allocation strategy.

pkg/providers/allocationstrategy/stages/sku_mix_rank.go
```diff
@@ -0,0 +1,374 @@
...
+			"hasDifferences", len(comparison.differences) > 0,
+			"differences", comparison.differences,
+		)
+
+		if s.mode != consts.ComputeRecommendationModeEnabled {
+			continue
+		}
+		for _, recommendation := range recommendations {
+			key := offeringKey(recommendation.VMSize, recommendation.CapacityType, group.key.placementScope, recommendation.Zone)
+			recommendedOfferings.Insert(key)
```
MaximilianoUribe, line 128: (AI-generated)

`placementChoices` are alternative scored deployment choices, each with its own ID and SKU split. Flattening all choices into one allow-list loses the distinction between the highest-scored choice and its fallbacks, and the logged first split ID might not correspond to the offering ultimately selected. Could we preserve choice precedence instead: offerings from the highest-scored choice first, followed by later choices, with local price ordering within each choice? Remaining local offerings could still be retained as fallback if that matches the desired fail-open behavior.

pkg/providers/capacityrecommendation/capacity_recommendation.go
MaximilianoUribe: Not sure if this would fail, but its worth mentioning because of the contract

pkg/providers/allocationstrategy/stages/sku_mix_rank.go
MaximilianoUribe: this is more about a comment correction more that about the implementation itself, I agree that on fail keeping the first 10 locals is the right choice imo

pkg/providers/allocationstrategy/stages/sku_mix_rank.go
MaximilianoUribe: this is a little complicated and maybe out of scope for this pr but it seems like an interesting idea to follow up on

pkg/providers/capacityrecommendation/capacity_recommendation.go
matthchr: yes omission has been validated against the live API. See:
https://github.com/Azure/karpenter-provider-azure/actions/runs/32533004094/job/96928690467

Here's an example response produced by this code:
```
2026-08-21T22:45:21.1482489Z         {"level":"DEBUG","time":"2026-08-21T22:42:52.337Z","logger":"controller","caller":"capacityrecommendation/capacity_recommendation.go:143","message":"received SKU Mix Placement API response","commit":"a4212fa","controller":"nodeclaim.lifecycle","controllerGroup":"karpenter.sh","controllerKind":"NodeClaim","NodeClaim":{"name":"knaveash-2-ecsfo3bph5-mpssc"},"namespace":"","name":"knaveash-2-ecsfo3bph5-mpssc","reconcileID":"d981b33d-73bd-40c7-8f16-e6df1e654a5b","response":"{\"partialFulfillmentReason\":\"None\",\"placementChoices\":[{\"id\":\"2b51b7dc-1379-460d-af62-ddf13c9aad44\",\"score\":9,\"skuSplit\":[{\"capacity\":4,\"name\":\"Standard_D2als_v6\",\"priority\":\"Spot\",\"zone\":\"3\"},{\"capacity\":3,\"name\":\"Standard_D2als_v6\",\"priority\":\"Spot\",\"zone\":\"1\"},{\"capacity\":3,\"name\":\"Standard_D2als_v6\",\"priority\":\"Spot\",\"zone\":\"2\"}]},{\"id\":\"115316e7-c5c5-4553-87dc-ec16ce3b99d6\",\"score\":9,\"skuSplit\":[{\"capacity\":4,\"name\":\"Standard_D2als_v7\",\"priority\":\"Spot\",\"zone\":\"3\"},{\"capacity\":3,\"name\":\"Standard_D2als_v7\",\"priority\":\"Spot\",\"zone\":\"1\"},{\"capacity\":3,\"name\":\"Standard_D2als_v7\",\"priority\":\"Spot\",\"zone\":\"2\"}]},{\"id\":\"12bf3abf-d68a-4ca0-9e66-67a87b7506bd\",\"score\":9,\"skuSplit\":[{\"capacity\":3,\"name\":\"Standard_D2ls_v5\",\"priority\":\"Spot\",\"zone\":\"2\"},{\"capacity\":4,\"name\":\"Standard_D2ls_v5\",\"priority\":\"Spot\",\"zone\":\"3\"},{\"capacity\":3,\"name\":\"Standard_D2ls_v5\",\"priority\":\"Spot\",\"zone\":\"1\"}]},{\"id\":\"f2528494-eff9-4bf5-8434-d3de035685f6\",\"score\":9,\"skuSplit\":[{\"capacity\":3,\"name\":\"Standard_D2as_v5\",\"priority\":\"Spot\",\"zone\":\"2\"},{\"capacity\":4,\"name\":\"Standard_D2as_v5\",\"priority\":\"Spot\",\"zone\":\"3\"},{\"capacity\":3,\"name\":\"Standard_D2as_v5\",\"priority\":\"Spot\",\"zone\":\"1\"}]},{\"id\":\"b53d18c5-0a09-4287-9fc9-7da02007a1c9\",\"score\":9,\"skuSplit\":[{\"capacity\":4,\"name\":\"Standard_D2ls_v6\",\"priority\":\"Spot\",\"zone\":\"3\"},{\"capacity\":3,\"name\":\"Standard_D2ls_v6\",\"priority\":\"Spot\",\"zone\":\"2\"},{\"capacity\":3,\"name\":\"Standard_D2ls_v6\",\"priority\":\"Spot\",\"zone\":\"1\"}]},{\"id\":\"459c9fd9-c103-420e-9ec8-2b80445930a9\",\"score\":9,\"skuSplit\":[{\"capacity\":3,\"name\":\"Standard_D2as_v7\",\"priority\":\"Spot\",\"zone\":\"1\"},{\"capacity\":3,\"name\":\"Standard_D2as_v7\",\"priority\":\"Spot\",\"zone\":\"2\"},{\"capacity\":4,\"name\":\"Standard_D2as_v7\",\"priority\":\"Spot\",\"zone\":\"3\"}]},{\"id\":\"963adc5c-5aa2-4146-87bc-6c25e59595f7\",\"score\":9,\"skuSplit\":[{\"capacity\":4,\"name\":\"Standard_D2s_v3\",\"priority\":\"Spot\",\"zone\":\"3\"},{\"capacity\":3,\"name\":\"Standard_D2s_v3\",\"priority\":\"Spot\",\"zone\":\"1\"},{\"capacity\":3,\"name\":\"Standard_D2s_v3\",\"priority\":\"Spot\",\"zone\":\"2\"}]},{\"id\":\"e4dbdda2-8614-4d13-9e6b-a4f1f6f3c175\",\"score\":9,\"skuSplit\":[{\"capacity\":4,\"name\":\"Standard_D2alds_v6\",\"priority\":\"Spot\",\"zone\":\"3\"},{\"capacity\":3,\"name\":\"Standard_D2alds_v6\",\"priority\":\"Spot\",\"zone\":\"1\"},{\"capacity\":3,\"name\":\"Standard_D2alds_v6\",\"priority\":\"Spot\",\"zone\":\"2\"}]},{\"id\":\"a608ae7f-2149-49aa-8266-3bfec27a018e\",\"score\":9,\"skuSplit\":[{\"capacity\":7,\"name\":\"Standard_D2als_v6\",\"priority\":\"Spot\",\"zone\":\"3\"},{\"capacity\":3,\"name\":\"Standard_D2als_v6\",\"priority\":\"Spot\",\"zone\":\"2\"}]},{\"id\":\"3bfadc2e-64b0-4957-b42c-d9fe52e6cf35\",\"score\":9,\"skuSplit\":[{\"capacity\":6,\"name\":\"Standard_D2als_v6\",\"priority\":\"Spot\",\"zone\":\"3\"},{\"capacity\":4,\"name\":\"Standard_D2als_v6\",\"priority\":\"Spot\",\"zone\":\"2\"}]}],\"validUntil\":\"2026-08-21T22:43:52.2811946Z\"}"}
```

I will bring this documentation nit up with the owners of the SKU Split API.

matthchr: I sent a PR to fix the docs here: https://github.com/Azure/azure-rest-api-specs/pull/45891

pkg/providers/allocationstrategy/stages/sku_mix_rank.go
matthchr: This is a good callout actually -- I'll update the comment here to mention this.

I think it's OK to limit to 10 of each type because that's what we'll be doing when the API succeeds and practically speaking 10+ VM size attempts probably takes enough time that by the time we've gotten through them all deleting the NodeClaim and attempting a new one would be OK.

pkg/providers/allocationstrategy/stages/sku_mix_rank.go
matthchr: I am not 100% sure I am following the comment here but our required output to plug in to the rest of the Karpenter provisioning pipeline is to produce an ordered set of offerings/instanceTypes, so at some point we need to join things back together.

Maybe this is saying the same thing as I am saying in the comment below?
```
	// The net result here is that we're just using the SKU Mix API to filter offerings. It does NOT reorder/reprioritize them
	// currently due to the need to join the (up to) 4 recommendation groups into a single list. We have an outstanding ask to
	// the team to allow us to ask for spot + dedicated (at least) in a single list so that we can get back a true "price-prioritized"
	// list of recommendations without having to join the two lists together (which requires some external source of truth to order between
	// the two lists). Once we have that, we can remove the re-ranking step below and rely on the SKU API to provide the correct ordering.
```

I think in order to join the (up to 4) separate lists into a single list, we need some mechanism of ordering between the lists, and today that's price. There's an outstanding ask to the placement scores team to allow us to ask for spot + dedicated in the same request which hopefully they will action - assuming they do, we should be able to get a single list back and then we don't necessarily have to re-order it, so the APIs order could "win".

pkg/providers/capacityrecommendation/capacity_recommendation.go
```diff
@@ -0,0 +1,458 @@
...
+
+func (p *DefaultProvider) fetchAndCache(ctx context.Context, key string, input *RankingInput) ([]Recommendation, error) {
+	requestCtx, cancel := context.WithTimeout(ctx, requestTimeout)
+	defer cancel()
+
+	response, err := p.client.Post(requestCtx, p.location, toSKUMixPlacementRequest(input), nil)
+	if err != nil {
+		return nil, err
+	}
+	// TODO: Remove this or increase verbosity?
```
matthchr, line 142: Fix this before merge.

pkg/providers/allocationstrategy/stages/sku_mix_rank.go
```diff
@@ -0,0 +1,378 @@
...
+				// If our API failed, we want to fail open so add all items in the group to recommended
+				// Note that the group only has maxRecommendationVMSizes so even in the case the API fails
+				// we won't be recommending more than that number of VM sizes to the caller.
+				addLocalCandidates(group, recommendedOfferings)
+			}
+			continue
+		}
+
+		comparison := compareRecommendationGroup(group, recommendations)
+		log.FromContext(ctx).Info("compared SKU Mix Placement recommendations with local ranking",
```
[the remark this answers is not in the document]
matthchr: This is a fair comment, I will fix.

pkg/providers/allocationstrategy/stages/sku_mix_rank.go
```diff
@@ -0,0 +1,379 @@
...
+			ctx,
+			&capacityrecommendation.RankingInput{
+				VMSizes:      group.vmSizes,
+				Zones:        group.zones,
+				CapacityType: group.key.capacityType,
+				OSType:       group.key.osType,
+				Count:        initialRequestCount,
+			},
+		)
+		if err != nil {
```
theunrepentantgeek: If every group encounters an error calling `GetRecommendations()`, they'll all trigger **continue** on line 99 and all groups will be empty when the loop exits.

Is that intended behaviour?

pkg/providers/capacityrecommendation/capacity_recommendation.go
```diff
@@ -0,0 +1,458 @@
...
+	if input.CapacityType == karpv1.CapacityTypeSpot {
+		priority = armrecommender.SKUMixPlacementPrioritySpot
+	}
+	var osType armrecommender.SKUMixPlacementOSType
+	switch input.OSType {
+	case corev1.Linux:
+		osType = armrecommender.SKUMixPlacementOSTypeLinux
+	case corev1.Windows:
+		osType = armrecommender.SKUMixPlacementOSTypeWindows
+	}
```
theunrepentantgeek, line 223: Does this switch need a default/early-exit-error for the case where no value is specified?

If no value for `input.OSType` is a valid circumstance, consider commenting to that effect.

theunrepentantgeek (dismissed):

pkg/providers/allocationstrategy/stages/sku_mix_rank.go
matthchr: Yes, there are two cases:
Case 1: We are in log mode, in which case we don't reorder things anyway this entire stage is just for logging, so no issue just continuing.
Case 2: We are in Enabled mode, in which case we call `addLocalCandidates` before continuing and that ensures that we still consider all the candidates as if the API wasn't called at all.

pkg/providers/capacityrecommendation/capacity_recommendation.go
matthchr: Yes it is validated above in `validateInput` as well. I've added a comment.

pkg/providers/allocationstrategy/stages/sku_mix_rank.go
```diff
@@ -0,0 +1,379 @@
...
+	// currently due to the need to join the (up to) 4 recommendation groups into a single list. We have an outstanding ask to
+	// the team to allow us to ask for spot + dedicated (at least) in a single list so that we can get back a true "price-prioritized"
+	// list of recommendations without having to join the two lists together (which requires some external source of truth to order between
+	// the two lists). Once we have that, we can remove the re-ranking step below and rely on the SKU API to provide the correct ordering.
+	if s.mode == consts.ComputeRecommendationModeEnabled {
+		instanceOfferings = filterOfferingsByRecommended(instanceOfferings, recommendedOfferings)
+		// TODO: Currently we re-rank instance offerings because we don't 100% trust the SKU API response (see comments on placementChoiceIsBetter in capacityrecommendation.go).
+		// Once we have more confidence in the API, we can remove this re-ranking step and rely on the SKU API to provide the correct ordering.
+		// We will still need to join the (up to) 4 recommendation groups into a single list somehow though...
+		instanceOfferings = rankInstanceOfferings(instanceOfferings)
```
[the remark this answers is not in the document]
matthchr: This is wrong - this is currently required in order to join the separate lists together as otherwise we have no way to join in a standard way.

theunrepentantgeek (dismissed):

theunrepentantgeek (approved):

tallaxes (approved):

PR #1874, opened 2026-08-26
fix: isolate golangci-lint formatter state per worktree

comtalyst:
(AI-generated)

**Description**

The repository enables golangci-lint formatter fixes, but its default cache is shared across linked Git worktrees. Cached formatter state from one checkout can therefore be replayed while linting another checkout.

This change routes Make and pre-commit linting through one fail-closed wrapper. The wrapper stores cache state under the current worktree's absolute Git directory, preserves explicit Make command-line overrides, ignores ambient and `make -e` cache values, and runs each non-test Go module from its own directory.

This PR is stacked on https://github.com/Azure/karpenter-provider-azure/pull/1873 because it extends the same toolchain regression harness.

* Added primary/linked worktree, shell-metacharacter path, nested-module, ambient/explicit Make precedence, pre-commit, no-Git failure, linter failure propagation, and hostile Git-hook cases.
* Reproduced sibling worktree source truncation with the shared fix-enabled cache, then verified real linting with the isolated cache preserves sibling hashes.
* `make verify` passed at `eda74aa4` with `0 issues` and a clean worktree.
* `make presubmit` passed 39 suites at `eda74aa4` in `15m41.305127638s` with 68.8% composite coverage.
* CI passed at `eda74aa4`: https://github.com/Azure/karpenter-provider-azure/actions/runs/32992074192.
* All seven Kubernetes-version jobs passed at `eda74aa4`: https://github.com/Azure/karpenter-provider-azure/actions/runs/32992077215.

- [ ] Yes, issue opened: #
- [x] No

**Release Note**

```release-note
NONE
```

### DIFF
--- a/hack/golangci-cache-path.sh
+++ b/hack/golangci-cache-path.sh
@@ -1,0 +1,14 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+if (( $# != 0 )); then
+    echo "usage: $0" >&2
+    exit 2
+fi
+
+git_dir=$(git rev-parse --absolute-git-dir)
+if [[ -z "$git_dir" || ! -d "$git_dir" ]]; then
+    echo "unable to resolve an absolute Git directory" >&2
+    exit 1
+fi
+printf '%s/golangci-lint-cache\n' "$git_dir"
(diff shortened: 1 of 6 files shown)


hack/golangci-cache-path.sh
```diff
@@ -0,0 +1,14 @@
...
+    echo "usage: $0" >&2
+    exit 2
+fi
+
+git_dir=$(git rev-parse --absolute-git-dir)
+if [[ -z "$git_dir" || ! -d "$git_dir" ]]; then
+    echo "unable to resolve an absolute Git directory" >&2
+    exit 1
+fi
+printf '%s/golangci-lint-cache\n' "$git_dir"
```
theunrepentantgeek, line 14: I like the idea of isolating the cache folders (and will steal it for other projects 😁) but wouldn't it be better to use a hidden directory, like `.golangci-lint-cache` as the cache folder?

hack/toolchain_test.sh
theunrepentantgeek, line 1: I'm not sure about these enormous bash scripts that purport to be tests - they're going to be downright obtuse to debug if/when a test failure occurs. Is there a better way?

PR #1875, opened 2026-08-27
feat: Add NPS/Bootstrap support for Azure Container Linux

KyleGospo:
This is a continuation of #1841
Add `AzureContainerLinux` support to the NPS / bootstrapping-client provisioning path.

**Design flow**

1. **NodeClass selection**
   - Users select `spec.imageFamily: AzureContainerLinux`.
   - Admission validation enforces the 60 GB minimum OS disk, requires Secure Boot and vTPM, and rejects `artifactStreaming.enabled=true` (ACL's immutable `/usr` is incompatible with artifact streaming).
   - The validation reconciler rejects the nodeclass with a clear condition when the cluster is on Kubernetes < 1.34 or when the provider is running in `aksscriptless` mode (which cannot obtain Ignition from AgentBaker).

2. **Image resolution**
   - Karpenter selects the Azure Container Linux image definition from the Azure Linux gallery.
   - Architecture and FIPS determine one of four variants: AMD64, ARM64, AMD64 FIPS, or ARM64 FIPS.
   - Non-Trusted-Launch requests return no image (defense-in-depth alongside the CEL rules).

3. **Instance-type filtering**
   - Trusted Launch-incompatible SKUs are removed from consideration.
   - ARM64 is restricted to Cobalt v6 (`Standard_*p*_v6`) SKUs with an NVMe data disk — the only ARM64 SKUs where the ACL Ignition bootstrap has been validated.
   - GPU nodes use the Azure Container Linux compatibility list; supported NVIDIA SKUs remain available while unsupported NVIDIA and AMD SKUs are excluded.

4. **VM creation**
   - The VM is created through the Compute API with `SecurityType=TrustedLaunch`, `SecureBootEnabled=true`, and `VTpmEnabled=true`, using the resolved ACL image and bootstrapping via AgentBaker-generated Ignition.
   - The custom-data hydrator walks the Ignition doc and rewrites the `{{.TokenID}}.{{.TokenSecret}}` placeholder inside gzip-compressed `storage.files[]` entries so kubelet gets a valid TLS bootstrap token.

- Unit and acceptance suites pass locally on the changed packages (`pkg/apis/v1beta1`, `pkg/providers/imagefamily`, `pkg/providers/imagefamily/customscriptsbootstrap`, `pkg/providers/instancetype`, `pkg/providers/instance`, `pkg/controllers/nodeclass/status`).
- New reconciler tests cover both provision-mode compatibility (`aksscriptless` rejected; `bootstrappingclient` accepted) and Kubernetes version compatibility (1.30–1.33 rejected, 1.34+ accepted).
- New CEL test covers rejecting `artifactStreaming.enabled=true` alongside `imageFamily=AzureContainerLinux`.
- New unit test covers the nested-Ignition-gzip token replacement.
- Created a new AKS instance and confirmed working provisioning with ACL

- [ ] Yes, issue opened: #
- [x] No

**Release Note**

```release-note
Add Azure Container Linux support to AKSNodeClass on the NPS / bootstrapping-client provisioning path.


MaximilianoUribe:
did we test this with an actual node being provisioned here @KyleGospo?

KyleGospo:
> did we test this with an actual node being provisioned here Kyle Gospodnetich (@KyleGospo)?

Yes, this has since been tested with a full provisioning cycle in AKS.

PR #1884, opened 2026-09-02
Raise LocalDNS Preferred Kubernetes threshold to 1.37

saewoni:
## Summary

- Raise the LocalDNS `Preferred` Kubernetes version threshold from `1.99.0` to `1.37.0`.
- Update the status reconciler documentation to reflect the `1.37` gate.
- Update LocalDNS status tests for the `1.36`/`1.37` boundary and add explicit coverage for `1.37.0`.

This aligns the Karpenter provider with the planned AKS Kubernetes 1.37 LocalDNS default re-enablement.

## Test plan

- `git diff --check` passed.
- `go test ./pkg/controllers/nodeclass/status` attempted, but the local environment is missing `/usr/local/kubebuilder/bin/etcd`; envtest failed in `BeforeSuite` before running specs.

### DIFF
--- a/pkg/controllers/nodeclass/status/localdns.go
+++ b/pkg/controllers/nodeclass/status/localdns.go
@@ -45,7 +45,7 @@ var localDNSPreferredVersionThreshold = lo.Must(semver.ParseTolerant(localDNSPre
 const (
 	// localDNSPreferredK8sVersionThreshold is the minimum Kubernetes version
 	// required to auto-enable LocalDNS when Spec.LocalDNS.Mode=Preferred.
-	localDNSPreferredK8sVersionThreshold = "1.99.0"
+	localDNSPreferredK8sVersionThreshold = "1.37.0"
 
 	// konnectivityAgentPolicy{Name,Namespace} identify the AKS-managed
 	// NetworkPolicy that is allow-listed when scanning for conflicting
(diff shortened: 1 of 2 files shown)


yewmsft (approved):

yewmsft:
Heads-up from #1894, which adds the first e2e coverage for the Preferred -> Enabled path. Nothing to change here — this is all downstream of your change.

**The threshold is already wired to the e2e gate, so this PR needs no test edit.** #1894 exports the const as `LocalDNSPreferredK8sVersionThreshold` and the integration spec `should enable LocalDNS on an above-floor SKU under Mode=Preferred` compares it against the live `ServerVersion()` at run time rather than hardcoding a number. When this merges the gate re-aims itself to 1.37 on its own.

**What to watch: that spec will still skip after this lands.** The standalone RP tops out at k8s 1.36.4 today, so `1.37.0` is unreachable there just as `1.99.0` was. The spec moves from "skips because the threshold is deliberately unreachable" to "skips because no cluster is new enough yet" — a quieter state that is easy to stop noticing. Worth confirming it actually runs rather than skips the first time a >=1.37 cluster is available. Until then the Enabled path has no e2e coverage, only `TestPreferred_K8sAtThreshold_Enabled` here plus the manual snapshot matrix recorded in #1894.

**Merge order is a non-issue.** I test-merged this branch into #1894's head locally: both `localdns.go` and `localdns_test.go` auto-merge with zero conflicts. #1894 inserts above the const block and rewrites the sticky comment further down; neither overlaps your two hunks. Whichever lands first, the other needs no conflict rebase.

`TestPreferred_K8sAtThreshold_Enabled` is a good catch to add — at `1.99.0`/`1.98.0` there was no at-the-boundary case.

PR #1886, opened 2026-09-03
Use Microsoft Go for FIPS-compliant release builds

tallaxes:
Use the Microsoft build of Go for the release dependency/toolchain setup.

The existing FIPS release work switches the image to Azure Linux 3 and enables CGO. This change configures actions/setup-go to resolve the Go version from go.mod using Microsoft's official distribution endpoint, while preserving the pinned action SHA and updating the toolchain cache key.

This ensures release images are built with Microsoft Go's systemcrypto/OpenSSL backend on Linux.

### DIFF
--- a/.github/actions/install-deps/action.yaml
+++ b/.github/actions/install-deps/action.yaml
@@ -11,7 +11,7 @@ runs:
       id: setup-go
       with:
         go-version-file: go.mod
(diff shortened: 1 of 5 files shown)


tallaxes:
Superseded by #1887, which is based directly on main.

PR #1895, opened 2026-09-07
feat: Close validation gap with AKS checking supported Ubuntu version

theunrepentantgeek:
**Description**

This pull request introduces improved validation logic for image family and Kubernetes version compatibility in the Azure cloud provider, along with comprehensive test coverage for these scenarios. The main focus is to ensure that when a `NodeClass` specifies an incompatible image family for the discovered Kubernetes version, appropriate errors are surfaced early, and no disruptive drift or resource creation occurs. The changes also clarify that such validation failures do not cause drift for already-launched nodes.

```
make presubmit
```

- [ ] Yes, issue opened: # 
- [x] No

**Release Note**

```release-note
Validation will now block early for unsupported combinations of Kubernetes and Ubuntu instead of failing to reconcile.
```


theunrepentantgeek:
@copilot resolve the merge conflicts in this pull request

pkg/providers/imagefamily/compatibility.go
```diff
@@ -0,0 +1,219 @@
...
+// resolution. AzureLinux is likewise unpinned.
+var kubernetesVersionPinnedImageFamilies = map[string]kubernetesVersionPolicy{
+	v1beta1.Ubuntu2204ImageFamily: {
+		minimumVersion:     semver.MustParse("1.25.2"),
+		maximumVersion:     lo.ToPtr(semver.MustParse("1.37.0")),
+		fipsMaximumVersion: lo.ToPtr(semver.MustParse("1.39.0")),
+	},
+	v1beta1.Ubuntu2404ImageFamily: {
+		// Ubuntu2404 has no upper bound, and no FIPS-specific bound today.
+		minimumVersion: semver.MustParse("1.32.0"),
```
theunrepentantgeek, line 88: 🤖 TheUnrepentantRobot says:

**High · Compatibility** — AKS currently documents `Ubuntu2404` as supported only on Kubernetes 1.32–1.38, but this policy has no upper bound. On a 1.39+ cluster this validation therefore marks the NodeClass ready and both Machine API modes can submit an unsupported OS SKU, moving the failure back to provisioning. Please add an exclusive `1.39.0` maximum and cover rejection at that boundary; the existing “has no upper bound” test should be inverted.

pkg/controllers/nodeclass/status/validation.go
```diff
@@ -99,6 +104,42 @@ func (r *ValidationReconciler) Reconcile(ctx context.Context, nodeClass *v1beta1
...
+				logger.V(1).Info("image family compatibility validation failed", "error", err)
+				nodeClass.StatusConditions().SetFalse(
+					v1beta1.ConditionTypeValidationSucceeded,
+					ImageFamilyKubernetesVersionIncompatible,
+					err.Error(),
+				)
+				return reconcile.Result{}, nil
+			}
+
+			var malformedKubernetesVersionErr *imagefamily.MalformedDiscoveredKubernetesVersionError
```
theunrepentantgeek: 🤖 TheUnrepentantRobot says:

**Medium · Correctness** — This typed-error branch and the fallback immediately below produce the same reconcile result and the same wrapped error; only the log text differs. That makes `MalformedDiscoveredKubernetesVersionError`, its `Unwrap`, this `errors.As`, and their dedicated tests a sizeable distinction without behavioural meaning. Please keep the incompatibility type (it drives the condition), but let every other validation error flow through one ordinary return path. That removes a branch and a public error contract without changing behaviour.

pkg/providers/imagefamily/compatibility_unit_test.go
```diff
@@ -0,0 +1,146 @@
...
+	"github.com/Azure/karpenter-provider-azure/pkg/apis/v1beta1"
+)
+
+// TestKubernetesVersionPinnedImageFamilyRegistry pins the invariant that
+// kubernetesVersionPinnedImageFamilies is the only source of truth for pinned-family
+// policy: for every registered entry, the predicate, the accepted/rejected version
+// bounds, and the bounds reported on the typed error must all agree with the registry
+// itself. Adding a family (or changing a bound) without keeping the predicate and the
+// validator in step therefore fails here rather than silently diverging.
+func TestKubernetesVersionPinnedImageFamilyRegistry(t *testing.T) {
```
theunrepentantgeek: 🤖 TheUnrepentantRobot says:

**Medium · Test Coverage** — This 146-line test derives the family, bounds, and expected FIPS behaviour from the same registry and `upperBound` method used by the implementation. If a policy value is wrong—for example, if the Ubuntu2404 upper bound is omitted—the test faithfully repeats that mistake and passes. The black-box table in `compatibility_test.go` already pins concrete families, boundary versions, typed fields, and predicate results. Please delete this reflective suite rather than maintain two overlapping test models; fixing the separate assertion at line 127 would still leave the core test tautological.

pkg/providers/imagefamily/compatibility.go
theunrepentantgeek: The current release is v1.37, so I don't think we should be prejudging v1.39+ this early. Plus, it's highly unlikely for support to be removed, so a policy of "Supported by default" seems safe.

pkg/controllers/nodeclass/status/validation.go
theunrepentantgeek: Agreed, we can clean that up.

pkg/providers/imagefamily/compatibility_unit_test.go
theunrepentantgeek: Agreed.

pkg/providers/imagefamily/compatibility_unit_test.go
```diff
@@ -0,0 +1,146 @@
...
+	} else {
+		g.Expect(incompatibleErr.MaximumVersion).ToNot(BeNil())
+		g.Expect(incompatibleErr.MaximumVersion.String()).To(Equal(maximumVersion.String()))
+	}
+
+	expectedRange := fmt.Sprintf(">= %s", policy.minimumVersion)
+	if maximumVersion != nil {
+		expectedRange = fmt.Sprintf("%s and < %s", expectedRange, maximumVersion)
+	}
+	g.Expect(err.Error()).To(ContainSubstring("supported range is %s", expectedRange))
```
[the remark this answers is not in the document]
theunrepentantgeek: We should have a separate assertion for each component piece of string, plus we should use `g.Expect(err).To(MatchError(...))`.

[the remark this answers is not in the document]
theunrepentantgeek: 🤖 TheUnrepentantRobot says:

Removed the redundant reflective suite in a04163ec, including this assertion. The retained black-box malformed-version checks now use MatchError.

pkg/controllers/nodeclass/status/validation.go
theunrepentantgeek: 🤖 TheUnrepentantRobot says:

Removed MalformedDiscoveredKubernetesVersionError and the behaviorless errors.As branch in a04163ec. Non-incompatibility errors now use the single ordinary retry path.

pkg/providers/imagefamily/compatibility_unit_test.go
theunrepentantgeek: 🤖 TheUnrepentantRobot says:

Deleted the reflective compatibility_unit_test.go suite in a04163ec; the concrete black-box boundary coverage remains in compatibility_test.go.

pkg/providers/imagefamily/compatibility.go
```diff
@@ -0,0 +1,205 @@
...
+// FIPS does not change the reported range, so it is not flagged on the error either.
+func (p kubernetesVersionPolicy) upperBound(fips bool) (maximumVersion *semver.Version, fipsApplied bool) {
+	if fips && p.fipsMaximumVersion != nil {
+		return lo.ToPtr(*p.fipsMaximumVersion), true
+	}
+	if p.maximumVersion == nil {
+		return nil, false
+	}
+	return lo.ToPtr(*p.maximumVersion), false
+}
```
comtalyst, line 58: IMO it is simpler to unify this with `permits()` that abstract the concept of FIPS maximum version away from the caller. It would also prevent the potential confusion on having `maximumVersion` as a parameter of `permits()` at the same time as in `kubernetesVersionPolicy.maximumVersion`.

pkg/providers/imagefamily/compatibility.go
```diff
@@ -0,0 +1,205 @@
...
+// spec.imageFamily is explicitly version pinned, and so needs the discovered cluster
+// Kubernetes version in order to be validated.
+//
+// Callers use this to avoid making Kubernetes version readiness a precondition for
+// NodeClasses that no compatibility policy applies to: for those, an unavailable or
+// malformed Kubernetes version must not block the rest of validation.
+func RequiresKubernetesVersionCompatibility(nodeClass *v1beta1.AKSNodeClass) bool {
+	_, found := kubernetesVersionPolicyFor(nodeClass)
+	return found
+}
```
comtalyst, line 113: The alternative is to process every NodeClass instead of gating on `RequiresKubernetesVersionCompatibility()`, but implement the `permits()` in a way that it tolerates missing `minimumVersion`, `maximumVersion` etc., so that no-policy NodeClass is always allowed. 
The upside is it may save us some amount of code that was for the booleans (`RequiresKubernetesVersionCompatibility` and `kubernetesVersionPolicyFor`). Thoughts?

pkg/providers/imagefamily/compatibility.go
```diff
@@ -0,0 +1,205 @@
...
+// discovered cluster Kubernetes version does not support.
+//
+// This is a static, deterministic incompatibility: it can only change when the
+// NodeClass spec changes or the cluster Kubernetes version changes, both of
+// which trigger a fresh reconciliation.
+type ImageFamilyKubernetesVersionIncompatibleError struct {
+	// RequestedImageFamily is the value of spec.imageFamily that was validated.
+	RequestedImageFamily string
+	// KubernetesVersion is the discovered cluster Kubernetes version, as discovered (unparsed).
+	KubernetesVersion string
```
comtalyst, line 126:
```suggestion
	RequestedKubernetesVersion string
```
The downside is K8s version is cluster-level rather than "user"-requested, but that may change(?). 
The upside is it will be more clear when contrasting with `MinimumVersion` and `MaximumVersion`. And we could define "requested" to be "requested by the caller to this validation". 
Thoughts?

comtalyst (commented):
Looks correct overall. And, could you run and attach E2E results for sanity check?

go.mod
```diff
@@ -7,34 +7,34 @@ require (
 	github.com/Azure/aks-middleware v0.0.42
 	github.com/Azure/azure-kusto-go v0.16.1
 	github.com/Azure/azure-sdk-for-go v68.0.0+incompatible
-	github.com/Azure/azure-sdk-for-go-extensions v0.6.0
+	github.com/Azure/azure-sdk-for-go-extensions v0.7.0
```
matthchr, line 10: These updates probably dont need to be here? maybe need a merge or rebase w/ main?

pkg/providers/imagefamily/compatibility.go
```diff
@@ -0,0 +1,205 @@
...
+		return lo.ToPtr(*p.fipsMaximumVersion), true
+	}
+	if p.maximumVersion == nil {
+		return nil, false
+	}
+	return lo.ToPtr(*p.maximumVersion), false
+}
+
+// permits reports whether version falls within [minimumVersion, maximumVersion).
+func (p kubernetesVersionPolicy) permits(version semver.Version, maximumVersion *semver.Version) bool {
```
matthchr, line 61: Why is maxVersion a parameter but not minVers? When do we override maxVer?

pkg/providers/imagefamily/compatibility.go
```diff
@@ -0,0 +1,205 @@
...
+	}
+
+	version, err := parseKubernetesVersionTolerant(kubernetesVersion)
+	if err != nil {
+		return err
+	}
+
+	fips := lo.FromPtr(nodeClass.Spec.FIPSMode) == v1beta1.FIPSModeFIPS
+	maximumVersion, fipsApplied := policy.upperBound(fips)
+	if policy.permits(version, maximumVersion) {
```
matthchr, line 182: Would a better interface be:

```
policy, found := kubernetesVersionPolicyFor(lo.FromPtr(nodeClass.Spec.ImageFamily))
ok, reason := policy.permits(version, nodeClass)
```

and then maybe the error if `Reason` was set would use that direct reason ("With fips"), and otherwise would just talk about min/max?

I guess the problem with that is that maximumVersion changes with fips or not so you also need to get the max version back...

The thing that feels wonky to me here is that we're doing:
```
maximumVersion, fipsApplied := policy.upperBound(fips)
policy.permits(version, maximumVersion)
```

which in some ways feels like feeding information that `policy` already knows back to itself (it already knows the fips max version). So it seems strange to be passing maxVersion there in `permits` rather than something like `nodeClass` because it means that fact that FIPS matters leaks in a way that it wouldn't otherwise and I am mostly wondering if we can come up with a way to make things a bit less FIPS-centric and slightly more generic (encapsulate a bit more of the logic in policy, I guess?)

pkg/providers/imagefamily/compatibility.go
```diff
@@ -0,0 +1,205 @@
...
+	return &ImageFamilyKubernetesVersionIncompatibleError{
+		RequestedImageFamily: lo.FromPtr(nodeClass.Spec.ImageFamily),
+		KubernetesVersion:    kubernetesVersion,
+		FIPS:                 fipsApplied,
+		MinimumVersion:       policy.minimumVersion,
+		MaximumVersion:       maximumVersion,
+	}
+}
+
+func parseKubernetesVersionTolerant(kubernetesVersion string) (semver.Version, error) {
```
matthchr, line 195: minor: Not 100% sold we need this wrapper which just changes the error message slightly, but if we do it probably could be somewhere in utils and used elsewhere that we do the `strings.TrimPrefix(kubernetesVersion, "v")` thing too?

pkg/controllers/nodeclass/status/validation.go
```diff
@@ -125,6 +116,72 @@ func (r *ValidationReconciler) Reconcile(ctx context.Context, nodeClass *v1beta1
...
+	// remaining validations just because the Kubernetes version isn't available yet.
+	if !imagefamily.RequiresKubernetesVersionCompatibility(nodeClass) {
+		return false, nil
+	}
+
+	kubernetesVersion, err := nodeClass.GetKubernetesVersion()
+	if err != nil {
+		return false, fmt.Errorf("getting kubernetes version: %w", err)
+	}
+	if err := imagefamily.ValidateImageFamilyCompatibility(nodeClass, kubernetesVersion); err != nil {
```
matthchr, line 163: KubernetesVersion is on NodeClass, so do we really need to extract it here and pass it in, versus just extract it in the `ValidateImageFamilyCompatibility` function? (it could still call this same `GetKubernetesVersion()` helper I think?

Not necessarily saying that is better but I am curious what not doing that since as far as I can tell we don't actually use this value anywhere except to pass it to that function.

matthchr (commented):
Didn't quite get to review everything in a ton of depth but at least have some questions/comments

pkg/controllers/nodeclass/status/validation.go
theunrepentantgeek: It was to make the tests clearer, but I agree with you that we can dispense with that.

pkg/providers/imagefamily/compatibility.go
theunrepentantgeek: Huh.

TIL: `semver.ParseTolerant` already handles trimming any leading `v` ... so technically this now supports `vv1.37` which is just ... weird.

Removing this method, and taking a close look at the rest of the code too.

pkg/providers/imagefamily/compatibility.go
theunrepentantgeek: I split FIPS from non-FIPS policies and gave the policies descriptions which are used in the error messages. I think this is cleaner.

pkg/providers/imagefamily/compatibility.go
theunrepentantgeek: This was a part of the "loop" used to handle FIPS vs non-FIPS - which is now handled in a cleaner way.

pkg/providers/imagefamily/compatibility.go
theunrepentantgeek: I agree. I've split FIPS from non-FIPS policies and now the interface is just `permits(version) bool`.

pkg/providers/imagefamily/compatibility.go
```diff
@@ -0,0 +1,204 @@
...
+	"strings"
+
+	"github.com/blang/semver/v4"
+	"github.com/samber/lo"
+
+	"github.com/Azure/karpenter-provider-azure/pkg/apis/v1beta1"
+)
+
+// kubernetesVersionPolicy is the Kubernetes version window in which one explicitly
+// version-pinned spec.imageFamily value is usable.
```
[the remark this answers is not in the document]
theunrepentantgeek: Done.

pkg/controllers/nodeclass/status/validation.go
```diff
@@ -34,6 +35,10 @@ import (
 
 const (
 	DiskEncryptionSetRBACMissing = "DiskEncryptionSetRBACMissing"
+	// ImageFamilyKubernetesVersionIncompatible is the stable reason used when
+	// spec.imageFamily explicitly pins an OS version that the ready discovered
+	// Kubernetes version does not support.
```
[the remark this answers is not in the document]
theunrepentantgeek: Simplified.

pkg/providers/imagefamily/compatibility.go
theunrepentantgeek: Agreed. The only effect here was to call `kubernetesVersionPolicyFor()` multiple times, which seems wasteful.

I don't normally dislike predicates like this (I'm a big fan of them) but in this case idiomatic Go looks different.

pkg/providers/imagefamily/compatibility.go
theunrepentantgeek: Works for me.

go.mod
theunrepentantgeek: After `git merge main` these are now clean.

pkg/controllers/nodeclass/status/validation.go
matthchr: minor, style: could we put this into a function helper too with simular signature: (ctx, nodeClass) (bool, err), and then just have this controller be an iteration over 3 `[]validateFunc`'s? 1 for kata, 1 for imageCompatibility, and 1 for DiskEncryptionSetId?

pkg/controllers/nodeclass/status/validation_test.go
```diff
@@ -58,6 +62,22 @@ func createZoneOverride(zone string, forwardToVnetDNS bool) v1beta1.LocalDNSZone
...
 }
 
+func setKubernetesVersionReady(nodeClass *v1beta1.AKSNodeClass, kubernetesVersion string) {
+	nodeClass.Status.KubernetesVersion = lo.ToPtr(kubernetesVersion)
+	nodeClass.StatusConditions().SetTrue(v1beta1.ConditionTypeKubernetesVersionReady)
+}
+
+// newAuthorizationError returns the authorization failure the DES API surfaces when the
+// controlling identity lacks the Reader role, for the DES RBAC cases below.
+func newAuthorizationError(statusCode int) error {
```
matthchr: Is it really an authorization error if the status code is configurable?

pkg/controllers/controllers.go
```diff
@@ -81,7 +81,8 @@ func NewControllers(
 
 		// TODO: nodeclaim tagging
 		inplaceupdate.NewController(kubeClient, vmInstanceProvider, aksMachineInstanceProvider),
-		status.NewController[*v1beta1.AKSNodeClass](kubeClient, mgr.GetEventRecorderFor("karpenter")), //nolint:staticcheck // SA1019: will be replaced by mgr.GetEventRecorder once operatorpkg is updated
+		//nolint:staticcheck,nolintlint // SA1019: will be replaced by mgr.GetEventRecorder once operatorpkg is updated
```
matthchr, line 84: minor, why this change?

pkg/providers/imagefamily/compatibility.go
```diff
@@ -0,0 +1,182 @@
...
+// window each of them supports. Both RequiresKubernetesVersionCompatibility and
+// ValidateImageFamilyCompatibility are derived from it, so a family cannot be
+// treated as pinned without having bounds, or be bounded without being treated as
+// pinned.
+//
+// Validation is deliberately limited to the entries here: the generic "Ubuntu" family
+// (and an unset spec.imageFamily) is a rolling contract to run "an AKS-supported
+// Ubuntu", resolved per Kubernetes version by the provisioning path - the image
+// resolver for VM-based provisioning, and AKS itself for AKS Machine API provisioning
+// - so there is no version range we could enforce here without contradicting that
```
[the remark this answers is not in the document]
matthchr: Is this accurate? I think it is true, but I am not sure anything actually _breaks_ here as you're only applying the check at validation time and not at image provisioning time, so provisioning will still work?

pkg/providers/imagefamily/compatibility.go
```diff
@@ -0,0 +1,182 @@
...
+func kubernetesVersionPolicyFor(nodeClass *v1beta1.AKSNodeClass) (kubernetesVersionPolicy, bool) {
+	// If no nodeclass, there is no image family to check, so return false.
+	if nodeClass == nil {
+		return kubernetesVersionPolicy{}, false
+	}
+
+	// If the nodeclass specifies FIPS mode, check the FIPS-specific policy first.
+	usesFIPS := lo.FromPtr(nodeClass.Spec.FIPSMode) == v1beta1.FIPSModeFIPS
+	if usesFIPS {
+		if policy, found := kubernetesFIPSVersionPinnedImageFamilies[lo.FromPtr(nodeClass.Spec.ImageFamily)]; found {
```
matthchr, line 99: Should this be an error if we don't have a mapping? I think it's not possible to fall through here currently because we have CEL expressions blocking 2404 + FIPS, so erroring here might be a good idea so that if we do eventually support that we note to fix it here too.

pkg/controllers/nodeclass/status/validation.go
```diff
@@ -71,32 +75,18 @@ func NewValidationReconciler(
...
-			nodeClass.StatusConditions().SetFalse(
-				v1beta1.ConditionTypeValidationSucceeded,
-				KataRequiresAzureLinux3,
-				fmt.Sprintf("workloadRuntime KataVmIsolation requires Azure Linux 3 and Kubernetes 1.32 or newer; Kubernetes version %s resolves imageFamily AzureLinux to Azure Linux 2", kubernetesVersion),
-			)
-			return reconcile.Result{}, nil
-		}
+
+	// Validation for image family compatibility
+	if validationFailed, err := validateImageFamilyCompatibility(ctx, nodeClass); err != nil {
```
matthchr: If a user triggers this condition, it still triggers drift, although I think because NodeClaims won't launch (Ready condition is failing) it won't actually drift the nodes...

but I wonder if we should only mark drifted if the NodeClass is in a successful ready condition? @tallaxes what do you think?

Assuming we agree to make a change there, I dont think it needs to be done in this PR.

pkg/controllers/controllers.go
theunrepentantgeek: Needed to add `nolintlint` to forestall oscillation - one linter forcing removal of the line, another forcing it to be restored.

TBH, I'm not sure whether it's a `golangci-lint` caching issue, or different versions between different environments.

pkg/controllers/nodeclass/status/validation.go
theunrepentantgeek: Good suggestion, particularly given we have other validation coming after this.

pkg/controllers/nodeclass/status/validation_test.go
theunrepentantgeek: It's used as a factory for many different kinds of authorization error, but it's really a more generic helper. Renamed.

pkg/providers/imagefamily/compatibility.go
theunrepentantgeek: Are you saying the two sets of policies are mutually exclusive - that FIPS nodes have a completely separate set of constraints? I saw it more as a set of overrides to tune the rules for FIPS nodes, but the other rules still applied.

PR #1908, opened 2026-09-14
feat: add metrics for capacity recommendation provider

matthchr:
**Description**

* Unit tests

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


pkg/metrics/constants.go
```diff
@@ -31,4 +31,10 @@ const (
 	CapacityTypeLabel = "capacity_type"
 	NodePoolLabel     = "nodepool"
 	PhaseLabel        = "phase"
+	// capacity recommendation keys
+	PriorityLabel           = "priority"
```
tallaxes: We already have capacity_type ... Is there are good reason to deviate on key and values?

pkg/providers/capacityrecommendation/metrics.go
```diff
@@ -0,0 +1,154 @@
...
+		prometheus.CounterOpts{
+			Namespace: metrics.Namespace,
+			Subsystem: capacityRecommendationSubsystem,
+			Name:      "requests_total",
+			Help:      "Total number of SKU Mix Placement API requests completed.",
+		},
+		[]string{metrics.PriorityLabel, metrics.AllocationStrategyLabel, metrics.OSTypeLabel, metrics.PlacementScopeLabel, metrics.ResultLabel},
+	)
+
+	// SKUMixPlacementCacheHitMetric tracks SKU Mix Placement recommendation cache hits.
```
tallaxes: I presume the actual intent is to get (or calculate) a hit ratio? Can it be done reliably from this + other metrics (like API call count)? Not sure if every cache miss means an API call ... Otherwise may need to capture both hits and misses explicitly

tallaxes (dismissed):

theunrepentantgeek (dismissed):

pkg/metrics/constants.go
matthchr: No there is not (and that was one of the advantages of using this constants file too, so that we could catch overlap). Totally missed this though good catch.

matthchr: Fixed.

pkg/providers/capacityrecommendation/metrics.go
matthchr: Good call, I've fix this to log both hits and misses.

tallaxes (dismissed):

tallaxes (dismissed):

tallaxes (approved):

PR #1909, opened 2026-09-14
feat: add CapacityBuffer support

### LINKED ISSUE #5873: NAP supports CapacityBuffers
This feature would allow support for CapacityBuffer CRD in Node Auto Provisioning (NAP) or managed Karpenter. \
CapacityBuffer allows pre-provisioned spare capacity that workloads can be scheduled without waiting for new nodes to be provisioned.

This feature is rolled out in our OSS Azure Karpenter Provider [v1.14.0](https://github.com/Azure/karpenter-provider-azure/releases/tag/v1.14.0) and will require some additional work to enable in NAP.

See upstream documentation for more details: [Capacity Buffers - Karpenter](https://karpenter.sh/docs/concepts/capacitybuffers/)

ETA: October 2026

tallaxes:
**Description**

Fixes: https://github.com/Azure/AKS/issues/5873

Adds alpha CapacityBuffer support through Karpenter core rather than an Azure-specific controller or cloud-provider API.

- upgrades Karpenter core to v1.14.1 and packages the upstream `autoscaling.x-k8s.io/v1beta1` CRD in the provider, main Helm chart, and standalone CRD chart
- adds `settings.featureGates.capacityBuffer` to the public chart with a default of `false`, including missing legacy-value-safe rendering for self-hosted upgrades
- adds gate-conditional CapacityBuffer and PodTemplate RBAC and requires the v1beta1 API during startup only when the gate is enabled
- keeps generated self-hosted values default-disabled and has the dedicated in-cluster E2E suite temporarily enable the gate and exact required RBAC, restoring the original configuration afterward
- adds provider acceptance coverage for VM and AKS Machine API provisioning and a dedicated Azure E2E suite that locally mirrors the relevant upstream CapacityBuffer regression contract, with Azure-specific adaptations where upstream assumes fixed KWOK node shapes
- adds a bounded example and self-hosted alpha documentation covering installation, cost controls, supported workload shapes, and known limitations

The public self-hosted chart remains default-disabled. Karpenter v1.14.1 still contains the deletion/reference-loss trigger race tracked by kubernetes-sigs/karpenter#3258, so three deletion-path E2E regressions remain pending and managed default-on rollout should wait for the upstream fix. Live Machine API E2E was not run in this implementation session; provider acceptance coverage exercises the Machine API implementation.

* `make verify`
* `go test ./pkg/operator -ginkgo.focus=getRequiredGVKs -count=1`
* `go test ./pkg/cloudprovider -ginkgo.focus=CapacityBuffer -count=1`
* live `aksscriptless` AKS E2E coverage for 20 runnable CapacityBuffer scenarios, including a disabled-install dynamic enable/restore check, scale from zero, refill, sizing and limits, PodTemplate and scalableRef anti-affinity, zonal/regional/spot placement, restart reconstruction, drift, expiration, heterogeneous buffers, and bounded scale
* ✅ [E2E · Azure/karpenter-provider-azure@237d3cc](https://github.com/Azure/karpenter-provider-azure/actions/runs/34908891363)
* ✅ [E2E · Azure/karpenter-provider-azure@237d3cc](https://github.com/Azure/karpenter-provider-azure/actions/runs/34914332254) - aksmachineapi

- [ ] Yes, issue opened: #
- [ ] No

**Release Note**

```release-note
Adds alpha CapacityBuffer support. The self-hosted Helm feature gate remains disabled by default; install or upgrade the autoscaling.x-k8s.io/v1beta1 CapacityBuffer CRD before enabling settings.featureGates.capacityBuffer.
```


docs/capacity-buffer.md
```diff
@@ -0,0 +1,70 @@
...
+
+```yaml
+settings:
+  featureGates:
+    capacityBuffer: true
+```
+
+The main chart and standalone CRD chart both ship `autoscaling.x-k8s.io/v1beta1`. Keep the controller and CRD chart on the same release. When using the standalone CRD chart, upgrade it before enabling the controller gate.
+
+Enabling the gate also grants the controller permission to read CapacityBuffers and PodTemplates and update CapacityBuffer status.
```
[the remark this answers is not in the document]
tallaxes: The chart does ship CapacityBuffer CRD, possibly tripping on symlink?

pkg/operator/operator.go
```diff
@@ -444,7 +446,8 @@ func getRequiredGVKs() []schema.GroupVersionKind {
 			return false
 		}
 
-		return gvk.Group == karpapis.Group || gvk.Group == v1beta1.Group
+		return gvk.Group == karpapis.Group || gvk.Group == v1beta1.Group ||
+			(capacityBufferEnabled && gvk.Group == autoscalingv1beta1.Group)
```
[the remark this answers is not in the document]
tallaxes: Same as above, the CRD is shipped.

hack/validation/capacitybuffer.sh
```diff
@@ -0,0 +1,146 @@
...
+done
+
+core_crd="$(go list -m -f '{{.Dir}}' sigs.k8s.io/karpenter)/pkg/apis/crds/autoscaling.x-k8s.io_capacitybuffers.yaml"
+# The module dependency is authoritative. Generated/copied/chart paths must
+# retain its API version and exact schema.
+for crd in \
+    "$core_crd" \
+    pkg/apis/crds/autoscaling.x-k8s.io_capacitybuffers.yaml \
+    charts/karpenter-crd/templates/autoscaling.x-k8s.io_capacitybuffers.yaml \
+    charts/karpenter/crds/autoscaling.x-k8s.io_capacitybuffers.yaml; do
```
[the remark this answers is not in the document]
tallaxes: Same, the CRD is there/shipped

Makefile
```diff
@@ -95,12 +96,15 @@ verify: tidy download ## Verify code. Includes dependencies, linting, formatting
...
-	cp $(addprefix $(KARPENTER_CORE_DIR)/pkg/apis/crds/,$(KARPENTER_CORE_CRDS)) pkg/apis/crds
+	cp -f $(addprefix $(KARPENTER_CORE_DIR)/pkg/apis/crds/,$(KARPENTER_CORE_CRDS)) pkg/apis/crds
+	chmod u+w $(addprefix pkg/apis/crds/,$(KARPENTER_CORE_CRDS))
 	hack/validation/kubelet.sh
 	hack/validation/labels.sh
 	hack/validation/requirements.sh
 	hack/mutation/kubectl_get_ux.sh
-	cp $(addprefix pkg/apis/crds/,$(SUPPORTED_CRDS)) charts/karpenter-crd/templates
+	cp -f $(addprefix pkg/apis/crds/,$(SUPPORTED_CRDS)) charts/karpenter-crd/templates
+	chmod u+w $(addprefix charts/karpenter-crd/templates/,$(SUPPORTED_CRDS))
```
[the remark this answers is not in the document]
tallaxes: Incorrect; those that are not copied are symlinked.

Makefile
```diff
@@ -95,12 +96,15 @@ verify: tidy download ## Verify code. Includes dependencies, linting, formatting
...
+	hack/validation/capacitybuffer.sh
```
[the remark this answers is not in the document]
tallaxes: CI passes

hack/validation/capacitybuffer.sh
```diff
@@ -0,0 +1,145 @@
+#!/usr/bin/env bash
+set -euo pipefail
```
matthchr: This file looks like tests for the capacity buffer CRD + Helm chart, which is good to have. As far as I can tell, this is not actually editing the 

I'm curious about how this compares with the files in this same folder for labels or requirements which are actually a bit less validation and a bit more mutating?

matthchr: Would it make sense to (possibly in a different PR?) clarify which scripts are mutating and which are validating? Current directory structure makes it look like everything is validation when is actually not true - this proposed script is more in-line with validation than the others in the folder.

examples/v1/capacity-buffer.yaml
```diff
@@ -0,0 +1,80 @@
...
+apiVersion: v1
+kind: PodTemplate
+metadata:
+  name: general-purpose-buffer
+  namespace: default
+template:
+  spec:
+    containers:
+      - name: workload
+        image: mcr.microsoft.com/oss/kubernetes/pause:3.6
```
matthchr: Workload doesn't really matter here because it's not run, is that right? If so... put a comment to that effect here?

docs/capacity-buffer.md
```diff
@@ -0,0 +1,70 @@
+# CapacityBuffer
+
+Status: alpha, opt-in for self-hosted deployments
+
+CapacityBuffer lets Karpenter maintain spare schedulable capacity before real workloads need it. Karpenter resolves a workload shape, creates virtual Pods in memory, and provisions or preserves nodes for that demand. It does not create placeholder Pods in the Kubernetes API.
+
+## Enable The Feature
+
+CapacityBuffer is disabled by default. Enable it when installing or upgrading the self-hosted Karpenter chart:
```
matthchr: Is the focus here selfhosted or NAP?

Disabled by default in selfhosted but not true for NAP, is that right?

docs/capacity-buffer.md
```diff
@@ -0,0 +1,70 @@
...
+
+```yaml
+settings:
+  featureGates:
+    capacityBuffer: true
+```
+
+The main chart and standalone CRD chart both ship `autoscaling.x-k8s.io/v1beta1`. Keep the controller and CRD chart on the same release. When using the standalone CRD chart, upgrade it before enabling the controller gate.
```
matthchr: minor: We could maybe just cut this whole sentence?

docs/capacity-buffer.md
```diff
@@ -0,0 +1,70 @@
...
+> Karpenter core v1.14.1 has a known deletion/reference-loss defect tracked in
+> [kubernetes-sigs/karpenter#3258](https://github.com/kubernetes-sigs/karpenter/issues/3258).
+> On an otherwise idle cluster, stale in-memory buffer placement can delay
+> `WhenEmpty` disruption after a CapacityBuffer or its backing reference is
+> deleted. Keep the feature limited to alpha evaluation until a fixed core
+> revision is adopted. Restarting the controller rebuilds this in-memory state.
+
+## Create A Buffer
+
+Apply the bounded PodTemplate example:
```
matthchr:
```suggestion
Apply the CapacityBuffer example:
```
?

examples/v1/capacity-buffer.yaml
```diff
@@ -0,0 +1,80 @@
...
+      kubernetes.io/os: linux
+---
+apiVersion: autoscaling.x-k8s.io/v1beta1
+kind: CapacityBuffer
+metadata:
+  name: general-purpose-buffer
+  namespace: default
+spec:
+  provisioningStrategy: buffer.x-k8s.io/active-capacity
+  podTemplateRef:
```
matthchr: Could include a comment here indicating what the docs already state:
> A CapacityBuffer can instead reference a Deployment, ReplicaSet, or StatefulSet through `spec.scalableRef`. References must be in the same namespace. Karpenter periodically resolves scalable workload size and computes the requested buffer percentage.

Could even leave such a section commented out here?

docs/capacity-buffer.md
```diff
@@ -0,0 +1,70 @@
...
+
+```bash
+kubectl describe capacitybuffer general-purpose-buffer
+```
+
+Deleting or reducing a buffer removes virtual demand. After the known v1.14.1
+deletion issue above is resolved or the controller is restarted, excess nodes
+are removed according to normal NodePool disruption budgets and
+`consolidateAfter`; scale-down is not immediate.
+
```
matthchr: Document seems maybe missing a link to the upstream capacity buffer reference documentation?
https://karpenter.sh/docs/concepts/capacitybuffers/

examples/v1/capacity-buffer.yaml
```diff
@@ -0,0 +1,80 @@
...
+  spec:
+    containers:
+      - name: workload
+        image: mcr.microsoft.com/oss/kubernetes/pause:3.6
+        resources:
+          requests:
+            cpu: "1"
+            memory: 1Gi
+    nodeSelector:
+      capacity-buffer.example/enabled: "true"
```
matthchr: Is this label actually required? I assume no and that you could use any label? Do we "need" a particular capacity buffer specific label? I guess we must, as if we were not forcing these pods onto these capacity buffer nodes then they could be simulated onto other nodes which would be bad? May be worth clarifying that in a comment here?

charts/karpenter/templates/deployment.yaml
```diff
@@ -134,7 +134,7 @@ spec:
               value: "{{ join "," . }}"
           {{- end }}
             - name: FEATURE_GATES
-              value: "ReservedCapacity={{ .Values.settings.featureGates.reservedCapacity }},SpotToSpotConsolidation={{ .Values.settings.featureGates.spotToSpotConsolidation }},NodeRepair={{ .Values.settings.featureGates.nodeRepair }},NodeOverlay={{ .Values.settings.featureGates.nodeOverlay }},StaticCapacity={{ .Values.settings.featureGates.staticCapacity }}"
+              value: "CapacityBuffer={{ dig "capacityBuffer" false .Values.settings.featureGates }},ReservedCapacity={{ .Values.settings.featureGates.reservedCapacity }},SpotToSpotConsolidation={{ .Values.settings.featureGates.spotToSpotConsolidation }},NodeRepair={{ .Values.settings.featureGates.nodeRepair }},NodeOverlay={{ .Values.settings.featureGates.nodeOverlay }},StaticCapacity={{ .Values.settings.featureGates.staticCapacity }}"
```
matthchr: minor: inconsistent with other feature gate rendering.

Should we be using `dig` in the other cases too? I guess maybe at this point not fully necessary but the inconsistency here is somewhat jarring?

charts/karpenter/crds/autoscaling.x-k8s.io_capacitybuffers.yaml
```diff
@@ -0,0 +1 @@
+../../../pkg/apis/crds/autoscaling.x-k8s.io_capacitybuffers.yaml
```
matthchr: This doens't look right?

matthchr: Ooh wait this is a symlink ignore me =)

charts/karpenter-crd/templates/autoscaling.x-k8s.io_capacitybuffers.yaml
```diff
@@ -0,0 +1,263 @@
+---
+apiVersion: apiextensions.k8s.io/v1
```
matthchr: Missing a guard where we don't deploy it if the feature gate is disabled? Though I guess we don't have one for NodeOverlay either. Does AWS avoid deploying the CRD if the gate is disabled?

Fine leaving it as-is, was just curious.

matthchr: This is not a symlink but the karpenter/crds one is? Why?

pkg/cloudprovider/suite_aksmachineapi_features_test.go
```diff
@@ -82,6 +83,12 @@ var _ = Describe("CloudProvider", func() {
 			azureEnv.Reset(ctx)
 			azureEnvNonZonal.Reset(ctx)
 		})
+
+		runCapacityBufferTests(func() {
+			Expect(azureEnv.AKSMachinesAPI.AKSMachineCreateOrUpdateBehavior.CalledWithInput.Len()).To(Equal(1))
```
matthchr: minor: Doesn't this assume the structure of the underlying tests? Wouldn't it be better to have the function take a parameter like `expectedCalls int` and then just assert equal to that int (w/ the other always asserting 0 since no calls are ever expected?)

pkg/operator/operator.go
```diff
@@ -444,7 +446,8 @@ func getRequiredGVKs() []schema.GroupVersionKind {
 			return false
 		}
 
-		return gvk.Group == karpapis.Group || gvk.Group == v1beta1.Group
+		return gvk.Group == karpapis.Group || gvk.Group == v1beta1.Group ||
```
matthchr: minor, style, feel free to ignore: now 3 this could be a set/list and we could use a `lo.contains` or similar as opposed to doing the 3 checks manually.

Obviously not a big deal though

Makefile
```diff
@@ -95,12 +96,15 @@ verify: tidy download ## Verify code. Includes dependencies, linting, formatting
 	make az-swagger-generate-clients-raw
 	go generate ./...
 	hack/boilerplate.sh
-	cp $(addprefix $(KARPENTER_CORE_DIR)/pkg/apis/crds/,$(KARPENTER_CORE_CRDS)) pkg/apis/crds
```
matthchr: minor: Some comments here about the sections (copy from core dir... run mutating scripts... run validating scripts...) might help clarity.

docs/capacity-buffer.md
matthchr: 🤖 **Existing-upgrade CRD ordering:** Helm does not install newly added resources from a chart `crds/` directory during a normal `helm upgrade`. An existing installation that enables this gate could therefore start requiring the CapacityBuffer GVK without having installed the CRD, causing controller startup to time out and affecting all provisioning. Could this document require applying/upgrading the version-matched CRD before enabling the gate for every existing-installation upgrade path, not only when the standalone CRD chart is already in use?

pkg/apis/crds/autoscaling.x-k8s.io_capacitybuffers.yaml
```diff
@@ -0,0 +1,263 @@
...
+                description: |-
+                  provisioningStrategy defines how the buffer is utilized.
+                  "buffer.x-k8s.io/active-capacity" is the default strategy, where the buffer actively scales up the cluster by creating placeholder pods.
+                type: string
+              replicas:
+                description: |-
+                  replicas defines the desired number of buffer chunks to provision.
+                  If neither `replicas` nor `percentage` is set, as many chunks as fit within
+                  defined resource limits (if any) will be created. If both are set, the minimum
+                  of the two will be used.
```
matthchr: 🤖 **The installed API contract differs from the pinned runtime behavior.** This description says `replicas` and `percentage` use the minimum, but Karpenter v1.14.1 `computeReplicas` takes their maximum before applying `limits`; the new E2E test also expects `replicas: 1` plus 40% of five replicas to produce two chunks. The same CRD says an empty `scalableRef.apiGroup` means the core group, while the resolver defaults it to `apps`. Since this schema is user-facing and capacity affects cost, can these descriptions be corrected upstream (and documented locally until the corrected CRD is consumed)?

charts/karpenter/templates/clusterrole-core.yaml
```diff
@@ -41,6 +41,14 @@ rules:
   - apiGroups: ["apps"]
     resources: ["daemonsets", "deployments", "replicasets", "statefulsets"]
     verbs: ["list", "watch", "get"]
+  {{- if eq (toString (dig "capacityBuffer" false .Values.settings.featureGates)) "true" }}
```
matthchr: 🤖 **Helm and runtime can disagree about this gate.** This exact, case-sensitive comparison only enables RBAC for Boolean `true` or string `"true"`, while Karpenter parses `FEATURE_GATES` with `strconv.ParseBool`, which also accepts forms such as `True`, `TRUE`, `t`, and `1`. For example, `capacityBuffer: "True"` enables the controller but omits these permissions, resulting in authorization failures. Could the value be strictly schema-enforced as a Boolean, or normalized once in a shared helper used by both the Deployment and RBAC templates?

pkg/apis/crds/autoscaling.x-k8s.io_capacitybuffers.yaml
```diff
@@ -0,0 +1,263 @@
...
+                - kind
+                - name
+                type: object
+            type: object
+            x-kubernetes-validations:
+            - message: If podTemplateRef is set, replicas or limits must also be set
+              rule: '!has(self.podTemplateRef) || has(self.replicas) || has(self.limits)'
+            - message: You must define either PodTemplateRef or ScalableRef, but not
+                both
+              rule: '!(has(self.podTemplateRef) && has(self.scalableRef))'
```
matthchr: 🤖 **This validation does not enforce the documented “exactly one” requirement.** It rejects the case where both references are present, but permits an object with neither `podTemplateRef` nor `scalableRef`; that object is accepted by the API and only fails later during reconciliation. Could this be an XOR validation that requires exactly one reference? It would also be useful to reject `percentage` unless `scalableRef` is present.

docs/capacity-buffer.md
matthchr: I think that this is a fundamental limitation of Helm charts + crd folder. Not sure we can do anything about it here but it might be worth creating an item to determine if we should be doing something different than we are...? I guess maybe that rolls into the "controller should install its own CRDs" option we've talked about before which I think we do have an item tracking.

test/suites/capacitybuffer/capacitybuffer_test.go
```diff
@@ -0,0 +1,859 @@
...
+)
+
+// This suite mirrors the relevant behavioral contract from
+// sigs.k8s.io/karpenter/test/suites/regression/capacitybuffer_test.go so it can
+// run independently in Azure provider CI. Correspondence comments identify
+// direct replicas, Azure-specific adaptations, and additional provider coverage.
+var _ = Describe("CapacityBuffer", func() {
+	BeforeEach(func() {
+		if !env.InClusterController {
+			Skip("CapacityBuffer tests require the controller to be running in-cluster")
```
matthchr: Do they? Aren't we going to enable the feature gate for it in NAP? So we could run them there too, couldn't we?

test/suites/capacitybuffer/capacitybuffer_test.go
```diff
@@ -0,0 +1,859 @@
...
+		}
+		env.ExpectUpdated(nodePool)
+
+		env.EventuallyExpectRegisteredNodeClaimCount("==", 1)
+		EventuallyExpectCapacityBufferProvisioned(env, env.Client, buffer)
+	})
+
+	// Upstream correspondence: "should not provision buffer capacity beyond NodePool CPU limit".
+	// Azure adaptation: assert aggregate NodeClaim CPU stays within the limit rather than assuming two 2-CPU nodes.
+	It("should partially satisfy a buffer without exceeding a nonzero NodePool CPU limit", func() {
```
matthchr: minor: Curious why we named our tests differently, if they are basically mirror upstream tests we could have them be named exactly the same? (Doesn't mean we get rid of the upstream correspondence comment, that seems useful either way).

test/suites/capacitybuffer/capacitybuffer_test.go
```diff
@@ -0,0 +1,859 @@
...
+func provisioningTransitionTime(buffer *autoscalingv1beta1.CapacityBuffer) metav1.Time {
+	GinkgoHelper()
+	current := &autoscalingv1beta1.CapacityBuffer{}
+	Expect(env.Client.Get(env, client.ObjectKeyFromObject(buffer), current)).To(Succeed())
+	condition := apimeta.FindStatusCondition(current.Status.Conditions, autoscalingv1beta1.ProvisioningCondition)
+	Expect(condition).ToNot(BeNil())
+	return condition.LastTransitionTime
+}
+
+func eventuallyExpectNodeClaimDeleted(nodeClaim *karpv1.NodeClaim, timeout time.Duration) {
```
matthchr: Is there not an existing helper that already does this?

matthchr: env.EventuallyExpectNotFound(nodeClaim) already exists. You could just use that or make a variant with a timeout, which seems to be the only difference?

PR #1921, opened 2026-09-18
feat: support compute quota categories

matthchr:
* If category is enabled, use it.
* If not, don't.
* Note that I have not currently tested this on a real subscription with categories enabled because we don't currently have one, but given that it definitely won't work without this change I think this is at least a good start and I can do some more testing in parallel

* Unit tests
* E2E test (regression): https://github.com/Azure/karpenter-provider-azure/actions/runs/35289721399

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note

```


pkg/providers/quota/quota.go
```diff
@@ -62,57 +74,133 @@ type Provider interface {
...
 }
 
 func (p *DefaultProvider) Update(ctx context.Context) error {
-	freshUsages := map[string]*armcompute.Usage{}
+	freshUsages, err := p.listUsages(ctx)
+	if err != nil {
+		return err
+	}
+	// Build mapping of family to quota (for quota categories)
+	freshFamilyToQuotaName := map[string]string{}
```
matthchr, line 110: I thought about having this in its own controller but it seemed overkill and the throttling limits on this API are high (just ARM GET throttling from what I can tell, which is something like 1200/5m). Since we'll call 1x/10m we're way under the limits there and it seems OK to just do it here.

pkg/controllers/quota/suite_test.go
```diff
@@ -28,9 +28,12 @@ import (
 	clock "k8s.io/utils/clock/testing"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/compute/armcompute/v7"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/computelimit/armcomputelimit"
```
matthchr, line 31: I need to do some testing to ensure that we have permissions to call this API before this merges.

tallaxes: If extra permissions are needed, lets also document this for self-hosted use (if applicable)

tallaxes (approved):

pkg/controllers/quota/suite_test.go
matthchr: I'm like 90% sure they aren't needed based on my current testing but just need to confirm

matthchr: Confirmed things work as-is, no additional perms needed.

PR #1934, opened 2026-09-21
fix: garbage collect AKS machines with missing backing VMs

GuetaHen:
Related issue: none (standalone bug fix).

**Description**

### Failure sequence and circular cleanup dependency

An AKS Machine resource and its backing Compute VM are separate resources. The Machine can remain `Succeeded`, retaining the original VM resource ID, after the VM has been deleted. A successful Machine GET therefore does not necessarily prove that the VM still exists.

1. The backing VM is deleted, for example by Azure Spot eviction with the Delete eviction policy.
2. The kubelet disappears and the Kubernetes Node becomes NotReady or Unknown, but the AKS Machine remains `Succeeded` and the registered NodeClaim remains present.
3. Core NodeClaim garbage collection calls `CloudProvider.List`. The Azure provider builds that list from AKS Machine records without checking each backing VM, so the stale Machine's provider ID still appears. Core therefore keeps the NodeClaim.
4. Azure instance garbage collection finds that same NodeClaim and treats the Machine as owned, so it keeps the stale Machine.
5. Neither collector removes its side of the relationship: the Machine keeps the NodeClaim alive, and the NodeClaim keeps the Machine alive.

This is a logical cleanup deadlock, not a controller/thread lockup; unrelated nodes can still be deleted normally. If termination has already started, `CloudProvider.Get` can also return the stale Machine instead of NotFound, so the normal missing-instance shortcut is not taken. Pods bound to the dead node can remain Terminating because no kubelet remains to complete shutdown.

### Why Spot exposes this most often, but is not the only affected capacity type

Spot eviction can remove the Compute VM before Karpenter completes its normal drain and deletion sequence. This external deletion can leave the AKS Machine and Kubernetes objects behind, and repeated evictions can accumulate the stale records. This does not mean every Spot eviction produces a stale Machine.

On-demand VMs are not subject to Spot reclamation, so that trigger is absent during their normal lifecycle. However, deleting an on-demand VM outside the normal lifecycle can produce the same Machine/VM mismatch. The bug is not GPU-specific or inherently Spot-only, and this fix deliberately does not filter by capacity type.

Existing node repair is not a reliable escape from this state: its greater-than-20%-unhealthy safety gate can block repair when many nodes in a pool are unhealthy. Pre-eviction handling also depends on receiving and processing `PreemptionScheduled=True` before the VM disappears. Post-deletion garbage collection must work independently of that signal and should not require weakening the repair safety threshold.

### Targeted fix and safety boundary

Extend the existing instance garbage collector to close this cleanup gap. For an AKS Machine with exactly one matching, registered NodeClaim and exactly one Node whose Ready condition is False or Unknown, query the backing VM through the existing VM provider. Only a typed VM-not-found result allows the existing Machine and Node deletion path to run. Existing core controllers then finish Node and NodeClaim finalization.

Ready nodes, unregistered claims, missing or ambiguous mappings, and existing VMs are preserved. Authorization, throttling, and service errors propagate without deleting anything. Healthy nodes do not incur extra Compute lookups. Provider-ID maps are built once per reconcile rather than repeatedly scanning the cluster.

This uses the existing controller, VM client, and deletion/finalization paths. It adds no controller, configuration, CRD, permission, or operational component, and does not change node-repair thresholds, disruption budgets, capacity-type selection, or the five-minute grace period for ordinary ownerless-instance cleanup. It covers both AKS Machine create modes and existing Machines managed in VM mode; ordinary VM cleanup is unchanged.

### Scope and related work

This fixes recovery from the stale-Machine/missing-VM mismatch; it does not fix the AKS service-side reason that the Machine remains `Succeeded`. It also does not prevent Spot eviction, guarantee replacement capacity, change Spot-to-on-demand scheduling, or repair live GPU nodes that fail to advertise their GPU resources.

The pre-eviction proposal in #1829 is complementary: it uses the Spot notice deadline before the VM is reclaimed. This PR handles cleanup after the VM is gone, including when that notice was absent or could not be acted on. It intentionally retains a narrow scope rather than adding an interruption controller or combining independent scheduling and initialization fixes.

* Existing garbage-collection suite before the change: 20/20 passed.
* Red regression run against the original implementation: 9 failures among 33 cases, demonstrating the missing cleanup path and error-handling expectations.
* Final `go test ./pkg/controllers/nodeclaim/garbagecollection ./pkg/controllers -count=1 -timeout=15m -v`: 37/37 garbage-collection specs passed, none skipped; controller wiring compiled. Run with Go 1.26.6, CGO disabled, and Kubernetes 1.35 envtest.
* Coverage includes both Machine API modes, mixed-mode management, Spot and on-demand orphan Machines, Ready/unregistered/ambiguous/missing-node safeguards, existing VMs, HTTP 403/429/503 failures, deletion failure/retry, and the real core termination/lifecycle controllers completing finalizers despite a do-not-disrupt pod on a deleted VM.
* `go vet ./pkg/controllers/nodeclaim/garbagecollection ./pkg/controllers`, `gofmt`, and `git diff --check` passed.
* `make verify` was attempted but could not complete because the Windows checkout's shell-script line endings produce `/usr/bin/env: 'bash\r': No such file or directory` under WSL. No dependency changes are included. Full presubmit and live Azure E2E have not been run; this patch has not been deployed to a live cluster.

- [ ] Yes, issue opened: # 
- [ ] No

**Release Note**

```release-note
Fix cleanup of AKS Machine resources whose backing VMs have been deleted, including after Spot eviction, so registered unhealthy nodes and their NodeClaims can finish normal termination.
```


tallaxes:
@GuetaHen, thank you for the proposed fix, we are taking a look at the issue, experimenting with a slightly different approach to cleanup

tallaxes:
@GuetaHen, closing in favor of https://github.com/Azure/karpenter-provider-azure/pull/1937 - merged, and will be in the next release (but feel free to pre-test if you environment allows it). Again, thank you for reporting and suggesting a fix!

